PR #120, opened 2014-11-20
[POC] Use the Section VO instead of a custom entity

bdunogier:
> Proof of concept

Feeds the Section form with the Section Value Object instead of using a custom object.

Requirements aren't respected, but the existing @require annotations could be replaced with the ones from Form. A next step would be to have 2 distinct forms for Create and Update,
so that they can have their own ValueObject.

Tested manually from `/pjax/section/create`. It works :-)

### Approach
@dpobel told me that this is the approach that was tried a couple month back, and that we didn't go for it for "reasons", one of them being form validation (see below).

#### Form validation
The most common way to validate forms is through

#### Existing contributions
- netgen/NetgenEzFormsBundle, written by @pspanja. Has a data wrapper for Content & User, and create/update forms for it. Uses a registry to convert fields.
- heliopsis/ezforms-bundle, written by @bchoquet-heliopsis. More mature than Netgen's solution.

### DIFF
--- a/Controller/SectionController.php
+++ b/Controller/SectionController.php
@@ -10,7 +10,7 @@ namespace EzSystems\PlatformUIBundle\Controller;
 
 use eZ\Publish\API\Repository\Exceptions\NotFoundException;
 use eZ\Publish\API\Repository\Exceptions\UnauthorizedException;
-use EzSystems\PlatformUIBundle\Entity\Section;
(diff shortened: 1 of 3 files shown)


dpobel:
Closing as Repository form has been introduced in the meantime and Section form management is being refactored to use it in https://github.com/ezsystems/PlatformUIBundle/pull/288

PR #145, opened 2014-12-30
EZS-41: Changed the way the user redirection urls are generated

sunpietro:
Changed the way the user redirection urls are generated in the `contentCreateViewService` allowing users to overwrite the default values of the redirection attributes using plugins or another services.


sunpietro:
Closing the PR as it's not needed anymore.

PR #147, opened 2014-12-31
EZP-23813: As an editor, I want to be able to fill the Media fields

dpobel:
JIRA: https://jira.ez.no/browse/EZP-23813

# Description

This patch adds the field edit view for the Media (video/audio) fields.
Screencast: http://youtu.be/UgPLrlGjq6w

# Known issues/limitations

* [The REST API has a bug preventing the Media field from being changed](https://jira.ez.no/browse/EZP-23837), you need the rough patch available in https://github.com/ezsystems/ezpublish-kernel/pull/1121 to do what is presented in the screencast
* As for Image, it's not possible to change the Media settings (loop, autoplay, ...) without providing a file (https://jira.ez.no/browse/EZP-23845)
* In the field definition settings, it's possible to set the *Media player type* to Flash, QuickTime, RealPlayer (!!!), Silverlight (!!), Windows Media Player, Html5 Video, Html5 Audio. Only the last two options are really handled right now. Actually, as in legacy, you can put any kind of file in the Media fields but the edit preview will only work for file format recognized by the browser.
* No drag and drop yet (https://jira.ez.no/browse/EZP-23789)

# Tests

unit tests + manual tests


andrerom:
+1

PR #149, opened 2015-01-06
EZP-23745 time edit for old browsers

StephaneDiot:
## Description

This is the time edit field for browsers not supporting time HTML5 input (like firefox browser for example). Time has to be fill with a HH:MM(:SS) pattern.

## Link

jira: https://jira.ez.no/browse/EZP-23745

### DIFF
--- a/Resources/config/yui.yml
+++ b/Resources/config/yui.yml
@@ -206,7 +206,7 @@ system:
                     type: 'template'
                     path: %ez_platformui.public_dir%/templates/fields/view/dateandtime.hbt
                 ez-time-editview:
-                    requires: ['ez-fieldeditview', 'event-valuechange', 'timeeditview-ez-template', 'datatype-date-format']
+                    requires: ['ez-fieldeditview', 'event-valuechange', 'timeeditview-ez-template', 'datatype-date-format', 'datatype-date']
                     path: %ez_platformui.public_dir%/js/views/fields/ez-time-editview.js
                 timeeditview-ez-template:
                     type: 'template'
--- a/Resources/public/js/views/fields/ez-time-editview.js
+++ b/Resources/public/js/views/fields/ez-time-editview.js
@@ -27,6 +27,55 @@ YUI.add('ez-time-editview', function (Y) {
             '.ez-time-input-ui input': {
                 'blur': 'validate',
                 'valuechange': 'validate',
+            },
+        },
+
+        /**
+         * Check if browser supports time input
+         *
+         * @method _detectInputTimeSupport
+         * @private
+         */
+        _detectInputTimeSupport: function () {
+            var i = document.createElement("input");
+
+            i.setAttribute("type", "time");
+            return i.type === "time";
+        },
+
+        /**
+         * Validation for browsers supporting time input
+         *
+         * @protected
+         * @method _supportedTimeInputValidate
+         */
+        _supportedTimeInputValidate: function () {
+            var validity = this._getInputValidity();
+
+            if ( validity.valueMissing ) {
+                this.set('errorStatus', 'This field is required');
+            } else if ( validity.badInput ) {
+                this.set('errorStatus', 'This is not a valid input');
+            } else {
+                this.set('errorStatus', false);
+            }
+        },
+
+        /**
+         * Validation for browsers NOT supporting time input
+         *
+         * @protected
+         * @method _unsupportedTimeInputValidate
+         */
+        _unsupportedTimeInputValidate: function () {
+            var validity = this._getInputValidity();
+
+            if ( validity.valueMissing ) {
+                this.set('errorStatus', 'This field is required');
+            } else if ( validity.patternMismatch ) {
+                this.set('errorStatus', 'This time is invalid, enter a correct time: HH:MM(:SS)');
+            } else {
+                this.set('errorStatus', false);
             }
         },
 
@@ -36,14 +85,10 @@ YUI.add('ez-time-editview', function (Y) {
          * @method validate
          */
         validate: function () {
-            var validity = this._getInputValidity();
-
-            if ( validity.valueMissing ) {
-                this.set('errorStatus', 'This field is required');
-            } else if ( validity.badInput ) {
-                this.set('errorStatus', 'This is not a valid input');
+            if (this.get('supportsTimeInput')) {
+                this._supportedTimeInputValidate();
             } else {
-                this.set('errorStatus', false);
+                this._unsupportedTimeInputValidate();
             }
         },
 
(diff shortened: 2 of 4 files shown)


andrerom:
Note: if these things are not related to this PR, please fix them in seperate commit as @ezrobot is back online after the ci upgrade in december so it reports cs issues merged to master in between.

Resources/public/js/views/fields/ez-time-editview.js
```diff
@@ -57,14 +104,35 @@ YUI.add('ez-time-editview', function (Y) {
...
             var def = this.get('fieldDefinition'),
                 field = this.get('field'),
-                time = '';
+                time = '',
+                hours = '',
+                minutes = '',
+                seconds = '',
+                isEmpty = true;
 
+            if (!this.get('supportsTimeInput')) {
```
dpobel: this can be simplified. No matter if the time input is supported, you always send 4 variables to the template `isRequired`, `supportsTimeInput`, `html5InputTime`, `useSeconds`. But you have to make sure that `useSeconds` is taken into account when formatting the `html5InputTime` value

Resources/public/js/views/fields/ez-time-editview.js
```diff
@@ -104,13 +172,48 @@ YUI.add('ez-time-editview', function (Y) {
...
+            var valueOfInput,
+                time,
+                hoursTimestamp,
+                minutesTimestamp,
+                secondsTimestamp = 0;
 
-            if (valueOfInput) {
-                return valueOfInput/1000;
+            if (this.get('supportsTimeInput')) {
+                valueOfInput = this._getInputNode().get('valueAsNumber');
```
dpobel: you should adopt the same organization as in `validate`, ie something like:

```js
if ( this.get('supportsTimeInput') ) {
    return this._supportedTimeInputGetFieldValue();
} else {
    return this._unsupportedTimeInputGetFieldValue();
}
```

Resources/public/js/views/fields/ez-time-editview.js
```diff
@@ -104,13 +172,48 @@ YUI.add('ez-time-editview', function (Y) {
...
+            } else {
+
+                time = this._getInputNode().get('value').split(":");
+                hoursTimestamp = Y.Number.parse(time[0])*3600;
+                minutesTimestamp = Y.Number.parse(time[1])*60;
+                if ( time.length > 2 ) {
+                    secondsTimestamp = Y.Number.parse(time[2]);
+                }
+                time = hoursTimestamp + minutesTimestamp + secondsTimestamp;
+                if (time !== 0){
```
dpobel: bug here :) 0 is midnight so this method can return 0
And in any case it's useless to transform a number of seconds in ... a number of seconds :p

Resources/public/js/views/fields/ez-time-editview.js
```diff
@@ -104,13 +172,48 @@ YUI.add('ez-time-editview', function (Y) {
...
+            if (this.get('supportsTimeInput')) {
+                valueOfInput = this._getInputNode().get('valueAsNumber');
+                if (valueOfInput) {
+                    return valueOfInput/1000;
+                }
+                return null;
+            } else {
+
+                time = this._getInputNode().get('value').split(":");
+                hoursTimestamp = Y.Number.parse(time[0])*3600;
```
dpobel: you should first check that time has at least 2 entries and instead of using Y.Number.parse, you can use `parseInt(time[0], 10)`

Resources/public/js/views/fields/ez-time-editview.js
```diff
@@ -48,6 +63,38 @@ YUI.add('ez-time-editview', function (Y) {
...
+            }
+        },
+
+        /**
+         * Validates the current input of time field
+         *
+         * @method validate
+         */
+        validate: function () {
+
```
dpobel: nitpick CS: useless empty line

Resources/public/templates/fields/edit/time.hbt
```diff
@@ -9,11 +9,38 @@
...
+        <div class="ez-editfield-input"><div class="ez-time-input-ui">
+            {{#if supportsTimeInput}}
+                <input type="time" {{#if useSeconds}}step="1"{{/if}}
+                     value="{{ html5InputTime }}"
+                     class="ez-validated-input"
+                     id="ez-field-{{ content.contentId }}-{{ fieldDefinition.identifier }}"
+                      {{#if isRequired}} required{{/if}}
+                >
+            {{else}}
+                <input type="text"
```
dpobel: with https://github.com/ezsystems/PlatformUIBundle/pull/149/files#r22789925 this can also be simplified. The input can always have `type="time"` even if time is not supported. When time is not supported, you'll just have to add the `pattern` and the `placeholder` (which are different depending on `useSeconds`).

Resources/public/js/views/fields/ez-time-editview.js
```diff
@@ -104,13 +172,48 @@ YUI.add('ez-time-editview', function (Y) {
...
-            if (valueOfInput) {
-                return valueOfInput/1000;
+            if (this.get('supportsTimeInput')) {
+                valueOfInput = this._getInputNode().get('valueAsNumber');
+                if (valueOfInput) {
+                    return valueOfInput/1000;
+                }
+                return null;
+            } else {
+
```
dpobel: nitpick cs: useless empty line

Resources/config/yui.yml
```diff
@@ -206,7 +206,7 @@ system:
                     type: 'template'
                     path: %ez_platformui.public_dir%/templates/fields/view/dateandtime.hbt
                 ez-time-editview:
-                    requires: ['ez-fieldeditview', 'event-valuechange', 'timeeditview-ez-template', 'datatype-date-format']
+                    requires: ['ez-fieldeditview', 'event-valuechange', 'timeeditview-ez-template', 'datatype-date-format', 'datatype-number', 'datatype-date']
```
dpobel: datatype-number is useless now (also remove it from the html test file)

dpobel:
besides the datatype-number module thing, +1

Resources/public/js/views/fields/ez-time-editview.js
```diff
@@ -57,16 +102,23 @@ YUI.add('ez-time-editview', function (Y) {
...
             }
-
             return {
                 "isRequired": def.isRequired,
-                "html5InputTime": time,
-                "useSeconds": def.fieldSettings.useSeconds
+                "supportsTimeInput": this.get('supportsTimeInput'),
+                "time": time,
+                "useSeconds": def.fieldSettings.useSeconds,
+                "isEmpty": isEmpty
```
dpobel: this is variable is useless, no ?

PR #153, opened 2015-01-16
EZP-23789: As an editor, I want to be able to drag and drop a file to fill the Image, BinaryFile and Media fields

dpobel:
JIRA: https://jira.ez.no/browse/EZP-23789

# Description

This patch adds the ability to drag and drop a file to fill the BinaryFile, Media and Image fields. In addition, a mime type check has been added on the Image (also works with the regular file selection).

Screencast: http://youtu.be/RX40EiedAzk

Tasks:

* [x] Implement basic drag and drop
* [x] Make sure it does not crash when dropping some text
* [x] Add validation on mime type in the case of Image (should also be done for "regular" file choice)
* [x] Check in IE10 http://caniuse.com/#feat=dragndrop
* [x] Unit tests
* [x] Improve general visual behavior

# Tests

unit tests and manual tests (IE10, Firefox and Chrome)

### DIFF
--- a/Resources/public/js/views/fields/ez-binarybase-editview.js
+++ b/Resources/public/js/views/fields/ez-binarybase-editview.js
@@ -15,6 +15,7 @@ YUI.add('ez-binarybase-editview', function (Y) {
 
     var HAS_WARNING = 'has-warning',
         IS_EMPTY = 'is-field-empty',
+        DRAGGING = 'is-dragging-file',
         L = Y.Lang,
         OVER_SIZE_TPL = "The file '{name}' was refused because its size is greater than the maximum allowed size ({max})",
         win = Y.config.win,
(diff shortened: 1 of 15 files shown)


dpobel:
ping @sunpietro @StephaneDiot

Resources/public/js/views/fields/ez-binarybase-editview.js
```diff
@@ -230,21 +236,50 @@ YUI.add('ez-binarybase-editview', function (Y) {
...
+        },
+
+        /**
          * Checks whether the size is valid according to the field definition
-         * configuration
+         * configuration. If the file can not be accepted, a warning message is
+         * set in the `warning` attribute.
          *
-         * @param {Number} size
+         * @param {File} file the File object to be stored in the field
```
yannickroger: is it a File we are talking about ? :)

dpobel: "File" != "file" :p
(FYI "file" is parameter name, File is a build in class https://developer.mozilla.org/en-US/docs/Web/API/File)

PR #154, opened 2015-01-19
FIX/EZP-23882 Allows to set keyword field to empty

StephaneDiot:
## Description

Setting a keyword field to empty was returning [""], this was a problem because the view was displaying an empty string instead of the classic "This field is empty". Now it returns an empty array which is what we want.
However it seems that there is a bug in the REST API because the fieldValue is not correctly updated when we return an empty array ( bug has been reported at: https://jira.ez.no/browse/EZP-23912 )...

## Link

Jira: https://jira.ez.no/browse/EZP-23882?filter=-1


StephaneDiot:
ping @dpobel

andrerom:
+1

dpobel:
+1

PR #155, opened 2015-01-21
EZP-22247 Enrich routes with config

StephaneDiot:
## Description

The goal here was to be able to enrich the routes of the application with some config elements. This is currently needed to implement the country view, but it's a generic way to do it.

## Link 

jira: https://jira.ez.no/browse/EZP-22247?filter=-1

### DIFF
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -138,9 +138,33 @@ YUI.add('ez-platformuiapp', function (Y) {
                     oldService.setNextViewServiceParameters(newService);
                 }
             });
-
+            this._routeConfig();
             // Listening for events fired on child views
             this.views.errorView.instance.addTarget(this);
+        },
+
+        /**
+         * Check if there is config to add to the application routes
+         *
+         * @protected
+         * @method _routeConfig
+         */
+        _routeConfig: function () {
+            if (this.get('routeConfig')) {
+                Y.Array.each(this.get('routes'), Y.bind(this._enrichRoute, this));
+            }
+        },
+
+        /**
+         * Enrich the route with the route configuration
+         *
+         * @protected
+         * @method _enrichRoute
+         */
+        _enrichRoute: function (route, index) {
+            if (this.get('routeConfig')[route.name]) {
+                this.get('routes')[index].config = this.get('routeConfig')[route.name];
+            }
         },
 
         /**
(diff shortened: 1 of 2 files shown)


StephaneDiot:
ping @dpobel

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -138,12 +138,37 @@ YUI.add('ez-platformuiapp', function (Y) {
                     oldService.setNextViewServiceParameters(newService);
                 }
             });
-
+            if (this.get('routeConfig')) {
```
dpobel: I would put that in a dedicated method with a good name so that you directly know what it's about

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -138,12 +138,37 @@ YUI.add('ez-platformuiapp', function (Y) {
...
+        },
+
+        /**
+         * Enrich the route with the route configuration
+         *
+         * @protected
+         * @method _enrichRoute
+         */
+        _enrichRoute: function (route, index) {
+            if (this._findRouteConfig(route.name)) {
```
dpobel: the `_findRouteConfig` is not really necessary, you can directly check here `this.get('routeConfig')[route.name]`

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -708,6 +733,20 @@ YUI.add('ez-platformuiapp', function (Y) {
                 value: false
             },
 
+
+            /**
+             * Routes configuration
```
dpobel: A bit of explanations on the expected format would not hurt

Resources/views/PlatformUI/shell.html.twig
```diff
@@ -67,7 +67,8 @@
                     capi: new Y.eZ.CAPI(
                         "{{ app.request.uriForPath('/')|trim('/') }}",
                         new Y.eZ.SessionAuthAgent()
-                    )
+                    ),
+                    routeConfig: {}
```
dpobel: at this stage, you should not even add that.

Tests/js/apps/assets/ez-platformuiapp-tests.js
```diff
@@ -1561,6 +1561,63 @@ YUI.add('ez-platformuiapp-tests', function (Y) {
...
+            this.app.destroy();
+        },
+
+        "Should enrich route with config": function () {
+            var that = this;
+
+            Y.Array.each(this.app.get('routes'), function (value) {
+                if (value.name == 'loginForm') {
+                    Y.Assert.areSame(value.config, that.app.get('routeConfig').loginForm, 'Route loginForm should be enrich with config');
+                } else {
```
dpobel: this else block should not be in this test.

Tests/js/apps/assets/ez-platformuiapp-tests.js
```diff
@@ -1561,6 +1561,63 @@ YUI.add('ez-platformuiapp-tests', function (Y) {
...
+                    Y.Assert.isUndefined(value.config, 'Route should not have been enrich if routeConfig does NOT match');
+                }
+            });
+        },
+
+        "Should Not enrich route with config if routeConfig does NOT match": function () {
+            var that = this;
+
+            Y.Array.each(this.app.get('routes'), function (value) {
+                if (value.name == 'loginForm') {
```
dpobel: in this test, you should actually that the `value.config` is never the same as `that.app.get('routeConfig').doesNotMatchAnything` (on the loginForm route and the others)

dpobel:
+1

PR #156, opened 2015-01-21
EZP-22247 Give configuration to view service

StephaneDiot:
## Description

The goal here is to give to the view service the route configuration. This is currently needed for the implementation of country field view.

## Link

Jira: https://jira.ez.no/browse/EZP-22247

### DIFF
--- a/Resources/public/js/views/services/ez-viewservice.js
+++ b/Resources/public/js/views/services/ez-viewservice.js
@@ -193,7 +193,18 @@ YUI.add('ez-viewservice', function (Y) {
              */
             app: {
                 writeOnce: "initOnly"
-            }
+            },
+
+            /**
+             * The configuration coming from the matched route in the application
+             *
+             * @attribute config
+             * @type mixed
+             * @initOnly
+             */
+            config: {
+                writeOnce: "initOnly"
+            },
         }
     });
 });
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -482,6 +482,7 @@ YUI.add('ez-platformuiapp', function (Y) {
                     request: req,
                     response: res,
                     plugins: Y.eZ.PluginRegistry.getPlugins(ServiceContructor.NAME),
+                    config: route.config,
                 });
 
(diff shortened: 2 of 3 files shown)


StephaneDiot:
ping @dpobel

Resources/public/js/views/services/ez-viewservice.js
```diff
@@ -193,7 +193,18 @@ YUI.add('ez-viewservice', function (Y) {
...
             app: {
                 writeOnce: "initOnly"
-            }
+            },
+
+            /**
+             * The config is an element configured for the route
+             *
+             * @attribute app
+             * @type mixte
```
dpobel: `mixed` :)

Resources/public/js/views/services/ez-viewservice.js
```diff
@@ -193,7 +193,18 @@ YUI.add('ez-viewservice', function (Y) {
              */
             app: {
                 writeOnce: "initOnly"
-            }
+            },
+
+            /**
+             * The config is an element configured for the route
+             *
+             * @attribute app
```
dpobel: `@attribute config`

Resources/public/js/views/services/ez-viewservice.js
```diff
@@ -193,7 +193,18 @@ YUI.add('ez-viewservice', function (Y) {
              */
             app: {
                 writeOnce: "initOnly"
-            }
+            },
+
+            /**
+             * The config is an element configured for the route
```
dpobel: I would put something like `The configuration coming from the matched route in the application.`

dpobel:
besides the issues on the attribute's documentation, +1

sunpietro:
Why the view service needs routing config? The description is not clear.

dpobel:
Just answered to @sunpietro in our daily but for potential others reviewers, the idea is to be able to configure some kind of "static" data in a route and to inject that while instantiating the application.

The use case here is to inject the country list supported by eZ Platform so that the country view and edit view can have the correct country names. But this can be reused for others purposes, basically as long as you have a setting in eZ Platform and you want to inject it in the PlatformUI application, this system can be used.

dpobel:
+1

PR #157, opened 2015-01-23
EZP-23817: As a developer, I want to have a universal discovery widget to pick a content

dpobel:
JIRA: https://jira.ez.no/browse/EZP-23817

# Description

This PR provides the foundations of the **Universal Discovery Wigdet**. The universal discovery widget is interface the editors will use to pick one or several contents in the repository. For instance, it'll be used when the user wants to add some content(s) in a relation or assign a section to a content.

From a technical point of view, it will be implemented as a *side view* (like the top or left menus) but with a way for any view to trigger it (with an event) and to potentially decide what to do when the user cancels or confirms his selection.

The following screencast shows this code in action with the custom code in https://github.com/ezsystems/PlatformUIBundle/commit/47a85a45851644beaac5cdf6813edda37830cd69. This custom code just adds 2 buttons in the dashboard that allows to trigger the (empty) universal discovery widget with some settings (custom title, selection mode and the event handlers with an example of preventing the user from closing the universal discovery widget by clicking on Cancel).

Screencast: https://www.youtube.com/watch?v=s_H-TCa-Xy8

Of course, what happens inside the universal discovery will be done later :)

Tasks:

* [x] Change the side views handling to allow "on demand" show/hide of a sideview
* [x] Allow to configure the side view while showing it with the app's API
* [x] Start implementing the universal discovery view
* [x] Wire up the application and universal discovery view with an app plugin

# Tests

unit tests + manual tests

### DIFF
--- a/Resources/public/css/theme/views/universaldiscovery.css
+++ b/Resources/public/css/theme/views/universaldiscovery.css
@@ -1,0 +1,26 @@
+/**
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+.is-universaldiscovery-hidden .ez-universaldiscovery-container {
+    display: block;
+
+    background: rgba(0, 0, 0, 0);
+    -webkit-transform: translateY(-100%);
+            transform: translateY(-100%);
+}
+
+.ez-universaldiscovery-container {
+    background: rgba(0, 0, 0, 0.4);
+
+    -webkit-transform: translateY(0);
+            transform: translateY(0);
+    -webkit-transition: all 0.3s;
+            transition: all 0.3s;
+}
+
+.ez-universaldiscovery-container .ez-view-universaldiscoveryview {
+    background: #fff;
+    border: 1px solid #aaa;
+}
--- a/Resources/public/js/views/ez-universaldiscoveryview.js
+++ b/Resources/public/js/views/ez-universaldiscoveryview.js
@@ -1,0 +1,193 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-universaldiscoveryview', function (Y) {
+    "use strict";
+    /**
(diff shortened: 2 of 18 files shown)


dpobel:
ping @StephaneDiot

StephaneDiot:
+1 FWIW

Resources/public/css/theme/views/universaldiscovery.css
```diff
@@ -0,0 +1,26 @@
+/**
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+.is-universaldiscovery-hidden .ez-universaldiscovery-container {
+    display: block;
+
+    background: rgba(0, 0, 0, 0);
+    -webkit-transform: translateY(-100%);
```
sunpietro: Wouldn't it be better to use `transform: translate3d(0, -100%, 0);` to use GPU for animations?

Resources/public/js/views/ez-universaldiscoveryview.js
```diff
@@ -0,0 +1,193 @@
...
+
+            '.ez-universaldiscovery-cancel': {
+                'tap': '_cancel',
+            },
+        },
+
+        initializer: function () {
+            this.on('contentDiscoveredHandlerChange', function (e) {
+                this._syncEventHandler(DISCOVERED, e.prevVal, e.newVal);
+            });
```
sunpietro: nitpick: you could write it like this: `this.on('contentDiscoveredHandlerChange', this._syncEventHandler, this, DISCOVERED)` and make use of event param in the `this._syncEventHandler` method.

dpobel: which event name ?

sunpietro: The `contentDiscoveredHandlerChange`

dpobel: well, the only way to do that, would be to change the attribute name `contentDiscoveredHandler`, any idea ?

PR #160, opened 2015-01-29
EZP-23817: As a developer, I want to have a universal discovery widget to pick a content #2

dpobel:
JIRA: https://jira.ez.no/browse/EZP-23817 (Step 2 after https://github.com/ezsystems/PlatformUIBundle/pull/157)

# Description

This patch completes the implementation of the universal discovery view so that it can use several discovery methods and those methods are presented in a tab based interface. For now, the universal discovery view references only the *browse* method (which is empty for now).

The following short screencast shows how the whole interface will behave with 2 methods. The method visible by default can be decided by the code triggering the universal discovery widget:

https://www.youtube.com/watch?v=wmvTA9MxhBc&

Tasks

* [x] Add the tab based interface to potentially have multiple discovery methods
* [x] Start implementing the *browse* discovery method

# Tests

manual tests + unit tests


Resources/public/js/views/ez-universaldiscoveryview.js
```diff
@@ -44,6 +52,72 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
...
+         *
+         * @method _updateVisibleMethod
+         * @protected
+         * @param {EventFacade} e
+         */
+        _updateVisibleMethod: function (e) {
+            var identifier = e.tabId.replace(/^#/, ''),
+                newlyVisibleMethod;
+
+            console.log('_updateVisibleMethod', identifier);
```
dpobel: self note: remove that!

dpobel:
ping @StephaneDiot

StephaneDiot:
+1 FWIW

andrerom:
review ping @sunpietro

Resources/public/js/views/ez-universaldiscoveryview.js
```diff
@@ -44,6 +52,71 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
...
+
+            this._selectTab(
+                this._getTabLabel(container.one('[href="' + htmlId + '"]')),
+                htmlId,
+                container
+            );
+        },
+
+        /**
+         * Updates the visible flag of the method views depending on the value
```
sunpietro: The description is unclear to me. What the `method` means in the descripton above?

Resources/public/js/views/universaldiscovery/ez-universaldiscoverybrowseview.js
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @namespace eZ
+     * @class UniversalDiscoveryBrowseView
+     * @constructor
+     * @extends eZ.UniversalDiscoveryMethodBaseView
+     */
+    Y.eZ.UniversalDiscoveryBrowseView = Y.Base.create('universalDiscoveryBrowseView', Y.eZ.UniversalDiscoveryMethodBaseView, [], {
+        initializer: function () {
+            this._set('title', 'Browse');
+            this._set('identifier', 'browse');
```
sunpietro: Why don't you set them as default values in the `ATTRS` object?

Tests/js/views/assets/ez-universaldiscoveryview-tests.js
```diff
@@ -252,9 +301,206 @@ YUI.add('ez-universaldiscoveryview-tests', function (Y) {
...
+            );
+        },
+    });
+
+    tabTest = new Y.Test.Case({
+        name: "eZ Universal Discovery View render test",
+
+        setUp: function () {
+            this.method1 = new Y.eZ.UniversalDiscoveryMethodBaseView();
+            this.method1._set("title", "Method 1");
```
sunpietro: The same as there: https://github.com/ezsystems/PlatformUIBundle/pull/160/files#r24815866

sunpietro: Why don't set `title` and `identifier` when initializing `Y.eZ.UniversalDiscoveryMethodBaseView`?
```
new Y.eZ.UniversalDiscoveryMethodBaseView({
  title: 'Method 1',
  identifier: 'method1'
});
```

Resources/public/js/views/ez-universaldiscoveryview.js
dpobel: https://github.com/ezsystems/PlatformUIBundle/pull/160/files#diff-c94326efd4810afcf1f696312d405c97R306
The universal discovery view has several *methods*. A method is view which provides a way to discover content. At first, we'll implement the *browse* method which basically displays the content as a tree, later we'll provide the *recent* method to display the last visited contents and after we can provide others (search, bookmark, ...)

Resources/public/js/views/universaldiscovery/ez-universaldiscoverybrowseview.js
dpobel: to avoid having to redefine the same attribute but that does not make any real difference.

Tests/js/views/assets/ez-universaldiscoveryview-tests.js
dpobel: those attributes are readOnly (on purpose)

Resources/public/js/views/ez-universaldiscoveryview.js
sunpietro: For me, they are more like actions than methods. But it's up to you.

Tests/js/views/assets/ez-universaldiscoveryview-tests.js
sunpietro: I don't like the idea. They are `readOnly` but you change them anyway.

PR #161, opened 2015-01-30
Fix EZP-23922: Improve Navigation hub to prevent accidental hovers

dpobel:
JIRA: https://jira.ez.no/browse/EZP-23922 (Feedback from Gareth Arnott, a community member)

# Description

This patch removes the ability to switch from a zone ("Create", "Deliver", "Optimize") to another by hovering over the zone name. This feature was actually painful because the user might hover over a zone while he tries to reach a navigation item.

# Tests

Unit test + manual test


andrerom:
+1

sunpietro:
+1

PR #162, opened 2015-01-30
EZP-23098: As a developer I want to have the ability to extend the navigation hub

dpobel:
JIRA: https://jira.ez.no/browse/EZP-23098

# Description

This patch makes the navigation hub dynamic and makes it extensible. So now, a navigation item can be highlighted depending on what the PlatformUIApp is displaying and of course, it also works with items provided by external bundles (like *Contents by types* in the *Deliver* zone).

https://www.youtube.com/watch?v=ScmuoIPoUHE

Tasks:

* [x] Make the zone list dynamic
* [x] Make the navigation dynamic
* [x] Inject the navigation item when building the navigation hub
* [x] Handle the *selected* navigation item and zone
* [ ] ~~Transform the eZ Exchange and Administration links to be zones so the administration also gets its navigation items~~ (post poned in [a different issue/PR](https://jira.ez.no/browse/EZP-23986) to limit the size of the PR)

# Tests

unit tests and manual tests


andrerom:
yeay :)

andrerom:
Final decision made ;)

dpobel:
yeah but the PR is already quite big, the additional changes will be done in [EZP-23896](https://jira.ez.no/browse/EZP-23986).

Resources/public/js/views/ez-navigationhubview.js
```diff
@@ -372,14 +420,130 @@ YUI.add('ez-navigationhubview', function (Y) {
...
+            Y.Array.each(value, function (struct) {
+                var ViewConstructor, view;
+
+                if ( struct instanceof Y.eZ.NavigationItemView ) {
+                    view = struct;
+                } else {
+                    ViewConstructor = struct.Constructor;
+                    view = new ViewConstructor(struct.config || {});
+                }
+                view.addTarget(that);
```
dpobel: self note: this needs to be tested

Resources/public/js/views/ez-navigationhubview.js
```diff
@@ -159,12 +186,33 @@ YUI.add('ez-navigationhubview', function (Y) {
...
+
+                Y.Array.each(that.get(key + 'NavigationItems'), function (view) {
+                    after.insert(view.render().get('container'), 'after');
+                    after = view.get('container');
+                });
+            });
+        },
+
         _onActiveUpdate: function (e) {
             var fixed = this.get('navigationFixed');
```
dpobel: self note: need to forward the active flag to the navigation item views

dpobel:
ready to be reviewed
ping @andrerom @yannickroger @sunpietro @StephaneDiot

andrerom:
+1

Resources/public/js/views/ez-navigationhubview.js
```diff
@@ -159,15 +186,37 @@ YUI.add('ez-navigationhubview', function (Y) {
...
+         * @method _renderNavigationItems
+         * @protected
+         */
+        _renderNavigationItems: function () {
+            var that = this;
+
+            Y.Object.each(this.get('zones'), function (zone, key) {
+                var after = that._getNavigationNode(key).one('.ez-logo');
+
+                Y.Array.each(that.get(key + 'NavigationItems'), function (view) {
```
sunpietro: What do you think about using [this.getAttrs()](http://yuilibrary.com/yui/docs/api/classes/View.html#method_getAttrs)? You'll make only one call to get all the attributes and then you could iterate over them.

Resources/public/js/views/ez-navigationhubview.js
```diff
@@ -182,6 +231,21 @@ YUI.add('ez-navigationhubview', function (Y) {
...
         /**
+         * Sets the active flag on the navigation item views
+         *
+         * @protected
+         * @method _setNavigationItemActive
+         * @param {Boolean} active
+         */
+        _setNavigationItemActive: function (active) {
+            Y.Object.each(this.get('zones'), function (zone, key) {
+                Y.Array.each(this.get(key + 'NavigationItems'), function (view) {
```
sunpietro: The same as there: https://github.com/ezsystems/PlatformUIBundle/pull/162/files#r24313897

Resources/public/js/views/navigation/ez-navigationitemsubtreeview.js
```diff
@@ -0,0 +1,48 @@
...
+
+        /**
+         * Checks whether the navigation item is selected or not based on the
+         * location id in the route.
+         *
+         * @method match
+         * @param {Object} route the matched route object
+         * @return {Boolean}
+         */
+        match: function (route) {
```
sunpietro: The function name is too generic. It should describe what is being matched using that function.

Resources/public/js/views/services/ez-navigationhubviewservice.js
```diff
@@ -48,7 +75,144 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
...
+         * initialization of the navigation hub. As a result, in a navigation
+         * hub view service plugin, this method should be called only in or from
+         * the initializer method.
+         *
+         * @method removeNavigationItem
+         * @param {String} identifier the identifier of the navigation item to
+         * remove
+         * @param {String} zone the identifier of the zone
+         */
+        removeNavigationItem: function (identifier, zone) {
```
sunpietro: What is the use case for this method?

Tests/js/views/navigation/assets/ez-navigationitemview-tests.js
```diff
@@ -0,0 +1,145 @@
...
+                route: this.route,
+            });
+        },
+
+        tearDown: function () {
+            this.view.destroy();
+            delete this.view;
+        },
+
+        "Should use a li as a container": function () {
```
sunpietro: It should be `Should use a li - HTML tag, as a container`

Resources/public/js/views/ez-navigationhubview.js
dpobel: that would not change anything in the code structure. getAttrs would return something like:
```js
{platformNavigationItems: [/* array of items */, studioNavigationItems: [/* array of items */, /* ... */}
```

Resources/public/js/views/navigation/ez-navigationitemsubtreeview.js
dpobel: suggestion then ?

Resources/public/js/views/services/ez-navigationhubviewservice.js
dpobel: an extension that wants to remove a default navigation item. You'll use that in Studio.

Resources/public/js/views/navigation/ez-navigationitemsubtreeview.js
sunpietro: maybe `selectedRouteMatch`?

Resources/public/js/views/ez-navigationhubview.js
sunpietro: Ok, now I see. For few elements only, the performance progress will be unnoticeable.

Resources/public/js/views/services/ez-navigationhubviewservice.js
sunpietro: Ok, good to know.

sunpietro:
What about getting an information/attributes from currently active view service by the navigation elements? I mean, some element might change it's behaviour basing on the location id of the content.

Resources/public/js/views/navigation/ez-navigationitemsubtreeview.js
bdunogier: matchesRoute ?

dpobel:
> What about getting an information/attributes from currently active view service by the navigation elements? I mean, some element might change it's behaviour basing on the location id of the content.

the active view and the navigation items are clearly separated to avoid a strong coupling between those components. However, the navigation item views receive [the *matched route*](https://github.com/ezsystems/PlatformUIBundle/pull/162/files#diff-ce457bdf640ad0e239a06ae1c9b8a148R85) through the `match` method and it's up to the navigation item view to decide what to do with that. The default navigation item view only [compares the route name](https://github.com/ezsystems/PlatformUIBundle/pull/162/files#diff-606b4b396febee16c0ad69658f06f653R58) to decide if it is selected or not, the navigation item subtree view also [compares the id of the route](https://github.com/ezsystems/PlatformUIBundle/pull/166/files#diff-f4c2e7ac9bd2834a43b99851b7000caaR48) to detect in which subtree the user is. You can of course implement your own navigation item to implement any logic you want. For instance, in https://github.com/ezsystems/PlatformUIBundle/pull/166, the navigation item parameter view is added to recognize a route based on its name and on the value of a given parameter.

sunpietro:
If I understood correctly, the navigation items will not know what content is currently visible in the `locationView`. Am I right? If so, I will not be able to detect what content should be visible in Studio.
Correct me if I'm wrong.

dpobel:
The navigation items receive the matched route ie the route + the parameters that were matched. For instance, when using the `viewLocation` route, the navigation items receive something like:

```js
{
    // from the route
    "name": "viewLocation",
    "path": "/view/:id",
    "sideViews {
        "discoveryBar": true,
        "navigationHub":true
    },
    "view": "locationViewView",
    // the parameters that are being used
    "parameters": {
        "id": "/api/ezp/v2/content/locations/1/2/94/95"
    }
}
```

so the navigation items is notified that the application will display the location which id is "/api/ezp/v2/content/locations/1/2/94/95".

Resources/public/js/views/navigation/ez-navigationitemsubtreeview.js
dpobel: usually, method names begin with an infinitive verb so I'll change it for `matchRoute`. (`selected` has nothing to do in the name of the method as the `selected` attribute handling is a pure detail of implementation in the default navigation item view and that might be different depending on the item implementation)

StephaneDiot:
+1 FWIW

PR #164, opened 2015-02-03
Fix EZP-23905: JS error displaying a content with an empty Url field

yannickroger:
Link: https://jira.ez.no/browse/EZP-23905

## Test
Manual and update of unit test

### DIFF
--- a/Resources/public/js/views/fields/ez-url-view.js
+++ b/Resources/public/js/views/fields/ez-url-view.js
@@ -32,7 +32,7 @@ YUI.add('ez-url-view', function (Y) {
         _getFieldValue: function () {
             var value = this.get('field').fieldValue, res;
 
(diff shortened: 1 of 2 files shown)


andrerom:
+1

Resources/public/js/views/fields/ez-url-view.js
```diff
@@ -32,7 +32,7 @@ YUI.add('ez-url-view', function (Y) {
         _getFieldValue: function () {
             var value = this.get('field').fieldValue, res;
 
-            if ( !value.link ) {
+            if ( value === null || !value.link ) {
```
dpobel: I would even write:
```js
if ( !value || !value.link ) {
```

dpobel:
besides the inline, +1
good job :)

PR #165, opened 2015-02-03
Fixing error on keyword view

StephaneDiot:
## Description

There was a little error in the keyword field view configuration file (yui.yml)... The fieldeditview module was loaded instead of the fieldview... So well, this should be better now !


StephaneDiot:
ping @dpobel

dpobel:
+1

andrerom:
+1

PR #166, opened 2015-02-05
EZP-23986: As an editor, I want to have a navigation hub to navigate through the PlatformUI

dpobel:
JIRA: https://jira.ez.no/browse/EZP-23986

# Description

Completing the navigation hub implementation (follow up of https://github.com/ezsystems/PlatformUIBundle/pull/162).

Main changes:

* names of the zone
* admin is a zone with a specific style
* the navigation menu disappears if no zone is selected (like on the dashboard)
* the navigation appears if the selected zone has more than one item (Platform zone or Admin zone in the screencast)
* if the zone has exactly one navigation item, choosing the zone is like clicking on the item (Studio zone in the screencast)

Screencast:
https://www.youtube.com/watch?v=z1HAeN25T7M

Tasks:

* [x] Rename the *Create*, *Deliver*, *Optimize* zones to *Platform*, *Studio* and *Studio Plus*
* [x] Remove the hint on each zone
* [x] Hide/Remove *eZ Exchange* for now
* [x] Change *Administration* to be a new zone
* [x] Add the navigation items (dashboard, section, system info) for the administration
* [x] Add the custom styles for the Administration zone
* [x] Hide the navigation no zone is selected (for instance on the dashboard)
* [x] Display the navigation only when needed
* [ ] ~~Improve the transition when the navigation menu appears/disappears~~ (out of inspiration [for now](https://jira.ez.no/browse/EZP-24004))
* [x] When choosing a zone, load the route associated with the first navigation item ?


Resources/public/js/views/navigation/ez-navigationitemparameterview.js
```diff
@@ -0,0 +1,75 @@
...
+    }, {
+        /**
+         * Stores the identifier of the parameter to check in the matched route
+         * and in the `route` attribute
+         *
+         * @attribute matchParameter
+         * @type {String}
+         * @required
+         */
+        matchParameter: {
```
dpobel: self note: missing ATTRS level!

Resources/public/js/views/ez-navigationhubview.js
```diff
@@ -492,75 +571,80 @@ YUI.add('ez-navigationhubview', function (Y) {
...
-                    },
-                    'optimize': {
-                        name: 'Optimize',
-                        hint: 'Statistics & Analytics',
-                        identifier: 'optimize',
-                    },
+                    'platform': 'Platform',
+                    'studio': 'Studio',
+                    'studioplus': 'Studio Plus',
+                    'admin': 'Admin. Panel',
```
andrerom: Think we can remove the `.`, not part of the screenshoot I have seen so far and as PM said that was not needed.

andrerom:
+1

dpobel:
@andrerom any reason to have downgraded your "+1" to a "+0.9" (which does not help)

andrerom:
Just that I'm not so familiar with this part, so if I put a +1 it is not really a indication I have properly reviewed it. But I see your point that it can be misunderstood as meaning I'm saying we could do better or similar.

StephaneDiot:
+1 FWIW

PR #167, opened 2015-02-05
Update dashboard

andrerom:
- [x] Remove test link
- [x] Fix Software name
- [x] Write intro text

### DIFF
--- a/Resources/public/css/theme/views/dashboard.css
+++ b/Resources/public/css/theme/views/dashboard.css
@@ -5,6 +5,6 @@
 
 .ez-view-dashboardview {
     /* for test purpose */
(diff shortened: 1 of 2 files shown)


dpobel:
*Good luck, and thanks for trying the alpha!* :-D

To make it looking a tiny bit nicer, I would change https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/css/theme/views/dashboard.css#L8 for `padding: 1em;`

Other than that, +1 of course

yannickroger:
+1

andrerom:
Done, I can not see it as have issues getting install up and running but I took your word for it :)

PR #171, opened 2015-02-16
EZP-22247: As an editor, I want the country field to be displayed in the raw content view

StephaneDiot:
## Description

The topic here is to be able to display a field view with the country field filled in it.

## Screenshot

image

## Link

Jira: https://jira.ez.no/browse/EZP-22247


StephaneDiot:
ping @dpobel

andrerom:
+1

Resources/public/js/views/fields/ez-country-view.js
```diff
@@ -0,0 +1,86 @@
...
+                isMultiple = this.get('fieldDefinition').fieldSettings.isMultiple,
+                res = [];
+
+            if (field && fieldValue) {
+                Y.Array.each(fieldValue, function (country) {
+                    if (config[country]) {
+                        if(isMultiple) {
+                            res.push(config[country].Name);
+                        } else {
+                            res = config[country].Name;
```
dpobel: I would threat this case outside of loop to decrease a bit the imbrication level. So basically, the code could be:
```js
if (field && fieldValue) {
    if (!isMuliple ) {
        return config[fieldValue[0]].Name;
    }
    Y.Array.each(fieldValue, function (country) {
        // building a res when isMultiple
    })
}
return res;
```

Resources/public/js/views/fields/ez-country-view.js
```diff
@@ -0,0 +1,86 @@
...
+    Y.eZ.CountryView = Y.Base.create('countryView', Y.eZ.FieldView, [], {
+
+        /**
+         * Returns the value to be used in the template. If the value is not
+         * filled, it returns undefined otherwise an object with a `link` and a
+         * `text` entries.
+         *
+         * @method _getFieldValue
+         * @protected
+         * @return Object
```
dpobel: `@return {Array|String}`

Resources/public/js/views/fields/ez-country-view.js
```diff
@@ -0,0 +1,86 @@
...
+     *
+     * @namespace eZ
+     * @class CountryView
+     * @constructor
+     * @extends eZ.FieldView
+     */
+    Y.eZ.CountryView = Y.Base.create('countryView', Y.eZ.FieldView, [], {
+
+        /**
+         * Returns the value to be used in the template. If the value is not
```
dpobel: this needs to be updated

Resources/public/js/views/fields/ez-country-view.js
```diff
@@ -0,0 +1,86 @@
...
+                        console.warn("Unknown country code: " + country);
+                        res.push(country);
+                    }
+                });
+            }
+            return res;
+        },
+
+        _isFieldEmpty: function () {
+            return (this._getFieldValue().length === 0);
```
dpobel: it works because you are lucky :-) since `_getFieldValue` can return a string and here you don't really need to run the logic of this method, I would write something like:
```js
_isFieldEmpty: function () {
    var fieldValue = this.get('field').fieldValue;

    return (!fieldValue || fieldValue.length === 0);
}
```

StephaneDiot:
updates done ! ping @dpobel

Resources/public/js/views/fields/ez-country-view.js
```diff
@@ -0,0 +1,100 @@
...
+
+        /**
+         * Returns the current country name.
+         * If the country is missing, it will return a warning and the country code.
+         *
+         * @method _getCountry
+         * @protected
+         * @return {String}
+         */
+        _getCountry: function (country) {
```
dpobel: I would call it `_getCountryName`

Resources/public/js/views/fields/ez-country-view.js
```diff
@@ -0,0 +1,100 @@
...
+                });
+            }
+            return res;
+        },
+
+        /**
+         * Returns the current country name.
+         * If the country is missing, it will return a warning and the country code.
+         *
+         * @method _getCountry
```
dpobel: missing doc on the parameter

Resources/public/js/views/fields/ez-country-view.js
```diff
@@ -0,0 +1,100 @@
...
+                Y.Array.each(fieldValue, function (country) {
+                    res.push(that._getCountry(country));
+                });
+            }
+            return res;
+        },
+
+        /**
+         * Returns the current country name.
+         * If the country is missing, it will return a warning and the country code.
```
dpobel: it does not *return* a warning but this is an implementation detail, you don't need to write that in the doc of the method.

StephaneDiot:
done ! ping @dpobel

PR #172, opened 2015-02-16
Fix EZP-23917: PHP warning and notice in the server tab in the system info

glye:
Set missing decimal number precision parameter to 1. My VM's memory now shows as 854.5 MB which is a bit TMI, but with zero precision 1.5 GB would show as 2 GB, so I prefer this way.

https://jira.ez.no/browse/EZP-23917

### DIFF
--- a/Resources/views/SystemInfo/info.html.twig
+++ b/Resources/views/SystemInfo/info.html.twig
@@ -66,7 +66,7 @@
                         {% endif %}
                     </dd>
                     <dt>{{ 'memory'|trans }}</dt>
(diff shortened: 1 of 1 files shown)


glye:
About the failing check, Jenkins says "gunnstein lye is missing the Overall/Read permission". How rude!

pspanja:
+1, though we should maybe also consider adding default value for the `$precision` parameter.

andrerom:
+1

@glye Ignore the @ezrobot failure, it is caused by missing re installing node.js / npm / grunt after reinstall of jenkins, and @vidarl / @dpobel  might need to sync on that soon :)

PR #173, opened 2015-02-17
EZP-23817: As a developer, I want to have a universal discovery widget to pick a content #3

dpobel:
JIRA: https://jira.ez.no/browse/EZP-23817

# Description

This patch provides a first implementation of the *browse discovery method* allowing to a pick a content in the content in the repository. As you can see in the screencast, some parts are unfinished (the selected item display, on the right we are supposed to see the sub items, ...) but the core feature is here and the universal discovery can now be integrated to fill a Relation or to assign a section to a subtree for instance.

There's a remaining display issue in Firefox when the tree is to high, I'm still working on that without success so far.

**Screencast:**
https://www.youtube.com/watch?v=anU7VVOiIrM

## Tasks:

* [x] Add the tree to select a content
* [x] Display the selected content in a dedicated view
* [x] Handle the selection in the universal discovery view and provide it the `contentDiscovered` handler(s)
* [x] Style the discovery view
* [ ] Fix the *overflow* bug in Firefox

# Test

unit tests and manual tests (Chrome, Firefox, IE10 and IE11)


dpobel:
ping @StephaneDiot

StephaneDiot:
+1 FWIW

PR #175, opened 2015-02-19
Fix EZP-24035: Fatal error when accessing system info in admin panel tab

lolautruche:
> https://jira.ez.no/browse/EZP-24035

Missing dependency since legacy removal.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,6 +14,7 @@
         }
     },
     "require": {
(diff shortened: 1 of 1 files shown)


glye:
+1

yannickroger:
+1

sunpietro:
+1

PR #178, opened 2015-02-20
Implement EZP-23993: Clarify and document how security issues should be reported

glye:
https://jira.ez.no/browse/EZP-23993

Ref ezsystems/ezpublish-kernel#1201


dpobel:
kind of duplication of the second paragraph in the README, no ? I would prefer you complete it instead.

glye:
@dpobel D'oh. Fixed.

dpobel:
+1

PR #181, opened 2015-03-02
PlatformUI BDD core implementations

miguelcleverti:
# PlatformUI BDD base
Base implementations for platformUI BDD
Needed to run the PlatformUI bundle scenarios

### DIFF
--- a/Features/Context/PlatformUI.php
+++ b/Features/Context/PlatformUI.php
@@ -1,0 +1,327 @@
+<?php
+
+/**
+ * File containing the main context class for PlatformUI.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+
+namespace EzSystems\PlatformUIBundle\Features\Context;
+
+use EzSystems\BehatBundle\Context\Browser\Context;
+use Behat\Gherkin\Node\TableNode;
+use PHPUnit_Framework_Assert as Assertion;
+
+class PlatformUI extends Context
+{
+    use SubContext\CommonActions;
+
+    /**
+     * PlatformUI relative URL path
+     *
+     * @var string
+     */
+    private $platformUiUri;
+    protected $shouldBeLoggedIn;
+    protected $driver;
+
+    /**
+     * @Given I go to homepage
+     */
+    public function goToPlatformUi()
+    {
+        $this->visit( $this->platformUiUri );
+    }
+
+    /**
+     * @Given I go to PlatformUI app with username :user and password :password
+     */
+    public function goToPlatformUiAndLogIn( $username, $password )
+    {
+        // Given I go to PlatformUI app
+        $this->goToPlatformUi();
+        //wait fos JS
+        $this->waitForJs();
+        // And I fill in "Username" with "admin"
+        $this->fillFieldWithValue( 'Username', $username );
+        //And I fill in "Password" with "publish"
+        $this->fillFieldWithValue( 'Password', $password );
+        //And I click on the "Login" button
+        $this->iClickAtButton( 'Login' );
+        //wait fos JS
+        $this->waitForJs();
+        //Then I should be logged in
+        $this->iShouldBeLoggedIn();
+        //Catches Js errors
+        $this->activateJsErrorHandler();
+    }
+
+    /**
+     * @Given I click (on) the logo
+     */
+    public function iClickLogo()
+    {
+        $this->clickLogo();
+    }
+
+    /**
+     * @Given I click (on) the tab :tab
+     */
+    public function iClickTab( $tab )
+    {
+        $this->clickTab( $tab );
+    }
+
+     /**
+     * @Given I click (on) the menu zone :zone
+     */
+    public function iClickMenuZone( $zone )
+    {
+        $this->clickMenuZone( $zone );
+    }
+
+    /**
+     * @Given I over (on) the menu zone :zone
+     */
+    public function iOverMenuZone( $zone )
+    {
+        $this->overMenuZone( $zone );
+    }
+
+    /**
+     * @Given I click (on) the menu link :link
+     */
+    public function iClickMenuLink( $link )
+    {
+        $this->clickMenuLink( $link );
+    }
+
+    /**
+     * @Given I click (on) the sub-menu option :subMenu
+     */
+    public function iClickSubMenuOption( $subMenu )
+    {
+        $this->clickSubMenuLink( $subMenu );
+    }
+
+     /**
+     * @Given I click (on) the side menu option :sideMenuOption
+     */
+    public function iclickSideMenuOption( $sideMenuOption )
+    {
+        $this->clickSideMenuLink( $sideMenuOption );
+    }
+
+    /**
+     * @Given I click (on) the content type :contentType
+     */
+    public function iclickContentTypeOption( $contentType )
+    {
+        $this->clickContentType( $contentType );
+    }
+
+    /**
+     * @Given I click (on) the content tree with path :path
+     */
+    public function iClickContentTreePath( $path )
+    {
+        $this->clickContentTreeLink( $path );
+    }
+
+    /**
+     * @Given I upload the file :path
+     */
+    public function iUploadTheFile( $path )
+    {
+        $this->attachFilePrepare( $path, '#ez-field--file' );
+    }
+
+    /**
+     * @Given I fill in :field subform with:
+     */
+    public function iFillSubForm( $field, TableNode $values )
+    {
+        $arrayValues = array();
+        foreach ( $values as $value )
+        {
+            foreach ( array_keys( $value ) as $key )
+            {
+                $arrayValues[$key][] = $value[$key];
+            }
+        }
+        $arrayValues['size'] = count( array_keys( $arrayValues[$key] ) );
+        $this->fillSubFormList( $field, $arrayValues );
+    }
+
+    /**
+     * @Given I upload the image :path
+     */
+    public function iUploadTheImage( $path )
+    {
+        $this->attachFilePrepare( $path, '#ez-field--image' );
+    }
+
+    /**
+     * @Given I drag and drop a file :file to upload
+     */
+    public function iDragAndDropFile( $path )
+    {
+        $this->dragAndDropFile( $path, 'ez-button-upload' );
+    }
+
+    /**
+     * @Given I logout
+     */
+    public function iLogout()
+    {
+        $this->shouldBeLoggedIn = false;
+        $this->iClickAtLink( "Logout" );
+    }
+
+    /**
+     * @Then I see Content :contentName of type :contentType
+     */
+    public function contentExists( $contentName, $contentType )
+    {
+        $contentId = $this->getThisContentId();
+        $content = $this->getContentManager()->loadContentWithLocationId( $contentId );
+        $contentInfo = $content->contentInfo;
+        $contentTypeName = $this->getContentManager()->getContentType( $content );
+        Assertion::assertEquals( $contentName, $contentInfo->name, "Content has wrong name" );
+        Assertion::assertEquals( $contentType, $contentTypeName, "Content has wrong type" );
+    }
+
+    /**
+     * @Then I should see the following elements:
+     */
+    public function iSeeElementsVisible( TableNode $elements )
+    {
+        foreach ( $elements as $element )
+        {
+            $name = array_values( $element )[0];
+            $isVisible = $this->isElementVisible( $name, '.ez-zone-name, .ez-navigation-item' );
+            Assertion::assertTrue( $isVisible, "Element $name not found" );
+        }
+    }
+
+    /**
+     * @Then I shouldn't see the following elements:
+     * @Then I should not see the following elements:
+     */
+    public function iSeeElementsNotVisible( TableNode $elements )
+    {
+        foreach ( $elements as $element )
+        {
+            $name = array_values( $element )[0];
+            $isVisible = $this->isElementVisible( $name, '.ez-zone-name, .ez-navigation-item' );
+            Assertion::assertFalse( $isVisible, "Element $name found" );
+        }
+    }
+
+    /**
+     * @Then I should see (an) element :element with value :value
+     */
+    public function iSeeElementsWithValue( $element, $value )
+    {
+        $valueActual = $this->checksElementSiblingValue( $element, '.ez-fieldview-label' );
+        //check later
+        Assertion::assertEquals( preg_replace( '/\s+/', '', $value ), preg_replace( '/\s+/', '', $valueActual ) );
+    }
+
+    /**
+     * @Then I should see (an) element :element with (an) image
+     */
+    public function iSeeElementImage( $element )
+    {
+        //Temporary solution ONLY
+        sleep( 2 );
+        $imgUrl = $this->checksImagePresent( $element, '.ez-fieldview-label' );
+        $isImage = exif_imagetype( $imgUrl );
+        Assertion::assertNotFalse( $isImage, "Not an image" );
+    }
+
+    /**
+     * @Then I should see (an) element :element with (an) file :file
+     */
+    public function iSeeElementFile( $element, $file )
+    {
+        $url = $this->getFileUrl( $element, '.ez-fieldview-label' );
+        $fileContentActual = file_get_contents( $url );
+        $file = rtrim( realpath( $this->getMinkParameter( 'files_path' ) ), DIRECTORY_SEPARATOR ).DIRECTORY_SEPARATOR.$file;
+        $fileContentExpected = file_get_contents( $file );
+        Assertion::assertEquals( $fileContentActual, $fileContentExpected );
+    }
+
+    /**
+     * @Then I should see elements with the following names:
+     */
+    public function iSeeElements( TableNode $elements )
+    {
+        foreach ( $elements as $element )
+        {
+            $found = false;
+            $name = array_values( $element )[0];
+            $found = $this->checksElementsByText( $name, '.ez-selection-filter-item' );
+            Assertion::assertNotNull( $found, "Element: $name not found" );
+        }
+    }
+
+    /**
+     * @Then I should be logged in
+     */
+    public function iShouldBeLoggedIn()
+    {
+        $this->shouldBeLoggedIn = true;
+
+        $jsCode = <<<JSCODE
+        YUI().use('node', function(Y){
+            window.Y = Y;
+        });
+
+        var node = Y.one('.ez-loginform');
+        return (node == null);
+JSCODE;
+
+        $isLoggedIn = $this->evalJavascript( $jsCode, false );
+        Assertion::assertTrue( $isLoggedIn, "Not logged in" );
+    }
+
+    /**
+     * Initialize class
+     *
+     * @param string $uri
+     */
+    public function __construct( $uri, $driver )
+    {
+        parent::__construct();
+        $this->platformUiUri = $uri;
+        $driver = "EzSystems\PlatformUIBundle\Features\DriverJS\\" . $driver;
+        $this->driver = new $driver;
+    }
+
+    /**
+     * Runs a empty Javascript between step so that the next step is only executed when the previous Javascript finnished
+     *
+     * @AfterStep
+     */
+    public function waitForJs()
+    {
+        $this->waitForLastJs();
+    }
+
+    /**
+     * Checks if the user is still logged in
+     *
+     * @AfterStep
+     */
+    public function runAfterStep()
+    {
+        if ( $this->shouldBeLoggedIn )
+        {
+           $this->iShouldBeLoggedIn();
+        }
+    }
+
+}
--- a/Features/Context/PlatformUI.php
+++ b/Features/Context/PlatformUI.php
@@ -76,7 +76,7 @@ class PlatformUI extends Context
         $this->clickTab( $tab );
     }
 
-     /**
+    /**
      * @Given I click (on) the navigation zone :zone
      */
     public function iClickNavigationZone( $zone )
@@ -100,12 +100,12 @@ class PlatformUI extends Context
         $this->clickNavigationItem( $subMenu );
     }
 
-     /**
-     * @Given I click (on) the actionbar action :sideMenuOption
+    /**
+     * @Given I click (on) the actionbar action :action
      */
-    public function iclickSideMenuOption( $sideMenuOption )
+    public function iclickActionBar( $action )
     {
-        $this->clickActionBarAction( $sideMenuOption );
+        $this->clickActionBarAction( $action );
     }
 
     /**
@@ -300,8 +300,18 @@ class PlatformUI extends Context
     {
         if ( $this->shouldBeLoggedIn )
         {
-           $this->iShouldBeLoggedIn();
+            $this->iShouldBeLoggedIn();
         }
+    }
+
+    /**
+     * Checks if the user is still logged in
+     *
+     * @AfterScenario
+     */
+    public function loggOutAfterScenario()
+    {
+        $this->iLogout();
     }
 
 }
--- a/Features/Context/PlatformUI.php
+++ b/Features/Context/PlatformUI.php
@@ -1,7 +1,9 @@
 <?php
 
 /**
- * File containing the main context class for PlatformUI.
+ * File containing the main context class for PlatformUI
+ * This flie contains the mapping of the BDD sentences to the functions that implement them
+ * Also contains the necessary initializations
  *
  * @copyright Copyright (C) eZ Systems AS. All rights reserved.
  * @license For full copyright and license information view LICENSE file distributed with this source code.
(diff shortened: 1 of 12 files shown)


Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    /**
+     * Clicks on the PlatformUI logo
+     *
+     * @param string $text
+     */
+    protected function clickLogo()
+    {
+        $selector = 'img[alt="Logo eZ"]';
```
dpobel: to make it a bit more robust, I would use `.ez-logo img` as a selector instead (the `alt` value is more a data than a markup thing)

miguelcleverti: yes, you are right I didn't spend much time on the logo sentence because it was not a priority at the time but I will fix it.

Features/Context/PlatformUI.php
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    /**
+     * @Then I should be logged in
+     */
+    public function iShouldBeLoggedIn()
+    {
+        $this->shouldBeLoggedIn = true;
+
+        $jsCode = <<<JSCODE
```
dpobel: eh what is that ? if it works, you are lucky, since the use callback function can be called asynchronously.
if you just want to check if an element is in the page, you don't need YUI at all, something like:

```js
return (document.querySelector('.ez-loginform') === null);
```

should do the trick. But that does not really check that you are logged but more that the login form disappeared.

Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+    {
+        $this->clickElementsByText( $text, ".ez-tabs-label a[href]" );
+    }
+
+    /**
+     * Click on a PlatformUI menu zone
+     *
+     * @param string $text
+     */
+    protected function clickMenuZone( $text )
```
dpobel: to fit the domain, this method should be called `clickNavigationZone`

Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    /**
+     * Click on a PlatformUI menu link
+     *
+     * @param string $text
+     */
+    protected function clickMenuLink( $text )
+    {
+        $this->clickElementsByText( $text, '.ez-link a[href]' );
```
dpobel: the `ez-link` links is gone for quite some time now, so I guess this method can be removed

Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    /**
+     * click on a PlatformUI sub-menu option
+     *
+     * @param string $text
+     */
+    protected function clickSubMenuLink( $text )
```
dpobel: to fit with the domain, this should be called `clickNavigationItem` (as the class suggests)

Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+    {
+        $this->clickElementsByText( $text, '.ez-navigation-item' );
+    }
+
+    /**
+     * click on a PlatformUI side menu option
+     *
+     * @param string $text
+     */
+    protected function clickSideMenuLink( $text )
```
dpobel: same here, `clickActionBarAction` or something like that, *side menu* does not refer to anything in PlatformUI.

Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+    {
+        $this->clickElementsByText( $text, '.ez-action p.action-label' );
```
dpobel: not sure you really need to be that specific, `.ez-action` should be enough

Features/Context/PlatformUI.php
miguelcleverti: Yes it is more simpler this way. This check was implemented because there were problems with platformUI losing the log in, this was a quick way for we to check that everything was going okay. I will take a note to improve this.

Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    /**
+     * click on a PlatformUI side menu content type
+     *
+     * @param string $text
+     */
+    protected function clickContentType( $text )
+    {
+        $this->clickElementsByText( $text, '.ez-selection-filter-item' );
```
dpobel: this is not specific enough, the selection filter can be used in others contexts `.ez-contenttypeselector-types .ez-selection-filter-item` would be ok I guess

Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+        $jsCode = "return BDD.fillListPair( $jsArgs )";
+        $this->execJavascript( $jsCode );
+    }
+
+    /**
+     * Finds an HTML element by class and the text value and clicks it
+     *
+     * @param string $name
+     */
+    protected function clickElementsByText( $text, $selector )
```
dpobel: since you are clicking on one element, it should be called `clickElementByText`

Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+        $jsCode = "return BDD.clickElementByText( $jsArgs );";
+        $this->execJavascript( $jsCode );
+    }
+
+    /**
+     * Finds an HTML element by class and the text value and returns it
+     *
+     * @param string $name
+     */
+    protected function checksElementsByText( $text, $selector )
```
dpobel: `checksElementByText`

Features/Context/SubContext/CommonActions.php
miguelcleverti: This is actually a remnant of an earlier implementation that needed to be more specific, after the improvement this missed the clean up

Features/Context/SubContext/CommonActions.php
miguelcleverti: Yes, I inteded to remove this but forgot

Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+    /**
+     * Makes and element visible so a file can be attached to it
+     *
+     * @param string $path      file path relative to mink definitions
+     * @param string $selector  selector of the element do to make visible
+     */
+    protected function attachFilePrepare( $path, $selector )
+    {
+        $jsArgs = JsHelper::generateFuncArgs( $selector );
+        $jsCode = "return BDD.changeCssDisplay( $jsArgs );";
```
dpobel: hum that's a bit weird to have to unhide the file input to fill it, is that really needed ?

Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+     */
+    protected function attachFile( $path )
+    {
+        if ( $this->getMinkParameter( 'files_path' ) )
+        {
+            $fullPath = rtrim( realpath( $this->getMinkParameter( 'files_path' ) ), DIRECTORY_SEPARATOR ).DIRECTORY_SEPARATOR.$path;
+
+            if ( is_file( $fullPath ) )
+            {
+                $fileInput = 'input[type="file"]';
```
dpobel: if there's many file inputs, you might fill the wrong one.

Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+                $field = $this->getSession()->getPage()->find( 'css', $fileInput );
+
+                if ( null === $field )
+                {
+                    throw new Exception( "File input is not found" );
+                }
+            $field->attachFile( $fullPath );
```
dpobel: nitpick: wrong indentation level

Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+                $jsArgs = JsHelper::generateFuncArgs( $fileName, $fileType, $contentBase64, $selector );
+                $jsCode = "return BDD.simulateDropEventWithFile( $jsArgs );";
+                $jsFile = $this->execJavascript( $jsCode );
+            }
+        }
+        else throw new Exception( "File is not found at the given location" );
+    }
+
+    /**
+     * Finds an HTML element by class and the text value and return the value of the sibling
```
dpobel: wrong php doc

Features/Context/SubContext/CommonActions.php
miguelcleverti: Yes, I struggled with this for some time and the only way I found was to unhide the file input, the problem is that the webdriver can't attach a file to an hidden input.

Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+     *
+     * @param string $name
+     */
+    protected function checksImagePresent( $text, $selector )
```
dpobel: what is the `$text` variable for ?

Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+    {
+        $jsArgs = JsHelper::generateFuncArgs( $text, $selector );
+        $jsCode = "return BDD.getImgSrc( $jsArgs );";
+        return $this->evalJavascript( $jsCode );
+    }
+
+    /**
+     * Finds an HTML element by class and the text value and return the value of the sibling
```
dpobel: wrong php doc

Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+     */
+    protected function getFileUrl( $text, $selector )
+    {
+        $jsArgs = JsHelper::generateFuncArgs( $text, $selector );
+        $jsCode = "return BDD.getLinkUrl( $jsArgs );";
+        return $this->evalJavascript( $jsCode );
+    }
+
+    /**
+     * Get the current window Node Id
```
dpobel: *current window Node Id* ? this does not mean anything, I guess this is returning the **location** id that is currently displayed but this only works when displaying a location not in the whole PlatformUI so this should probably not be in a `CommonActions` trait.

Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+     *
+     * @return string
+     */
+    protected function getThisContentId()
```
dpobel: `getLocationId`

Features/Context/SubContext/CommonActions.php
miguelcleverti: Yes, this is a workaround because so far there is no information on platformUI about the content no ids, no type, etc. This is a supposed temporary solution for some test cases.

Features/Context/PlatformUI.php
```diff
@@ -0,0 +1,327 @@
...
+    {
+        $this->clickContentTreeLink( $path );
+    }
+
+    /**
+     * @Given I upload the file :path
+     */
+    public function iUploadTheFile( $path )
+    {
+        $this->attachFilePrepare( $path, '#ez-field--file' );
```
dpobel: this id is weird because it might not be unique across the page, so I guess there's a problem in the generated id in the page which should be something like `#ez-field-42-file` where 42 is the content id.

Features/Context/PlatformUI.php
```diff
@@ -0,0 +1,327 @@
...
+        $arrayValues['size'] = count( array_keys( $arrayValues[$key] ) );
+        $this->fillSubFormList( $field, $arrayValues );
+    }
+
+    /**
+     * @Given I upload the image :path
+     */
+    public function iUploadTheImage( $path )
+    {
+        $this->attachFilePrepare( $path, '#ez-field--image' );
```
dpobel: same here

Features/Context/PlatformUI.php
```diff
@@ -0,0 +1,327 @@
...
+        // Given I go to PlatformUI app
+        $this->goToPlatformUi();
+        //wait fos JS
+        $this->waitForJs();
+        // And I fill in "Username" with "admin"
+        $this->fillFieldWithValue( 'Username', $username );
+        //And I fill in "Password" with "publish"
+        $this->fillFieldWithValue( 'Password', $password );
+        //And I click on the "Login" button
+        $this->iClickAtButton( 'Login' );
```
dpobel: `iClickOnButton( 'Login' );`

Features/Context/PlatformUI.php
```diff
@@ -0,0 +1,327 @@
...
+        $this->iClickAtLink( "Logout" );
+    }
+
+    /**
+     * @Then I see Content :contentName of type :contentType
+     */
+    public function contentExists( $contentName, $contentType )
+    {
+        $contentId = $this->getThisContentId();
+        $content = $this->getContentManager()->loadContentWithLocationId( $contentId );
```
dpobel: `loadContentWithLocationId( $contentId )` this is really confusing (same issue as in https://github.com/ezsystems/PlatformUIBundle/pull/181/files#r25683698)

Features/Context/SubContext/CommonActions.php
miguelcleverti: The text is needed to find the image, normally associated with a text label, example:
image

$text in red

Features/Context/PlatformUI.php
miguelcleverti: I missed this, it is suposed to use the class tag, but while I was testing I used the id and forgot to change.

Features/Context/PlatformUI.php
miguelcleverti: The iClickAtButton funtion belong to BehatBundle browser context, but I can change it.

Features/Lib/PlatformUI.js
```diff
@@ -0,0 +1,422 @@
...
+     * Finds an element by text and clicks on it
+     *
+     * @param {string} text
+     * @param {string} selector
+     */
+    this.clickElementByText = function(text, selector)
+    {
+        this.clickElement(
+                this.getElementByText(text, selector)
+                );
```
joaoinacio: CS

Features/Context/PlatformUI.php
miguelcleverti: Like I said before, so far the only way I found to test some cases was uses the window hash to retrieve the location id, the variable name is wrong it should be $locationId, in the future this should probably be replace by a better method

andrerom:
@miguelcleverti @pcardiga @dpobel Seems we might need to do a meeting to sync on approach, cause ideally the tests should not have to workaround test issues caued by how we have implemented the UI.

This is not a legacy UI, we can (and should) adjust how we generate the UI to simplify testing ;)

Features/Lib/PlatformUI.js
dpobel: on the CS thing, the whole file is wrong, the CS for JavaScript are different. For instance, this method should be written like:
```js
this.clickElementByText = function(text, selector) {
    this.clickElement(this.getElementByText(text, selector));
};
```
I guess there's enough JS example in this bundle to get it :-)

But, I'm a bit surprise to see such file here, shouldn't the BDD framework provide such helpers ?

pcardiga: @andrerom: I agree. There might exist some information that might be included in the UI that would easily the tests and we could sync on that.

miguelcleverti: BDD framework provides helpers for normal webpages, with AJAX and a full JS webpage things get a little more complicated. The JS was the only way we found to access certain elements and interact with them.

miguelcleverti:
@andrerom I agree, if the UI is generated with BDD in mind it will improve the overall quality of the tests and the UI itself

andrerom:
@miguelcleverti could you list up here the misc issues you encountered so far? (like not being able to use file upload because of hidden upload field* and so on).

_* A question that will pop up on this is if the tests really test the UI if it does not test the *faked* upload UI the user would work with, so might be we should make sure *that* is more testable somehow_.

miguelcleverti:
@andrerom 
- mentioned uplaod UI;
- little information about content (type, ids, etc);
- hidden elements cause wrong results with BDD base framework(behat), bypassed with JS;
- the Image API does not seem to have a loading tag;
- fields validation trigger could be improved for easier testing;
- application is loading tag are good overall but a global one("something is loading") would be even better;
- HTML generated structure could be greatly improved, following a simple consistent logic making the association between elements more explicit(most important IMO);
-Sahi can't publish contents, problems with cookies.

These are some of the main problems I have found. If we can improve this I believe it would benefit the UI. Also none of these are functionality related so it should be no problem.

Features/Context/SubContext/CommonActions.php
bdunogier: Then it should be called "label", or "imageLabel", should it not ?

Features/Context/PlatformUI.php
bdunogier: All @dpobel is referring to here is the naming, imho. And yes, it should be changed :-)

Features/Context/PlatformUI.php
```diff
@@ -0,0 +1,327 @@
+<?php
+
+/**
+ * File containing the main context class for PlatformUI.
```
bdunogier: "This file is part of the PlatformUI package."

Features/Context/PlatformUI.php
```diff
@@ -0,0 +1,327 @@
...
+ * @version //autogentag//
+ */
+
+namespace EzSystems\PlatformUIBundle\Features\Context;
+
+use EzSystems\BehatBundle\Context\Browser\Context;
+use Behat\Gherkin\Node\TableNode;
+use PHPUnit_Framework_Assert as Assertion;
+
+class PlatformUI extends Context
```
bdunogier: Please add a summary of what the class does, a couple lines (it looks like an important one).

Features/Context/SubContext/CommonActions.php
```diff
@@ -0,0 +1,363 @@
...
+                $fileType = finfo_file( $finfo, $fullPath );
+                $contents = file_get_contents( $fullPath );
+                $contentBase64 = base64_encode( $contents );
+
+                $jsArgs = JsHelper::generateFuncArgs( $fileName, $fileType, $contentBase64, $selector );
+                $jsCode = "return BDD.simulateDropEventWithFile( $jsArgs );";
+                $jsFile = $this->execJavascript( $jsCode );
+            }
+        }
+        else throw new Exception( "File is not found at the given location" );
```
bdunogier: What file ? What location ? Try to think of what you'd like to be given as info when you encounter this exception in a month.

While I'm at it, CS:

```php
}
else
{
    throw...
}
```

Features/Context/SubContext/CommonActions.php
miguelcleverti: I will change it, I guess $label is more explicit.

bdunogier: Well, you did find the right answer by yourself. If you can use a word to describe it in the sentence, then it's usually the right word. And if the word you have doesn't make sense when used directly in a sentence, then it is not right.

miguelcleverti: makes sense :-)

miguelcleverti:
I fixed most of the issues commented here and improved the documentations in geral, also I added a doc folder with some relevant documents on installing/using the PlatformUI BDD

Features/doc/README.md
```diff
@@ -0,0 +1,28 @@
+## Pre-requisites:
+* [Behat bundle](https://github.com/ezsystems/BehatBundle) is installed and configured properly;
+* [PlatformUI bundle](https://github.com/ezsystems/PlatformUIBundle) is installed and configured properly.
+
+## Installation:
+* Update behat.yml, on the installation root, with PlatformUI parameters( check example behat.yml provided );
```
dpobel: instead of documenting that, we should directly do it in the ezplatform's behat.yml.

dpobel: or wouldn't it be possible to have a behat.yml in PlatformUIBundle that is just imported in the ezplatform's behat.yml ?

miguelcleverti:
Yes I also think it's better to add a platformUI only behat.yml but I think that some documentation should still be present since there are some parameters dependent of the enviroment. So it should be something like "Update the behat.yml on platformUI directory" and the example behat.yml could be deleted.

dpobel:
Documentation rarely hurts :-) That said, my point is more that the test should be directly ready to be run without any changes (except maybe the base_url). This would also ease the integration with Travis or any other CI platform.

Instead of adding lots of comments (still I will add some), I tried to experiment different things to remove most of PlatformUI.js which is more a hack than anything and sometimes does wrong/weird things that tends to fail from time to time. The result can be seen in https://github.com/ezsystems/PlatformUIBundle/commit/524c8b1f95de3ce8f78129a91104b9763d3193a2. This is an incomplete but I hope it gives you path to go:

1. `waitForJs` should not be used as is. As we discussed, it waits for way to many things. So I added a `waitForInitialApplicationLoading` and `waitForApplicationPageLoading` to wait for the app initial loading and to wait for the *page* loading. As, you can see, this does not rely at all on any custom JavaScript code but only on what [Mink provides to parse the HTML code of the page](http://mink.behat.org/en/latest/guides/traversing-pages.html).
2. Likewise, I modified `iShouldBeLoggedIn`, `clickLogo`, `clickTab`, `clickNavigationZone` to remove the need to inject the custom JavaScript code and to use Mink's API. For the last two method, we need to use `xpath` since we are looking for a label in the page.
3. I removed `activateJsErrorHandler`. Actually, I wonder if we should really test the JavaScript errors here. Most likely, if a JavaScript happens, something else should fail. But if it happens that we want to track the JavaScript error, then this should be done in a more global way and we should provide more details (here basically only the error message is provided).

(Some scenarios are commented just because I did not have time to look at them but some are outdated like the first commented one)

Features/Context/PlatformUI.php
```diff
@@ -0,0 +1,423 @@
...
+     */
+    protected $driver;
+
+    /**
+     * Evaluate javascript code and return result.
+     *
+     * @param   string  $jsCode
+     * @return  mixed
+     */
+    protected function evalJavascript( $jsCode, $addJsHelper = true )
```
dpobel: most likely we should be able to get rid of this method and `execJavascript` but one thing though, with `$addJsHelper` set to true (default value), the whole `PlatformUI.js` is read and injected in the page again and again.

Features/Context/PlatformUI.php
```diff
@@ -0,0 +1,423 @@
...
+        return $this->getSession()->executeScript( $jsCode );
+    }
+
+    /**
+     * Waits for Javascript to finnish by running a empty Javascript
+     * (In Sahi it's possible to have the same result by running an empty javascript only)
+     */
+    protected function waitForLastJs()
+    {
+        //Needs to be here
```
dpobel: this is very confusing and instead we should be able to read the *why* it needs to be here.

Features/Context/PlatformUI.php
```diff
@@ -0,0 +1,423 @@
...
+        $jsCode = "return BDD.isSomethingLoading();";
+
+        while ( $this->evalJavascript( $jsCode ) )
+        {
+            usleep( 100 * 1000 ); // 100ms
+        }
+    }
+
+    /**
+     * Waits for Javascript to finnish by running a empty Javascript
```
dpobel: This is an obvious copy paste error but please pay attention to that.

Features/Context/PlatformUI.php
```diff
@@ -0,0 +1,423 @@
...
+        $contentId = $this->getLocationId();
+        $content = $this->getContentManager()->loadContentWithLocationId( $contentId );
+        $contentInfo = $content->contentInfo;
+        $contentTypeName = $this->getContentManager()->getContentType( $content );
+        Assertion::assertEquals( $contentName, $contentInfo->name, "Content has wrong name" );
+        Assertion::assertEquals( $contentType, $contentTypeName, "Content has wrong type" );
+    }
+
+    /**
+     * @Given I fill in :field subform with:
```
dpobel: what is a *subform* ? Most likely, the original sentence should be written differently

Features/DriverJS/Sahi.php
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * Adapts javascript to run on Sahi webDriver
+     *
+     * @param string    $code javacript code to adapt
+     *
+     * @return string   adapted javascript code
+     */
+    public function wrappedFunction( $code )
+    {
+        $jsCode = "(function(){
```
dpobel: actually, if we ever need to inject some custom JavaScript code, we should always put it inside such a function. So to me, the whole `DriverJSInterface` thing seems useless.

Features/Lib/PlatformUI.js
```diff
@@ -0,0 +1,407 @@
...
+ */
+
+/**
+ * Container for Javascript functions
+ *
+ * @returns {PlatformUIHelper}
+ */
+function PlatformUIHelper(){
+    //YUI global object with node-event-simulate module
+    var Y = YUI().use('node-event-simulate', 'stylesheet');
```
dpobel: This is wrong, sometimes fails and makes the test unreliable because the YUI module loading is an asynchronous process so it happens that those 2 modules are not loaded while we try to use what they provide.

But in any case, I find weird to rely on YUI in the test and hopefully, this complete file can be removed.

Features/Lib/PlatformUI.js
```diff
@@ -0,0 +1,407 @@
...
+        window.document.getElementsByClassName(selector)[0].dispatchEvent(event);
+    }
+
+    /**
+     * Makes an element visible by changing its CSS style
+     *
+     * @param {String} selector     element selector
+     */
+    this.changeCssDisplay = function( selector ){
+        Y.StyleSheet('NewStyle').set( selector, { display: 'inline' });
```
dpobel: For the record, you don't to use a dynamically generated stylesheet for that.
`Y.one(selector).setStyle('display', 'inline');` would do exactly the same (in more reliable manner)

Features/Standard/platformui.feature
```diff
@@ -0,0 +1,132 @@
...
+    @javascript
+    Scenario: As a admin User, I want to login to PlatformUI
+        Given I go to PlatformUI app
+        And I fill in "Username" with "admin"
+        And I fill in "Password" with "publish"
+        When I click on the "Login" button
+        Then I should see "Welcome to eZ Platform" title
+
+    @javascript
+    Scenario: As a admin User, I want to login to PlatformUI
```
dpobel: isn't it exactly the same as the previous scenario ?

Features/Standard/platformui.feature
```diff
@@ -0,0 +1,132 @@
...
+
+    @javascript
+    Scenario: As a admin User, I want to see Activity tab
+        Given I go to PlatformUI app with username "admin" and password "publish"
+        And I click on "Root location" link
+        When I click on the tab "Activity"
+        Then I should see "Activity content" text
+
+    @javascript
+    Scenario: As a admin User, I want to use side menu
```
dpobel: the name of scenario is bad it seems, it has nothing to do with a *side menu* (side menu has no meaning, I guess you are referring the *action bar* or *discovery bar* but none of those relevant anyway)

Features/Standard/platformui.feature
```diff
@@ -0,0 +1,132 @@
...
+        Given I go to PlatformUI app with username "admin" and password "publish"
+        And I click on "Root location" link
+        When I click on the logo
+        Then I should see "Welcome to eZ Platform" title
+
+    @javascript
+    Scenario: As a admin User, I want to see navigate menu
```
dpobel: *I want to use the navigation zone menu*

dpobel:
One last thing, the way the PR are split makes the review very complicated. A quite big part of the patch is kind of dead code if I don't go to others PR and to be honest this will be very complicated to merge it this way.

PR #201, opened 2015-03-02
EZP-23905 JsErrorEmptyUrlField BDD scenarios

miguelcleverti:
# EZP-23905 BDD scenario
BDD scenarios for issue #164 
jira: https://jira.ez.no/browse/EZP-23905

requires: #181


andrerom:
-1 No regression tests in BDD please.

We should not add a new scenario for every bug we get in, if we do that we will get so many BDD scenarios with time it will take a day to run them.

So lets put this one on hold.

PR #202, opened 2015-03-02
EZP-21650 AuthorsFieldType BDD scenario

miguelcleverti:
# EZP-2477 BDD scenario
BDD scenarios for issue EZP-21650
jira: https://jira.ez.no/browse/EZP-21650

requires: #181

### DIFF
--- a/Features/Stories/Fieldtypes/EZP-2477-AuthorsFieldType.feature
+++ b/Features/Stories/Fieldtypes/EZP-2477-AuthorsFieldType.feature
@@ -1,0 +1,97 @@
+Feature: Test the validations done on Author fieldtype from Editorial Interface
+
+    @javascript
+    Scenario: Message when using empty author field
+        Given I go to PlatformUI app with username "admin" and password "publish"
+        And a Content Type exists with identifier "Author" in Group with identifier "Content" with fields:
+            | identifier | Type     | Name   |
+            | name       | ezstring | Name   |
+            | author     | ezauthor | Author |
+        And I click on the menu zone "Platform"
+        And I click on the "Content structure" link
+        And I click on the side menu option "Create a content"
+        And I click on the content type "Author"
+        And I fill in "Name" with "AuthorTest"
+        When I click on the side menu option "Publish"
+        Then I should see an element "Author" with value "This field is empty"
+
+    @javascript
+    Scenario: Fill more than one author 
+        Given I go to PlatformUI app with username "admin" and password "publish"
+        And a Content Type exists with identifier "Author" in Group with identifier "Content" with fields:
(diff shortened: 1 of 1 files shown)


Features/Stories/Fieldtypes/EZP-2477-AuthorsFieldType.feature
```diff
@@ -0,0 +1,97 @@
+Feature: Test the validations done on Author fieldtype from Editorial Interface
+
+    @javascript
+    Scenario: Message when using empty author field
+        Given I go to PlatformUI app with username "admin" and password "publish"
```
pcardiga: All the administrator logins could be replaced by sentences such:
        Given I am logged in as an administrator

And the implementation would login as admin/publish.
Option: We could have a before-scenario hook where the login is made. Either that or including this line would fit the purpose o the login.

Features/Stories/Fieldtypes/EZP-2477-AuthorsFieldType.feature
```diff
@@ -0,0 +1,97 @@
...
+        And a Content Type exists with identifier "Author" in Group with identifier "Content" with fields:
+            | identifier | Type     | Name   |
+            | name       | ezstring | Name   |
+            | author     | ezauthor | Author |
+        And I click on the menu zone "Platform"
+        And I click on the "Content structure" link
+        And I click on the side menu option "Create a content"
+        And I click on the content type "Author"
+        And I fill in "Name" with "AuthorTest"
```
pcardiga: We could shorten the sentences:
"
        And I click on the menu zone "Platform"
        And I click on the "Content structure" link
        And I click on the side menu option "Create a content"
        And I click on the content type "Author"
        And I fill in "Name" with "AuthorTest"
"

With 
        And I create a content of content type "Author" with
            | name       | author | email |
            | AuthorTest |        |       | 

Where the creation of the content would follow all the steps described before.

andrerom: +1

Features/Stories/Fieldtypes/EZP-2477-AuthorsFieldType.feature
andrerom: +1 (more or less, but we did try something like this last year, lets make sure it does not grow into the same abstract level)

andrerom:
Besides those commenets:
@pcardiga We need to define what we actually want to cover in BDD, cause all of these PR's focus on the issue, while they should focus on covering the features (see files names), this will lead to way to many BDD tests and also would be bad way to organize them,

PR #177, opened 2015-02-20
EZP-21715: As an editor, I want to be able edit country fields. 1rst part: Give route config to fieldEditViews

StephaneDiot:
## Description

In this first part we want to give configuration from the application to the field edit views.

## Link

https://jira.ez.no/browse/EZP-21715

## Tests

unit tests

PR #203, opened 2015-03-03
EZP-21715: As an editor, I want to be able edit country fields

StephaneDiot:
## Description

This is the second part of the edit part of the contry field (first part was mostly done to import the country list from the application : https://github.com/ezsystems/PlatformUIBundle/pull/177).

## Video

youtube: http://youtu.be/cpLNMrW65lA

## Link 

jira: https://jira.ez.no/browse/EZP-21715

### DIFF
--- a/Resources/public/css/theme/modules/selection-filter.css
+++ b/Resources/public/css/theme/modules/selection-filter.css
@@ -19,3 +19,31 @@
     background: #528036;
     color: #fff;
 }
+
+.ez-selection-list {
+    background: #fff;
+    border: 1px solid #ccc;
+    border-top: 0;
+    border-bottom-right-radius: 5px;
+    border-bottom-left-radius: 5px;
+
+    box-sizing: border-box;
+    box-shadow: 4px 4px 5px 0 rgba(120, 120, 120, 0.75);
+
+    -webkit-transform: scaleY(1);
+            transform: scaleY(1);
+
+    -webkit-transform-origin: center top;
+            transform-origin: center top;
+
+    -webkit-transition: -webkit-transform 0.2s ease;
+            transition: transform 0.2s ease;
+}
+
+.is-list-hidden .ez-selection-values {
+    border: 1px solid #ccc;
+}
+
+.is-list-hidden.is-error .ez-selection-values {
+    border: 1px solid #BF3E33;
+}
--- a/Resources/public/js/views/fields/ez-country-editview.js
+++ b/Resources/public/js/views/fields/ez-country-editview.js
@@ -1,0 +1,133 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-country-editview', function (Y) {
+    "use strict";
+    /**
+     * Provides the field edit view for the country fields
+     *
+     * @module ez-country-editview
+     */
+
+    Y.namespace('eZ');
+
+    var FIELDTYPE_IDENTIFIER = 'ezcountry';
+
+    /**
+     * Country edit view
+     *
+     * @namespace eZ
+     * @class CountryEditView
+     * @constructor
+     * @extends eZ.SelectionEditView
+     */
+    Y.eZ.CountryEditView = Y.Base.create('countryEditView', Y.eZ.SelectionEditView, [], {
+
+        initializer: function () {
+            this.containerTemplate = '<div class="' +
+                this._generateViewClassName(this._getName()) + ' ' +
+                this._generateViewClassName(Y.eZ.SelectionEditView.NAME) + '"/>';
+        },
+
+        /**
+         * Returns the country name corresponding to the given alpha2Code.
+         * If the country is missing, it will return a country code and send a warning.
+         *
+         * @method _getCountryName
+         * @param {String} alpha2Code The country code (Alpha2)
+         * @protected
+         * @return {String}
+         */
+        _getCountryName: function (alpha2Code) {
+            var countryList = this.get('config');
+
+            if (countryList[alpha2Code]) {
+                return countryList[alpha2Code].Name;
+            } else {
+                console.warn("Unknown country code: " + alpha2Code);
+                return alpha2Code;
+            }
+        },
+
+        _getSelectionFilter: function () {
+            var container = this.get('container'),
+                selectedObjectArray = this._getSelectedTextValues(),
+                input = container.one('.ez-selection-filter-input'),
+                source = this._getSource(),
+                selected = [];
+
+            Y.Array.each(selectedObjectArray, function (selectedObject) {
+                selected.push(selectedObject.text);
+            });
+            return new Y.eZ.SelectionFilterView({
+                container: input.get('parentNode'),
+                inputNode: input,
+                listNode: this.get('container').one('.ez-selection-options'),
+                selected: selected,
+                source: source,
+                filter: true,
+                resultFilters: 'startsWith',
+                resultHighlighter: 'startsWith',
+                isMultiple: this.get('isMultiple'),
+                resultTextLocator: function (sourceElement) {
+                    return sourceElement.Name;
+                },
+                resultAttributesFormatter: function (sourceElement) {
+                    return {
+                        alpha2: sourceElement.Alpha2,
+                        text: sourceElement.Name,
+                    };
+                },
+            });
+        },
+
+        _getSource: function () {
+            var countriesArray = [];
+
+            Y.Object.each(this.get('config'), function (country) {
+                countriesArray.push(country);
+            });
+            return countriesArray;
+        },
+
+        _getSelectedTextValues: function () {
+            var field = this.get('field'),
+                valueIndexes = [],
+                that =  this,
+                res = [];
+
+            if ( field && field.fieldValue ) {
+                valueIndexes = field.fieldValue;
+            }
+            Y.Array.each(valueIndexes, function (index) {
+                res.push({text: that._getCountryName(index), alpha2: index});
+            });
+            return res;
+        },
+
+        _getFieldValue: function () {
+            var res = [];
+
+            Y.Array.each(this.get('values'), function(value) {
+                res.push(value.alpha2);
+            });
+            return res;
+        }
+    },  {
+        ATTRS: {
+            /**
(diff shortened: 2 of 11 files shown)


Resources/public/js/views/fields/ez-country-editview.js
```diff
@@ -0,0 +1,225 @@
...
+    /**
+     * Country edit view
+     *
+     * @namespace eZ
+     * @class CountryEditView
+     * @constructor
+     * @extends eZ.SelectionEditView
+     */
+    Y.eZ.CountryEditView = Y.Base.create('countryEditView', Y.eZ.SelectionEditView, [], {
+        VALUES_TPL: '<li class="ez-selection-value" />',
```
dpobel: this not needed, since this is already defined in Y.eZ.SelectionEditView

Resources/public/js/views/fields/ez-country-editview.js
```diff
@@ -0,0 +1,225 @@
...
+
+        /**
+         * Default implementation of the field edit view render. By default, it
+         * injects the field definition, the field, the content and the content
+         * type
+         *
+         * @method render
+         * @return {eZ.FieldEditView}
+         */
+        render: function () {
```
dpobel: it should not be needed to redefine this method as well

Resources/public/js/views/fields/ez-country-editview.js
```diff
@@ -0,0 +1,225 @@
...
+         * Returns the current country name.
+         * If the country is missing, it will return a country code and send a warning.
+         *
+         * @method _getCountryName
+         * @param {String} alpha2Code The country code (Alpha2)
+         * @protected
+         * @return {String}
+         */
+        _getCountryName: function (alpha2Code) {
+            var config = this.get('config');
```
dpobel: you should call this variable `countryList` or something like that

Resources/public/js/views/fields/ez-country-editview.js
```diff
@@ -0,0 +1,225 @@
...
+                filter: true,
+                resultFilters: 'startsWith',
+                resultHighlighter: 'startsWith',
+                isMultiple: this.get('isMultiple'),
+                resultTextLocator: function (sourceElement) {
+                    return sourceElement.Name;
+                },
+                resultAttributesFormatter: function (sourceElement) {
+                    return {
+                        id: sourceElement.Alpha2,
```
dpobel: instead of `id`, you should call it `alpha2` or `alpha2code`

Resources/public/js/views/fields/ez-country-editview.js
```diff
@@ -0,0 +1,225 @@
...
+                        text: sourceElement.Name,
+                    };
+                },
+            });
+        },
+
+        _addSelection: function (text, id) {
```
dpobel: it should take only the alpha2 code

Resources/public/js/views/fields/ez-country-editview.js
```diff
@@ -0,0 +1,225 @@
...
+
+            if (config[alpha2Code]) {
+                return config[alpha2Code].Name;
+            } else {
+                console.warn("Unknown country code: " + alpha2Code);
+                return alpha2Code;
+            }
+        },
+
+        _initSelectionFilter: function () {
```
dpobel: to avoid overriding this whole method, you should probably change the arguments of `_addSelection`/`_removeSelection`

Resources/public/js/views/fields/ez-country-editview.js
```diff
@@ -0,0 +1,225 @@
...
+        ATTRS: {
+            /**
+             * The text and id values of the selected options
+             *
+             * @attribute values
+             * @readonly
+             * @default []
+             * @type Array
+             */
+            values: {
```
dpobel: you don't need to redefine the attribute but it's good to "override" the comment since the content of the attribute is specific to the country edit view.

Resources/public/js/views/fields/ez-country-editview.js
```diff
@@ -0,0 +1,225 @@
...
+                "isMultiple": this.get('isMultiple'),
+            };
+        },
+        _getFieldValue: function () {
+            var res = [];
+
+            Y.Array.each(this.get('values'), function(value) {
+                res.push(value.id);
+            });
+            return res;
```
dpobel: when no country is selected, it returns an empty array, does the REST API handle that correctly ?

StephaneDiot: It does :) !

Resources/public/css/theme/modules/selection-filter.css
```diff
@@ -19,3 +19,31 @@
...
+    border: 1px solid #ccc;
+    border-top: 0;
+    border-bottom-right-radius: 5px;
+    border-bottom-left-radius: 5px;
+
+    box-sizing: border-box;
+    box-shadow: 4px 4px 5px 0 rgba(120, 120, 120, 0.75);
+
+    -webkit-transform: scaleY(1);
+    transform: scaleY(1);
```
dpobel: please keep the indentation while moving the CSS rule ie here, we should have:
```css
-webkit-transform: scaleY(1);
        transform: scaleY(1);
```

Resources/public/css/views/fields/edit/country.css
```diff
@@ -0,0 +1,55 @@
+/**
```
dpobel: as far as I can see, `views/fields/edit/country.css` and `theme/views/fields/edit/country.css` are almost the same as the selection.css files. To avoid that and since CountryEditView extends SelectionEditView, you should probably force the container of the CountryEditView instance to get the class `ez-view-selectioneditview` in addition to `ez-view-countryeditview`.

Resources/public/js/views/fields/ez-country-editview.js
```diff
@@ -0,0 +1,161 @@
...
+        },
+
+        _addSelection: function (attributes) {
+            var values = this.get('values');
+
+            values.push({text:attributes.text , alpha2: attributes.alpha2});
+            this._set('values', values, {"added": attributes.text});
+        },
+
+        _removeSelection: function (text, node) {
```
dpobel: to be consistent with `_addSelection`, `_removeSelection` should probably take the attributes in parameter (and the search through `values` should be made against the alpha2 code as well).

Resources/public/js/views/fields/ez-country-editview.js
```diff
@@ -0,0 +1,161 @@
...
+                res.push({text: that._getCountryName(index), alpha2: index});
+            });
+            return res;
+        },
+
+        _variables: function () {
+            var def = this.get('fieldDefinition'),
+                selected = [];
+
+            Y.Array.each(this.get('values'), function(value) {
```
dpobel: in the case of country, the `selected` template variable should probably contain the *country object* (ie the name + alpha2 code) so basically you can write:

```js
return {
    "isRequired": def.isRequired,
    "selected": this.get('values'),
    "isMultiple": this.get('isMultiple'),
};
```

Resources/public/js/views/fields/ez-selection-editview.js
```diff
@@ -162,8 +186,8 @@ YUI.add('ez-selection-editview', function (Y) {
         _addSelection: function (value) {
```
dpobel: for readability sake, you should rename the parameter

Tests/js/views/fields/assets/ez-selection-editview-tests.js
```diff
@@ -597,11 +597,11 @@ YUI.add('ez-selection-editview-tests', function (Y) {
...
 
             _assertCorrectFieldValue: function (fieldValue, msg) {
-                Y.Assert.areEqual(options.indexOf(this.newValue[0]), fieldValue[0], msg);
-                Y.Assert.areEqual(options.indexOf(this.newValue[1]), fieldValue[1], msg);
+                Y.Assert.areEqual(options.indexOf(this.newValue[0].text), fieldValue[0], msg);
+                Y.Assert.areEqual(options.indexOf(this.newValue[1].text), fieldValue[1], msg);
             },
 
-            newValue: ['Vergongeat', 'Croix Rousse']
+            newValue: [{text: 'Vergongeat'}, {text:'Croix Rousse'}]
```
dpobel: weird to have to make this change, normally from an outside point of view, SelectionEditView should not have been changed

dpobel:
Travis job failed because of a CS issue... Please, always run `grunt test` before pushing if the PR is supposed to be ready.

Besides that:
on f14281af1a3b71ec50d9590e0e7c6125e22df01b:
* the commit message is bad, it should be something like: `Fix a JS error in SelectionFilterView when unselecting a filtered out option`
* the new code branch is not covered in the SelectionFilterView. Since you are fixing a bug, you should add a regression test.

on 95a7f6df20e23cf9504b74a4465c5bcafc0b3286:
* you should add a corresponding unit test to check that the container of the CountryEditView is getting the `ez-view-selectioneditview class

Resources/public/js/views/fields/ez-country-editview.js
```diff
@@ -0,0 +1,131 @@
...
+     * @extends eZ.SelectionEditView
+     */
+    Y.eZ.CountryEditView = Y.Base.create('countryEditView', Y.eZ.SelectionEditView, [], {
+
+        initializer: function () {
+            this.containerTemplate = '<div class="' + this._generateViewClassName(this._getName()) + ' ' + this._generateViewClassName(Y.eZ.SelectionEditView.NAME) +'"/>';
+        },
+
+        /**
+         * Returns the current country name.
```
dpobel: *the current country name* ? More *the country name corresponding to the given alpha2Code*

Resources/public/js/views/fields/ez-selection-editview.js
```diff
@@ -125,6 +117,49 @@ YUI.add('ez-selection-editview', function (Y) {
             );
         },
 
+        _unselectSelectionFilter: function () {
```
dpobel: this method is badly named and not documented. It should have a name like `_addSelectionFilterUnselectEventHandler` but is it necessary to add such method ? If you feel it is, you should rename it, do the same for the `select` event and properly document those methods.

StephaneDiot:
ping @dpobel

dpobel:
+1

PR #205, opened 2015-03-10
EZP-24079: Sub items prototype

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24079

# Description

This is a very quick and dirty implementation of the sub items list. Don't worry, it's not meant to be merged as is but it's more to test the server side rendered strategy. Of course, it's working but there's a workaround to handle the fact that the Symfony application has no clue about the routing of the application (trick with the `data-route` attributes introduced in fe8d3e7fe7ac8b89759c61dd5633c72b8d2582f7).

It's worth mentioning that at some point, the controller to generate the sub items list will really look like [the REST views system](https://github.com/ezsystems/ezpublish-kernel/blob/master/doc/specifications/rest/REST-API-V2.rst#views) with a template applied to its result.


dpobel:
See https://github.com/ezsystems/PlatformUIBundle/pull/337

PR #207, opened 2015-03-11
[POC] Improve the tree loading performances by using the REST view

dpobel:
Related to:
* [EZP-23741 Research how to minimize REST requests required to build content tree](https://jira.ez.no/browse/EZP-23741)
* [EZP-24079 Prototype a basic sub-items list](https://jira.ez.no/browse/EZP-24079)

# Description

Trying to use the REST views to load the list of contents under a given location. This pattern is used while building the content tree and will be used to load the sub-items of a content if we don't choose the server side sub items rendering (see https://github.com/ezsystems/PlatformUIBundle/pull/205).

# Gain

By using the REST views, the number of HTTP requests is decreased. For instance, when loading a tree level with 10 contents, with the previous approach **32 HTTP requests** are needed while with the REST view approach, **21 HTTP requests** are send: 1 for the content list + 2 per content to load the location and the content type (note: the content type requests can *easily* be optimized by introducing a content type cache)

# Problem

As far as I can tell, the current patch has one issue regarding the content with several locations because currently the main location is taken without checking if it is in the subtree we are loading. The location is needed to know whether the content has some sub-items.
This can be fixed with different approaches but in *my ideal world*, I'd like to have a smarter REST view system ;-)

### DIFF
--- a/Resources/public/js/views/services/plugins/ez-contenttreeplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-contenttreeplugin.js
@@ -49,45 +49,47 @@ YUI.add('ez-contenttreeplugin', function (Y) {
         _loadNode: function (node, callback) {
             var capi = this.get('host').get('capi'),
                 options = {api: capi},
-                contentService = capi.getContentService();
+                contentService = capi.getContentService(),
+                viewStruct = contentService.newViewCreateStruct("children " + node.data.location.get('locationId'));
 
-            contentService.loadLocationChildren(node.id, function (err, response) {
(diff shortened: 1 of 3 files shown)


andrerom:
As discussed I think we need to start to think about introducing location search, with content inline in the response, and this should imho be part of "REST v2.1" effort where we also make sure ESI is used to keep resources unique.

q: How does this perform? It is search, so it is not cached so much, however ESI will help here I think.

dpobel:
Closing in favor of https://github.com/ezsystems/PlatformUIBundle/pull/389

PR #208, opened 2015-03-11
EZP-23788: Fix the field value when trying to update only the alt text

dpobel:
JIRA: https://jira.ez.no/browse/EZP-23788

# Description

Change needed after https://github.com/ezsystems/ezpublish-kernel/pull/1209 so that it's now possible to update only the alternative text without re-uploading the image. In such case, the REST API now expects the client to send an object based on the field value it provided.

# Tests

unit tests updated + manual tests.


dpobel:
ping @bdunogier @yannickroger @andrerom

PR #209, opened 2015-03-12
EZP-24091: Removed useless CSS rule after upgrade to Pure 0.6

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24091

# Description

Very small CSS clean up after [the upgrade to Pure 0.6](https://github.com/ezsystems/PlatformUIAssetsBundle/pull/5). The class `.pure-g-r` was removed in Pure 0.5 (and it was not used anyway).


andrerom:
+1

PR #210, opened 2015-03-16
EZP-24065: As a developer, I want to be able to assign a subtree to a section

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24065

# Description

[EZP-24065](https://jira.ez.no/browse/EZP-24065) is originally about assigning a section to a subtree but unfortunately, the REST API only provides a resource to assign a section to a content without reflecting this operation to its subtree(s). As a result, this patch only allows the developer to assign a section to some contents but the descendants of those contents in the tree won't be affected.

Screencast: http://youtu.be/29eLEuC4U-w

## Tasks:

* [x] add a custom view when dealing with the section
* [x] trigger the universal discovery widget when the using the "Assign to a subtree" button
* [x] patch the JavaScript REST client https://github.com/ezsystems/ez-js-rest-client/pull/54
* [x] patch the REST API https://github.com/ezsystems/ezpublish-kernel/pull/1212
* [x] assign the section to the contents
* [x] add some feedback to the interface

# Tests

manual tests + unit tests


dpobel:
ping @andrerom @StephaneDiot @yannickroger

andrerom:
+1

yannickroger:
+1

PR #211, opened 2015-03-17
EZP-23770 Refactored BDD scenarios

miguelcleverti:
#### Includes the following stories: 
* QA-207 - BDD scenarios for Checkbox Fieldtype
* QA-201 - BDD scenarios for Authors Fieldtype
* QA-221 - BDD scenarios for Email Fieldtype
* QA-223 - BDD scenarios for Float Fieldtype
* QA-228 - BDD scenarios for Integer Fieldtype
* QA-248 - BDD scenarios for URL Fieldtype
* QA-213 - BDD scenarios for Date Fieldtype
* QA-215 - BDD scenarios for Time Fieldtype
* QA-211 - BDD scenarios for DateAndTime Fieldtype
* QA-426 - BDD scenarios for Drag & Drop for Binary, image and video files

### DIFF
--- a/Features/Stories/Fieldtypes/AuthorsFieldType.feature
+++ b/Features/Stories/Fieldtypes/AuthorsFieldType.feature
@@ -1,96 +1,180 @@
-Feature: Test the validations done on Author fieldtype from Editorial Interface - Authors fieldtype
+Feature: Test the validations done on fields from PlatformUI - Authors fieldtype
+    In order to validate the authors fieldtype
+    As an Editor  user
+    I need to be able to create, update and delete content with authors fieldtypes
+
+    Background:
+        Given I am logged in as an Editor in PlatformUI
+
+    ##
+    # Validate the existence of expected fields from a field type when creating a content
+    ##
+    @javascript
+    Scenario: A Content of a Content Type that has an author fieldtype must have an author and an email fields
+        Given a Content Type with an "authors" Field exists
+        When I create a content of this Content Type
+        Then I should have an "name" field
+        And I should have an "email" field
 
     @javascript
-    Scenario: Message when using empty author field
-        Given I am logged in as admin on PlatformUI
-        And a Content Type exists with identifier "Author" in Group with identifier "Content" with fields:
-            | identifier | Type     | Name   |
-            | name       | ezstring | Name   |
-            | author     | ezauthor | Author |
-        And I create a content of content type "Author" with:
-            | Name       |
-            | AuthorTest |
-        And I fill in "Author:" subform with:
-            | Name: | Email: |
-            |       |        |
-        When I click on the actionbar action "Publish"
-        Then I should see an element "Author" with value "This field is empty"
+    Scenario: A Content of a Content Type that has an author fieldtype must be prefilled with user's name and email
+        Given a Content Type with an "authors" Field exists
+        When I create a content of this Content Type
+        Then I should have an "name" field filled with Editor's Username
+        And I should have an "email" field filled with Editor's Email
 
     @javascript
-    Scenario: Fill more than one author
-        Given I am logged in as admin on PlatformUI
-        And a Content Type exists with identifier "Author" in Group with identifier "Content" with fields:
-            | identifier | Type     | Name   |
-            | name       | ezstring | Name   |
-            | author     | ezauthor | Author |
-        And I create a content of content type "Author" with:
-            | Name       |
-            | AuthorTest |
-        And I clicked on "Add another author" button
-        And I fill in "Author:" subform with:
-            | Name:   | Email:              |
-            | Author1 | author1@authors.pt  |
-            | Author2 | author2@authors.com |
-        When I click on the actionbar action "Publish"
-        Then I should see an element "Author" with value "Author1 <author1@authors.pt> Author2 <author2@authors.com>"
+    Scenario: A Content of a Content Type that has an author fieldtype must have a "Remove this author" and a "Add another author" buttons
+        Given a Content Type with an "authors" Field exists
+        When I create a content of this Content Type
+        Then I should have an "Remove this author" button
+        And I should have an "Add another author" button
 
     @javascript
-    Scenario Outline: Fill three authors then remove one then publish
-        Given I am logged in as admin on PlatformUI
-        And a Content Type exists with identifier "Author" in Group with identifier "Content" with fields:
-            | identifier | Type     | Name   |
-            | name       | ezstring | Name   |
-            | author     | ezauthor | Author |
-        And I create a content of content type "Author" with:
-            | Name       |
-            | AuthorTest |
-        And I clicked on "Add another author" button
-        And I clicked on "Add another author" button
-        And I fill in "Author:" subform with:
-            | Name:   | Email:              |
-            | Author1 | author1@authors.pt  |
-            | Author2 | author2@authors.com |
-            | Author3 | author3@authors.no  |
-        And I click on the "Remove this author" button number <index>
-        When I click on the actionbar action "Publish"
-        Then I should see an element "Author" with value "<value>"
-
-        Examples:
-            | index | value                                                      |
-            | 1     | Author2 <author2@authors.com> Author3 <author3@authors.no> |
-            | 2     | Author1 <author1@authors.pt> Author3 <author3@authors.no>  |
-            | 3     | Author1 <author1@authors.pt> Author2 <author2@authors.no>  |
+    Scenario: When editing a Content the label of an author field must have the same name of the field type from the respective Content Type
+        Given a Content Type with an "authors" with "Name" "Participants" exists
+        When I create a content of this Content Type
+        Then I should have a "Participants:" label related with the "eame" and "email" fields
+        And I should have a "Name:" label related with the "name" field
+        And I should have an "Email:" label related with the "email" field
 
     @javascript
-    Scenario: Fill one author then remove it then publish
-        Given I am logged in as admin on PlatformUI
-        And a Content Type exists with identifier "Author" in Group with identifier "Content" with fields:
-            | identifier | Type     | Name   |
-            | name       | ezstring | Name   |
-            | author     | ezauthor | Author |
-        And I create a content of content type "Author" with:
-            | Name       |
-            | AuthorTest |
-        And I fill in "Author:" subform with:
-            | Name:   | Email:             |
-            | Author1 | author1@authors.pt |
-        And I clicked on "Add another author" button
-        And I click on the "Remove this author" button number "1"
-        When I click on the actionbar action "Publish"
-        Then I should see an element "Author" with value "This field is empty"
+    Scenario: When editing a Content with an author field the "Remove this author" button is disabled when exists one author
+        Given a Content Type with an "authors" Field exists
+        When I create a content of this Content Type
+        Then all the "Remove this author" buttons are "disabled"
 
     @javascript
-    Scenario: Unable to publish with wrong email format
-        Given I am logged in as admin on PlatformUI
-        And a Content Type exists with identifier "Author" in Group with identifier "Content" with fields:
-            | identifier | Type     | Name   |
-            | name       | ezstring | Name   |
-            | author     | ezauthor | Author |
-        And I create a content of content type "Author" with:
-            | Name       |
-            | AuthorTest |
-        And I fill in "Author:" subform with:
-            | Name:   | Email:            |
-            | Author1 | author1authors.pt |
-        When I click on the actionbar action "Publish"
-        Then I should see 'New "Author"' title
+    Scenario: When editing a Content the "Remove this author" button is enabled when exists more than one author
+        Given a Content Type with an "authors" Field exists
+        When I create a content of this Content Type
+        And I add two authors
+        Then all the "Remove this author" buttons are "enabled"
+
+    @javascript
+    Scenario: When editing a Content with an author field with two authors the "Remove this author" button is disabled again when deleting one author
+        Given a Content Type with an "authors" Field exists
+        And a Content of this Content Type exists with "two" authors
+        When I remove the "First" author
+        Then all the "Remove this author" buttons are "disabled"
+
+    @javascript
+    Scenario: The label of an mandatory author field of a Content must have an * as suffix on the field's label's' name
+        Given a Content Type with a "required" "authors" with "Name" "Participants" exists
+        When I create a content of this Content Type
+        Then I should have a "Participants*:" label related with the "name" and "email" fields
+
+    ##
+    # Creating Content using a Content Type that has an Authors Fieldtype
+    ##
+    @javascript
+    Scenario: Publishing a valid authors Field works when using one author
+        Given a Content Type with an "authors" Field exists
+        When I create a content of this Content Type
+        And I set "Paul" as the "name" Field Value of the "first" author
+        And I set "paul@acme.com" as the "email" Field Value of the "first" author
+        And I publish the content
+        Then the Content is successfully published
+
+    @javascript
+    Scenario: Publishing a valid authors Field works when using more than one author
+        Given a Content Type with an "authors" Field exists
+        When I create a content of this Content Type
+        And I set "Paul" as the "name" Field Value of the "first" author
+        And I set "paul@acme.com" as the "email" Field Value of the "first" author
+        And I add another author
+        And I set "Ana" as the "name" Field Value of the "second" author
+        And I set "ana@acme.com" as the "email" Field Value of the "second" author
+        And I publish the content
+        Then the Content is successfully published
+
+    @javascript
+    Scenario: Publishing an authors Field fails validation when using an invalid email
+        Given a Content Type with an "authors" Field exists
+        When I create a content of this Content Type
+        And I set "Paul" as the "name" Field Value of the "first" author
+        And I set "paul" as the "email" Field Value of the "first" author
+        And I publish the content
+        Then the Content is not published
+        And I see a "!" before the "email" Field Value
+
+    @javascript
+    Scenario: Publishing an required authors Field fails validation when using an empty author
+        Given a Content Type with a "requited" "authors" exists
+        When I create a content of this Content Type
+        And I set an empty value as the "name" Field Value of the "first" author
+        And I set an empty value as the "email" Field Value of the "first" author
+        And I publish the content
+        Then Publishing fails
+        And I see a "This field is required" before the "name" Field Value
+        And I see a "This field is required" before the "email" Field Value
+
+    ##
+    # Update Content using a Content Type that has an Authors Fieldtype
+    ##
+    @javascript
+    Scenario: Updating an author Field works
+        Given a Content Type with an "authors" Field exists
+        And a Content of this Content Type exists
+        When I update this content setting "Paul" as the "name" Field Value of the "first" author
+        And I update this content setting "paul@acme.com" as the "email" Field Value of the "first" author
+        And I publish the content
+        Then the Content is successfully published
+
+    @javascript
+    Scenario: Updating an authors Field fails validation when using an invalid email
+        Given a Content Type with an "authors" Field exists
+        And a Content of this Content Type exists
+        When I set "paul" as the "email" Field Value of the "first" author
+        And I publish the content
+        Then the Content is not published
+        And I see a "!" before the "email" Field Value
+
+    @javascript
+    Scenario: Publishing a required authors Field fails validation when using an empty author
+        Given a Content Type with a "requited" "authors" Field exists
+        And a Content of this Content Type exists
+        When I set an empty value as the "name" Field Value of the "first" author
+        And I set an empty value as the "email" Field Value of the "first" author
+        And I publish the content
+        Then the Content is not published
+        And I see a "!" before the "name" Field Value
+        And I see a "!" before the "email" Field Value
+
+    ##
+    # Delete Content using a Content Type that has an Authors Fieldtype
+    ##
+    @javascript
+    Scenario: Deleting a content that has an author field
+        Given a Content Type with an "authors" Field exists
+        And a Content of this Content Type exists
+        When I delete this Content
+        Then the Content is successfully deleted
+
+    @javascript
+    Scenario: Deleting one author from the authors list
+        Given a Content Type with an "authors" Field exists
+        And a Content of this Content Type exists with "two" authors
+        When I delete the "first" author
+        Then the author is successfully deleted
+
+    ##
+    # Viewing content that has an Authors fieldtype
+    ##
+    @javascript
+    Scenario: Viewing a Content that has an author fieldtype should return "This field is empty" when the value is empty
+        Given a Content Type with an "authors" Field exists
+        And a Content of this Content Type exists with "name" and "email" Fields Values set to empty
+        When I view this Content
+        Then I should see a field with value "This field is empty"
+
+    @javascript
+    Scenario: Viewing a Content that has more than one author should return the expected authors
+        Given a Content Type with an "authors" Field exists
+        And a Content of this Content Type exists with "<name>" and "<email>" Fields Values
+            | name | email         |
+            | Paul | paul@acme.com |
+            | Ana  | ana@acme.com  |
+        When I view this Content
+        Then I should see a field with value "Paul <paul@acme.com>"
+        And I should see a field with value "Ana <ana@acme.com>"
--- a/Features/Stories/Fieldtypes/AuthorsFieldType.feature
+++ b/Features/Stories/Fieldtypes/AuthorsFieldType.feature
@@ -1,0 +1,96 @@
+Feature: Test the validations done on Author fieldtype from Editorial Interface - Authors fieldtype
+
+    @javascript
+    Scenario: Message when using empty author field
+        Given I am logged in as admin on PlatformUI
+        And a Content Type exists with identifier "Author" in Group with identifier "Content" with fields:
+            | identifier | Type     | Name   |
+            | name       | ezstring | Name   |
+            | author     | ezauthor | Author |
+        And I create a content of content type "Author" with:
+            | Name       |
+            | AuthorTest |
+        And I fill in "Author:" subform with:
+            | Name: | Email: |
+            |       |        |
+        When I click on the actionbar action "Publish"
+        Then I should see an element "Author" with value "This field is empty"
+
+    @javascript
+    Scenario: Fill more than one author
+        Given I am logged in as admin on PlatformUI
+        And a Content Type exists with identifier "Author" in Group with identifier "Content" with fields:
+            | identifier | Type     | Name   |
+            | name       | ezstring | Name   |
+            | author     | ezauthor | Author |
+        And I create a content of content type "Author" with:
+            | Name       |
+            | AuthorTest |
+        And I clicked on "Add another author" button
+        And I fill in "Author:" subform with:
+            | Name:   | Email:              |
+            | Author1 | author1@authors.pt  |
+            | Author2 | author2@authors.com |
+        When I click on the actionbar action "Publish"
+        Then I should see an element "Author" with value "Author1 <author1@authors.pt> Author2 <author2@authors.com>"
+
+    @javascript
+    Scenario Outline: Fill three authors then remove one then publish
+        Given I am logged in as admin on PlatformUI
+        And a Content Type exists with identifier "Author" in Group with identifier "Content" with fields:
+            | identifier | Type     | Name   |
+            | name       | ezstring | Name   |
+            | author     | ezauthor | Author |
+        And I create a content of content type "Author" with:
+            | Name       |
+            | AuthorTest |
+        And I clicked on "Add another author" button
+        And I clicked on "Add another author" button
+        And I fill in "Author:" subform with:
+            | Name:   | Email:              |
+            | Author1 | author1@authors.pt  |
+            | Author2 | author2@authors.com |
+            | Author3 | author3@authors.no  |
+        And I click on the "Remove this author" button number <index>
+        When I click on the actionbar action "Publish"
+        Then I should see an element "Author" with value "<value>"
+
+        Examples:
+            | index | value                                                      |
+            | 1     | Author2 <author2@authors.com> Author3 <author3@authors.no> |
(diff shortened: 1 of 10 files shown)


joaoinacio:
Nitpick: `s/Features\/Stories\/Fieldtypes/Features\/FieldTypes/g` 8-)

andrerom:
You shouldn't use use "/" as separator when humans should be able to parse it.. :)

But yes, as it should be feature centered, "Stories" should be removed. However "Features" is the folder we put BDD in, so in needs to be on root, hence `Features/Fieldtypes/` or rather `Features/Content/Fieldtypes`

Features/Stories/Fieldtypes/CheckboxFieldType.feature
```diff
@@ -0,0 +1,30 @@
+Feature: Test the validations done on fields from Editorial Interface - Checkbox fieldtype
+
+    @javascript
+    Scenario: Publish with checkbox checked
+        Given I am logged in as admin on PlatformUI
+        And a Content Type exists with identifier "Checkbox" in Group with identifier "Content" with fields:
```
bdunogier: Does it matter that it is part of the Content Type Group Content ?

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,96 @@
...
+            | identifier | Type     | Name   |
+            | name       | ezstring | Name   |
+            | author     | ezauthor | Author |
+        And I create a content of content type "Author" with:
+            | Name       |
+            | AuthorTest |
+        And I fill in "Author:" subform with:
+            | Name:   | Email:             |
+            | Author1 | author1@authors.pt |
+        And I clicked on "Add another author" button
```
bdunogier: "click" without "ed"

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,96 @@
...
+        And I click on the "Remove this author" button number "1"
+        When I click on the actionbar action "Publish"
+        Then I should see an element "Author" with value "This field is empty"
```
bdunogier: Why element ? "Field" would be more accurate.

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,96 @@
...
+
+    @javascript
+    Scenario: Unable to publish with wrong email format
```
bdunogier: What about "Content with an invalid email in an Author Field fails validation" ? I think it's a better description of what is covered here.

Features/Stories/Fieldtypes/CheckboxFieldType.feature
pcardiga: If the Content Type don't exist, it will be created. Then the Content Type Group.

bdunogier: It's not the point. The fact that it is part of this or that content type group has no incidence on this scenario.

pcardiga: In BDD scenarios, 3 phases exists: pre-condition, action, validation/verification.
To be able to do the action, the pre-condition has to be filled, i.e., the content type has to exist or be created. 
To do that, 3 options might be followed: 
-- either Content Type Group is referred in the sentence so that Content Type might be created within it if it don't exist, 
-- or one can assume that content type always belong to the same content type group, i.e, "Content"
-- Or join the two previous approaches, having two different sentences: one omitting Content Type Group (assuming the "Content" is used) and other referring Content Type Group for the other situations.

Features/Stories/Fieldtypes/AuthorsFieldType.feature
andrerom: What is the terminology used in Behat/Mink on this? _(so we don't duplicate the domain language)_

pcardiga: Behat/Mink uses the "element" term, but if we all agree that "Field" is more noticeable, I don't see any problem in using this term.

andrerom: Lets go with `Field`, it is what MinkExtension uses:
https://github.com/Behat/MinkExtension/blob/master/src/Behat/MinkExtension/Context/MinkContext.php#L101

Elements also there, but for generic html elements, so can also be used but then we miss out on the specific Form handling. Which seems to include sentences for field, check, checkbook, select, file upload field and button :)

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,96 @@
...
+    Scenario Outline: Fill three authors then remove one then publish
+        Given I am logged in as admin on PlatformUI
+        And a Content Type exists with identifier "Author" in Group with identifier "Content" with fields:
+            | identifier | Type     | Name   |
+            | name       | ezstring | Name   |
+            | author     | ezauthor | Author |
+        And I create a content of content type "Author" with:
+            | Name       |
+            | AuthorTest |
+        And I clicked on "Add another author" button
```
andrerom: can reuse: https://github.com/Behat/MinkExtension/blob/master/src/Behat/MinkExtension/Context/MinkContext.php#L77

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,96 @@
...
+        And I clicked on "Add another author" button
+        And I fill in "Author:" subform with:
```
andrerom: Ideally should reuse: https://github.com/Behat/MinkExtension/blob/master/src/Behat/MinkExtension/Context/MinkContext.php#L113

However you might need to ask @dpobel to use predictable form field names for this if not already the case.

Features/Stories/Fieldtypes/AuthorsFieldType.feature
bdunogier: Since we were in a Field Scenario, I thought this sentence was testing the the Value it had after publishing.

Element is fine, since it is HTML (and it's not a Form Element anymore, missing that Field is actually invalid here).

Features/Stories/Fieldtypes/CheckboxFieldType.feature
bdunogier: The current one is okay.

Features/Stories/Fieldtypes/EmailFieldType.feature
```diff
@@ -0,0 +1,78 @@
...
+            | name       | ezstring | Name  |
+            | email      | ezemail  | Email |
+        And I create a content of content type "Email" with:
+            | Name      | Email    |
+            | EmailTest | failtest |
+        When I click on the actionbar action "Publish"
+        Then I should see "The value should be a valid email address" text
+
+    @javascript
+    Scenario Outline: Use of invalid email field
```
bdunogier: Isn't this the same one than the one above ?

Features/Stories/Fieldtypes/FloatFieldType.feature
```diff
@@ -0,0 +1,87 @@
...
+        And a Content Type exists with identifier "Float3" in Group with identifier "Content" with fields:
+            | identifier | Type     | Name  | validator          |
+            | name       | ezstring | Name  | false              |
+            | float      | ezfloat  | Float | FloatValue:2.0~3.1 |
+        And I create a content of content type "Float3" with:
+            | Name   | Float   |
+            | <name> | <float> |
+        When I click on the actionbar action "Publish"
+        Then I should see "FloatTest" title
+        And I should see an element 'Float' with value "<float>""
```
bdunogier: Two quotes at the end of the line

Features/Stories/Fieldtypes/UrlFieldType.feature
```diff
@@ -0,0 +1,58 @@
+Feature: Test the validations done on fields from PlatformUI - URL fieldtype
+
+    @javascript
+    Scenario Outline: Fill one Url field with valid url
+        Given I go to PlatformUI app with username "admin" and password "publish"
```
bdunogier: I don't think we need the username and password.

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,96 @@
...
+        Given I am logged in as admin on PlatformUI
+        And a Content Type exists with identifier "Author" in Group with identifier "Content" with fields:
+            | identifier | Type     | Name   |
+            | name       | ezstring | Name   |
+            | author     | ezauthor | Author |
+        And I create a content of content type "Author" with:
+            | Name       |
+            | AuthorTest |
+        And I fill in "Author:" subform with:
+            | Name: | Email: |
```
bdunogier: Is it worth using this syntax if we only fill one field ? It would be more readable as one sentence, Like `I fill in the "Author" subform's "name" with an empty string`

Features/Stories/Fieldtypes/CheckboxFieldType.feature
miguelcleverti: It is possible to use the default Content Type 'Content' if not specified, the sentence is just missing I will add it ASAP.

Features/Stories/Fieldtypes/AuthorsFieldType.feature
miguelcleverti: We are using EzSystems/BehatBundle Browser Context (https://github.com/ezsystems/BehatBundle/blob/master/Context/Browser/SubContext/CommonActions.php) which itself uses that when possible

Features/Stories/Fieldtypes/AuthorsFieldType.feature
andrerom: > ignore me :)

you are not ignored :)

I see we had a small understanding, there is no field here of course as it is empty, so for now element, and if we find a way to introduce a abstract term for `Field` we can explore that as followup maybe? My concern on that is that Fields are vastly different, so if you try to hide that from the BDD for everything then we cause a lot of potential fragile magic to work behind the scenes. But anyway, give me a ping on what you guys agree on during today meeting.

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,96 @@
...
+    @javascript
+    Scenario: Message when using empty author field
+        Given I am logged in as admin on PlatformUI
+        And a Content Type exists with identifier "Author" in Group with identifier "Content" with fields:
+            | identifier | Type     | Name   |
+            | name       | ezstring | Name   |
+            | author     | ezauthor | Author |
+        And I create a content of content type "Author" with:
+            | Name       |
+            | AuthorTest |
```
dpobel: to ease the usage of the test report, it would be nice to use more meaningful names, "AuthorTest" appears 5 times in this PR. Here, the name could be "Empty author test" for instance.

bdunogier: This should be covered by the updated direction we are going to prototype.

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,180 @@
...
+        Given I am logged in as an Editor in PlatformUI
+
+    ##
+    # Validate the existence of expected fields from a field type when creating a content
+    ##
+    @javascript
+    Scenario: A Content of a Content Type that has an author fieldtype must have an author and an email fields
+        Given a Content Type with an "authors" Field exists
+        When I create a content of this Content Type
+        Then I should have an "name" field
```
bdunogier: `a "name" field` ?

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,180 @@
...
+        Given a Content Type with an "authors" Field exists
+        When I create a content of this Content Type
+        Then I should have an "name" field filled with Editor's Username
+        And I should have an "email" field filled with Editor's Email
+
+    @javascript
+    Scenario: A Content of a Content Type that has an author fieldtype must have a "Remove this author" and a "Add another author" buttons
+        Given a Content Type with an "authors" Field exists
+        When I create a content of this Content Type
+        Then I should have an "Remove this author" button
```
bdunogier: Do we know how we will deal with UI translations when they're added ?

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,180 @@
...
+        And I should have an "Add another author" button
+
+    @javascript
+    Scenario: When editing a Content the label of an author field must have the same name of the field type from the respective Content Type
```
bdunogier: "same name _than_ the field" ?

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,180 @@
...
+        Given a Content Type with an "authors" with "Name" "Participants" exists
+        When I create a content of this Content Type
+        Then I should have a "Participants:" label related with the "eame" and "email" fields
```
bdunogier: typo: name

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,180 @@
...
+
+    @javascript
+    Scenario: Publishing a required authors Field fails validation when using an empty author
+        Given a Content Type with a "requited" "authors" Field exists
+        And a Content of this Content Type exists
+        When I set an empty value as the "name" Field Value of the "first" author
+        And I set an empty value as the "email" Field Value of the "first" author
+        And I publish the content
+        Then the Content is not published
+        And I see a "!" before the "name" Field Value
```
bdunogier: I'd say that's implementation, not behaviour.

Features/Stories/Fieldtypes/CheckboxFieldType.feature
```diff
@@ -0,0 +1,78 @@
...
+        Then the Content is successfully published
+
+    ##
+    # Update Content using a Content Type that has a Checkbox Field Type
+    ##
+    @javascript
+    Scenario: Updating a Checkbox field using a valid checkbox Field works
+        Given a Content Type with a "checkbox" Field exists
+        And a Content of this Content Type exists with "checkbox" Field Value checked
+        When I update this content unchecking the Field Value
```
bdunogier: Why does this sentence contain both the editing action and the action of checking, why create above uses two sentences ?

I'd have expected:
When I edit this Content
 And I uncheck the field value

Features/Stories/Fieldtypes/CheckboxFieldType.feature
```diff
@@ -0,0 +1,78 @@
...
+    Scenario: Viewing a Content that has a Checkbox fieldtype should show the expected value when the checkbox is checked
+        Given a Content Type with a "checkbox" Field exists
+        And a Content of this Content Type exists with "checkbox" Field Value checked
+        When I view this Content
+        Then I should see a field with value "Yes"
+
+    @javascript
+    Scenario: Viewing a Content that has a Checkbox fieldtype should show the expected value when the checkbox is unchecked
+        Given a Content Type with a "checkbox" Field exists
+        And a Content of this Content Type exists with "checkbox" Field Value unchecked
```
bdunogier: Could we separate the check that a Content of this Type exists from the value check ?

And a Content of this Content Type exists
And the "checkbox" Field Value is unchecked

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,180 @@
...
+        Given a Content Type with an "authors" Field exists
+        And a Content of this Content Type exists with "two" authors
+        When I remove the "First" author
+        Then all the "Remove this author" buttons are "disabled"
+
+    @javascript
+    Scenario: The label of an mandatory author field of a Content must have an * as suffix on the field's label's' name
+        Given a Content Type with a "required" "authors" with "Name" "Participants" exists
+        When I create a content of this Content Type
+        Then I should have a "Participants*:" label related with the "name" and "email" fields
```
bdunogier: imho the * is an implementation thing. It's used for ALL fieldtypes, right ?

I'd rather have something like "Then the "Participants" field should be marked as mandatory.

Also, the scenario doesn't mention what page this should appear on.

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,180 @@
...
+        And I set "Paul" as the "name" Field Value of the "first" author
+        And I set "paul@acme.com" as the "email" Field Value of the "first" author
+        And I publish the content
+        Then the Content is successfully published
+
+    @javascript
+    Scenario: Publishing a valid authors Field works when using more than one author
+        Given a Content Type with an "authors" Field exists
+        When I create a content of this Content Type
+        And I set "Paul" as the "name" Field Value of the "first" author
```
bdunogier: Would it be much more complicated if we used such a sentence ?

`And I add "Paul <paul@acme.com>" as the "first" author`

It could under the hood translate into the two sentences you have here.

Features/Stories/Fieldtypes/AuthorsFieldType.feature
pcardiga: I'll replace the message by "This field is required", as is on other fieldtypes

Features/Stories/Fieldtypes/CheckboxFieldType.feature
pcardiga: Updated and splitted edit from the update action. This was also updated in other fieldtypes, since the updating sentences were using the edit+action in the same phrase.

Features/Stories/Fieldtypes/CheckboxFieldType.feature
pcardiga: The problem I see is when to do the publishing action to avoid creating multiple versions. 
To avoid that, we could either
- include/set all necessary properties in the same sentence (as we have now) or
- Have several sentences to create the object, but only to publish it in the end. 
With second approach, question is when do we know it's the end of the content definition? Until now, the action (the "when" part of BDD sentences)  is done with a "When I create|update|delete|view <something>". We could state that when we are going to do some action on the object then the object must exist and then we publish it..

bdunogier: Aaah yes.

Well, what bothers me a bit with this approach is that we end up with weird looking sentences, but it's a fair deal, I guess. Having to mention the "end of a chain" in this context sounds a bit tedious.

Features/Stories/Fieldtypes/AuthorsFieldType.feature
bdunogier: "is marked as required" ? Sounds a bit more right to me.

bdunogier:
One overall remark: I think we could shorten "of this content type" with "of this type in most sentences", at least in the Context of FieldType testing. What do you think ? It shouldn't conflict with anything else.

Features/Stories/Fieldtypes/AuthorsFieldType.feature
pcardiga: The scenario should appear during the create action. All create/update scenarios have a 3 steps action: (1) create|edit; (2) set|verify values ; (3) publish action if necessary. In current case, since the publish action was not performed, then we are on the editing page

bdunogier: Hmmm, ok. It kinda feels weird, but it makes sense.

Features/Stories/Fieldtypes/AuthorsFieldType.feature
pcardiga: Sorry. I explained myself wrong. I'll replace the sentence by 
"And I see a "This field is required" before the "name" Field Value". Like this, we not only validate that the message appears, as we validate the message text.

bdunogier: OK.

Features/Stories/Fieldtypes/AuthorsFieldType.feature
pcardiga: Let's go with "set" instead of "add". 
" And I **set** "Paul <paul@acme.com>" as the "first" author "

Since we have an action before (create | edit), using "set" allows us to reuse the same sentence in both cases

bdunogier: Well, I'd say that it depends if you specify the order

```
And I add "Paul <paul@acme.com>" to "author"
And I add "Maria <maria@acme.com>" to "author"
```

vs

```
And I set "Paul <paul@acme.com>" as the "second" "author"
And I set "Maria <maria@acme.com>" as the "second" "author"
```

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,180 @@
...
+    @javascript
+    Scenario: A Content of a Content Type that has an author fieldtype must be prefilled with user's name and email
+        Given a Content Type with an "authors" Field exists
+        When I create a content of this Content Type
+        Then I should have an "name" field filled with Editor's Username
+        And I should have an "email" field filled with Editor's Email
+
+    @javascript
+    Scenario: A Content of a Content Type that has an author fieldtype must have a "Remove this author" and a "Add another author" buttons
+        Given a Content Type with an "authors" Field exists
```
crevillo: with an "authors" Field exists? sounds a bit weird to me...

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,180 @@
...
+        When I create a content of this Content Type
+        Then I should have an "Remove this author" button
+        And I should have an "Add another author" button
```
crevillo: ```should have``` => ```should see```?

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,180 @@
...
+
+    @javascript
+    Scenario: When editing a Content the label of an author field must have the same name of the field type from the respective Content Type
+        Given a Content Type with an "authors" with "Name" "Participants" exists
```
crevillo: again this with ... exists part looks a bit weird to me. but not an expert on this topic...

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,180 @@
...
+        When I create a content of this Content Type
+        Then I should have a "Participants:" label related with the "eame" and "email" fields
+        And I should have a "Name:" label related with the "name" field
+        And I should have an "Email:" label related with the "email" field
+
+    @javascript
+    Scenario: When editing a Content with an author field the "Remove this author" button is disabled when exists one author
```
crevillo: exists *only* an author maybe?

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,180 @@
...
+        When I create a content of this Content Type
+        And I add two authors
+        Then all the "Remove this author" buttons are "enabled"
+
+    @javascript
+    Scenario: When editing a Content with an author field with two authors the "Remove this author" button is disabled again when deleting one author
+        Given a Content Type with an "authors" Field exists
+        And a Content of this Content Type exists with "two" authors
+        When I remove the "First" author
+        Then all the "Remove this author" buttons are "disabled"
```
crevillo: but should we disable the remove button if we have only one author and the field isn't mandatory? Or that goes to another sceneario?

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,180 @@
...
+        When I create a content of this Content Type
+        And I set "Paul" as the "name" Field Value of the "first" author
+        And I set "paul" as the "email" Field Value of the "first" author
+        And I publish the content
+        Then the Content is not published
+        And I see a "!" before the "email" Field Value
+
+    @javascript
+    Scenario: Publishing an required authors Field fails validation when using an empty author
+        Given a Content Type with a "requited" "authors" exists
```
crevillo: typo, ```requited``` should be ```required```

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,180 @@
...
+        Given a Content Type with an "authors" Field exists
+        And a Content of this Content Type exists
+        When I set "paul" as the "email" Field Value of the "first" author
+        And I publish the content
+        Then the Content is not published
+        And I see a "!" before the "email" Field Value
+
+    @javascript
+    Scenario: Publishing a required authors Field fails validation when using an empty author
+        Given a Content Type with a "requited" "authors" Field exists
```
crevillo: same typo as above

Features/Stories/Fieldtypes/AuthorsFieldType.feature
pcardiga: The buttons refers to the possibility of having a place to include authors. And, accordingly specs, one place/field must always be present.

Features/Stories/Fieldtypes/AuthorsFieldType.feature
pcardiga: Typos fixed.
Thanks Carlos for taking time to give feedback.

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,175 @@
...
+        Given a Content Type with an "authors" Field exists
+        When I create a content of this type
+        Then I should see a "Remove this author" button
+        And I should see an "Add another author" button
+
+    @javascript
+    Scenario: When editing a Content the label of an author field must have the same name than field type from the respective Content Type
+        Given a Content Type with an "authors" with "Name" "Participants" exists
+        When I create a content of this type
+        Then I should see a "Participants:" label related with the "name" and "email" fields
```
dpobel: I'm not a fan of the usage *related* here and in the lines below because the *relation* between "Participants" and the "name" and "email" fields is different from the *relation* between "Name:" and the "name" field and *related* is not precise enough.

Actually, I would replace *Then I should see a "Participants:" label related with the "name" and "email" fields* by something like *Then I should see "Participants" as the field definition name*

And for the other case, I would replace *And I should see a "Name:" label related with the "name" field* by *And I should see a form field labelled with "Name:"*
and replace *And I should see an "Email:" label related with the "email" field* by *And I should see a form field labelled with "Email:"*

Features/Stories/Fieldtypes/AuthorsFieldType.feature
```diff
@@ -0,0 +1,175 @@
...
+    @javascript
+    Scenario: A Content of a Content Type that has an author fieldtype must have a "Remove this author" and a "Add another author" buttons
+        Given a Content Type with an "authors" Field exists
+        When I create a content of this type
+        Then I should see a "Remove this author" button
+        And I should see an "Add another author" button
+
+    @javascript
+    Scenario: When editing a Content the label of an author field must have the same name than field type from the respective Content Type
+        Given a Content Type with an "authors" with "Name" "Participants" exists
```
dpobel: s/"Name"/field definition name/

Features/Stories/Fieldtypes/CheckboxFieldType.feature
```diff
@@ -0,0 +1,79 @@
...
+        Then the "Do you agree?" field should be marked as mandatory
+
+    ##
+    # Creating Content using a Content Type that has a Checkbox Field Type
+    ##
+    @javascript
+    Scenario: Publishing a valid checkbox Field works
+        Given a Content Type with a "checkbox" Field exists
+        When I create a content of this type
+        And I check the Field Value
```
dpobel: "I check the Field Value" ? I would rather write *I check the "checkbox" field*

Features/Stories/Fieldtypes/CheckboxFieldType.feature
```diff
@@ -0,0 +1,79 @@
...
+    Scenario: Updating a Checkbox field using a valid checkbox Field works
+        Given a Content Type with a "checkbox" Field exists
+        And a Content of this type exists with "checkbox" Field Value checked
+        When I edit this content
+        And I uncheck the Field Value
+        And I publish the content
+        Then the Content is successfully published
+
+    ##
+    # Delete Content using a Content Type that has a Checkbox Field Type
```
dpobel: I would not add such scenario (here and in the others field types). Removing a content is barely related to its fields.

dpobel:
Why do you add so many scenarios for so different Field types in one PR. I can understand that Integer, Float and maybe Email are grouped together as feature wise, they are very close. But given the complexity of Author, BinaryFile or the Date and time related field types deserve their own pull requests, otherwise it's very complicated to review.

And I think we should focus our effort on **one** not too simple field type first (ie not Checkbox but maybe Email ou Integer for instance) to have a good base to derive it for the others.

pcardiga:
On the [this approach](https://github.com/ezsystems/PlatformUIBundle/pull/181), one pull request was made for each fieldtype. Being requested to drop this approach, then only one PR was made for everything. This was the reason to have so many fieldtypes in one PR.

dpobel:
@pcardiga yeah but the case of the others PR was different because we tried to do the scenario and the implementation at the same time but spread in different PR which was impossible to manage.
Now, that we are only focusing on the scenario, we should also only focus first on one field type to set the base of the topic and once this is done, we can work on several others field types in separated pull requests.

Look at the implementation PRs, this is what we are doing all the time when implementing the various field type views or any other feature, the golden rule is one feature per PR otherwise it's a mess.

pcardiga:
@dpobel ok, no problem on this. We'll drop this PR and create a new one **only** with integer scenarios, already with suggested updates. Then we'll go from there and only work on the others when the integer scenarios are merged. This new PR will refer the current one so that the suggestions made in here aren't lost, since some of them are transverse to all fieldtypes.

dpobel:
:+1:

miguelcleverti:
Closed in favor of #234

PR #213, opened 2015-03-26
EZP-24039: Impossible to uncheck a Checkbox field

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24039

# Description

As the title suggests, it's impossible to uncheck a Checkbox field. cb7dcad fixes this by making sure we correctly check the state of the checkbox input. While working on it, I also realized that the checkbox state was also wrongly set because it is based on the field definition default value instead of the actual field value. So the second commit fixes that as well.

# Tests

manual + unit tests


andrerom:
+1

yannickroger:
+1

PR #218, opened 2015-04-09
EZP-24063: As an editor, I want to be able to remove some contents from the universal discovery widget content selection

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24063

# Description

This patch adds the ability to remove a content from the universal discovery widget selection. When the editor wants to do that, he has to confirm the removal. This confirmation is done through a new side view called *confirmBox* which will then be reusable in different context.

Screencast: http://youtu.be/2Bmuz81AKAg

# Tasks

* [x] add the remove button and the corresponding behavior on the full selection display
* [x] handle the "Choose this content" state after using the remove button
* [x] add the confirm box

# Test

unit tests + manual test

### DIFF
--- a/Resources/config/css.yml
+++ b/Resources/config/css.yml
@@ -10,6 +10,7 @@ system:
             # layout stylesheets
                 - '@eZPlatformUIBundle/Resources/public/css/layout.css'
                 - '@eZPlatformUIBundle/Resources/public/css/views/navigationhub.css'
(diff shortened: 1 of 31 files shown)


dpobel:
ready to be reviewed
ping @StephaneDiot @yannickroger @andrerom

yannickroger:
+1

andrerom:
+1

PR #220, opened 2015-04-24
EZP-24298: Display an iframe in the studio and studio+ description

yannickroger:
Link: https://jira.ez.no/browse/EZP-24298

## Description
Replace static content by an iframe pointing to a dedicated page on ez.no for the studio and studio+ description.

## Tests
Manual and automated

## TODO
- [x] Update iframe size when window size channge
- [x] Fix unit test
- [x] Add unit tests
- [x] Fix newly added unit test (dummy-iframe url) -> had to update nodejs on my local env
- [x] Add behavior to studio+
- [x] Rebase cleanup commit if needed


Resources/public/css/theme/views/studiopresentation.css
```diff
@@ -4,7 +4,10 @@
  */
 
 .ez-view-studiopresentationview {
-    /* for test purpose */
-    padding: 1em;
+    padding: 0px;
     background: #fff;
 }
+
+.ez-view-studiopresentationview iframe{
```
dpobel: usually we avoid styling based on the tag name, so you should add a class on the iframe (.ez-studiopresentation-content for instance) and use it here.
CS nitpick: space before `{`
nitpick 2: you don't need to add the unit when setting something to 0.

Resources/public/js/extensions/ez-height-fit.js
```diff
@@ -0,0 +1,44 @@
...
+    "use strict";
+    /**
+     * The height fit extension
+     *
+     * @module ez-height-fit
+     */
+    Y.namespace('eZ');
+
+    /**
+     * Views extension providing the concept of expanded state. When `expanded`,
```
dpobel: needs to be updated

Resources/public/js/extensions/ez-height-fit.js
```diff
@@ -0,0 +1,44 @@
...
+     * view accordingly.
+     *
+     * @namespace eZ
+     * @class HeightFit
+     * @extensionfor Y.View
+     */
+    Y.eZ.HeightFit = Y.Base.create('heightFitExtension', Y.View, [], {
+        initializer: function () {
+            this.after('activeChange', function () {
+                this._uiSetHeight();
```
dpobel: you should call `_uiSetHeight` only when active becomes true.

Resources/public/js/views/ez-studiopresentationview.js
```diff
@@ -1,3 +1,4 @@
+
```
dpobel: not the most useful change ever :p

Resources/public/templates/studiopresentation.hbt
```diff
@@ -1,22 +1 @@
...
-
-<p>
-Quisque ut tincidunt elit. Pellentesque nulla felis, vestibulum non eros vitae, lacinia luctus odio.
-    Etiam dapibus metus eget bibendum mattis. Donec lectus risus, hendrerit non condimentum et,
-    venenatis nec neque. Etiam sed varius arcu. Nulla ullamcorper luctus augue, blandit tempor neque
-    lobortis sit amet. Mauris ac eleifend augue. Etiam condimentum auctor ex, nec vehicula nunc aliquam
-    et. Quisque sagittis sed mi maximus rutrum. Proin eu enim non neque viverra iaculis. In hac habitasse
-    platea dictumst. Aenean euismod nunc urna, id iaculis nisl gravida vel.
-</p>
+<iframe class="studio-presentation-iframe" src="about:blank" width="100%" height="100%"></iframe>
```
dpobel: please always prefix the class names with `ez-`. Usually, we also avoid the tag name in the class name. And setting the size of the iframe like that is very HTML4ish, please use the CSS for that.

Tests/js/views/ez-studiopresentationview.html
```diff
@@ -27,13 +27,17 @@
...
-                requires: ['ez-templatebasedview'],
+                requires: ['ez-templatebasedview', 'ez-height-fit'],
                 fullpath: "../../../Resources/public/js/views/ez-studiopresentationview.js"
             },
             "ez-templatebasedview": {
                 requires: ['ez-view', 'handlebars', 'template'],
                 fullpath: "../../../Resources/public/js/views/ez-templatebasedview.js"
             },
+            "ez-height-fit": {
+                requires: ['ez-view'],
```
dpobel: AFAICS, ez-height-fit actually depends on the module `view` not `ez-view`. I think you also need to add `node-screen` (because of the usage of `getY`) and `node-style` (because of the usage of `setStyle`)

Resources/public/css/theme/views/studiopresentation.css
```diff
@@ -4,7 +4,10 @@
  */
 
 .ez-view-studiopresentationview {
-    /* for test purpose */
-    padding: 1em;
+    padding: 0px;
```
dpobel: not sure you really need that but if it's the case, this should go in the layout stylesheet not in a theme one.

Resources/config/yui.yml
```diff
@@ -653,6 +653,9 @@ system:
                 ez-expandable:
                     requires: ['view']
                     path: %ez_platformui.public_dir%/js/extensions/ez-expandable.js
+                ez-height-fit:
+                    requires: ['view']
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/220/files#r29037735 (node-screen and node-style)

Resources/public/js/views/ez-studiopresentationview.js
yannickroger: I think we should ping the whole team and make sure :) (I fixed it)

Resources/public/css/theme/views/studiopresentation.css
```diff
@@ -4,7 +4,11 @@
...
 
 .ez-view-studiopresentationview {
-    /* for test purpose */
-    padding: 1em;
     background: #fff;
 }
+
+.ez-view-studiopresentationview .ez-studiopresentation-content {
+    border: 0;
+    width: 100%;
```
StephaneDiot: Shouldn't it be in the corresponding layout stylesheet ?

Resources/public/js/extensions/ez-height-fit.js
```diff
@@ -0,0 +1,45 @@
...
+    "use strict";
+    /**
+     * The height fit extension
+     *
+     * @module ez-height-fit
+     */
+    Y.namespace('eZ');
+
+    /**
+     * Views extension providing the concent of height fit . When, `fitted to height`, the view
```
dpobel: *concent* ?

Resources/public/js/extensions/ez-height-fit.js
```diff
@@ -0,0 +1,45 @@
...
+     * container scales to match the window's size new dimension.
```
dpobel: it's more *viewport's height* instead of *window's size*

Resources/public/js/views/ez-studiopresentationview.js
```diff
@@ -19,7 +19,14 @@ YUI.add('ez-studiopresentationview', function (Y) {
      * @constructor
      * @extends eZ.TemplateBasedView
      */
-    Y.eZ.StudioPresentationView = Y.Base.create('studioPresentationView', Y.eZ.TemplateBasedView, [], {
+    Y.eZ.StudioPresentationView = Y.Base.create('studioPresentationView', Y.eZ.TemplateBasedView, [Y.eZ.HeightFit], {
+
+        initializer: function () {
+            this.after('activeChange', function () {
+                this.get('container').one('.ez-studiopresentation-content').set('src', this.get('iframeSource'));
```
dpobel: I would put that in a dedicated method call `_setIFrameSource` so that you can write:
```js
this.after('activeChange', this._setIframeSource);
```
which is way more readable.

Tests/js/views/assets/ez-studiopresentationview-tests.js
```diff
@@ -9,7 +9,9 @@ YUI.add('ez-studiopresentationview-tests', function (Y) {
         name: "eZ Studio Presentation View test",
 
         setUp: function () {
-            this.view = new Y.eZ.StudioPresentationView();
+            this.view = new Y.eZ.StudioPresentationView({container: '.container'});
+            this.dummyIframeFile = 'assets/dummy-iframe.hml';
```
dpobel: most likely, the added HTML is not reachable with this URL

Tests/js/views/assets/ez-studiopresentationview-tests.js
```diff
@@ -28,6 +30,31 @@ YUI.add('ez-studiopresentationview-tests', function (Y) {
...
+                container.getStyle('height'),
+                "The height of the container should have been adapted to the windows size"
+            );
+        },
+
+        "Test iframe link": function () {
+            var iframeSource = this.view.get('container').one('.ez-studiopresentation-content').get('src');
+            this.view.render();
+
+            Y.Assert.isTrue(
```
dpobel: if this test is passing, something is wrong, the iframe src attribute is supposed to be changed on the `activeChange` event and the `active` attribute is not changed anywhere here.

dpobel: ok get it, you fetch the iframeSource before the view is rendered and since `tearDown` does not reset the container, you get the iframeSource of the previous test...

Resources/public/css/views/studiopresentation.css
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+.ez-view-studiopresentationview .ez-studiopresentation-content {
+    border: 0;
```
dpobel: the border rule should go to the theme stylesheet

Resources/public/js/views/ez-studiopresentationview.js
```diff
@@ -28,7 +33,29 @@ YUI.add('ez-studiopresentationview', function (Y) {
          */
         render: function () {
             this.get('container').setHTML(this.template());
+            this._attachedViewEvents.push(Y.on("windowresize", Y.bind(this._uiSetHeight, this)));
+
             return this;
+        },
+
+        _setIFrameSource: function () {
```
dpobel: missing doc bloc

Tests/js/views/assets/ez-studiopresentationview-tests.js
```diff
@@ -28,6 +30,32 @@ YUI.add('ez-studiopresentationview-tests', function (Y) {
...
+                container.get('winHeight') - container.getY() + "px",
+                container.getStyle('height'),
+                "The height of the container should have been adapted to the windows size"
+            );
+        },
+
+        "Test iframe link": function () {
+            this.view.render();
+            this.view.set('active', true);
+            var iframeSource = this.view.get('container').one('.ez-studiopresentation-content').get('src');
```
dpobel: CS: the variable declaration always comes first.

Resources/public/js/extensions/ez-height-fit.js
dpobel: you forgot to fix *concent* :)

dpobel:
the theme stylesheet is not added in css.yml, so it's not included in the page and you forgot to fix https://github.com/ezsystems/PlatformUIBundle/pull/220/files#r29148270 (*concent*)

yannickroger:
the theme stylesheet was added in my previous PR: https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/config/css.yml#L89

yannickroger:

dpobel:
+1

Resources/public/templates/studiopluspresentation.hbt
```diff
@@ -1,22 +1 @@
...
-
-<p>
-Quisque ut tincidunt elit. Pellentesque nulla felis, vestibulum non eros vitae, lacinia luctus odio.
-    Etiam dapibus metus eget bibendum mattis. Donec lectus risus, hendrerit non condimentum et,
-    venenatis nec neque. Etiam sed varius arcu. Nulla ullamcorper luctus augue, blandit tempor neque
-    lobortis sit amet. Mauris ac eleifend augue. Etiam condimentum auctor ex, nec vehicula nunc aliquam
-    et. Quisque sagittis sed mi maximus rutrum. Proin eu enim non neque viverra iaculis. In hac habitasse
-    platea dictumst. Aenean euismod nunc urna, id iaculis nisl gravida vel.
-</p>
+<iframe class="ez-studiopluspresentation-content" src="about:blank"></iframe>
```
andrerom: cs: nw

rolandbenedetti: not sure yet what will be in there but the URL are defined
URLs for iFrame:
http://ez.no/in-product-studio-teaser
http://ez.no/in-product-studioplus-teaser
for now they just redirect somewhere else.

yannickroger: Url is added by `Resources/public/js/views/ez-studiopluspresentationview.js`
The goal is to load the content of the iframe after the application is loaded to respond much faster to the user's click.

dpobel: @andrerom we can add the `seamless` attribute but no browser is supporting it http://caniuse.com/#search=seamless

PR #221, opened 2015-04-27
Fix EZP-20347: Implement an API to get the readable name of a field type from its identifier

lolautruche:
> https://jira.ez.no/browse/EZP-20347

Idea is to use Translation component for this.
Note that YAML format is used on purpose, in order to differenciate from a "real" translation file.

Example to get human readable name for `ezstring`:

```php
/** \Symfony\Component\Translation\TranslatorInterface $translator */
echo $translator->trans('ezstring.name', [], 'fieldtypes');
```

### DIFF
--- a/Resources/translations/fieldtypes.en.yml
+++ b/Resources/translations/fieldtypes.en.yml
@@ -1,0 +1,27 @@
+ezauthor: { name: "Authors" }
+ezbinaryfile: { name: "File" }
+ezboolean: { name: "Checkbox" }
+ezcountry: { name: "Country" }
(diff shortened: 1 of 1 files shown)


lolautruche:
Review ping @dpobel @yannickroger @bdunogier

dpobel:
Binary file is missing :p
+1

lolautruche:
Fixed, thanks :smile:

yannickroger:
+1

andrerom:
shouldn't we prefix the namespace a bit more for our use?
ref `fieldtypes`

PR #223, opened 2015-04-28
Fix EZP-24126: Numeric fields with value 0 are considered empty

yannickroger:
Link: https://jira.ez.no/browse/EZP-24126

## Description
Numerics fields (Integer and Floats) were using the default `_isFieldEmpty` method that considers 0 as empty. A custom one has been implemented for those fields.

## Tests
Manual and unit test

## TODO
- [x] Add floats ?
- [x] Rename issue and commits
- [x] Rebase commits

### DIFF
--- a/Resources/public/js/views/fields/ez-float-view.js
+++ b/Resources/public/js/views/fields/ez-float-view.js
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-float-view', function (Y) {
+    "use strict";
+    /**
+     * Provides the Float field view
+     *
+     * @module ez-float-view
+     */
+    Y.namespace('eZ');
+
+    /**
+     * The Float field view
+     *
+     * @namespace eZ
+     * @class FloatView
+     * @constructor
+     * @extends eZ.FieldView
+     */
+    Y.eZ.FloatView = Y.Base.create('floatView', Y.eZ.FieldView, [], {
+        _isFieldEmpty: function () {
+            return (this.get('field').fieldValue === null);
+        },
+
+        /**
+         * Overrides the name to use the generic field view template
+         *
+         * @method _getName
+         * @protected
+         * @return String
+         */
+        _getName: function () {
+            return Y.eZ.FieldView.NAME;
+        },
+    });
+
+    Y.eZ.FieldView.registerFieldView('ezfloat', Y.eZ.FloatView);
+});
--- a/Resources/public/js/views/fields/ez-integer-view.js
+++ b/Resources/public/js/views/fields/ez-integer-view.js
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-integer-view', function (Y) {
+    "use strict";
+    /**
+     * Provides the Integer field view
+     *
+     * @module ez-integer-view
+     */
+    Y.namespace('eZ');
+
+    /**
+     * The Integer field view
+     *
+     * @namespace eZ
+     * @class IntegerView
+     * @constructor
+     * @extends eZ.FieldView
+     */
+    Y.eZ.IntegerView = Y.Base.create('integerView', Y.eZ.FieldView, [], {
+        _isFieldEmpty: function () {
+            return (this.get('field').fieldValue === null);
+        },
+
+        /**
+         * Overrides the name to use the generic field view template
+         *
+         * @method _getName
+         * @protected
+         * @return String
+         */
+        _getName: function () {
+            return Y.eZ.FieldView.NAME;
+        },
+    });
+
+    Y.eZ.FieldView.registerFieldView('ezinteger', Y.eZ.IntegerView);
+});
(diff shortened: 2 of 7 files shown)


yannickroger:
Ready for reviews. Ping @dpobel and @StephaneDiot

Resources/public/js/views/fields/ez-float-view.js
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * The Float field view
+     *
+     * @namespace eZ
+     * @class FloatView
+     * @constructor
+     * @extends eZ.FieldView
+     */
+    Y.eZ.FloatView = Y.Base.create('floatView', Y.eZ.FieldView, [], {
+        _isFieldEmpty: function () {
```
StephaneDiot: missing doc ?

Resources/public/js/views/fields/ez-integer-view.js
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * The Integer field view
+     *
+     * @namespace eZ
+     * @class IntegerView
+     * @constructor
+     * @extends eZ.FieldView
+     */
+    Y.eZ.IntegerView = Y.Base.create('integerView', Y.eZ.FieldView, [], {
```
StephaneDiot: same here

Resources/public/templates/fields/view/float.hbt
```diff
@@ -0,0 +1,6 @@
+<div class="ez-fieldview-row pure-g">
+    <div class="ez-fieldview-label pure-u">
+        <p class="ez-fieldview-name"><strong>{{ fieldDefinition.names.[eng-GB] }}</strong></p>
+    </div>
```
StephaneDiot: line is too long ;)  i know it s niptick

Resources/public/templates/fields/view/float.hbt
```diff
@@ -0,0 +1,6 @@
+<div class="ez-fieldview-row pure-g">
+    <div class="ez-fieldview-label pure-u">
+        <p class="ez-fieldview-name"><strong>{{ fieldDefinition.names.[eng-GB] }}</strong></p>
```
crevillo: is the ```eng-GB``` hardcoded for some reason? (asking from my ignorance about the whole thing)

Resources/public/templates/fields/view/integer.hbt
```diff
@@ -0,0 +1,6 @@
+<div class="ez-fieldview-row pure-g">
+    <div class="ez-fieldview-label pure-u">
+        <p class="ez-fieldview-name"><strong>{{ fieldDefinition.names.[eng-GB] }}</strong></p>
+    </div>
+    <div class="ez-fieldview-value pure-u"><div class="ez-fieldview-value-content">{{#if isEmpty}}This field is empty{{else}}{{ value }}{{/if}}</div></div>
```
StephaneDiot: same here

Resources/public/js/views/fields/ez-float-view.js
dpobel: not really, the documentation of the method in the parent class is valid.

Resources/public/templates/fields/view/float.hbt
dpobel: because we don't handle the i18n yet

Resources/public/templates/fields/view/float.hbt
dpobel: indeed,

```handlebars
<div class="ez-fieldview-value pure-u"><div class="ez-fieldview-value-content">
{{#if isEmpty}}
    This field is empty
{{else}}
    {{ value }}
{{/if}}
</div></div>
```
is much easier to read.
btw @StephaneDiot you should comment on the correct line :p

Resources/public/templates/fields/view/float.hbt
crevillo: ah, ok.

dpobel:
+1 after fixing the too long line issue reported by @StephaneDiot

yannickroger:
@StephaneDiot @dpobel the too long line issue is a copy paste from `field.hbt` do you want me to also fix this file in a separate commit ?

dpobel:
actually you should not duplicate the template but make sure the FloatView and IntegerView use the default one. To do that, you should defined the `_getName` in FloatView and IntegerView so that those views pick the FieldView template and not a custom one.

yannickroger:

Tests/js/views/fields/assets/ez-float-view-tests.js
```diff
@@ -0,0 +1,69 @@
...
+                    {fieldValue: 0}, false,
+                    "The field should not be seen as empty"
+                );
+            },
+
+            "Test template override": function (){
+                var container = this.view.get('container');
+                this.view.render();
+
+                Y.Assert.areSame(container.one("#marcelobielsa").getHTML(), "Bielsa No Se Va");
```
dpobel: not sure *Bielsa No Se Va* is a good failure message, I would prefer something like: *The default fieldview template should have been used*

dpobel: a better assertion would be:
```js
Y.Assert.isObject(
    container.one('#marcelobielsa'),
    "The fieldview template should have been used"
);
```
the important thing is to always add a failure message.

Tests/js/views/fields/assets/ez-integer-view-tests.js
```diff
@@ -0,0 +1,69 @@
...
+                    {fieldValue: 0}, false,
+                    "The field should not be seen as empty"
+                );
+            },
+
+            "Test template override": function (){
+                var container = this.view.get('container');
+                this.view.render();
+
+                Y.Assert.areSame(container.one("#marcelobielsa").getHTML(), "Bielsa No Se Va");
```
dpobel: same here

yannickroger:
@dpobel fixed. Thanks

PR #224, opened 2015-04-28
EZP-23992: As an editor, I want a notification system for asynchronous operations

dpobel:
JIRA: https://jira.ez.no/browse/EZP-23992

# Description

This patch adds a notification system to the Platform UI Application. With this patch, any component in the application can launch a notification with a custom event and this notification will be displayed in the notification hub just below the navigation hub. As the navigation hub menu, the notifications stay fixed on the screen when the user scrolls down. The implementation of this feature makes me change the way the navigation hub and the *main view zone* are positioned. As of this patch, the navigation hub is absolutely positioned and a top margin on the *main view zone* is adjusted so that nothing overlaps, this top margin is also changed when a notification is displayed.

For now, a notification is a very simple object with a text, a state and an identifier. This identifier allows to update the properties of an already existing notification.

**Screencast: http://youtu.be/fdoRhYnykz4**

In this screencast, the dashboard randomly launches some notifications and randomly updates their state and and text. The notification system is also roughly used to provide some feedback to the user when a draft is saved. This feature will be properly implemented in [EZP-23512](https://jira.ez.no/browse/EZP-23512).

## Tasks

* [x] Create the Notification Model and NotificationList ModelList
* [x] Add the NotificationView to display one Notification Model
* [x] Add the NotificationHubView and the corresponding ViewService to handle the list of notification
* [x] Add a plugin and configure the application to handle the notification
* [x] Make sure the position of the notification hub view does not overlap any part of the application
* [x] Improve/Add some transitions when adding/removing a notification
* [ ] ~~Auto hide the notification~~ postponed to [EZP-24327](https://jira.ez.no/browse/EZP-24327)

# Tests

manual tests + unit tests


dpobel:
ping @StephaneDiot

Resources/public/css/theme/views/notificationhub.css
```diff
@@ -0,0 +1,9 @@
+/**
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+.ez-view-notificationhubview {
+    background: #fff; /* TODO remove */
```
dpobel: self note: remove

StephaneDiot:
+1 FWIW

PR #225, opened 2015-04-29
EZP-24293: Rename the navigation zone "Studio Plus" to "Studio+"

yannickroger:
Link: https://jira.ez.no/browse/EZP-24293

## Description
Just changed the title

## Test
Manual tests


andrerom:
+1, but wasn't it other places also? or is the value reused?

yannickroger:
@andrerom it only had to be changed in the navigation zone. Other occurrences are either in the code (routing, filenames, etc.) or will be removed by #220

dpobel:
+1

PR #226, opened 2015-05-11
EZP-24327: As an editor, I want the notifications to automatically disappear

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24327

# Description

This patch adds the ability for a component triggering a notification to decide the duration after which the notification should disappear. While working on that, I also noticed that after some time, an offset in the notifications position appears because the height of the notifications was sometimes incorrectly computed due to the CSS transtions. So 6e9db7a also fixes that.

**Screencast: https://www.youtube.com/watch?v=5G7KKNgW1OQ**

# Tests

manual tests + unit tests

### DIFF
--- a/Resources/public/js/views/ez-notificationview.js
+++ b/Resources/public/js/views/ez-notificationview.js
@@ -23,26 +23,35 @@ YUI.add('ez-notificationview', function (Y) {
     Y.eZ.NotificationView = Y.Base.create('notificationView', Y.eZ.TemplateBasedView, [], {
         events: {
             '.ez-notification-close': {
-                'tap': '_closeNotification',
+                'tap': function (e) {
+                    e.preventDefault();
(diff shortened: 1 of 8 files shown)


dpobel:
ping @StephaneDiot

Resources/public/js/views/ez-notificationview.js
```diff
@@ -23,26 +23,35 @@ YUI.add('ez-notificationview', function (Y) {
...
         _closeNotification: function (e) {
-            e.preventDefault();
             this.get('notification').destroy();
         },
 
         initializer: function () {
             var notification = this.get('notification');
 
+            /**
+             * Stores the atuo hide timer returned by Y.later.
```
StephaneDiot: typo : "auto"

StephaneDiot:
+1

PR #228, opened 2015-05-12
EZP-24335: Notifications displayed when assigning section to contents

mhyndle:
JIRA: https://jira.ez.no/browse/EZP-24335

# Description
Adding notification when assigning section to contents.

### DIFF
--- a/Resources/public/js/views/serverside/ez-sectionserversideview.js
+++ b/Resources/public/js/views/serverside/ez-sectionserversideview.js
@@ -48,6 +48,7 @@ YUI.add('ez-sectionserversideview', function (Y) {
                     multiple: true,
                     data: {
                         sectionId: button.getAttribute('data-section-rest-id'),
+                        sectionName: button.getAttribute('data-section-name'),
                         afterUpdateCallback: unsetLoading,
                     },
                 },
--- a/Resources/public/js/views/services/ez-sectionserversideviewservice.js
+++ b/Resources/public/js/views/services/ez-sectionserversideviewservice.js
@@ -39,6 +39,10 @@ YUI.add('ez-sectionserversideviewservice', function (Y) {
                 tasks = new Y.Parallel(),
                 contentService = this.get('capi').getContentService();
 
+            this._assignSectionNotificationStarted(
+                data.sectionId, data.sectionName, e.selection
+            );
+
             Y.Array.each(e.selection, function (struct) {
                 var update = contentService.newContentMetadataUpdateStruct();
 
@@ -48,7 +52,96 @@ YUI.add('ez-sectionserversideviewservice', function (Y) {
                     struct.content.get('id'), update, tasks.add()
                 );
             });
-            tasks.done(Y.bind(data.afterUpdateCallback, this));
+
+            tasks.done(
+                Y.bind(this._assignSectionCallback, this,
+                    data.sectionId, data.sectionName, e.selection,
+                    Y.bind(data.afterUpdateCallback, this)
+                )
+            );
+        },
+
+        /**
+         * Assign section callback trigerring notification and calling given callback
+         * 
+         * @method _assignSectionCallback
+         * @protected
+         * @param {String} sectionId the section id
+         * @param {String} sectionName the section name
+         * @param {Array} contents the array of contents to which section is being assigned to
+         * @param {Function} callback the callback to call when other tasks are done
+         */
+        _assignSectionCallback: function (sectionId, sectionName, contents, callback) {
+            this._assignSectionNotificationDone(sectionId, sectionName, contents);
+            callback();
+        },
+
+        /**
+         * Notification changed to *started* before assigning section to contents
+         * 
+         * @method _assignSectionNotificationStarted
+         * @protected
+         * @param {String} sectionId the section id
+         * @param {String} sectionName the section name
+         * @param {Array} contents the array of contents to which section is being assigned to
+         */
+        _assignSectionNotificationStarted: function (sectionId, sectionName, contents) {
+            var notificationIdentifier = this._getAssignSectionNotificationIdentifier(
+                    'assign-section', sectionId, contents
+                );
+
+            this.fire('notify', {
+                notification: {
+                    identifier: notificationIdentifier,
+                    text: 'Assigning the section "' + sectionName + '" to ' + contents.length + ' contents',
+                    state: 'started',
+                    timeout: 0
+                },
+            });
+        },
+
+        /**
+         * Notification changed to *done* after assigning section to contents
+         * 
+         * @method _assignSectionNotificationDone
+         * @protected
+         * @param {String} sectionId the section id
+         * @param {String} sectionName the section name
+         * @param {Array} contents the array of contents to which section is being assigned to
+         */
+        _assignSectionNotificationDone: function (sectionId, sectionName, contents) {
+            var notificationIdentifier = this._getAssignSectionNotificationIdentifier(
+                    'assign-section', sectionId, contents
+                );
+
+            this.fire('notify', {
+                notification: {
+                    identifier: notificationIdentifier,
+                    text: 'Section "' + sectionName + '" assigned to ' + contents.length + ' contents',
+                    state: 'done',
+                    timeout: 5
+                },
+            });
+        },
+
+        /**
+         * Generates identifier for notifications which is unique basing on
+         * section ID and IDs of contents which section is being assigned to.
+         * 
+         * @method _getAssignSectionNotificationIdentifier
+         * @protected
+         * @param {String} action custom string describing action which is being taken
+         * @param {String} sectionId the section id
+         * @param {Array} contents the array of contents to which section is being assigned to
+         * @return {String} unique notification identifier based on passed parameters
+         */
+        _getAssignSectionNotificationIdentifier: function (action, sectionId, contents) {
+            var contentIds = [];
+
+            Y.Array.each(contents, function (struct) {
+                contentIds.push(struct.content.get('id'));
+            });
+            return action + '-' + sectionId + '-' + contentIds.join('_');
         },
     });
 });
--- a/Tests/js/views/serverside/assets/ez-sectionserversideview-tests.js
+++ b/Tests/js/views/serverside/assets/ez-sectionserversideview-tests.js
@@ -51,6 +51,11 @@ YUI.add('ez-sectionserversideview-tests', function (Y) {
                         e.config.data.sectionId,
                         "The section id should be available in the config data"
                     );
+                    Assert.areEqual(
+                        button.getAttribute('data-section-name'),
+                        e.config.data.sectionName,
+                        "The section name should be available in the config data"
+                    );
                     Assert.areSame(
                         e.config.cancelDiscoverHandler,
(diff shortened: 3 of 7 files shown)


Resources/public/js/views/serverside/ez-sectionserversideview.js
```diff
@@ -48,11 +49,32 @@ YUI.add('ez-sectionserversideview', function (Y) {
...
+        /**
+         * Assign section to contents callback. It sets notifications to done
+         * state and unsets loading state from assign button.
+         * 
+         * @method _assignSectionCallback
+         * @protected
+         * @param {Y.Node} button
+         */
+        _assignSectionCallback: function(button){
+            var unsetLoading = Y.bind(this._uiUnsetAssignSectionLoading, this, button);
```
dpobel: you don't need to do that here, you can directly call `_uiUnsetAssignSectionLoading` with:

```js
this._uiUnsetAssignSectionLoading(button);
```

Resources/public/js/views/serverside/ez-sectionserversideview.js
```diff
@@ -48,11 +49,32 @@ YUI.add('ez-sectionserversideview', function (Y) {
...
+        
+        /**
+         * Assign section to contents callback. It sets notifications to done
+         * state and unsets loading state from assign button.
+         * 
+         * @method _assignSectionCallback
+         * @protected
+         * @param {Y.Node} button
+         */
+        _assignSectionCallback: function(button){
```
dpobel: CS: there's always a space after `function` and always a space before `{`

Resources/public/js/views/services/ez-sectionserversideviewservice.js
```diff
@@ -39,6 +39,14 @@ YUI.add('ez-sectionserversideviewservice', function (Y) {
                 tasks = new Y.Parallel(),
                 contentService = this.get('capi').getContentService();
 
+            this.fire('notify', {
+                notification: {
+                    identifier: 'assign-section',
```
dpobel: you should add a method to generate the identifier so that you don't repeat it (I would even create a method to fire the event as well) and in addition, the identifier should be more *unique* so that if the user assign a section to others contents, he gets another notification bar. So you should include the section id and the content ids in it.

dpobel: I've just realized that the two notifiy events are fired in 2 different components. I think they should be both fired from the view service.

mhyndle:
Changes included in: mhyndle/PlatformUIBundle@05b94ac : 
* notifications about assigning section to contents placed in separate functions
* function "generating" unique identifier for notifications when assigning section to contents basing on section ID and contents
* place all functions related to notifications in view service instead of view

Resources/public/js/views/serverside/ez-sectionserversideview.js
```diff
@@ -53,7 +53,7 @@ YUI.add('ez-sectionserversideview', function (Y) {
                 },
             });
         },
-
+
```
andrerom: cs ;)

Resources/public/js/views/services/ez-sectionserversideviewservice.js
```diff
@@ -48,7 +50,81 @@ YUI.add('ez-sectionserversideviewservice', function (Y) {
...
+         * @protected
+         * @param {String} sectionId the section id
+         * @param {Array} contents the array of contents to which section is being assgined to
+         */
+        _assignSectionNotificationStarted: function (sectionId, contents) {
+            var notificationIdentifier = this._getAssignSectionNotificationIdentifier('assign-section', sectionId, contents);
+            this.fire('notify', {
+                notification: {
+                    identifier: notificationIdentifier,
+                    text: 'Assigning section to contents',
```
dpobel: it would be nice to include the section name and the number of affected contents in the text of the notifications (so here and in the 'done' notification), so that the user sees something like 'Assigning the section "Standard" to 4 contents'

Resources/public/js/views/services/ez-sectionserversideviewservice.js
```diff
@@ -48,7 +50,81 @@ YUI.add('ez-sectionserversideviewservice', function (Y) {
...
+         * Notification changed to *done* after assigning section to contents
+         * 
+         * @method _assignSectionNotificationDone
+         * @protected
+         * @param {String} sectionId the section id
+         * @param {Array} contents the array of contents to which section is being assgined to
+         */
+        _assignSectionNotificationDone: function (sectionId, contents) {
+            var notificationIdentifier = this._getAssignSectionNotificationIdentifier('assign-section', sectionId, contents);
+            this.fire('notify', {
```
dpobel: nitpick CS: please add an empty line between the variable declarations and the actual code

Resources/public/js/views/services/ez-sectionserversideviewservice.js
```diff
@@ -48,7 +50,81 @@ YUI.add('ez-sectionserversideviewservice', function (Y) {
...
+         * @protected
+         * @param {String} action custom string describing action which is being taken
+         * @param {String} sectionId the section id
+         * @param {Array} contents the array of contents to which section is being assgined to
+         * @return {String} unique notification identifier based on passed parameters
+         */
+        _getAssignSectionNotificationIdentifier: function (action, sectionId, contents) {
+            var contentIds = [];
+            Y.Array.each(contents, function (struct) {
+                contentIds.push( struct.content.get('id') );
```
dpobel: CS: no space after `(` nor before `)`

Resources/public/js/views/services/ez-sectionserversideviewservice.js
```diff
@@ -48,7 +50,81 @@ YUI.add('ez-sectionserversideviewservice', function (Y) {
...
+            });
+            return action+'-'+sectionId+'-'+contentIds.join('_');
```
dpobel: CS: spaces before and after `+`

dpobel:
Nice! Other than the inline comments, this patch is breaking the unit tests and you need to add new tests on the notifications.

mhyndle:
I think that last thing is left - include section's name in notification message.

dpobel:
To include the section name you have different options but I think the easiest is to add it in the `data` object send when firing the `notify` event in https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/serverside/ez-sectionserversideview.js#L49. And to easily retrieve it in the view, you should probably add a data attribute (like for the section id) in https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/views/Section/list.html.twig#L20 and https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/views/Section/view.html.twig#L68

mhyndle:

Resources/public/js/views/services/ez-sectionserversideviewservice.js
```diff
@@ -48,7 +50,89 @@ YUI.add('ez-sectionserversideviewservice', function (Y) {
...
+         * @param {String} sectionName the section name
+         * @param {Array} contents the array of contents to which section is being assgined to
+         */
+        _assignSectionNotificationDone: function (sectionId, sectionName, contents) {
+            var notificationIdentifier = this._getAssignSectionNotificationIdentifier('assign-section', sectionId, contents);
+
+            this.fire('notify', {
+                notification: {
+                    identifier: notificationIdentifier,
+                    text: 'Section "' + sectionName + '" assigned to ' + contents.length + ' content' + ((contents.length>1) ? 's' : ''),
```
dpobel: you don't need to handle the plural here, we'll need to dig in this topic when we'll implement i18n/l10n, so the string can directly be: `'Section "' + sectionName + '" assigned to ' + contents.length + ' contents'`

dpobel:
Almost there! You just need to adapt the tests of eZ.SectionServerSideView (https://github.com/ezsystems/PlatformUIBundle/blob/master/Tests/js/views/serverside/assets/ez-sectionserversideview-tests.js#L30) to make sure the name of the section is retrieved and added in the `data` object. After that and the inline comment on the plural handling, that's good. Good job :+1:

mhyndle:
Added fixes according to you note @dpobel

andrerom:
:tada:

Tests/js/views/serverside/assets/ez-sectionserversideview-tests.js
```diff
@@ -51,6 +51,11 @@ YUI.add('ez-sectionserversideview-tests', function (Y) {
                         e.config.data.sectionId,
                         "The section id should be available in the config data"
                     );
+                    Assert.areEqual(
+                        button.getAttribute('data-section-name'),
```
dpobel: to be relevant, you should add a `data-section-name` attribute on the button generated by the template in https://github.com/ezsystems/PlatformUIBundle/blob/master/Tests/js/views/serverside/ez-sectionserversideview.html#L9

mhyndle:
ping @dpobel

dpobel:
+1
@yannickroger or @StephaneDiot can you review this PR as well ?

Resources/public/js/views/services/ez-sectionserversideviewservice.js
```diff
@@ -48,7 +50,89 @@ YUI.add('ez-sectionserversideviewservice', function (Y) {
                     struct.content.get('id'), update, tasks.add()
                 );
             });
-            tasks.done(Y.bind(data.afterUpdateCallback, this));
+            
+            tasks.done(
+                Y.bind(this._assignSectionCallback, this, data.sectionId, data.sectionName, e.selection, Y.bind(data.afterUpdateCallback, this))
```
yannickroger: nitpick: line is too long

Resources/public/js/views/services/ez-sectionserversideviewservice.js
```diff
@@ -48,7 +50,89 @@ YUI.add('ez-sectionserversideviewservice', function (Y) {
...
+        },
+
+        /**
+         * Assign section callback trigerring notification and calling given callback
+         * 
+         * @method _assignSectionCallback
+         * @protected
+         * @param {String} sectionId the section id
+         * @param {String} sectionName the section name
+         * @param {Array} contents the array of contents to which section is being assgined to
```
yannickroger: typo unless "ass gined" is something I don't know about :)

Resources/public/js/views/services/ez-sectionserversideviewservice.js
```diff
@@ -48,7 +50,89 @@ YUI.add('ez-sectionserversideviewservice', function (Y) {
...
+
+        /**
+         * Generates identifier for notifications which is unique basing on
+         * section ID and IDs of contents which section is being assigned to.
+         * 
+         * @method _getAssignSectionNotificationIdentifier
+         * @protected
+         * @param {String} action custom string describing action which is being taken
+         * @param {String} sectionId the section id
+         * @param {Array} contents the array of contents to which section is being assgined to
```
yannickroger: typo++

yannickroger:
except for nitpick +1

mhyndle:
RE @yannickroger : Fixed typos and shortened lines (in correct way I hope).

StephaneDiot:
+1

PR #229, opened 2015-05-18
EZP-24187: As an editor, I want to be able to move a content

StephaneDiot:
Jira: https://jira.ez.no/browse/EZP-24187

## Description

The goal of the story is to be able to move a content and its subtree to a new location. This will be made by a button from the action bar view. Notifications are sent when the move operation begin, when the content is successfully moved, and when an error occurs.

## Tasks

* [x] Basic implementation
* [x] Tests
* [x] CSS


crevillo:
side question: when in legacy we had a kinda limit of nodes that could be moved and a message in the screen when you selected more than this limit of nodes. Should this be taken in account here or probably another story?

dpobel:
I think in legacy the copy and remove operations were indeed "limited" on the number of contents they can affect but AFAIK the move did not have this kind of limitation. And indeed, we need to keep that in mind for future stories.

crevillo:
@dpobel you're right sorry. there is not that limitation in move.

StephaneDiot:
ready to be reviewed : ping @dpobel @yannickroger

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,78 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+         * @protected
+         * @param {EventFacade} e
+         */
+        _moveContent: function (e) {
+            var app = this.get('app'),
+                initialActiveView = app.get('activeView'),
+                parentLocationId = e.selection.location.get('id'),
+                locationId = this.get('location').get('id'),
+                that = this,
+                notificationSuffixe = parentLocationId + locationId;
```
dpobel: instead of preparing a suffix (without a *e*), why don't you store here the identifier of the notification you'll use below ?

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,78 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+            that.fire('notify', {
+                notification: {
+                    text: 'The move operation of your content has begun',
+                    identifier: 'move-notification-' + notificationSuffixe,
+                    state: 'started',
+                    timeout: 5,
+                }
+            });
+
+            this.get('capi').getContentService().moveSubtree(locationId, parentLocationId, function (error, response) {
```
dpobel: you should add a `move` method to the Location model and use it here to write something like:
```js
this.get('location').move({api: this.get('capi')}, parentLocationId, function (error, response) {
   // ...
});
```

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,78 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+         * Fire 'notify' event
+         *
+         * @method _fireNotify
+         * @protected
+         * @param {String} text the text shown durring the notification
+         * @param {String} identifier the identifier of the notification
+         * @param {String} state the state of the notification
+         * @param {Integer} timeout the number of second, the notification will be shown
+         */
+        _fireNotify: function (text, identifier, state, timeout) {
```
dpobel: didn't you change the name yesterday ? `_fireNotification` or `_notify` would be a better name

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,78 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+         * @protected
+         * @param {EventFacade} e
+         */
+        _moveContent: function (e) {
+            var app = this.get('app'),
+                initialActiveView = app.get('activeView'),
+                parentLocationId = e.selection.location.get('id'),
+                locationId = this.get('location').get('id'),
+                that = this,
+                notificationIdentifier =  'move-notification-' + parentLocationId + locationId;
```
dpobel: you should add '-' between the `parentLocationId` and `locationId`

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -46,14 +46,31 @@ YUI.add('ez-locationmodel-tests', function (Y) {
...
+            this.contentServiceMock = new Y.Mock();
         },
 
         tearDown: function () {
             this.model.destroy();
             delete this.model;
-        }
+        },
+
+        "Should have a move method": function () {
```
dpobel: bad name, the location model should indeed have a move method but the important thing is that it moves the location... `"Should move the location"` would be better and you should put this test in its own test case dedicated to the move method.

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -46,14 +46,31 @@ YUI.add('ez-locationmodel-tests', function (Y) {
...
+            var callback = function () {},
+                parentLocationId = '4/5/6';
 
+            Y.Mock.expect(this.capiMock, {
+                method: 'getContentService',
+                returns: this.contentServiceMock
+            });
+            this.options = {api: this.capiMock};
```
dpobel: why putting the options as a property of the test case ? A local variable would be better suited.

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -46,14 +46,31 @@ YUI.add('ez-locationmodel-tests', function (Y) {
...
+            Y.Mock.expect(this.contentServiceMock, {
+                method: 'moveSubtree',
+                args: [this.locationId, parentLocationId, callback],
+            });
+            this.model.move(this.options, parentLocationId, callback);
```
dpobel: You should call the `Mock.verify` on the CAPI and content service mocks.

Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
```diff
@@ -430,6 +430,216 @@ YUI.add('ez-locationviewviewservice-tests', function (Y) {
...
+                method: 'get',
+                args: ['id'],
+                returns: this.locationId
+            });
+
+            Y.Mock.expect(this.locationMock, {
+                method: 'move',
+                args: [Y.Mock.Value.Object, this.parentLocationId, Y.Mock.Value.Function],
+                run: function (options, parentLocationId, callback) {
+                    options.api = that.capiMock;
```
dpobel: ??? Maybe you need to add an assertion here but for sure, you should not assign anything.

Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
```diff
@@ -430,6 +430,216 @@ YUI.add('ez-locationviewviewservice-tests', function (Y) {
...
+            });
+
+            this.service.on('contentDiscover', function (e) {
+                e.config.contentDiscoveredHandler.call(this, fakeEventFacade);
+            });
+
+            this.service.once('notify', function (e) {
+                notified = true;
+                that.activeView = new Y.View({location: 'newLocation/1/2/3/4'});
+                Y.Mock.expect(that.app, {
```
dpobel: it's a bit late to add expectations, usually expectations are added in the setUp or in the beginning of a test.

Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
```diff
@@ -430,6 +430,216 @@ YUI.add('ez-locationviewviewservice-tests', function (Y) {
...
+            this.service.fire('whatever:moveAction');
+        },
+    });
+
+    moveContentTest = new Y.Test.Case({
+        name: "eZ Location View View Service move content tests",
+
+        setUp: function () {
+            var that = this;
+            this.capiMock = new Y.Test.Mock();
```
dpobel: side note: you can write `Y.Mock` or you can even define the shortcut at the beginning of the test to get rid of `Y.`.

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,78 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+                parentLocationId = e.selection.location.get('id'),
+                locationId = this.get('location').get('id'),
+                that = this,
+                notificationIdentifier =  'move-notification-' + parentLocationId + '-' + locationId;
+
+            this._notify('The move operation of your content has begun', notificationIdentifier, 'started', 5);
+
+            this.get('location').move({api: this.get('capi')}, parentLocationId, function (error, response) {
+                if (error) {
+                    that._notify('An error occured while moving your content', notificationIdentifier, 'error', 5);
```
dpobel: usually error notification should not automatically disappear, so the timeout should 0 here

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,78 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+         */
+        _moveContent: function (e) {
+            var app = this.get('app'),
+                initialActiveView = app.get('activeView'),
+                parentLocationId = e.selection.location.get('id'),
+                locationId = this.get('location').get('id'),
+                that = this,
+                notificationIdentifier =  'move-notification-' + parentLocationId + '-' + locationId;
+
+            this._notify('The move operation of your content has begun', notificationIdentifier, 'started', 5);
```
dpobel: Should be something like: `"'" + contentName + "' is being moved under '" + parentContentName + "'"`

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,78 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+
+            this.get('location').move({api: this.get('capi')}, parentLocationId, function (error, response) {
+                if (error) {
+                    that._notify('An error occured while moving your content', notificationIdentifier, 'error', 5);
+                    return;
+                }
+
+                that._notify('Your content has been successfully moved', notificationIdentifier, 'done', 5);
```
dpobel: "'" + contentName + "' has been successfully moved under '" + parentContentName + "'"

dpobel:
besides the few inline comments, +1

Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
```diff
@@ -430,6 +430,219 @@ YUI.add('ez-locationviewviewservice-tests', function (Y) {
             Y.Mock.verify(app);
             Y.Mock.verify(contentMock);
         },
+
+        "Should launch the universal discovery widget when receiving an moveAction event": function () {
```
dpobel: I would put this test in the moveContentTest instead (yeah `eventTest` is badly named)

Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
```diff
@@ -430,6 +430,219 @@ YUI.add('ez-locationviewviewservice-tests', function (Y) {
...
+            this.service.on('contentDiscover', function (e) {
+                Y.Assert.isObject(e.config, "contentDiscover config should be an object");
```
dpobel: and same boolean flag issue as in https://github.com/ezsystems/PlatformUIBundle/pull/237/files#r31313336

dpobel:
+1 (once travis agrees)

PR #231, opened 2015-05-21
EZP-24385: As an editor, I want to be able to remove a content

mhyndle:
Jira: https://jira.ez.no/browse/EZP-24385

## Description
The goal of that story is to create button in the action bar view which allows sending content to trash. User should be asked to confirm sending content to trash. Notifications are sent when sending content to trash process is started, when the content is successfully sent to trash and when an error occurs. After sending content to trash app should navigate to location view of parent location.
## Tasks
- [x] Implementation
 - [x] confirmBox
 - [x] get parent location info (it is used for navigate to after successful sending to trash)
 - [x] notifications
 - [x] navigate to parent location after successful sending content to trash
- [x] Tests
- [x] CSS
 - [x] button icon
 - [x] color of text on button

### DIFF
--- a/Resources/public/js/models/ez-locationmodel.js
+++ b/Resources/public/js/models/ez-locationmodel.js
@@ -53,12 +53,17 @@ YUI.add('ez-locationmodel', function (Y) {
          * @param {Function} callback
          */
         trash: function (options, callback) {
-            var api = options.api,
+            var trashPath,
+                api = options.api,
                 location = this;
 
             api.getContentService().loadRoot(function (error, response) {
-                var trashPath = response.document.Root.trash._href;
+                if (error) {
+                    callback(error);
+                    return;
+                }
 
+                trashPath = response.document.Root.trash._href;
                 api.getContentService().moveSubtree(
                     location.get('id'), trashPath, callback
                 );
--- a/Resources/public/js/models/ez-locationmodel.js
+++ b/Resources/public/js/models/ez-locationmodel.js
@@ -42,6 +42,27 @@ YUI.add('ez-locationmodel', function (Y) {
             } else {
                 callback("Only read operation is supported at the moment");
             }
+        },
+
+        /**
+         * Moves location to trash. Callback is directly passed to ContentService.moveSubtree location
+         *
+         * @method trash
+         * @param {Object} options the options for moving to trash
+         * @param {Object} options.api (required) the JS REST client instance
+         * @param {Function} callback
+         */
+        trash: function (options, callback) {
+            var api = options.api,
+                location = this;
+
+            api.getContentService().loadRoot(function (error, response) {
+                var trashPath = response.document.Root.trash._href;
+
+                api.getContentService().moveSubtree(
+                    location.get('id'), trashPath, callback
+                );
+            });
         }
     }, {
         REST_STRUCT_ROOT: "Location",
--- a/Resources/public/js/views/services/ez-locationviewviewservice.js
+++ b/Resources/public/js/views/services/ez-locationviewviewservice.js
@@ -44,7 +44,8 @@ YUI.add('ez-locationviewviewservice', function (Y) {
         },
 
         /**
-         * _sendContentToTrashConfirmBox event handler on send to trash button
+         * `sendToTrashAction` event handler, 
+         * it asks confirmation to the user before sending the location to the trash.
          *
          * @method _sendContentToTrashConfirmBox
          * @protected
@@ -67,15 +68,15 @@ YUI.add('ez-locationviewviewservice', function (Y) {
          *
          * @method _sendToTrash
          * @protected
-         * @param {Object} content
          */
-        _sendToTrash: function (content) {
-            var options,
-                that = this,
+        _sendToTrash: function () {
+            var that = this,
                 location = this.get('location'),
                 locationId = location.get('id'),
                 path = this.get('path'),
-                contentName = content.get('name');
+                content = this.get('content'),
+                contentName = content.get('name'),
+                parentLocation = path[path.length - 1].location;
 
             this._notify(
                 'Sending "' + contentName + '" to Trash',
@@ -84,12 +85,7 @@ YUI.add('ez-locationviewviewservice', function (Y) {
                 0
             );
 
-            options = {
-                parentLocation: path[path.length - 1].location,
-                contentName: contentName
-            };
-
-            location.trash({api: this.get('capi')}, Y.bind(that._afterSendToTrashCallback, that, options));
+            location.trash({api: this.get('capi')}, Y.bind(that._afterSendToTrashCallback, that, parentLocation, contentName));
         },
 
         /**
@@ -97,15 +93,14 @@ YUI.add('ez-locationviewviewservice', function (Y) {
          *
          * @method _afterSendToTrashCallback
          * @protected
-         * @param {Object} options the options for sending to trash
+         * @param {eZ.Location} parentLocation the parent location to which app will navigate to
+         * @param {String} contentName the name of the content
          * @param {Boolean} error
          */
-        _afterSendToTrashCallback: function (options, error) {
+        _afterSendToTrashCallback: function (parentLocation, contentName, error) {
             var app = this.get('app'),
                 location = this.get('location'),
-                locationId = location.get('id'),
-                parentLocation = options.parentLocation,
-                contentName = options.contentName;
+                locationId = location.get('id');
 
             if (error) {
                 this._notify(
--- a/Resources/public/js/views/services/ez-locationviewviewservice.js
+++ b/Resources/public/js/views/services/ez-locationviewviewservice.js
@@ -61,6 +61,7 @@ YUI.add('ez-locationviewviewservice', function (Y) {
                 trashLocation = '/api/ezp/v2/content/trash',
                 contentName = e.content.get('name');
 
+            e.preventDefault();
             this._set('trackOutsideEvents', false);
             this.fire('confirmBoxOpen', {
                 config: {
--- a/Resources/public/js/views/services/ez-locationviewviewservice.js
+++ b/Resources/public/js/views/services/ez-locationviewviewservice.js
@@ -24,6 +24,7 @@ YUI.add('ez-locationviewviewservice', function (Y) {
     Y.eZ.LocationViewViewService = Y.Base.create('locationViewViewService', Y.eZ.ViewService, [], {
         initializer: function () {
             this.on('*:editAction', this._editContent);
+            this.on('*:sendToTrashAction', this._sendContentToTrash);
         },
 
         /**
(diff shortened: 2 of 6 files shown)


crevillo:
Questions: 
* As i asked for moving, should be warn the user if trying to send to trash a content (location) having lot of descedants
 * A question coming from the past, when we send Content to trash, do we plan to delete all the locations content may have? what are the plans for this?

glye:
Re: @crevillo's question, in addition to the problem of user deleting more than they intended to, for legacy we sometimes have the problem that moving/deleting huge subtrees to trash can time out, leading to a more or less broken state. The solution for legacy was doing such jobs later in cronjobs. Has this been considered for new stack? (Possibly not a real problem for moving to trash, but might be for deleting)

dpobel:
@crevillo @glye those questions are for the product management, not for us :) ping @rolandbenedetti

mhyndle:
I'm just wondering if "Send to Trash" button should be displayed when being in rootLocation context as there are two things related to that:
* loading parent location for rootLocation is triggering error
* rootLocation shouldn't be available for sending to trash - maybe another safety mechanism should be implemented for avoiding it? Currently I was able to move rootLocation to trash with 
```javascript
getContentService().moveSubtree('/api/ezp/v2/content/locations/1/2', '/api/ezp/v2/content/trash', callback)
```

ping @dpobel

dpobel:
You were able to do that because you are using the admin account, but with a limited user account, the REST API should check the permission and deny the action if the user does not have the right to remove the content. It's of course possible to completely deny the removal of some special contents, but given the state of the development of PlatformUI I prefer to not add any special rules yet.

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,136 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+                location = this.get('location'),
+                locationId = location.get('id'),
+                contentName = e.content.get('name');
+
+            e.preventDefault();
+            this._set('trackOutsideEvents', false);
+            this.fire('confirmBoxOpen', {
+                config: {
+                    title: "Are you sure you want to send this content to trash?",
+                    confirmHandler: Y.bind(function () {
```
dpobel: this function is too big to be anonymous, you should create a method and use it instead.

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,136 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+                        var tasks, parentLocation, trashPath, endLoadParent, endLoadTrashPath;
+
+                        that._sendContentToTrashNotificationStarted(locationId, contentName);
```
dpobel: since you are *binding* the function to the correct context, you don't need to use the *that/this* trick.

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,136 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+
+                        tasks = new Y.Parallel();
+                        endLoadParent = tasks.add();
+                        that._loadParent(location, function (error, parentLocationResponse) {
+                            if (error) {
+                                that._sendContentToTrashNotificationError(locationId, contentName);
+                                that._error("Failed to load parent location of " + locationId);
```
dpobel: you should not fire a notification and call `_error`, it's either one or the other as `_error` will soon also use the notification system. In this context, you should fire the notifiy event and not call `_error`.

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,136 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+        _sendContentToTrash: function (e) {
+            var app = this.get('app'),
+                that = this,
+                contentService = this.get('capi').getContentService(),
+                location = this.get('location'),
+                locationId = location.get('id'),
+                contentName = e.content.get('name');
+
+            e.preventDefault();
+            this._set('trackOutsideEvents', false);
```
dpobel: there's `trackOutsideEvents` attribute on LocationViewViewService, I guess it's a copy paste from UniversalDiscoveryConfirmedListView, so this should be removed (and below as well).

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,136 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+                                that._error("Failed to contact the REST API");
+                                return;
+                            }
+
+                            trashPath = loadRootResponse.document.Root.trash._href;
+                            endLoadTrashPath();
+                        });
+
+                        tasks.done(function () {
+                            contentService.moveSubtree(locationId, trashPath, function (err, response) {
```
dpobel: I think you should add a `trash` method to the location model and this method would discover the trash path, so that here, you would have only the parent loading and then the `trash` call.

mhyndle:
ping @dpobel - I have done some changes according to you comments

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,150 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+        _sendToTrash: function (e) {
+            var tasks, parentLocation, endLoadParent,
+                that = this,
+                location = this.get('location'),
+                locationId = location.get('id'),
+                contentName = e.content.get('name');
+
+            this._sendContentToTrashNotificationStarted(locationId, contentName);
+
+            tasks = new Y.Parallel();
```
mhyndle: I'm wondering if it's OK to using `Y.Parallel` here for `_loadParent()` only or maybe I should call `location.trash()` in callback of `_loadParent()`?

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,150 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+         */
+        _sendContentToTrashConfirmBox: function (e) {
+            e.preventDefault();
+            this.fire('confirmBoxOpen', {
+                config: {
+                    title: "Are you sure you want to send this content to trash?",
+                    confirmHandler: Y.bind(function () {
+                        this._sendToTrash(e);
+                    }, this),
+                    cancelHandler: Y.bind(function () {
```
dpobel: you can remove that since you are generating an empty function

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -42,6 +42,27 @@ YUI.add('ez-locationmodel', function (Y) {
...
+         * @param {Object} options the options for moving to trash
+         * @param {Object} options.api (required) the JS REST client instance
+         * @param {Function} callback
+         */
+        trash: function (options, callback) {
+            var api = options.api,
+                location = this;
+
+            api.getContentService().loadRoot(function (error, response) {
+                var trashPath = response.document.Root.trash._href;
```
dpobel: you should first check the error before using the response, for what ever reason, the loadRoot call might fail

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,150 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+         * @protected
+         * @param {Object} e event facade of the sendToTrashAction event
+         */
+        _sendContentToTrashConfirmBox: function (e) {
+            e.preventDefault();
+            this.fire('confirmBoxOpen', {
+                config: {
+                    title: "Are you sure you want to send this content to trash?",
+                    confirmHandler: Y.bind(function () {
+                        this._sendToTrash(e);
```
dpobel: instead of passing the event facade, I would pass what you really need (the content AFAICS) so that this method can maybe be reused else where.

Resources/public/js/views/services/ez-locationviewviewservice.js
dpobel: indeed you don't need to use Y.Parallel here. Y.Parallel is useful when you want to run several asynchronous operations at the same time and do something when they are all done, here you can indeed call `location.trash` in the callback of `_loadParent`.

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,150 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+
+        /**
+         * Notification changed to *started* before sending content to trash
+         *
+         * @method _sendContentToTrashNotificationStarted
+         * @protected
+         * @param {String} locationId
+         * @param {String} contentName
+         */
+        _sendContentToTrashNotificationStarted: function (locationId, contentName) {
```
dpobel: instead of having those 3 methods that do pretty much the same, we should have a generic `_notify` method for that. @StephaneDiot is adding one in https://github.com/ezsystems/PlatformUIBundle/pull/229 so you should sync to use the same names/parameters so that when the second to merge his PR has a trivial conflict to fix.

mhyndle:

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,110 @@ YUI.add('ez-locationviewviewservice', function (Y) {
         },
 
         /**
+         * _sendContentToTrashConfirmBox event handler on send to trash button
```
dpobel: Should be something like: "`sendToTrashAction` event handler, it asks confirmation to the user before sending the location to the trash."

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,110 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+        },
+
+        /**
+         * Sends location to trash, triggering loading parent location and notifications
+         *
+         * @method _sendToTrash
+         * @protected
+         * @param {Object} content
+         */
+        _sendToTrash: function (content) {
```
dpobel: actually, the content in parameter is the same as the one you can retrieve with `this.get('content')`, so you can omit this parameter.

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -43,6 +44,110 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+
+            location.trash({api: this.get('capi')}, Y.bind(that._afterSendToTrashCallback, that, options));
+        },
+
+        /**
+         * Send to trash callback triggering notifications and making app to navigate to parent location
+         *
+         * @method _afterSendToTrashCallback
+         * @protected
+         * @param {Object} options the options for sending to trash
```
dpobel: you should better describe what you expect in this object with something like:
```
* @param {Object} options
* @param {eZ.Location} options.parentLocation
* @param {String} options.contentName
```
(or use individual parameters since options is supposed to contain only the parentLocation and the contentName).

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -56,7 +57,99 @@ YUI.add('ez-locationmodel-tests', function (Y) {
...
+
+            Mock.expect(this.model, {
+                method: 'get',
+                args: [Mock.Value.String],
+                returns: locationId
+            });
+
+            Mock.expect(contentServiceMock, {
+                method: 'moveSubtree',
+                args: [Mock.Value.String, Mock.Value.String, Mock.Value.Function],
```
dpobel: this is a case where the arguments are important, so here instead of `Mock.Value.String` you should put the actual expected values (`locationId`  and `loadRootResponse.document.Root.trash._href`).

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -56,7 +57,99 @@ YUI.add('ez-locationmodel-tests', function (Y) {
...
+                run: function (id, trashPath, cb) {
+                    cb(false);
+                }
+            });
+
+            this.model.trash({
+                api: capiMock
+            }, function (error) {
+                Assert.isTrue(
```
dpobel: `Assert.isFalse(error, "blabla");`

Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
```diff
@@ -478,9 +479,316 @@ YUI.add('ez-locationviewviewservice-tests', function (Y) {
...
+                run: function (options, callback){
+                    callback(true);
+                }
+            });
+
+            this.service.on('confirmBoxOpen', function (e) {
+                e.config.confirmHandler.apply(this);
+            });
+
+            this.service.on('notify', function (e) {
```
dpobel: `this.service.once` would be more suited I think

Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
```diff
@@ -478,9 +479,316 @@ YUI.add('ez-locationviewviewservice-tests', function (Y) {
...
+                method: 'trash',
+                args: [Mock.Value.Object, Mock.Value.Function],
+                run: function (options, callback){
+                    callback(false);
+                }
+            });
+
+            Mock.expect(this.app, {
+                method: 'navigateTo',
+                args: [Mock.Value.String, Mock.Value.Object]
```
dpobel: here as well, the parameter are important, you should put the first to `'viewLocation`' and add a `run` function in the expectation to check that the second parameter contains the expected property.

Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
```diff
@@ -478,9 +479,316 @@ YUI.add('ez-locationviewviewservice-tests', function (Y) {
...
+                method: 'trash',
+                args: [Mock.Value.Object, Mock.Value.Function],
+                run: function (options, callback){
+                    callback(false);
+                }
+            });
+
+            Mock.expect(this.app, {
+                method: 'navigateTo',
+                args: [Mock.Value.String, Mock.Value.Object],
```
dpobel: instead of checking the routeName in run, you can do it directly in the expectation:
```js
args: ['viewLocation', Mock.Value.Object],
```

Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
```diff
@@ -478,9 +479,316 @@ YUI.add('ez-locationviewviewservice-tests', function (Y) {
...
+                        "Id of location passed to navigateTo method should be the parent location id"
+                    );
+                }
+            });
+
+            this.service.on('confirmBoxOpen', function (e) {
+                e.config.confirmHandler.apply(this);
+            });
+
+            this.service.fire('whatever:sendToTrashAction', {content: content});
```
dpobel: you should also add a `Mock.verify(this.app)` here to check that we really redirect the user.

Tests/js/models/assets/ez-locationmodel-tests.js
mhyndle: Yep but it should be done for previous assertion (Line 116) ;)

dpobel:
+1

StephaneDiot:
+1

PR #232, opened 2015-05-22
EZP-23814: As an editor, I want to see the content of the RichText fields

dpobel:
JIRA: https://jira.ez.no/browse/EZP-23814

# Description

This patch adds a field view for the RichText fields. For now, it's pretty simple as it only *re-serializes* the `xhtml5edit` version of the field to display it.

richtext_display

## Open questions

But a RichText field can contain a lot more than a few headings, paragraphs and lists so as with the good old XmlText it is far more complicated than that. At least, I can see 3 topics that needs some thought from the PM and/or from the Engineering/community:

1. Links resolution
2. Custom tags
3. Embed

ping @rolandbenedetti @andrerom @bdunogier @lolautruche @yannickroger @pspanja and whoever wants to jump in :)

### Links resolution

I guess that when a RichText field contains a link to a content or a location in the repository, the user should be able to use it by a click to view the corresponding content or with an intermediate UI maybe. But in any case, with the current `xhtml5edit` version it's very complicated to achieve because the target of a link is expressed either with `ezlocation://2` or `ezcontent://59` but those ids *does not mean anything* from a PlatformUI/REST API perspective. A solution would be to enrich the RichText `xhtml5edit` version to have for instance a `data-location-rest-id` (and a `data-language-code` after [EZP-24384](https://jira.ez.no/browse/EZP-24384)) attribute on all links so that the RichText field view can easily compute the corresponding link in PlatformUI to the correct Location. The question is now, where/how such addition should be done?

### Custom tags

For the custom tags, the question is how should they be rendered when displaying a RichText field in PlatformUI. Let's take the example of a Youtube custom tag, should we display a kind of box representing the custom tag or should we display the fully rendered custom tag ie the Youtube video player ?

If we choose the later, most likely that means this rendering should be done server side and I think, we should maybe have a dedicated RichText text format instead of `xhtml5edit` (unless we do the same in while editing the content). Again, the question is where this is supposed to be implemented?  If we add more RichText format, this also reopens the question of how to request with the REST API a given RichText format because currently, we always provide the `xml` and `xhtml5edit` versions which is a bit brutal :)

### Embed

For embeds, it's basically a mix of the two above cases. To allow the navigation in PlatformUI I need a richer (!) RichText version but I also need to know whether the embed have to be fully rendered or not. Note that for embed, we'll most likely need to partially render them anyway if we want the editors to see the images embed in a RichText field. Again, we have to add the code for that, but where ?

# Tests

manual + unit tests and it works in simple use cases.


dpobel:
ping @StephaneDiot @yannickroger @mhyndle for the review (but of course your input is welcome on the open questions as well)

andrerom:
To answer some of the questions, I think rest layer should probably fix up relations to be correct from REST perspective, however then we need to do it on input as well.

As for preview of custom tags and embed, some kind of box here, full preview does not make sense in this view. So similar to what we end up with during editing I would say.

StephaneDiot:
+1

yannickroger:
+1

pspanja:
@dpobel 
> But in any case, with the current xhtml5edit version it's very complicated to achieve because the target of a link is expressed either with ezlocation://2 or ezcontent://59 but those ids does not mean anything from a PlatformUI/REST API perspective.

`xhtml5edit` is **editing** format, it was not intended to be used for viewing. Conversion from internal format to HTML5 already exists. In Twig this is exposed through a filter, and my initial idea was to implement a REST resource that would do that. But it could also be implemented in the same way edit format is exposed: https://github.com/ezsystems/ezpublish-kernel/pull/1211.

Regarding generating links for a translation, shouldn't that really be resolved for a siteaccess language configuration?

Note that Docbook to HTML5 converter is aggregate of separate converters, so for different (non-rendered) representation of links, embeds and custom tags one would only need to define an aggregate converter with different dedicated converters for those.

pspanja:
@dpobel 
> But in any case, with the current xhtml5edit version it's very complicated to achieve because the target of a link is expressed either with ezlocation://2 or ezcontent://59 but those ids does not mean anything from a PlatformUI/REST API perspective.

I guess we really need a path string there?

andrerom:
> I guess we really need a path string there? For Location anyway.

yes, well mentioned above I think REST layer for this field type could change any Content / Location  links / relations from the internal ezcontent / ezlocation format to the format used by REST. But if we do such a change it should also be done for Relation(s) field types.

dpobel:
> xhtml5edit is editing format, it was not intended to be used for viewing. [...]

I know, but it's the only one exposed for now (and btw even for editing, current xhtml5edit won't be usable as it is, partly with the same issues). If we can avoid having an other HTTP request for that, that would be nice and I'm not for embedding all different formats by default as well. I guess we should move forward on the "enrich REST output" feature (used to be called "custom media types" as well)

> Regarding generating links for a translation, shouldn't that really be resolved for a siteaccess language configuration?

AFAIK, the REST API is not really siteaccess aware (and the PlatformUIApp somehow follows that) and I'm interested in the main language which is independent from the configuration.

> Note that Docbook to HTML5 converter is aggregate of separate converters, so for different (non-rendered) representation of links, embeds and custom tags one would only need to define an aggregate converter with different dedicated converters for those.

is there a documentation for that somewhere ?

> I guess we really need a path string there? For Location anyway.

I need the *REST id* ie the `ezpublish_rest_loadLocation` route for the pointed location and as written by André, replacing `href="ezlocation://2"` by `href="/api/ezp/v2/content/locations/1/2"` would perfectly work in this context.

dpobel:
anyway, I'm merging that as a first step and I'll create some follow up stories for the open questions as I need some clear specifications from the PM

pspanja:
I guess what I'm not clear with is why would we need REST specific enrichments for rendered output... it should be simply rendered and displayed, or am I missing something?

>  is there a documentation for that somewhere ?

Not yet, but we can sync about it.

PR #235, opened 2015-05-27
EZP-24378: Prototype of Rich Text editor with Alloy Editor

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24378

# Description

**This PR is not meant to be merged as it is, it's a HUGE work in progress**

This is a prototype of the RichText field edit view based on [AlloyEditor](http://alloyeditor.com/).

For the simplest RichText field features, it's working quite nicely but as soon as I want to add more complicated markup, it does not work because the parser seems to be super strict. For instance, it refuses the value if it has a `data-` attribute somewhere (while that's perfectly valid in (X)HTML5).

Screencast: http://youtu.be/yUZM_FCcS_I

### DIFF
--- a/Resources/public/js/views/fields/ez-richtext-editview.js
+++ b/Resources/public/js/views/fields/ez-richtext-editview.js
@@ -1,0 +1,104 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-richtext-editview', function (Y) {
+    "use strict";
+    /**
+     * Provides the field edit view for the RichText (ezrichtext) fields
+     *
+     * @module ez-richtext-editview
+     */
+    Y.namespace('eZ');
+
+    var FIELDTYPE_IDENTIFIER = 'ezrichtext';
+
+    /**
+     * Rich Text edit view
+     *
+     * @namespace eZ
+     * @class RichTextEditView
+     * @constructor
+     * @extends eZ.FieldEditView
+     */
+    Y.eZ.RichTextEditView = Y.Base.create('richTextEditView', Y.eZ.FieldEditView, [], {
+        events: {
+            '.ez-richtext-input-ui textarea': {
+                'blur': 'validate',
+                'valuechange': 'validate'
+            }
+        },
+
+        initializer: function () {
+            this.after('activeChange', function (e) {
+                if ( this.get('active') ) {
+                    this._editor = Y.eZ.AlloyEditor.editable(
+                        this._getEditableArea().get('id')
+                    );
+                } else {
+                    this._editor.destroy();
+                }
+            });
+        },
+
+        /**
+         * Validates the current input of the xml text
+         *
+         * @method validate
+         */
+        validate: function () {
+            // TODO
+            this.set('errorStatus', false);
+        },
+
+        _getEditableArea: function () {
+            return this.get('container').one('.ez-richtext-editable');
+        },
+
+        /**
+         * Defines the variables to be imported in the field edit template for xml
+         * text.
+         *
+         * @protected
+         * @method _variables
+         * @return {Object} containing isRequired entry
+         */
+        _variables: function () {
+            return {
+                "isRequired": this.get('fieldDefinition').isRequired,
+                "xhtml": this._serializeFieldValue(),
+            };
+        },
+
+        _serializeFieldValue: function () {
+            var doc = (new DOMParser()).parseFromString(this.get('field').fieldValue.xhtml5edit, "text/xml");
+
+            // TODO error handling
+            //doc.documentElement.setAttribute('xmlns', 'http://ez.no/namespaces/ezpublish5/xhtml5/edit');
+            return (new XMLSerializer()).serializeToString(doc.documentElement);
+        },
+
+        /**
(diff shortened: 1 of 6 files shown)


Resources/public/js/views/fields/ez-richtext-editview.js
```diff
@@ -0,0 +1,104 @@
...
+         *
+         * @method _getFieldValue
+         * @protected
+         * @return String
+         */
+        _getFieldValue: function () {
+            var val = {
+                    xml: this._getEditableArea()
+                            .one('section')
+                            .setAttribute('xmlns', 'http://ez.no/namespaces/ezpublish5/xhtml5/edit')
```
dpobel: @pspanja shouldn't the default namespace be `http://ez.no/namespaces/ezpublish5/xhtml5/edit` by default on xhtml5edit format ?

pspanja: We have multiple input formats + internal format as input. Format is defined by the XML namespace, so it should be always present. By that the correct converter is dispatched.

dpobel: I know that, the issue is that by default namespace is `http://ez.no/namespaces/ezpublish5/xhtml5/` not `http://ez.no/namespaces/ezpublish5/xhtml5/edit`, so my question is it normal ?

pspanja: Sorry, not quite following... where is it the default?

dpobel: The field value of my RichText field is:

```json
"fieldValue": {
    "xml": "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<section xmlns=\"http:\/\/docbook.org\/ns\/docbook\" xmlns:xlink=\"http:\/\/www.w3.org\/1999\/xlink\" xmlns:ezxhtml=\"http:\/\/ez.no\/xmlns\/ezpublish\/docbook\/xhtml\" xmlns:ezcustom=\"http:\/\/ez.no\/xmlns\/ezpublish\/docbook\/custom\" version=\"5.0-variant ezpublish-1.0\"><para>simple rich text<\/para><\/section>\n",
    "xhtml5edit": "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<section xmlns=\"http:\/\/ez.no\/namespaces\/ezpublish5\/xhtml5\"><p>simple rich text<\/p><\/section>\n"
}
```

as you can see, in the `xhtml5edit` version the section element has an `xmlns` attribute set to `http://ez.no/namespaces/ezpublish5/xhtml5`. If I use this exact document to set the value of a RichText field, the parser kicks me out but if I set the xmlns to `http://ez.no/namespaces/ezpublish5/xhtml5/edit` as it's done where I first commented, it's working. So my question is: is it normal to have the xmlns set to `http://ez.no/namespaces/ezpublish5/xhtml5` when using the xhtml5edit version ?

pspanja: No, that's a mistake in configuration :)

See https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Bundle/EzPublishCoreBundle/Resources/config/fieldtype_services.yml#L230, it should use:

https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/FieldType/RichText/Resources/stylesheets/docbook/xhtml5/edit/xhtml5.xsl

instead of:

https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/FieldType/RichText/Resources/stylesheets/docbook/xhtml5/output/xhtml5.xsl

Namespace is the only difference.

andrerom:
Side question: Which version of AlloyEditor did you use? 0.2.7+ ?

dpobel:
The one from today, 0.2.8 (indeed I did not create the corresponding PR in PlatformUIAssetsBundle)

andrerom:
@pspanja Should we adjust the RichText input handling to discard all attributes it does not recognize? Or at least all `data-` attributes (wdyt @dpobel ?)

pspanja:
I used these quite a bit for embeds and template (custom) tags... can AlloyEditor rather be configured to allow these? It is based on CKEditor so I would assume it can be done.

dpobel:
@pspanja you got it wrong, AlloyEditor is fine with `data-` attributes, but the RichText input is not.

pspanja:
@dpobel Ok :D 
We call the format XHTML5, but of course it must be only a subset. It has it's own schema, while it is not strictly necessary to validate it, as only end result of the conversion to the internal format must be validated. But why do does the input have stuff that is not handled?

pspanja:
Basically three options:

1. provide clean input
2. implement a XSL based converter that would do the cleanup it and configure it as first in the chain
3. disable validator for input

andrerom:
@pspanja how do you see us handling import scenarios in the future where html content might contain a lot for tags and attributes we would have to ignore?

pspanja:
@andrerom Depends, if the difference would be ignorable by conversion then disabling validator would work. Otherwise some kind of normalizing conversion would have to be implemented.
But, I would think having input format validation would help in this phase. As I already said validating input format is not really required if the internal format is validated before storing. I'll open a PR on kernel for disabling it (also fixing NS issue from https://github.com/ezsystems/PlatformUIBundle/pull/232).

PR #236, opened 2015-05-27
Fix EZP-24379: Integrate ContentType edition with PlatformUI

lolautruche:
> https://jira.ez.no/browse/EZP-24379

This PR adds ContentType management into PlatformUI.
This includes:

* List of ContentTypeGroups
* List of ContentTypes in a group
* Base view of a ContentType
* ContentType edit
* Integration in navigation hub
* Integration in admin dashboard

Depends on https://github.com/ezsystems/repository-forms/pull/16

PR #238, opened 2015-05-28
Fix EZP-24418: ServerSideView multiple submit buttons support

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24418

# Description

This patch makes sure we handle correctly the forms with multiple buttons. The used buttons is detected with the active element (ie the element having the focus) so that it's value is part of payload send to the browser while others buttons are just ignored.

# Tests

unit tests + manual tests with content type edit form provided in https://github.com/ezsystems/PlatformUIBundle/pull/236

### DIFF
--- a/Resources/public/js/views/ez-serversideview.js
+++ b/Resources/public/js/views/ez-serversideview.js
@@ -41,12 +41,90 @@ YUI.add('ez-serversideview', function (Y) {
              *
              * @event submitForm
              * @param {Node} form the Node object of the submitted form
+             * @param {Object} formData the serialzed form data including the
+             * used button to validate the form
              * @param {Event} originalEvent the original DOM submit event
              */
             this.fire('submitForm', {
                 form: e.target,
+                formData: this._serializeForm(e.target),
                 originalEvent: e,
             });
+        },
+
+        /**
+         * Serializes the form data so that it can be send with an AJAX request.
+         * It also checks which button was used (if any) to include it in the
+         * data.
+         *
+         * @method _serializeForm
+         * @param {Node} form
+         * @return {Object} key/value data of the form
+         */
+        _serializeForm: function (form) {
+            var focusedNode = form.get('ownerDocument').get('activeElement'),
+                data = {};
+
+            if ( this._isSubmitButton(focusedNode, form) ) {
+                data[focusedNode.getAttribute('name')] = "";
+            }
+
+            form.get('elements').each(function (field) {
+                var name = field.getAttribute('name'),
+                    type = field.get('type');
+
+                if ( !name ) {
+                    return;
+                }
+
+                /* jshint -W015 */
+                switch (type) {
+                    case 'button':
+                    case 'reset':
+                    case 'submit':
+                        break;
+                    case 'radio':
+                    case 'checkbox':
+                        if ( field.get('checked') ) {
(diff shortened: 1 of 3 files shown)


dpobel:
ping @lolautruche @yannickroger @mhyndle @StephaneDiot

Resources/public/js/views/ez-serversideview.js
```diff
@@ -41,15 +41,91 @@ YUI.add('ez-serversideview', function (Y) {
...
+                    case 'submit':
+                        break;
+                    case 'radio':
+                    case 'checkbox':
+                        if ( field.get('checked') ) {
+                            data[name] = field.get('value');
+                        }
+                        break;
+                    default:
+                        data[name] = field.get('value');
```
lolautruche: What about `textarea`?

Resources/public/js/views/ez-serversideview.js
```diff
@@ -41,15 +41,91 @@ YUI.add('ez-serversideview', function (Y) {
...
+                    case 'reset':
+                    case 'submit':
+                        break;
+                    case 'radio':
+                    case 'checkbox':
+                        if ( field.get('checked') ) {
+                            data[name] = field.get('value');
+                        }
+                        break;
+                    default:
```
lolautruche: Maybe add `select` as well? Or at least add a comment to make it clear that it works the same as `input`.

Resources/public/js/views/ez-serversideview.js
dpobel: indeed I forgot about this one but actually `textarea` (and `select`)  are handled in with the default case (thanks to `.get('value')`)

Resources/public/js/views/ez-serversideview.js
dpobel: good point, I'm adding a comment

dpobel: also added 2 unit tests for textarea and select

lolautruche:
+1

lolautruche:
Works like a charm! :sparkles:

Resources/public/js/views/ez-serversideview.js
```diff
@@ -41,15 +41,93 @@ YUI.add('ez-serversideview', function (Y) {
              *
              * @event submitForm
              * @param {Node} form the Node object of the submitted form
+             * @param {Object} formData the serialzed form data including the
```
yannickroger: typo

Resources/public/js/views/ez-serversideview.js
```diff
@@ -41,15 +41,93 @@ YUI.add('ez-serversideview', function (Y) {
...
              */
             this.fire('submitForm', {
                 form: e.target,
+                formData: this._serializeForm(e.target),
                 originalEvent: e,
             });
         },
 
         /**
+         * Serializes the form data so that it can be send with an AJAX request.
```
yannickroger: sent

yannickroger:
+1

andrerom:
besides comments, and besides it is a pity yui does not handle this, +1 :)

dpobel:
Manually merged in https://github.com/ezsystems/PlatformUIBundle/commit/3419e9c1876e7c8ab8d03fc0404448d8e2bd4723

PR #240, opened 2015-06-02
BDD Move Content Feature

pcardiga:
BDD scenarios for Move Content feature - https://jira.ez.no/browse/EZP-24187

The current BDD scenarios aim to reflect the behaviour of moving a content in PlatformUI
The situations reflected below are not yet reflected in BDD scenarios due clarification need
- [ ] Try to move an object that is read-only for the user
- [ ] Try to move an object to a read-only location
- [ ] Try to move one object that has one child inaccessible to the user.
- [ ] Try to move into a non-container
- [ ] Copy multiple objects
- [ ] Existence of Manipulate button action view on Action Bar respective sub-bar descendants on right panel
- [ ] Move objects to hidden locations

### DIFF
--- a/Features/CopyMoveDelete/MoveContent.feature
+++ b/Features/CopyMoveDelete/MoveContent.feature
@@ -14,10 +14,10 @@ Feature: Move content
         Given an "Older News" folder exists
         And a "Origin/News Flash" article exists
         And I am on "News Flash" full view
-        When I click "Move" button on the "Action Bar"
-        And I select "Older News" folder in the Universal Discovery Widget
-        And I click the "Confirm selection" button
-        Then I am notified that "News Flash' has been successfully moved under 'Older News'"
+        When I click the Move button on the Action Bar
+        And I select the "Older News" folder in the Universal Discovery Widget
+        And I confirm the selection
+        Then I am notified that "News Flash" has been moved under "Older News"
         And I see "Older News/News Flash" in the content tree
         And I do not see "Origin/News flash" in the content tree
 
@@ -25,7 +25,7 @@ Feature: Move content
     Scenario: Move one object that has children objects
         Given an "Older News" folder exists
         And a "Tomorrow news/News Flash" article exists
-        And I am on the "News Flash" full view
-        When I move the "Tomorrow news" into the "Older News" folder
+        And I am on "News Flash" full view
+        When I move "Tomorrow news" into the "Older News" folder
         Then "Tomorrow news" is moved
         And I see "Older News/Tomorrow news/News Flash" in the content tree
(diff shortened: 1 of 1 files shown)


pcardiga:
ping @bdunogier , @dpobel , @andrerom

Features/CopyMoveDelete/MoveContent.feature
```diff
@@ -0,0 +1,49 @@
+Feature: Move content using te role of Editor
```
andrerom: the

andrerom:
looks ok, but as mentioned yesterday we should skip scenarios that are currently broken because of API, example hide.

Anyway to mark as skipped?

pcardiga:
The scenarios related with hidden objects are going to be removed for now and included later when the feature is fixed.
In Behat 2.x, there was the possibility of having a tag, such as @skipped, and then configure behat.yml for the scenarios that have that tag be ignored during execution. But for some reason, this possibility is no longer present in Behat 3.0.

bdunogier:
According to the behat docs, tags should still be supported (at least from
the config file):
http://docs.behat.org/en/latest/guides/5.suites.html#suite-filters.

On Tue, Jun 16, 2015 at 11:29 AM, Paulo Cardiga <notifications@github.com>
wrote:

> The scenarios related with hidden objects are going to be removed for now
> and included later when the feature is fixed.
> In Behat 2.x, there was the possibility of having a tag, such as @skipped
> <https://github.com/skipped>, and then configure behat.yml for the
> scenarios that have that tag be ignored during execution. But for some
> reason, this possibility is no longer present in Behat 3.0.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ezsystems/PlatformUIBundle/pull/240#issuecomment-112362432>
> .
>



-- 
--
Bertrand Dunogier
Lead Engineer eZ extensibility
eZ Systems AS | skype:bertrand-dunogier | http://www.ez.no/

miguelcleverti:
I tried using the filters configuration in the behat.yml and it was telling me that the parameter was not recognized, I made some investigations and and found that it might have to do with the behat version. Iit is still possible to use it on the console command such as `php behat  (...) --tags '~@skipped'` which will run everything but the ones with the tag `@skipped`. I will try to investigate a bit more about this anyway.

pcardiga:
Scenarios updated:
ping @bdunogier, @dpobel, @andrerom

pcardiga:
Scenario sentences updated accordingly dicussion occurred in [Copy BDD PR](https://github.com/ezsystems/PlatformUIBundle/pull/243/ "PR for the BDD Copy feature")

pcardiga:
Scenario sentences updated again accordingly new dicussion occurred in [Copy BDD PR](https://github.com/ezsystems/PlatformUIBundle/pull/243/ "PR for the BDD Copy feature")

andrerom:
+1, ref https://github.com/ezsystems/PlatformUIBundle/pull/243#issuecomment-119960457

bdunogier:
+1

PR #243, opened 2015-06-02
BDD Copy Content Feature file

pcardiga:
BDD scenarios for Copy Content feature - https://jira.ez.no/browse/EZP-24188

The current BDD scenarios aim to reflect the behaviour of copying a content in PlatformUI
The situations reflected below are not yet reflected in BDD scenarios 

Todo:
- [ ] Implementation

Later follow up:
- Try to copy an object that is read-only for the user
- Try to copy an object to a read-only location
- Try to copy one object that has one child inaccessible to the user.
- Try to copy into a non-container
- Copy multiple objects
- Copy objects to hidden locations

### DIFF
--- a/Features/CopyMoveDelete/CopyContent.feature
+++ b/Features/CopyMoveDelete/CopyContent.feature
@@ -1,5 +1,5 @@
-Feature: Copy content using te role of Editor
-    In order to validate the copy action
+Feature: Copy content
+    In order to copy objects
     As an Editor user
     I need to be able to copy an object that I am viewing
 
@@ -13,34 +13,19 @@ Feature: Copy content using te role of Editor
     Scenario: Copy one object without children objects
         Given a "Destiny" folder exists
         And a "News Flash" article exists
-        When I copy the "News Flash" into "Destiny" folder
-        Then the "News Flash" is copied with message "'News Flash' has been successfully copied under 'Destiny'"
-        And I see "News Flash" as a child of "Destiny" folder
+        And I am viewing the "News Flash" article
+        When I click on "Copy" link on the "Action Bar"
+        And I select the "Destiny" folder in Universal Discovery Widget
+        And I click the "Confirm selection" button
+        Then I see the message "'News Flash' has been successfully copied under 'Destiny'"
+        And I see "Destiny/News Flash" in content tree
 
     @javascript
     Scenario: Copy one object that has children objects
         Given an "Destiny" folder exists
-        And an "Origin" folder exists
-        And a "News Flash" article exists as a child of "Origin" folder
-        When I copy the "News Flash" into "Destiny" folder
+        And an "Origin/News Flash" article exists
+        And I am viewing the "News Flash" article
+        When I copy the "News Flash" into the "Destiny" folder
         Then the "News Flash" is copied
-        And I see "News Flash" as a child of "Origin" folder
-        And I see "News Flash" as a child of "Destiny" folder
-
-    @javascript
-    Scenario: Content tree is updated after the copy of an object
-        Given an "Destiny" folder exists
-        And an "Origin" folder exists
-        And a "News Flash" article exists as a child of "Origin" folder
-        When I copy the "News Flash" into "Destiny" folder
-        Then the "News Flash" is copied
-        And I see "News flash" in content tree as child of "Destiny" folder
-
-    @javascript
-    Scenario: Copy one object to an hidden location
-        Given an "Destiny" folder exists
-        And "Destiny" is hidden
-        And a "News Flash" article exists
-        When I copy the "News Flash" into "Destiny" folder
-        Then the "News Flash" is copied
-        And I see "News Flash" as a child of "Origin" folder
+        And I see "Origin/News Flash" in content tree
+        And I see "Destiny/News Flash" in content tree
--- a/Features/CopyMoveDelete/CopyContent.feature
+++ b/Features/CopyMoveDelete/CopyContent.feature
@@ -1,0 +1,46 @@
+Feature: Copy content using te role of Editor
+    In order to validate the copy action
+    As an Editor user
+    I need to be able to copy an object that I am viewing
+
+    Background:
+        Given I am logged in as an Editor in PlatformUI
+
+    ##
+    #Copy objects to valid locations
+    ##
+    @javascript
+    Scenario: Copy one object without children objects
+        Given a "Destiny" folder exists
+        And a "News Flash" article exists
+        When I copy the "News Flash" into "Destiny" folder
+        Then the "News Flash" is copied with message "'News Flash' has been successfully copied under 'Destiny'"
+        And I see "News Flash" as a child of "Destiny" folder
+
+    @javascript
+    Scenario: Copy one object that has children objects
+        Given an "Destiny" folder exists
+        And an "Origin" folder exists
+        And a "News Flash" article exists as a child of "Origin" folder
+        When I copy the "News Flash" into "Destiny" folder
+        Then the "News Flash" is copied
+        And I see "News Flash" as a child of "Origin" folder
+        And I see "News Flash" as a child of "Destiny" folder
+
+    @javascript
+    Scenario: Content tree is updated after the copy of an object
+        Given an "Destiny" folder exists
+        And an "Origin" folder exists
+        And a "News Flash" article exists as a child of "Origin" folder
+        When I copy the "News Flash" into "Destiny" folder
+        Then the "News Flash" is copied
+        And I see "News flash" in content tree as child of "Destiny" folder
+
+    @javascript
+    Scenario: Copy one object to an hidden location
+        Given an "Destiny" folder exists
+        And "Destiny" is hidden
+        And a "News Flash" article exists
+        When I copy the "News Flash" into "Destiny" folder
+        Then the "News Flash" is copied
+        And I see "News Flash" as a child of "Origin" folder
--- a/Features/CopyMoveDelete/CopyContent.feature
+++ b/Features/CopyMoveDelete/CopyContent.feature
@@ -10,22 +10,24 @@ Feature: Copy content
     #Copy objects to valid locations
     ##
     @javascript
-    Scenario: Copy one object without children objects
-        Given a "Destiny" folder exists
+    Scenario: Copying a content using the "copy" button
+        Given a "Destination" folder exists
         And a "News Flash" article exists
-        And I am viewing the "News Flash" article
-        When I click on "Copy" link on the "Action Bar"
-        And I select the "Destiny" folder in Universal Discovery Widget
-        And I click the "Confirm selection" button
-        Then I see the message "'News Flash' has been successfully copied under 'Destiny'"
-        And I see "Destiny/News Flash" in content tree
+        And I am on "News Flash" full view
+        When I click "Copy" button on the "Action Bar"
+        And I select "Destination" folder in the Universal Discovery Widget
+        And I click "Confirm selection" button
+        Then I am notified that "'News Flash' has been successfully copied under 'Destination'"
+        And I see "Destination/News Flash" in the content tree
 
     @javascript
-    Scenario: Copy one object that has children objects
-        Given an "Destiny" folder exists
+    Scenario: Copy one object that has children objects using the "Copy Subtree" button
+        Given an "Destination" folder exists
         And an "Origin/News Flash" article exists
-        And I am viewing the "News Flash" article
-        When I copy the "News Flash" into the "Destiny" folder
-        Then the "News Flash" is copied
-        And I see "Origin/News Flash" in content tree
-        And I see "Destiny/News Flash" in content tree
+        And I am on "Origin" full view
+        When I click "Copy Subtree" button on the "Action Bar"
+        And I select "Destination" folder in the Universal Discovery Widget
+        And I click "Confirm selection" button
+        Then I am notified that "'Origin' has been successfully copied under 'Destination'"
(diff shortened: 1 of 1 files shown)


pcardiga:
ping @bdunogier , @dpobel , @andrerom

Features/CopyMoveDelete/CopyContent.feature
```diff
@@ -0,0 +1,46 @@
+Feature: Copy content using te role of Editor
+    In order to validate the copy action
```
bdunogier: This sounds a bit off... validating a feature isn't really something a *user* wants. It's something we/you want :-)

Features/CopyMoveDelete/CopyContent.feature
```diff
@@ -0,0 +1,46 @@
...
+        And an "Origin" folder exists
+        And a "News Flash" article exists as a child of "Origin" folder
+        When I copy the "News Flash" into "Destiny" folder
+        Then the "News Flash" is copied
+        And I see "News Flash" as a child of "Origin" folder
+        And I see "News Flash" as a child of "Destiny" folder
+
+    @javascript
+    Scenario: Content tree is updated after the copy of an object
+        Given an "Destiny" folder exists
```
bdunogier: I'm thinking that this is gonna happen a lot (folder/folder/article). I'd almost suggest you add a shortcut sentence:

```
Given that an article "Origin/News Flash" exists
```

The leaf in the sentence would be of the given content-type (article), and all the parents would just be folders. It's kind of how we refer to content when talking about issues, if you think about it.

Features/CopyMoveDelete/CopyContent.feature
```diff
@@ -0,0 +1,46 @@
...
+        And an "Origin" folder exists
+        And a "News Flash" article exists as a child of "Origin" folder
+        When I copy the "News Flash" into "Destiny" folder
```
bdunogier: "into the"

Features/CopyMoveDelete/CopyContent.feature
```diff
@@ -0,0 +1,46 @@
...
+        Then the "News Flash" is copied
```
bdunogier: Since this scenario is about the tree, I wouldn't even mention that the copy operation worked here.

Features/CopyMoveDelete/CopyContent.feature
```diff
@@ -0,0 +1,46 @@
...
+        Given I am logged in as an Editor in PlatformUI
+
+    ##
+    #Copy objects to valid locations
+    ##
+    @javascript
+    Scenario: Copy one object without children objects
+        Given a "Destiny" folder exists
+        And a "News Flash" article exists
+        When I copy the "News Flash" into "Destiny" folder
```
bdunogier: I might be off-topic, but I would have expected more details on the action of copying, from a user's perspective. The scenario does not say which page/view I'm standing on, and what I do to copy.

Features/CopyMoveDelete/CopyContent.feature
pcardiga: We thought about that but haven't follow that approach because intermediary containers might not be folders. That's why it's first defined "what is what". But you kind of have your point in here in terms of readability. What can be done is to assume that the type of container it's always a folder unless referred otherwise, and then refer to it's child with the full path  "Origin/News Flash". 
I'll update the scenarios using the above assumption.

Features/CopyMoveDelete/CopyContent.feature
pcardiga: That was the first approach, but I simplified it due readability reasons and because the "Manipulate" option and respective sub-bar are not yet contemplated in the Action Bar. Currently we only have the direct option (Copy, Move, Send to Trash).

Maybe the first copy scenario might have full details from user's perspective, using current implementation. When the implementation is updated to use the "Manipulate" menu, then we update the bdd scenario.
But i'm unsure about having these details in *all* scenarios. If so, we risk having too many details that might difficult the readability and understood of the real purpose of the scenarios objectives.

pcardiga:
Scenarios updated
ping @bdunogier, @dpobel, @andrerom

Features/CopyMoveDelete/CopyContent.feature
bdunogier: We must think by _Context_ here, in my opinion.

Depending on the context, there are terms/concepts/steps that will be very accurate, and others that will be much more abstract, with very little details.

I'd say we need at least one test, somewhere, that details the action of copying a content from its full view. The scenario title should reflect this: "Copying a content using the button on the detail view". In this one, it should be very clear where we stand and what actions we take to perform the copy and ensure that the result is clear.

If another scenario, outside the "Copy" context, requires a copy, then it should be a oneliner, like "I copy Content X to Location Y (using the full view button)", since it is not what we are interested in. _How_ we copy (e.g. full view button) might matter, or might not, it depends on the context again: if _how_ we copy matters to the scenario, then it should be specified. If we are testing how feature X reacts to a copy of a content, then we could even do the copy using the Public API, we don't need to click on the buttons.

Does it make sense ?

bdunogier:
Note: added [a reply](https://github.com/ezsystems/PlatformUIBundle/pull/243#discussion_r32893297) to a comment that got hidden.

Features/CopyMoveDelete/CopyContent.feature
```diff
@@ -0,0 +1,31 @@
...
+
+    Background:
+        Given I am logged in as an Editor in PlatformUI
+
+    ##
+    #Copy objects to valid locations
+    ##
+    @javascript
+    Scenario: Copy one object without children objects
+        Given a "Destiny" folder exists
```
bdunogier: I gotta ask... "destiny" ? Is it like the Fate folder ? :-)

Jokes aside, did you mean "Destination" ?

Features/CopyMoveDelete/CopyContent.feature
```diff
@@ -0,0 +1,31 @@
...
+        And a "News Flash" article exists
+        And I am viewing the "News Flash" article
```
bdunogier: I'd use "button" instead of "link" here.

Features/CopyMoveDelete/CopyContent.feature
```diff
@@ -0,0 +1,31 @@
...
+        When I click on "Copy" link on the "Action Bar"
```
bdunogier: I'd use "button" rather than "link".

Features/CopyMoveDelete/CopyContent.feature
```diff
@@ -0,0 +1,31 @@
...
+        And I select the "Destiny" folder in Universal Discovery Widget
```
bdunogier: "in *the* Universal Discover Widget" ?

Features/CopyMoveDelete/CopyContent.feature
```diff
@@ -0,0 +1,31 @@
...
+        And I click the "Confirm selection" button
+        Then I see the message "'News Flash' has been successfully copied under 'Destiny'"
```
bdunogier: "message" => "notification" ? Ping @dpobel

Features/CopyMoveDelete/CopyContent.feature
```diff
@@ -0,0 +1,31 @@
...
+        And I see "Destiny/News Flash" in content tree
+
+    @javascript
+    Scenario: Copy one object that has children objects
+        Given an "Destiny" folder exists
+        And an "Origin/News Flash" article exists
+        And I am viewing the "News Flash" article
```
bdunogier: Would it be more meaningful as "I am on the "News Flash" full view" (or whatever it is called in PlatformUI) ?

Features/CopyMoveDelete/CopyContent.feature
```diff
@@ -0,0 +1,31 @@
...
+    I need to be able to copy an object that I am viewing
+
+    Background:
+        Given I am logged in as an Editor in PlatformUI
+
+    ##
+    #Copy objects to valid locations
+    ##
+    @javascript
+    Scenario: Copy one object without children objects
```
bdunogier: children aren't supposed to matter in copy, are they ? Children aren't copied by this operation, that would be subtree copy (unless of course the feature has evolved, in which case ignore).

bdunogier:
I think we're getting there :-)

Features/CopyMoveDelete/CopyContent.feature
pcardiga: Opss... and you asked well. Fate or Destiny might have been some unconscious thought related with making BDD sentences...

Yes, i mean "Destination".

Features/CopyMoveDelete/CopyContent.feature
pcardiga: Yes, what is being shown is a notification information. i'll update the sentence(s) to reflect this.

bdunogier: Thinking about it, I think there are various levels of messages in the notification pane. Maybe we should plan sentences for them, since these are supposed to be used a lot.

Features/CopyMoveDelete/CopyContent.feature
pcardiga: The _"I am viewing"_ came from specs, that says _"As an editor, I want to copy, move, or delete the object that I am viewing."_. But that was a criteria. 
But we will follow your suggestion and will update these sentences. All Given actions will be treated as _finished_ tasks and not so much as _in progress_,

Features/CopyMoveDelete/CopyContent.feature
pcardiga: We could use similar sentences such:
- Then I see a notification informing that "<message>"
- Then I see a notification warning that "<message>"
- Then I see a notification error informing that "<message>"
- ...

Features/CopyMoveDelete/CopyContent.feature
```diff
@@ -0,0 +1,31 @@
...
+        And I click the "Confirm selection" button
+        Then I see the message "'News Flash' has been successfully copied under 'Destiny'"
+        And I see "Destiny/News Flash" in content tree
+
+    @javascript
+    Scenario: Copy one object that has children objects
+        Given an "Destiny" folder exists
+        And an "Origin/News Flash" article exists
+        And I am viewing the "News Flash" article
+        When I copy the "News Flash" into the "Destiny" folder
```
bdunogier: I think we should either skip "the" when only referring to the name. Either:
- `When I copy the "News Flash" article to the "Destination" folder`
- or `When I copy "News flash" to the "Destination" folder`

Also applies to the one below.

Features/CopyMoveDelete/CopyContent.feature
```diff
@@ -0,0 +1,31 @@
...
+        Then the "News Flash" is copied
+        And I see "Origin/News Flash" in content tree
```
bdunogier: "In _the_ content tree" ?

Features/CopyMoveDelete/CopyContent.feature
bdunogier: I've talked with @dpobel, and we've concluded that it would be more correct like this:

`Then I am notified that "News Flash" was successfully copied to "Destination"`

At the moment, there are 3 types of notifications:
- something started
- something terminated
- an error occured

pcardiga: I prefer to leave the notification sentence due BDD implementation reasons
`Then I am notified that "'News Flash' has been successfully copied under 'Destination'"`

Otherwise, it would be less practical to know, without building some kind of interpretation method, that the notification sentence says... what we are saying it says.

pcardiga:
Scenario sentences updated accordingly some of the comments

Features/CopyMoveDelete/CopyContent.feature
bdunogier: There is one major blocker: those will be translated. 

Yes, it means more implementation, but not that much, in my opinion:

`I am notified that "News Flash" has been copied under "Destination"`

the implementation can then know about the exact sentence pattern it expects to find, and can use a generic method from the Notification context: 

```php
/**
 * @then /^I am notified that :contentName was copied to :destinationName$/
 */
function iAmNotifiedThatContentWasCopiedTo( $contentName, $destinationName )
{
  $notificationMessage = sprintf(
    "'%s' has been successfully copied under '%s'",
    $contentName, $destinationName
  );
  $this->iAmNotifiedThat( $notificationMessage );
}
```

The implementation details (the exact text) is hidden, and the user/scenarist only cares about the fact that he does get notified about the copy of the content under its destination.

I know it adds a bit of code, but frankly not that much, and more importantly: it is *very* simple code. Just more methods, but with almost no complexity.

pcardiga: But then again, we have the translations...
So that function you referred would be an incomplete solution. We would have to have some king of mapping or dictionary for **every** sentence/button/option/notification subject to change with translations.  
```PHP
(...)
$notificationMessage = sprintf(
    "'%s%s%s'",
    $contentName, $MessageTranslatedToLangXPTO, $destinationName
  );
(...)
```
This also applies for the validation of "Copy", "Move", "Send to Trash", etc.

Translations: Is that really the path we want to follow now?

pcardiga:
Scenario sentences updated

andrerom:
+1 we have imminent need to get overall coverage of main use cases of *all* features to avoid regressions as we a re moving quite fast now and seems to cause some regressions all the time, that would be caught if we had basic BDD in place covering all features.

So I'm +1 on this, and moving the open tasks here into QA tasks to build up a query able backlog of bdd work for "detailed scenarios".

bdunogier:
+1

PR #248, opened 2015-06-04
[Symfony 2.7] Removed deprecation warnings

lolautruche:
* Used configureOptions() instead of setDefaultOptions in form types
* Fixed deprecations in routing.yml
* Removed usage of security context

Tested locally, with unit tests and in browser.

### DIFF
--- a/Form/Type/SectionType.php
+++ b/Form/Type/SectionType.php
@@ -11,7 +11,7 @@ namespace EzSystems\PlatformUIBundle\Form\Type;
 
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\FormBuilderInterface;
-use Symfony\Component\OptionsResolver\OptionsResolverInterface;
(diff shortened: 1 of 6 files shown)


Form/Type/SectionType.php
```diff
@@ -11,6 +11,7 @@
 
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\OptionsResolver\OptionsResolverInterface;
```
andrerom: seems this can be removed now in most cases here

andrerom:
+1

dpobel:
+1

Form/Type/SectionType.php
lolautruche: Thanks. Fixed.

PR #233, opened 2015-05-22
EZP-24384: Add a language parameter in the viewLocation route

yannickroger:
Link: https://jira.ez.no/browse/EZP-24384

## Description
See issue.

## Test
Manual and unit tests

## TODO
- [x] Load rootlocation and media location
- [x] Load parent content
- [x] Add new usages added in the meantime.
- [x] add the languageCode to the object returned by getViewParameters and add the corresponding attribute the LocationView. see #251 
- [x] Fix tests
- [x] Add tests
- [x] Find out why bug was not found by unit test
- [x] Rebase

PR #251, opened 2015-06-10
EZP-23765: As a PlatformUI user I want to view different language ver…

mhyndle:
JIRA: https://jira.ez.no/browse/EZP-23765

# Description
The goal of the story is to allow user switch viewed language of content by using dropdown list with existing translations. After choosing other translation app should navigate to the same location view but with selected language parameter. There will be language indicator in upper right corner of "Content" section.

Screencast: https://youtu.be/MYP8XpRlSZc
# Tasks
- [x] Implementation
- [x] CSS
- [x] Tests
- [x] Rebase

# Related issues:
* https://jira.ez.no/browse/EZP-24384 - language parameter in the viewLocation route


Resources/public/templates/rawcontent.hbt
```diff
@@ -1,4 +1,5 @@
 <h2 class="ez-raw-content-title"><a href="#">Content</a></h2>
+<div class="ez-language-switcher"></div>
```
dpobel: usually we add the `-container` on such element so it should be called `ez-language-switcher-container`.

dpobel: There's also an inconsistency here, the language switcher is a sub view of the LocationView but its container is added to the RawContentView template. It's either a sub view of the LocationView and its container is added to the LocationView template or a sub view of the RawContentView and it's ok to add the container in the rawcontent.hbt template.

Resources/public/js/views/ez-locationviewview.js
```diff
@@ -213,6 +217,21 @@ YUI.add('ez-locationviewview', function (Y) {
...
+             * The language switcher view instance
+             *
+             * @attribute languageSwitcherView
+             * @type eZ.LanguageSwitcherView
+             */
+            languageSwitcherView: {
+                valueFn: function () {
+                    return new Y.eZ.LanguageSwitcherView({
+                        content: this.get('content'),
+                        location: this.get('location')
```
dpobel: you should add the parent view as a bubble target of the languageSwitcher so that if it fires a event, this event can be handled at a higher level (parent view, view service, application, ...). To do that, you can write something like:
```js
return new Y.eZ.LanguageSwitcherView({
    content: this.get('content'),
    location: this.get('location'),
    bubbleTargets: this,
});
```

Resources/public/js/views/ez-languageswitcherview.js
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * The raw content view
+     *
+     * @namespace eZ
+     * @class LanguageSwitcherView
+     * @constructor
+     * @extends eZ.TemplateBasedView
+     */
+    Y.eZ.LanguageSwitcherView = Y.Base.create('languageSwitcherView', Y.eZ.TemplateBasedView, [Y.eZ.AccordionElement], {
```
dpobel: for the language switcher, I would rather use the Y.eZ.Expandable extension instead, that way you already get the logic of adding/removing the `is-expanded` class and given that the additional languages list will be absolutely positioned we don't need the maths done in accordion element to achieve a nice transition.

Resources/public/js/views/ez-languageswitcherview.js
```diff
@@ -0,0 +1,120 @@
...
+
+        /**
+         * Returns array with translations other than already being viewed
+         *
+         * @method _getOtherTranslationsList
+         * @private
+         * @return {Array} other translations
+         */
+        _getOtherTranslationsList: function () {
+            //todo: get list of existing translations for current content (miss current translation?)
```
dpobel: actually, the languages list is available in the version not in the content. When the content is loaded, the version info are also available (under Content.CurrentVersion.Version in the REST Response), so you should probably change the Content model to store that part somewhere and add a method to the Content model to get the corresponding Version.

Resources/public/js/views/ez-languageswitcherview.js
```diff
@@ -0,0 +1,120 @@
...
+         *
+         * @method render
+         * @return {eZ.LanguageSwitcherView} the view itself
+         */
+        render: function () {
+            var container = this.get('container');
+
+            container.setHTML(this.template({
+                content: this.get('content').toJSON(),
+                //todo: get current translation and pass it here
```
dpobel: the current language is the one in the URI (currently added in https://github.com/ezsystems/PlatformUIBundle/pull/233) so it should probably be passed from the LocationViewViewService to the LocationView and then passed to the LanguageSwitcherView.

@yannickroger in https://github.com/ezsystems/PlatformUIBundle/pull/233, can you add the `languageCode` to the object returned by `getViewParameters` and add the corresponding attribute the LocationView.

@mhyndle for now, you can add the languageCode attribute to the locationView and hardcode it.

Resources/public/js/views/ez-languageswitcherview.js
```diff
@@ -0,0 +1,120 @@
...
+            var container = this.get('container');
+
+            container.one('.ez-expandable-area').toggleClass(EXPANDED_LANGUAGE_LIST);
+        },
+
+        _switchLanguage: function (e) {
+            var location = this.get('location'),
+                languageCode = e.target._node.getAttribute('data-language-code');
+
+            this.get('app').navigateTo(
```
dpobel: you don't need to do that, instead the language switcher view should generate the links (there's a `path` template function to generate proper links) so that we let the browser do the navigation.

Tests/js/views/assets/ez-languageswitcherview-tests.js
```diff
@@ -0,0 +1,366 @@
...
+            otherTranslationsLink = container.one('.ez-dropdown-list-indicator'),
+
+            otherTranslationsLink.simulateGesture('tap', function () {
+                that.resume(function () {
+                    Assert.isTrue(
+                        this.view.get('expanded'),
+                        'The expanded should state should be true'
+                    );
+                });
+            });
```
mhyndle: I have no idea why travis don't want to pass it...

mhyndle:
Ready for review

Resources/public/css/theme/views/languageswitcher.css
```diff
@@ -0,0 +1,46 @@
+/**
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+.ez-language-switcher-container {
```
dpobel: the empty rules can be removed

Resources/public/js/views/ez-languageswitcherview.js
```diff
@@ -0,0 +1,158 @@
...
+         * @method _getOtherTranslationsList
+         * @private
+         * @return {Array} other translations
+         */
+        _getOtherTranslationsList: function () {
+            var translations = this._getTranslationsList(),
+                languageCode = this.get('languageCode'),
+                idx;
+
+            idx = Y.Array.indexOf(translations, languageCode);
```
dpobel: this could be better written with something like:
```js
var languageCode = this.get('languageCode');
 
return Y.Array.reject(this._getTranslationList(), function (value) {
    return value === languageCode;
});
```

Resources/public/js/models/ez-contentmodel.js
```diff
@@ -316,6 +317,17 @@ YUI.add('ez-contentmodel', function (Y) {
...
+            },
+
+            /**
+             * The version info of the content
+             *
+             * @attribute versionInfo
+             * @type Object
+             * @default {}
+             */
+            versionInfo: {
```
dpobel: I would store the whole version instead of just the version info. You should also call it `currentVersion`

dpobel: and in addition, I would add a getter function on that attribute so that `get('currentVersion`) returns an eZ.Version object.

Resources/public/js/views/ez-languageswitcherview.js
```diff
@@ -0,0 +1,158 @@
...
+            return translations;
+        },
+
+        /**
+         * Returns array with all translations of content
+         *
+         * @return {Array} translations
+         */
+        _getTranslationsList: function () {
+            return this.get('content').get('versionInfo').languageCodes.split(',');
```
dpobel: this should be implemented in the version model instead

Resources/public/templates/languageswitcher.hbt
```diff
@@ -0,0 +1,23 @@
...
+            <span class="ez-language-flag flag-icon flag-icon-{{{ getFlagCode currentTranslation }}}"></span>
+            {{ currentTranslation }}
+        </a>
+    {{else}}
+        <span class="ez-language-flag flag-icon flag-icon-{{{ getFlagCode currentTranslation }}}"></span>
+        {{ currentTranslation }}
+    {{/if}}
+</div>
+{{#if otherTranslations}}
+    <div class="ez-expandable-area">
```
dpobel: semantically, this should be a list (ul/li).

Resources/public/templates/languageswitcher.hbt
```diff
@@ -0,0 +1,23 @@
+<div class="ez-selected-language">
+    {{#if otherTranslations}}
+        <a href="#" class="ez-dropdown-list-indicator" data-test="{{ location.id }}">
```
dpobel: `data-test` ?

Tests/js/views/ez-rawcontentview.html
```diff
@@ -19,6 +20,32 @@
 
 <script type="text/x-handlebars-template" id="fieldview-ez-template">fieldview</script>
 
+<script type="text/x-handlebars-template" id="languageswitcherview-ez-template">
+    <div class="ez-selected-language">
```
dpobel: usually in test we put a simplified and minimal version of the template. In this case, this would also avoid having to declare the ez-getflagcodehelper dependency.

Tests/js/views/assets/ez-languageswitcherview-tests.js
dpobel: the issue appears only on Travis, not on your own system ?

Resources/public/css/theme/views/languageswitcher.css
```diff
@@ -0,0 +1,46 @@
...
+}
+
+.ez-language-switcher-container .ez-dropdown-list-indicator {
```
dpobel: the `.ez-language-switcher-container` element does not really belong to the language switcher so you should not style anything with it. Instead, you should use the automatically generated class on the view container which should be something like `.ez-view-languageswitcherview`

Tests/js/views/assets/ez-languageswitcherview-tests.js
mhyndle: Actually after updating npm dependencies and bower my test says the same so I'll dig into it bit more.

Resources/public/js/models/ez-contentmodel.js
```diff
@@ -316,6 +317,21 @@ YUI.add('ez-contentmodel', function (Y) {
...
+
+            /**
+             * The current version of the content
+             *
+             * @attribute currentVersion
+             * @type eZ.Version
+             */
+            currentVersion: {
+                getter: function (value) {
+                    var version = new Y.eZ.Version();
```
dpobel: so now the module ez-contentmodel should require ez-versionmodel in yui.yml

Tests/js/models/assets/ez-contentmodel-tests.js
```diff
@@ -388,6 +417,8 @@ YUI.add('ez-contentmodel-tests', function (Y) {
         setUp: function () {
             var that = this;
 
+            Y.eZ.Version = Y.Base.create('versionModel', Y.eZ.RestModel, [], {});
```
dpobel: actually you should load the ez-versionmodel module as it's now real dependency of ez-contentmodel (same as https://github.com/ezsystems/PlatformUIBundle/pull/251/files#r33257614 actually)

Tests/js/views/ez-languageswitcherview.html
```diff
@@ -0,0 +1,90 @@
...
+
+<div class="click-somewhere"></div>
+
+<div class="container"></div>
+
+<script type="text/x-handlebars-template" id="languageswitcherview-ez-template">
+    <div class="ez-selected-language">
+        {{#if otherTranslations}}
+            <a href="#" class="ez-dropdown-list-indicator" data-test="{{ location.id }}">
+                <span class="ez-language-flag flag-icon flag-icon-{{{ getFlagCode currentTranslation }}}"></span>
```
dpobel: here as well, the template can be simplified. Please remove any call to getFlagCode (it's not needed for the test) and this allows to remove the helper dependency below.

Tests/js/views/ez-languageswitcherview.html
```diff
@@ -0,0 +1,90 @@
...
+            },
+            "ez-view": {
+                requires: ['view', 'base-pluginhost', 'ez-pluginregistry'],
+                fullpath: "../../../Resources/public/js/views/ez-view.js"
+            },
+            'ez-expandable': {
+                requires: ['view'],
+                fullpath: "../../../Resources/public/js/extensions/ez-expandable.js"
+            },
+            'ez-texthelper': {
```
dpobel: does not seem to be needed anywhere

Tests/js/views/ez-rawcontentview.html
```diff
@@ -19,6 +20,32 @@
...
+        {{#if otherTranslations}}
+            <a href="#" class="ez-dropdown-list-indicator" data-test="{{ location.id }}">
+                <span class="ez-language-flag flag-icon"></span>
```
dpobel: the span can be removed as well

dpobel: actually in this test I think the template for the language switch can even be an empty string

Resources/public/css/views/languageswitcher.css
```diff
@@ -0,0 +1,37 @@
...
+    margin: 0;
+    padding: 0;
+    list-style: none;
+}
+
+.ez-view-languageswitcherview.is-expanded .ez-expandable-area {
+    display: block;
+}
+
+.ez-view-languageswitcherview .ez-content-translation a {
```
dpobel: please do not style based on tag name to avoid the strong coupling between the CSS and the markup. Instead you should add a class on the links and use it here (and in the theme stylesheet)

dpobel:
Nice work :)
Besides the inline comments, I've noticed a few issues:

* The flag of the current translation and the flags in the list are not aligned flags
* A CSS transition when the list (dis)appears would be nice. Since the list is absolutely positioned and appears like a vertical list, a transition on a scaleY transform seems appropriate and quite easy to do. So basically, in the theme stylesheet you should set the list with `display: block` but hide it with the `transform: scaleY(0)`, then when it's supposed to be displayed you have to set `transform: scaleY(1)` and to a get a nice transition

  ```css
    -webkit-transition: all 0.3s ease;
            transition: all 0.3s ease;
  ```

  would do the trick.
* The getter for eZ.Content currentVersion attribute is not really tested. You should check that it returns an instance of eZ.Version with its attributes consistent with the *raw* value of the currentVersion attribute.
* eZ.Version.getTranslationsList is not tested
* the RawContentView instantiated in LocationViewView now receives some new parameters, this should be tested
* The ez-getflagcodehelper is not tested at all.

mhyndle:
I think I haven't miss anything from @dpobel 's comments, so please review it again (I will rebase commits after positive review). Added the screencast which is presenting the behaviour of the language switcher, however navigating to translations is not working as it's waiting for this one: https://jira.ez.no/browse/EZP-24384. However, there shouldn't be need to fix anything after that in language switcher as `path` helper is already called with `languageCode` parameter.

Resources/public/js/models/ez-contentmodel.js
mhyndle: I have added `ez-versionmodel` to dependencies of `ez-contentmodel` in yui.yml

Tests/js/models/assets/ez-contentmodel-tests.js
mhyndle: I have added `ez-versionmodel` dependancy for `ez-contentmodel` test in `Tests/js/models/ez-contentmodel.html`

Resources/public/css/theme/views/languageswitcher.css
```diff
@@ -0,0 +1,47 @@
...
+}
+
+.ez-view-languageswitcherview .ez-expandable-area {
+    border: 1px solid #ccc;
+    box-shadow: 0 1px 5px rgba(0, 0, 0, 0.3);
+    background: #fff;
+    display: block;
+    transform: scaleY(0);
+    transform-origin: 50% 0%;
+    -webkit-transform-origin: 50% 0%;
```
dpobel: please order, group and align the rules like (and `-webkit-transform` is missing and `-webkit-transition` is not needed):
```css
        transform: scaleY(0);
-webkit-transform: scaleY(0);
        transform-origin: 50% 0%;
-webkit-transform-origin: 50% 0%;
transition: all 0.3s ease;
```

Resources/public/css/theme/views/languageswitcher.css
```diff
@@ -0,0 +1,47 @@
...
+    -webkit-transition: all 0.3s ease;
+    transition: all 0.3s ease;
+}
+
+.ez-view-languageswitcherview.is-expanded .ez-expandable-area {
+    transform: scaleY(1);
```
dpobel: missing `-webkit-transform` here as well

Tests/js/models/assets/ez-contentmodel-tests.js
dpobel: so the fake `Y.eZ.Version` can be removed, no ?

Resources/public/css/theme/views/languageswitcher.css
dpobel: after testing it, 0.3s is maybe a bit slow in this case, I think 0.2s would give a better result.

dpobel:
almost there! One last nitpick on the translation list CSS, again on the alignement but this time for the small arrow:

flags

Besides that and the last inline comments, +1

crevillo:
to flag or not to flag, that'ts the question... :)

mhyndle:
Rebased

dpobel:
+1

PR #254, opened 2015-06-11
EZP-23907: disable create content button when in context of non conta…

mhyndle:
JIRA: https://jira.ez.no/browse/EZP-23907
# Description
When viewing content that is not container, "Create content" button should be disabled as it shouldn't be allowed to create content inside non-container contents. The goal is to check if contentType of content is container and if it's not, disable "Crate content" button.

# Tasks
- [x] Implementation
- [x] Manual tests

### DIFF
--- a/Resources/public/js/views/ez-actionbarview.js
+++ b/Resources/public/js/views/ez-actionbarview.js
@@ -59,7 +59,7 @@ YUI.add('ez-actionbarview', function (Y) {
                         }),
                         new Y.eZ.CreateContentActionView({
                             actionId: 'createContent',
-                            disabled: false,
+                            disabled: !this.get('contentType').get('isContainer'),
                             label: 'Create a content',
                             priority: 210
                         }),
@@ -71,6 +71,17 @@ YUI.add('ez-actionbarview', function (Y) {
                         }),
                     ];
                 }
+            },
+
+            /**
+             * The content type of the content at the current location
+             *
+             * @attribute contentType
+             * @type Y.eZ.ContentType
+             * @writeOnce
+             */
+            contentType: {
+                writeOnce: "initOnly",
             },
         }
     });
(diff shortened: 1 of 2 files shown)


crevillo:
maybe for another task, but it would be good to disable this create content button if contentType is container but you don't have rights to create content under that location.

mhyndle:
I agree, it's out of the scope for this JIRA bug, so probably new issue should be created for that.

andrerom:
@crevillo we have a epic tracking that: https://jira.ez.no/browse/EZP-23738
TL;DR; Will require changes / additions to REST API, so this is not within scope for July Milestone (Pollux), but rather as soon as possible after.

crevillo:
@andrerom problem is i don't know how to get notifications about new issues added to tracker.

andrerom:
well, are you sure you want that? :) (notifications on all JIRA issues)

What we could do is tweak the dashboards, it already shows the last issues in open state.
And a change on share.ez.no to reflect the epics instead of the old roadmap solution is coming.

brookinsconsulting:
@andrerom 

We @brookinsconsulting would seriously very much like to get in on that offer for 'notifications on all JIRA issues'!

Also thank you for sharing the info about epics on share.ez.no

Cheers,
Brookins Consulting

andrerom:
Good morning Graham, we'll keep that in mind.

mhyndle:
Did manual tests. I'm not sure if there should be tests scenarios written for it as there is not much logical operations in it. I suppose that it should be tested somehow but have no idea how to do it.

andrerom:
+1

Resources/public/js/views/ez-actionbarview.js
```diff
@@ -59,7 +59,7 @@ YUI.add('ez-actionbarview', function (Y) {
                         }),
                         new Y.eZ.CreateContentActionView({
                             actionId: 'createContent',
-                            disabled: false,
+                            disabled: !this.get('contentType').get('isContainer'),
```
dpobel: I would move this logic to the CreateContentActionView meaning it should receive the contentType and set its `disabled` flag depending on it.

dpobel:
@andrerom: I guess you merge it by mistake ?

@mhyndle this should be unit tested. You need to write 3 tests:

1. one to check that the LocationViewView instantiates the ActionBarView with the contentType
1. one to check that the ActionBarView instantiates the CreateContentActionView with the contentType
1. one to check that the disabled flag is correctly set depending on the contentType

PR #255, opened 2015-06-15
EZP-23907: PlatformUI allows to create a content under non container content

mhyndle:
JIRA: https://jira.ez.no/browse/EZP-23907
Continuation of https://github.com/ezsystems/PlatformUIBundle/pull/254 as it was merged too early.

# Description
When viewing content that is not container, "Create content" button should be disabled as it shouldn't be allowed to create content inside non-container contents. The goal is to check if contentType of content is container and if it's not, disable "Create content" button.

# Tasks
- [x] Implementation
- [x] Tests

### DIFF
--- a/Resources/public/js/views/actions/ez-createcontentactionview.js
+++ b/Resources/public/js/views/actions/ez-createcontentactionview.js
@@ -23,6 +23,7 @@ YUI.add('ez-createcontentactionview', function (Y) {
      */
     Y.eZ.CreateContentActionView = Y.Base.create('createContentActionView', Y.eZ.ButtonActionView, [Y.eZ.Expandable], {
         initializer: function () {
+            this.set('disabled', !this.get('contentType').get('isContainer'));
             this.get('contentTypeSelectorView').addTarget(this);
             this.after({
                 'contentTypeGroupsChange': this._renderContentTypeSelector,
@@ -127,6 +128,17 @@ YUI.add('ez-createcontentactionview', function (Y) {
                 valueFn: function () {
                     return new Y.eZ.ContentTypeSelectorView();
                 }
+            },
+
+            /**
+             * The content type of the content at the current location
+             *
+             * @attribute contentType
+             * @type Y.eZ.ContentType
+             * @writeOnce
+             */
+            contentType: {
+                writeOnce: "initOnly",
             },
         }
     });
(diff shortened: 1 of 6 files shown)


Resources/public/js/views/ez-actionbarview.js
```diff
@@ -59,9 +59,10 @@ YUI.add('ez-actionbarview', function (Y) {
                         }),
                         new Y.eZ.CreateContentActionView({
                             actionId: 'createContent',
-                            disabled: !this.get('contentType').get('isContainer'),
+                            disabled: false,
```
dpobel: I think you should even omit it since it will be computed in the view.

dpobel:

mhyndle:
I think I'm lacking last test, which I'm currently working on. Test should check if ActionBarView instantiates the CreateContentActionView with the contentType. And for this one I'm currently writing new ez-actionbarview-tests - can you confirm that's proper place for checking it or maybe I should use one of already existing tests?

dpobel:
@mhyndle that's perfectly right

Resources/public/js/views/actions/ez-createcontentactionview.js
```diff
@@ -23,6 +23,12 @@ YUI.add('ez-createcontentactionview', function (Y) {
      */
     Y.eZ.CreateContentActionView = Y.Base.create('createContentActionView', Y.eZ.ButtonActionView, [Y.eZ.Expandable], {
         initializer: function () {
+            if (!this.get('contentType').get('isContainer')) {
```
dpobel: can be simplifed in:
```js
this.set('disabled', !this.get('contentType').get('isContainer'));
```

mhyndle:
Ready for review

Tests/js/views/assets/ez-actionbarview-tests.js
```diff
@@ -0,0 +1,46 @@
...
+        },
+
+        tearDown: function () {
+            delete Y.eZ.CreateContentActionView;
+            this.view.destroy();
+        },
+
+        "Should instantiate createContentActionView with contentType": function () {
+            for (var i = 0; i < this.view.get('actionsList').length; i++){
+                if (this.view.get('actionsList')[i].get('actionId') === 'createContent'){
```
dpobel: there's a small issue here, if the `createContent` action is not found the test should fail.

dpobel:
+1

PR #260, opened 2015-06-24
Fix EZP-24450: As an admin developer, I want to push notifications using a Pjax controller

lolautruche:
> https://jira.ez.no/browse/EZP-24450

Added a new block in Pjax base template for notifications.
This block is automatically filled by flash session messages with `notification` as identifier.

### DIFF
--- a/Resources/views/pjax_admin.html.twig
+++ b/Resources/views/pjax_admin.html.twig
@@ -25,3 +25,13 @@
     </header>
     {% block content %}{% endblock %}
 </div>
+
+{% block notification %}
+<ul data-name="notification">
+    {% set notifications = app.session.flashBag.get("notification") %}
+    {% for notification in notifications %}
+        <li data-state="{{ notification.state }}">{{ notification.message }}</li>
+
+    {% endfor %}
+</ul>
+{% endblock %}
--- a/Form/Processor/ContentTypeFormProcessor.php
+++ b/Form/Processor/ContentTypeFormProcessor.php
@@ -10,6 +10,8 @@
 namespace EzSystems\PlatformUIBundle\Form\Processor;
 
 use eZ\Publish\API\Repository\Values\ContentType\ContentTypeDraft;
+use EzSystems\PlatformUIBundle\Notification\NotificationPoolInterface;
+use EzSystems\PlatformUIBundle\Notification\TranslatableNotificationMessage;
 use EzSystems\RepositoryForms\Event\FormActionEvent;
 use EzSystems\RepositoryForms\Event\RepositoryFormEvents;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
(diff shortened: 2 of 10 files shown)


lolautruche:
Review ping @andrerom @dpobel @bdunogier @yannickroger @glye

Resources/views/pjax_admin.html.twig
```diff
@@ -25,3 +25,19 @@
     </header>
     {% block content %}{% endblock %}
 </div>
+
+{% block notification %}
+{# Possible states are "error", "started", and "done" #}
+<div data-name="notification" data-state="{{ notification_state|default("done") }}">
+    {% set notifications = app.session.flashBag.get("notification") %}
+    {% if notifications|length == 1 %}
+        {{ notifications[0] }}
```
crevillo: i would enclose this in ```<p>``` tag

lolautruche: No this is a convention that will be used by JS code

crevillo: ok then...

Resources/views/pjax_admin.html.twig
```diff
@@ -25,3 +25,19 @@
     </header>
     {% block content %}{% endblock %}
 </div>
+
+{% block notification %}
+{# Possible states are "error", "started", and "done" #}
```
crevillo: unneded comma before and? :)

dpobel: actually I would not mention the possible states as we'll probably add new ones and this list will eventually become wrong

bdunogier: +1 with @dpobel. Maybe it could reference where the possible values should be looked for (and are maintained).

Resources/views/pjax_admin.html.twig
bdunogier: Don't feel sad, Carlos, conventions happen :-)

dpobel: actually this comment makes me think we need to think a bit more of the case where we have several messages vs. the case where there's only one.

lolautruche: You can already have several messages, but with only one state.
And Pjax is synchronous, so there should not be any problem with that AFAICS. However, we should somehow be able to trigger notifications using JS.

bdunogier:
+1 with [my comment](https://github.com/ezsystems/PlatformUIBundle/pull/260/files#r33164932) fixed and @dpobel's approval :-)

Resources/views/pjax_admin.html.twig
lolautruche: OK I will remove the comment.

> Maybe it could reference where the possible values should be looked for (and are maintained).

We currently don't have such place AFAIK.

Resources/views/pjax_admin.html.twig
```diff
@@ -25,3 +25,18 @@
     </header>
     {% block content %}{% endblock %}
 </div>
+
+{% block notification %}
+<div data-name="notification" data-state="{{ notification_state|default("done") }}">
```
dpobel: after thinking a bit about it, the `data-name="notification"` element should probably be a `ul` and each notification text should be in a `li` (no matter if there's one or more notifications)
But I'm also bothered by the fact there's only a state for the list of notifications, isn't it possible to get several messages with different state ? In addition, the `notification_state` variable seems a bit magic, how does it get there ?

lolautruche: Can be a `ul` yes. And yes, potentially there might be different messages in different states, even though it's not currently discriminated.

As for `notification_state`, it can be overriden in a child block. If not defined, it defaults to `done`. That's no magic at all :wink:

lolautruche:
PR updated :-)
I closed https://github.com/ezsystems/repository-forms/pull/33 as now everything is in this one.

dpobel:
+1

pspanja:
Looks good, +1.

PR #261, opened 2015-06-25
EZP-24502: implement a content type server side view (UDW for relations)

mhyndle:
Jira: https://jira.ez.no/browse/EZP-24502

# Description
The goal of this PR is to create the handler for button "Select Item" (for picking the default root location). Handler should call Universal Discovery Widget to allow choosing single location which will be passed back and set into the input form.

# Tasks
- [x] Implementation
- [x] Tests
- [x] Rebase

# Related issues:
* https://jira.ez.no/browse/EZP-24446 - https://github.com/ezsystems/repository-forms/pull/31
* https://jira.ez.no/browse/EZP-24447 - https://github.com/ezsystems/repository-forms/pull/32

### DIFF
--- a/Resources/config/yui.yml
+++ b/Resources/config/yui.yml
@@ -31,6 +31,7 @@ system:
                         - 'ez-locationviewview'
                         - 'ez-locationviewviewservice'
                         - 'ez-serversideview'
+                        - 'ez-contenttypeeditserversideview'
                         - 'ez-serversideviewservice'
                         - 'ez-sectionserversideview'
                         - 'ez-sectionserversideviewservice'
@@ -670,6 +671,9 @@ system:
                 ez-serversideview:
                     requires: ['ez-view', 'ez-tabs', 'ez-selection-table', 'event-tap']
                     path: %ez_platformui.public_dir%/js/views/ez-serversideview.js
+                ez-contenttypeeditserversideview:
+                    requires: ['ez-serversideview', 'event-tap']
+                    path: %ez_platformui.public_dir%/js/views/serverside/ez-contenttypeeditserversideview.js
                 ez-sectionserversideview:
                     requires: ['ez-serversideview', 'event-tap']
                     path: %ez_platformui.public_dir%/js/views/serverside/ez-sectionserversideview.js
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -119,6 +119,9 @@ YUI.add('ez-platformuiapp', function (Y) {
             sectionServerSideView: {
                 type: Y.eZ.SectionServerSideView,
             },
+            contentTypeEditServerSideView: {
+                type: Y.eZ.ContentTypeEditServerSideView,
+            },
         },
 
         /**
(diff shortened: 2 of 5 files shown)


Resources/public/js/views/serverside/ez-contenttypeserversideview.js
```diff
@@ -0,0 +1,95 @@
...
+        /**
+         * Puts picked location into field value
+         *
+         * @method _setRoot
+         * @protected
+         * @param {Y.Node} button
+         * @param {EventFacade} e
+         */
+        _setRoot: function (button, e) {
+            var selectionRoot = button.ancestor('.selection-root');
```
mhyndle: I'm wondering if it's proper way of choosing proper input for update with location id. Or maybe it should be independent from the HTML structure and I should use data-field-defnition-id attributes to select proper input?
ping @dpobel

Resources/config/yui.yml
```diff
@@ -31,6 +31,7 @@ system:
                         - 'ez-locationviewview'
                         - 'ez-locationviewviewservice'
                         - 'ez-serversideview'
+                        - 'ez-contenttypeserversideview'
```
dpobel: actually you should introduce the ContentTypeEditServerSideView as what you are developing is completely specific to the content type edition

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -766,7 +769,7 @@ YUI.add('ez-platformuiapp', function (Y) {
                     path: "/admin/:uri",
                     sideViews: {'navigationHub': true, 'discoveryBar': false},
                     service: Y.eZ.ServerSideViewService,
-                    view: "serverSideView",
+                    view: "contentTypeServerSideView",
```
dpobel: related to https://github.com/ezsystems/PlatformUIBundle/pull/261/files#r33333706 you should add a new route that matches only the content type edition

Resources/public/js/views/serverside/ez-contenttypeserversideview.js
```diff
@@ -0,0 +1,95 @@
...
+
+        /**
+         * Changes the state of the provided root selection button to be *loading* and
+         * to be disabled
+         *
+         * @method _uiSetPickRootLoading
+         * @protected
+         * @param {Y.Node} button
+         */
+        _uiSetPickRootLoading: function (button) {
```
dpobel: you don't really need that in this case, because once the user has chosen a location in the UDW, we are done, there's no asynchronous process after.

Resources/public/js/views/serverside/ez-contenttypeserversideview.js
dpobel: I think the button should have a data attribute that contains a selector to find the input to update at the end of the selection process. That way, we can use the very same code for both ezrelation and ezrelationlist

Resources/public/js/views/serverside/ez-contenttypeeditserversideview.js
```diff
@@ -0,0 +1,66 @@
...
+         * provided in `data-ralation-root-input-selector` attribute of button, for example
+         * <button data-ralation-root-input-selector="#id_of_input"></button>
+         *
+         * @method _setRoot
+         * @protected
+         * @param {Y.Node} button
+         * @param {EventFacade} e
+         */
+        _setRoot: function (button, e) {
+            var selectionRootInput = Y.one(button.getAttribute('data-ralation-root-input-selector'));
```
dpobel: `ralation` ? :)
You should not use `Y.one` but `this.get('container').one`

mhyndle:
Ready for review
ping @dpobel @andrerom @yannickroger @StephaneDiot

Resources/public/js/views/serverside/ez-contenttypeeditserversideview.js
```diff
@@ -0,0 +1,66 @@
...
+         *
+         * @method _setRoot
+         * @protected
+         * @param {Y.Node} button
+         * @param {EventFacade} e
+         */
+        _setRoot: function (button, e) {
+            var selectionRootInput = this.get('container').one(button.getAttribute('data-relation-root-input-selector'));
+
+            selectionRootInput.setAttribute('value', e.selection.location.get('locationId'));
```
dpobel: we should also probably display the content name near the button so that the user can see that his action was taken into account (btw the same should also be done if the root is already filled, ping @glye)

dpobel: I've also just realized that you are putting the `id` attribute in the hidden field, while it probably expects the `locationId` (ie `42` not `/api/ezp/v2/content/locations/1/2/42`)

mhyndle:
Added the behaviour of update  the content's name after selecting the content with UDW.
ping @dpobel

dpobel:
+1

mhyndle:
rebased

PR #262, opened 2015-06-30
EZP-24438: As a UI developer, I want to be able to add new tabs in the locationView

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24438

# Description

The patch makes the tabs in the location view extensible so we and the extension developers can implement new tabs and add/remove the displayed tabs.

## Tasks

* [x] base system
* [x] add an API for extension developers
* [x] add a `selected` attribute to the tab base class and make it in sync the tab state

# Tests

unit tests + manual tests


dpobel:
ready to be reviewed
ping @yannickroger @mhyndle @StephaneDiot

dpobel:
actually I've just realized I forgot to add a `selected` attribute to the location view tab view so that it's easy know when a tab becomes selected.

yannickroger:
+1

PR #263, opened 2015-06-30
EZP-24536: Parse and display the notifications send in the admin part

mhyndle:
Jira: https://jira.ez.no/browse/EZP-24536

# Description:
The goal of this story is to parse notifications taken from Pjax controller as HTML block and display them usin PlatformUI notification system.

# Tasks:
- [x] Implementation
- [x] Tests
 - [x] Manual Tests
 - [x] Unit tests
- [x] Rebase

# Related issues:
* https://jira.ez.no/browse/EZP-24450 - As an admin developer, I want to push notifications using a Pjax controller
 * PR: https://github.com/ezsystems/PlatformUIBundle/pull/260

### DIFF
--- a/Resources/public/js/views/services/ez-serversideviewservice.js
+++ b/Resources/public/js/views/services/ez-serversideviewservice.js
@@ -118,6 +118,7 @@ YUI.add('ez-serversideviewservice', function (Y) {
          */
         _parseResponse: function (response) {
             var frag = Y.Node.create(response.responseText),
+                that = this,
                 html, title;
 
             html = frag.one('[data-name="html"]');
@@ -129,6 +130,10 @@ YUI.add('ez-serversideviewservice', function (Y) {
             if ( title ) {
                 this.set('title', title.get('text'));
             }
+
+            frag.all('[data-name="notification"] li').each(function (notificationNode) {
+                that._responseNotify(notificationNode);
+            });
         },
 
         /**
@@ -158,6 +163,33 @@ YUI.add('ez-serversideviewservice', function (Y) {
                 link.setAttribute('href', app.routeUri('adminGenericRoute', {uri: href}));
             });
             return node;
+        },
+
+        /**
+         * Fires notify event basing on node
+         *
+         * @method _responseNotify
+         * @protected
+         * @param {Node} node
+         */
+        _responseNotify: function (node) {
+            var app = this.get('app'),
+                timeout = 5;
+
+            if (node.getAttribute('data-state') === 'error') {
+                timeout = 0;
+            }
+
+            // the app is not yet a bubble target of the view service,
+            // so we are using the app to fire the event
+            // see https://jira.ez.no/browse/EZP-23013
+            app.fire('notify', {
+                notification: {
+                    text: node.getContent(),
+                    state: node.getAttribute('data-state'),
+                    timeout: timeout
+                }
+            });
         },
 
         /**
--- a/Tests/js/views/services/assets/ez-serversideviewservice-tests.js
+++ b/Tests/js/views/services/assets/ez-serversideviewservice-tests.js
@@ -3,7 +3,7 @@
  * For full copyright and license information view LICENSE file distributed with this source code.
  */
 YUI.add('ez-serversideviewservice-tests', function (Y) {
-    var unitTest, rewriteTest, formTest,
+    var unitTest, rewriteTest, formTest, notificationTest,
         Mock = Y.Mock, Assert = Y.Assert;
 
     unitTest = new Y.Test.Case({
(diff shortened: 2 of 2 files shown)


Resources/public/js/views/services/ez-serversideviewservice.js
```diff
@@ -118,10 +118,12 @@ YUI.add('ez-serversideviewservice', function (Y) {
          */
         _parseResponse: function (response) {
             var frag = Y.Node.create(response.responseText),
-                html, title;
+                that = this,
+                html, title, notifications;
 
             html = frag.one('[data-name="html"]');
             title = frag.one('[data-name="title"]');
+            notifications = frag.one('[data-name="notification"]');
```
dpobel: you don't really need this Node actually, so instead of the `if` condition below, you can write something like:

```js
frag.all('[data-name="notification"] li').each(function (notificationNode) {
    that._responseNotify(notificationNode);
});

File: Resources/public/js/views/services/ez-serversideviewservice.js [hunk 2]
```diff
@@ -161,6 +168,25 @@ YUI.add('ez-serversideviewservice', function (Y) {
...
+         * @param {Node} node
+         */
+        _responseNotify: function (node) {
+            console.log(node);
+            this.fire('notify', {
+                notification: {
+                    text: node.getContent(),
+                    identifier: Math.random(),
+                    state: node.getAttribute('data-state'),
+                    timeout: 0,
```
dpobel: the timeout should be set to 0 only on error notification, for the others you can set it to 5 (we should introduce a constant for that at some point...)

mhyndle:
Ready for review
ping @dpobel @yannickroger @andrerom

Tests/js/views/services/assets/ez-serversideviewservice-tests.js
```diff
@@ -335,8 +335,138 @@ YUI.add('ez-serversideviewservice-tests', function (Y) {
...
+        _notificationLoad: function (notificationState, notificationText) {
+            var pjaxResponse = '<div data-name="title"></div>' +
+                    '<div data-name="html"></div>' +
+                    '<ul data-name="notification">' +
+                    '<li data-state="' + notificationState +'">' + notificationText + '</li>' +
+                    '</ul>',
+                uri = 'echo/get/html/?response='+ Y.config.win.encodeURIComponent(pjaxResponse),
+                request = {'params': {'uri': uri}};
+
+            this.service = new Y.eZ.ServerSideViewService({
```
dpobel: this should be done in the `setUp` to separate the initialization from the actual test

Tests/js/views/services/assets/ez-serversideviewservice-tests.js
```diff
@@ -335,8 +335,138 @@ YUI.add('ez-serversideviewservice-tests', function (Y) {
...
+                request: request,
+                app: this.app
+            });
+        },
+
+        tearDown: function () {
+            delete this.service;
+            delete this.app;
```
dpobel: you should also call `this.service.destroy()` and `this.app.destroy()`

Tests/js/views/services/assets/ez-serversideviewservice-tests.js
```diff
@@ -335,8 +335,138 @@ YUI.add('ez-serversideviewservice-tests', function (Y) {
...
     });
 
+    notificationTest = new Y.Test.Case({
+        name: "eZ Server Side View Service notification test",
+
+        setUp: function () {
+            this.baseUri = '/Tests/js/views/services/';
+            this.app = new Y.Base();
+
+            Mock.expect(this.app, {
```
dpobel: `this.app` is not mock, instead you should just do:
```js
this.app.set('baseUri', this.baseUri);
```

Resources/public/js/views/services/ez-serversideviewservice.js
```diff
@@ -161,6 +166,32 @@ YUI.add('ez-serversideviewservice', function (Y) {
         },
 
         /**
+         * Fires notify event basing on node.
+         * notify event does not bubble to app so triggering it directly on app instance
```
dpobel: it's an implementation details so I would put that inline right above the `app.fire` call. And the text should something like:

```js
// the app is not yet a bubble target of the view service,
// so we are using the app to fire the event
// see https://jira.ez.no/browse/EZP-23013
```

mhyndle:
fixes done
ping @dpobel 
(will rebase it after positive review)

Resources/public/js/views/services/ez-serversideviewservice.js
```diff
@@ -161,6 +166,33 @@ YUI.add('ez-serversideviewservice', function (Y) {
         },
 
         /**
+         * Fires notify event basing on node.
+         * the app is not yet a bubble target of the view service,
```
dpobel: you didn't get [my previous comment](https://github.com/ezsystems/PlatformUIBundle/pull/263#discussion_r33790718) right. This comment is about an implementation detail so it should not be in the doc block of the method but just above the corresponding code as inline comment (with `//`)

dpobel:
besides the last comment, +1

PR #265, opened 2015-07-01
EZP-24438: As a UI developer, I want to be able to add new tabs in the locationView

dpobel:
Same as https://github.com/ezsystems/PlatformUIBundle/pull/262 (was merged to quickly and then reverted so that it's merged after https://github.com/ezsystems/PlatformUIBundle/pull/251 to ease conflicts resolution)


yannickroger:
still +1 then

PR #267, opened 2015-07-02
Make to sure the prevent the tap event handle the button action view

dpobel:
# Description

Make sure to prevent default behavior when the user uses a Button Action View, otherwise, if the view is rendered inside a form, the form is submitted. (This happening for instance in the RichText focus mode).

# Tests

manual + unit tests

### DIFF
--- a/Tests/js/views/actions/assets/genericbuttonactionview-tests.js
+++ b/Tests/js/views/actions/assets/genericbuttonactionview-tests.js
@@ -74,6 +74,12 @@ YUI.add('ez-genericbuttonactionview-tests', function (Y) {
             var that = this,
                 actionFired = false;
 
+            this.view.get('container').once('tap', function (e) {
+                Y.Assert.isTrue(
+                    !!e.prevented,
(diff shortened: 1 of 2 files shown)


dpobel:
ping @yannickroger @mhyndle

Tests/js/views/actions/assets/genericbuttonactionview-tests.js
```diff
@@ -74,6 +74,12 @@ YUI.add('ez-genericbuttonactionview-tests', function (Y) {
             var that = this,
                 actionFired = false;
 
+            this.view.get('container').once('tap', function (e) {
+                Y.Assert.isTrue(
+                    !!e.prevented,
```
mhyndle: Why not just `e.prevented` (without `!!`)?

dpobel: because e.prevented is an integer (1 in this case), so I do that to be able to use Assert.isTrue

mhyndle:
+1

PR #269, opened 2015-07-02
Make sure to parse the current version only when there's one

dpobel:
# Description

This fixes a regression introduced in https://github.com/ezsystems/PlatformUIBundle/pull/251. Without this patch, it's impossible to create a new content because the Content model tries to parse a non existent version.

# Tests

manual + unit


dpobel:
ping @mhyndle

mhyndle:
+1

PR #270, opened 2015-07-02
Fixed a crash after publication if the discovery bar has never been instantiated

dpobel:
# Description

If you go directly to the edit content page (for instance with a browser bookmark), the publication will crash because the update tree plugin tries to use the discovery bar view but it hasn't been created yet.

This patch makes sure we handle this case.

# Tests

manual test + unit test


dpobel:
ping @yannickroger @mhyndle

mhyndle:
+1

PR #271, opened 2015-07-02
Make sure to forward the active state to the location view tab

dpobel:
# Description

The `active` state is not fowarded to the location view tabs (this was forgotten in https://github.com/ezsystems/PlatformUIBundle/pull/265) as a result, the asynchronous field view (image, relation, relation list) are broken.

This patch makes sure the `active` flag value is forwarded to the tabs.

# Tests

manual test + unit test


dpobel:
ping @yannickroger @mhyndle

yannickroger:
+1

mhyndle:
+1

PR #278, opened 2015-07-10
EZP-24549: edit content in given language

mhyndle:
Jira: https://jira.ez.no/browse/EZP-24549

# Description
When viewing the content we are viewing it in given language. When clicking the "Edit" button placed on Action bar we should edit the content in currently selected language.

# Tasks
- [x] Implementation
- [x] Tests
- [x] Rebase

### DIFF
--- a/Resources/public/js/views/services/ez-contenteditviewservice.js
+++ b/Resources/public/js/views/services/ez-contenteditviewservice.js
@@ -24,6 +24,9 @@ YUI.add('ez-contenteditviewservice', function (Y) {
     Y.eZ.ContentEditViewService = Y.Base.create('contentEditViewService', Y.eZ.ViewService, [], {
         initializer: function () {
             this.on('*:closeView', this._handleCloseView);
+            this.after('*:requestChange', this._setLanguageCode);
+
+            this._setLanguageCode();
         },
 
         /**
@@ -36,10 +39,11 @@ YUI.add('ez-contenteditviewservice', function (Y) {
          */
         _load: function (next) {
             var request = this.get('request'),
-                service = this;
+                service = this,
+                languageCode = this.get('languageCode');
 
             this.get('version').reset();
-            this._loadContent(request.params.id, function () {
+            this._loadContent(request.params.id, languageCode, function () {
                 var tasks,
                     version = service.get('version'),
                     content = service.get('content'),
@@ -62,16 +66,23 @@ YUI.add('ez-contenteditviewservice', function (Y) {
         },
 
         /**
-         * Loads a content by its id
+         * Loads a content by its id and language code
          *
          * @method _loadContent
          * @protected
          * @param {String} id
+         * @param {String} languageCode
          * @param {Function} callback
          */
-        _loadContent: function (id, callback) {
-            this._loadModel('content', id, "Could not load the content with id '" + id + "'", callback);
-        },
+        _loadContent: function (id, languageCode, callback) {
+            this._loadModel(
+                'content',
+                id,
+                {languageCode: languageCode},
+                "Could not load the content with id '" + id + "' and languageCode '" + languageCode + "'",
+                callback
+            );
+         },
 
         /**
          * Loads a content type by its id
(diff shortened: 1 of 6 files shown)


Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -116,14 +125,16 @@ YUI.add('ez-contenteditviewservice', function (Y) {
...
          */
-        _loadModel: function (attr, modelId, errorMsg, callback) {
-            var model = this.get(attr);
+        _loadModel: function (attr, modelId, options, errorMsg, callback) {
+            var model = this.get(attr),
+                loadOptions = {api: this.get('capi')};
 
             model.set('id', modelId);
-            model.load({api: this.get('capi')}, Y.bind(function (error) {
+            model.load(Y.merge(loadOptions,options), Y.bind(function (error) {
```
dpobel: nitpick: space after the `,`

Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -36,10 +36,12 @@ YUI.add('ez-contenteditviewservice', function (Y) {
          */
         _load: function (next) {
             var request = this.get('request'),
-                service = this;
+                service = this,
+                languageCode = request.params.languageCode;
 
+            this.set('languageCode', languageCode);
```
dpobel: this should not be done here, you should add an event handler on the `requestChange` event for that (and the corresponding test)

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -240,12 +243,13 @@ YUI.add('ez-locationviewviewservice', function (Y) {
                 type = this.get('contentType'),
                 discoveryService = this.get('capi').getDiscoveryService();
 
+            this.set('languageCode', request.params.languageCode);
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/278/files#r34654568

dpobel:
You should add a test (or an assertion ?) in the tests of the ContentEditViewService to check that the content.load method receives the expected language.

In addition, did you check that these changes do not break the ContentCreateViewService (it extends the ContentEditViewService) ?

Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -24,6 +24,7 @@ YUI.add('ez-contenteditviewservice', function (Y) {
     Y.eZ.ContentEditViewService = Y.Base.create('contentEditViewService', Y.eZ.ViewService, [], {
         initializer: function () {
             this.on('*:closeView', this._handleCloseView);
+            this.on('*:requestChange', this._handleRequestChange);
```
dpobel: I would rename `_handleRequestChange` to `_setLanguageCode` or something similar, that way you directly what happen when the request change.

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -26,6 +26,7 @@ YUI.add('ez-locationviewviewservice', function (Y) {
             this.on('*:editAction', this._editContent);
             this.on('*:sendToTrashAction', this._sendContentToTrashConfirmBox);
             this.on('*:moveAction', this._selectLocation);
+            this.on('*:requestChange', this._handleRequestChange);
```
dpobel: same as above

mhyndle:
About creating a content it seems that it's not broken. Content is being created with mainLanguage, for example: if you are displaying folder in french translation and main language is eng-GB, then creating content leads to create new content in mainLanguage not the one which parent content was being previewed in. I suppose that's correct behaviour.

Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -36,10 +39,11 @@ YUI.add('ez-contenteditviewservice', function (Y) {
          */
         _load: function (next) {
             var request = this.get('request'),
-                service = this;
+                service = this,
+                languageCode = request.params.languageCode;
```
dpobel: so you should use this.get('languageCode') not the request.

mhyndle:
ready for review

dpobel:
+1

@yannickroger @StephaneDiot please review

StephaneDiot:
+1

PR #279, opened 2015-07-13
[WIP] Fix EZP-24453: Implement ContentType removal

glye:
**WIP**

https://jira.ez.no/browse/EZP-24453

*TODO*
- [x] Delete content instances using ContentService
- [x] Ask for removal confirmation?
- [x] Remove $languageCode parameter, just use default lang when displaying result
- [ ] Show error message when content exists after it's all supposed to be removed
- [ ] Show error message when access denied

Blocked by https://jira.ez.no/browse/EZP-24621

### DIFF
--- a/Controller/ContentTypeController.php
+++ b/Controller/ContentTypeController.php
@@ -11,7 +11,9 @@
 namespace EzSystems\PlatformUIBundle\Controller;
 
 use eZ\Bundle\EzPublishCoreBundle\Controller;
+use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\API\Repository\ContentTypeService;
+use eZ\Publish\API\Repository\Exceptions\BadStateException;
 use eZ\Publish\API\Repository\Exceptions\NotFoundException;
 use eZ\Publish\API\Repository\Exceptions\UnauthorizedException;
 use eZ\Publish\API\Repository\SearchService;
@@ -30,6 +32,11 @@ class ContentTypeController extends Controller
      * @var ContentTypeService
      */
     private $contentTypeService;
+
+    /**
+     * @var ContentService For deleting content prior to deleting content type
+     */
+    private $contentService;
 
     /**
      * @var SearchService
@@ -58,13 +65,15 @@ class ContentTypeController extends Controller
         SearchService $searchService,
         UserService $userService,
         ActionDispatcherInterface $actionDispatcher,
-        FieldTypeFormMapperRegistryInterface $fieldTypeMapperRegistry
+        FieldTypeFormMapperRegistryInterface $fieldTypeMapperRegistry,
+        ContentService $contentService
     ) {
         $this->contentTypeService = $contentTypeService;
         $this->searchService = $searchService;
         $this->userService = $userService;
         $this->actionDispatcher = $actionDispatcher;
         $this->fieldTypeMapperRegistry = $fieldTypeMapperRegistry;
+        $this->contentService = $contentService;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


Controller/ContentTypeController.php
```diff
@@ -198,4 +208,82 @@ public function updateContentTypeAction(Request $request, $contentTypeId, $langu
...
+        $contentCount = $contentList->totalCount;
+
+        if ($contentCount > 100) { // unhardcode TODO
+            // Refuse: Too many content instances TODO
+            return $this->forward('eZPlatformUIBundle:Pjax:accessDenied');
+            /*return $this->redirectToRoute(
+                'admin_contenttypeView',
+                ['contentTypeId' => $contentTypeId, 'languageCode' => $languageCode]
+            );*/
+        }
```
lolautruche: We decided to ignore this for now. This should be taken care of by slots.

Controller/ContentTypeController.php
```diff
@@ -198,4 +208,82 @@ public function updateContentTypeAction(Request $request, $contentTypeId, $langu
             'languageCode' => $languageCode,
         ]);
     }
+
+    public function removeContentTypeAction(Request $request, $contentTypeId, $languageCode = null)
```
lolautruche: I think you can remove `$request` as you don't seem to use it.

Resources/views/ContentType/view_content_type.html.twig
```diff
@@ -82,6 +82,7 @@
 
             <div class="pure-u-1">
                 <a href="{{ path('admin_contenttypeUpdate', {'contentTypeId': content_type.id}) }}" class="pure-button ez-button{% if not can_edit %} pure-button-disabled{% endif %}" data-icon="&#xe606;">{{ 'content_type.edit'|trans }}</a>
+                <a href="{{ path('admin_contenttypeRemove', {'contentTypeId': content_type.id}) }}" class="pure-button ez-button{% if not can_remove %} pure-button-disabled{% endif %}" data-icon="&#xe606;">{{ 'content_type.remove'|trans }}</a>
```
dpobel: semantically this should be a `button` element not a `a`.

dpobel:
The removal confirmation should be done by the JavaScript application not by any server side code. Basically the PlatformUI App should intercept the form submission and launch the "confirm box". However, there should be somewhere in the code a piece of HTML containing the details to show to user. Something like the following code should be generated:

```html
<button class="ez-button-remove ez-button pure-button ez-confirm-action" data-confirmbox-title="Do you really want to remove the 'Foobar' content type?" data-confirmbox-details-selector=".ez-confirmbox-content-type-removal-details">Remove</button>
<!-- ... -->

<div class="ez-confirmbox-content-type-removal-details">
<p>The <b>34</b> 'Foobar' instances will be permanently removed by this operation.</p>
</div>
```

glye:
Updates:
@lolautruche Ignoring the "too many content instances" case
@lolautruche Removed `$request` parameter
@dpobel Using `button` instead of `a`. Though you'd probably want the same for the edit button above it?
@dpobel Added confirm box, thanks for the howto.

dpobel:
for the button vs. a, the a for edit is fine because you want the user to navigate somewhere else (here on the edit form). But for removing the content type, you need a button as it's not just navigation anymore, using the button will run something remotely.

for the confirmbox, of course this won't work yet, we'll have to implement some JS code for that.

glye:
Weird bug: If the CT has any content instances, then removing the CT will fail on removing the first instance:
`NotFoundException: Could not find 'content' with identifier 'id: 128'`
...even though the id is correct and the instance is in fact removed. WTF.

glye:
@lolautruche Issue for the exception: https://jira.ez.no/browse/EZP-24621
Does this look valid to you?

lolautruche:
We should ask to @pspanja and @andrerom 

Sounds like a real bug.

glye:
Rebased on master, CS, doc

glye:
Fixed in https://github.com/ezsystems/PlatformUIBundle/pull/338

PR #280, opened 2015-07-14
EZP-23767: create new translation basing on existing one

mhyndle:
Jira: https://jira.ez.no/browse/EZP-23767

# Description:
The goal of this issue is to add "New Translation" button under the list of existing content's translations (after clicking "Translations" action button). Clicking "New Translation" button should open a box which allow the user to choose the language for which he wants to create the translation. There should be also possibility to choose the existing content's translation on which user want to base his new translation.

# Tasks:
- [x] Implementation
- [x] CSS
- [x] Tests
- [x] Rebase

# Screencast
https://youtu.be/DWFGEttg4V4


Resources/public/css/theme/views/languageselectionbox.css
```diff
@@ -0,0 +1,75 @@
...
+    -webkit-transform: translateY(-100%);
+            transform: translateY(-100%);
+}
+
+.ez-languageselectionbox-container {
+    background: rgba(0, 0, 0, 0.6);
+
+    -webkit-transform: translateY(0);
+            transform: translateY(0);
+    -webkit-transition: all 0.3s;
```
dpobel: transition does not need a prefix

Resources/public/css/theme/views/languageselectionbox.css
```diff
@@ -0,0 +1,75 @@
...
+    padding-bottom: 0.2em;
+    color: #666;
+}
+
+.ez-view-languageselectionboxview .ez-languageselectionbox-languages {
+    display: block;
+    background: #fff;
+    border: solid 1px #aaa;
+    border-radius: 3px;
+    width: 414px;
```
dpobel: do we need a so precise width ? At least I would express it in a relative unit like `em` (30em or something like that)
In addition, it's a theme stylesheet, this should be added in the other stylesheet.

Resources/public/css/theme/views/languageselectionbox.css
```diff
@@ -0,0 +1,75 @@
...
+    overflow: hidden;
+    padding: 10px 0;
+}
+
+.ez-view-languageselectionboxview .ez-translation {
+    float: left;
+    width: 163px;
```
dpobel: most of those style should not be there as well.

Resources/public/fonts/icomoon.svg
```diff
@@ -53,5 +53,6 @@
...
+<glyph unicode="&#xe9ca;" d="M512 960c-282.77 0-512-229.23-512-512s229.23-512 512-512 512 229.23 512 512-229.23 512-512 512zM512-0.002c-62.958 0-122.872 13.012-177.23 36.452l233.148 262.29c5.206 5.858 8.082 13.422 8.082 21.26v96c0 17.674-14.326 32-32 32-112.99 0-232.204 117.462-233.374 118.626-6 6.002-14.14 9.374-22.626 9.374h-128c-17.672 0-32-14.328-32-32v-192c0-12.122 6.848-23.202 17.69-28.622l110.31-55.156v-187.886c-116.052 80.956-192 215.432-192 367.664 0 68.714 15.49 133.806 43.138 192h116.862c8.488 0 16.626 3.372 22.628 9.372l128 128c6 6.002 9.372 14.14 9.372 22.628v77.412c40.562 12.074 83.518 18.588 128 18.588 70.406 0 137.004-16.26 196.282-45.2-4.144-3.502-8.176-7.164-12.046-11.036-36.266-36.264-56.236-84.478-56.236-135.764s19.97-99.5 56.236-135.764c36.434-36.432 85.218-56.264 135.634-56.26 3.166 0 6.342 0.080 9.518 0.236 13.814-51.802 38.752-186.656-8.404-372.334-0.444-1.744-0.696-3.488-0.842-5.224-81.324-83.080-194.7-134.656-320.142-134.656z" />
```
dpobel: same as the other PR, the selection.json changes are missing (by the way, we'll have to be careful when merging those PR to not lose the new icons).

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -84,6 +84,12 @@ YUI.add('ez-platformuiapp', function (Y) {
                 container: '.ez-confirmbox-container',
                 hideClass: 'is-confirmbox-hidden',
             },
+            languageSelectionBox: {
+                type: Y.eZ.LanguageSelectionBoxView,
+                service: Y.eZ.LanguageSelectionBoxViewService,
+                container: '.ez-languageselectionbox-container',
+                hideClass: 'is-languageselectionbox-hidden',
```
dpobel: you have to add this class to the `ez-platformui-app` element in shell.html.twig

Resources/public/js/apps/plugins/ez-languageselectionboxplugin.js
```diff
@@ -0,0 +1,42 @@
...
+     * @namespace eZ.Plugin
+     * @class LanguageSelectionBox
+     * @constructor
+     * @extends Plugin.Base
+     */
+    Y.eZ.Plugin.LanguageSelectionBox = Y.Base.create('languageSelectionBoxPlugin', Y.Plugin.Base, [], {
+        initializer: function () {
+            var app = this.get('host');
+
+            app.on('*:languageSelectionBoxOpen', function (e) {
```
dpobel: I would rather use more meaningful event names (modeled after the ones used in the universal discovery widget instead) like: `languageSelect`, `languageSelected` and `cancelLanguageSelect`

Resources/public/js/views/actions/ez-translateactionview.js
```diff
@@ -0,0 +1,160 @@
...
+
+        /**
+         * Returns the hint for translate button
+         *
+         * @return {String}
+         */
+        _getTranslateButtonHint: function () {
+            var translations = this.get('content').get('currentVersion').getTranslationsList(),
+                countAll = translations.length,
+                moreTranslations = Y.clone(translations);
```
dpobel: in such case, it's better to use `Y.merge(translations);` instead of `Y.clone` as we don't need a deep cloning.

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,278 @@
...
+    /**
+     * The language selection box view.
+     *
+     * @namespace eZ
+     * @class LanguageSelectionBoxView
+     * @constructor
+     * @extends eZ.TemplateBasedView
+     */
+    Y.eZ.LanguageSelectionBoxView = Y.Base.create('languageSelectionBoxView', Y.eZ.TemplateBasedView, [], {
+        TRANSLATION_TPL: '<li class="ez-translation" />',
```
dpobel: why don't you generate that in the template ?

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,278 @@
...
+            translationsContainer.setContent('');
+
+            Y.Array.each(this.get('newTranslations'), function (languageCode) {
+                translationsContainer.append(that._getSingleTranslation(languageCode, 'ez-newtranslation'));
+            });
+        },
+
+        _getSingleTranslation: function (languageCode, className) {
+            var node = Y.Node.create(this.TRANSLATION_TPL),
+                flagCode = languageCode.substring(4, 6).toLowerCase();
```
dpobel: by generating [the languages list in the template](https://github.com/ezsystems/PlatformUIBundle/pull/280/files#r34758029) you'll be able to use the Handlebars helper instead of duplicating its code here.

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,278 @@
...
+         *
+         * @protected
+         * @method _closeLanguageSelectionBox
+         */
+        _closeLanguageSelectionBox: function () {
+            this.fire('languageSelectionBoxClose');
+        },
+
+        _getAvailableTranslations: function () {
+            return ['eng-GB', 'pol-PL', 'ger-DE', 'fre-FR', 'nno-NO', 'esl-ES',
```
dpobel: you should not hardcode the languages list here but in an attribute of the LanguageSelectionBoxViewService, that way when we'll have the real language list, the changes will be minimal and located only in this component.

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,278 @@
...
+        },
+
+        _getAvailableTranslations: function () {
+            return ['eng-GB', 'pol-PL', 'ger-DE', 'fre-FR', 'nno-NO', 'esl-ES', 
+                'jpn-JP', 'rus-RU', 'swe-SE', 'por-BR', 'cze-CZ', 'ara-SA'];
+        },
+
+        _getNewTranslations: function () {
+            var availableTranslations = this._getAvailableTranslations(),
+                translations = Y.clone(availableTranslations),
```
dpobel: this seems a bit complicated, to ease the code, you should index the list of languages by language code. That way, building the list of potential new language becomes super easy.

Tests/js/views/assets/ez-actionbarview-tests.js
```diff
@@ -47,6 +56,31 @@ YUI.add('ez-actionbarview-tests', function (Y) {
                 'The ActionBarView should contain createContent action'
             );
         },
+
+        "Should instantiate translateActionView": function () {
+            var translateActionFound = false;
+
+            for (var i = 0; i < this.view.get('actionsList').length; i++){
```
dpobel: the very same test logic is now repeated (and it's actually an higher level assertion), so you should introduce a protected method to check that a given action view is in the list with the value of some attributes.

Tests/js/views/actions/ez-translateactionview.html
```diff
@@ -0,0 +1,94 @@
...
+</button>
+<div class="ez-expandable-area">
+    <div class="ez-contenttranslations">
+        <h2 class="ez-contenttranslations-title">Existing Translations</h2>
+        <div class="ez-contenttranslations-content">
+            <ul class="ez-contenttranslations-list">
+                {{#each translations}}
+                <li class="ez-contenttranslation">
+                    <a class="ez-contenttranslation-view-link" href="#">
+                        <span class="ez-language-flag flag-icon"></span>
```
dpobel: please simplify the template to minimal needed elements for the test.

Tests/js/views/actions/assets/ez-translateactionview-tests.js
```diff
@@ -0,0 +1,273 @@
...
+        "Should display all translations in button hint": function () {
+            var translationsList = ['Diego Armando Maradona Franco', 'Dennis Nicolaas Maria Bergkamp'],
+                hintExpected = 'Diego Armando Maradona Franco, Dennis Nicolaas Maria Bergkamp';
+
+            Mock.expect(this.versionMock, {
+                method: 'getTranslationsList',
+                returns: translationsList
+            });
+
+            this.view = new Y.eZ.TranslateActionView({
```
dpobel: the view is instantiated in the `setUp`, you should not recreate a new instance here.

Resources/public/js/apps/ez-platformuiapp.js
mhyndle: It's already done

Resources/public/js/views/ez-languageselectionboxview.js
mhyndle: I tried it, but while I was basing on confirmBox and UDW the view is loaded on start and I have to fill it with languages dynamically. Or maybe there is another to reload the view?

dpobel: well you can always set an attribute on the view and trigger a render when this view changes

mhyndle:
I have rebased it to master so now it's not basing on another branch

Resources/public/css/theme/views/languageselectionbox.css
```diff
@@ -0,0 +1,59 @@
...
+    border: solid 1px #aaa;
+    border-radius: 3px;
+}
+
+.ez-view-languageselectionboxview .ez-translation {
+    cursor: pointer;
+}
+
+.ez-view-languageselectionboxview .ez-translation.is-selected,
+.ez-view-languageselectionboxview .ez-translation:hover {
```
dpobel: we should have a different color for the selection and for what is actually selected

Resources/public/js/views/actions/ez-translateactionview.js
```diff
@@ -112,6 +119,49 @@ YUI.add('ez-translateactionview', function (Y) {
...
+                    content: this.get('content'),
+                    languageCode: languageCode,
+                    isBaseLanguage: isBaseLanguage
+                };
+
+            if (isBaseLanguage && e.target.get('selectedBaseTranslation') !== null){
+                data.baseLanguageCode = e.target.get('selectedBaseTranslation');
+            }
+
+            this.fire('translate', data);
```
dpobel: you should call this event `translateContent` instead (to be consistent with `editContent`)

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,402 @@
...
+        },
+
+        /**
+         * Tap event handler on switcher dicatating if new translation will base
+         * on already existing one or not
+         *
+         * @method _baseTranslationToggle
+         * @protected
+         */
+        _baseTranslationToggle: function () {
```
dpobel: `_toggleBaseTranslation`

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,402 @@
...
+            this.set('baseTranslation', !this.get('baseTranslation'));
+        },
+
+        initializer: function () {
```
dpobel: please put the `initializer` at the beginning (right after the `events`)

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,402 @@
...
+         * @protected
+         */
+        _resetState: function () {
+            this.reset('title');
+            this.reset('confirmHandler');
+            this.reset('cancelHandler');
+            this.reset('baseTranslation');
+            this.reset('selectedNewTranslation');
+            this.reset('selectedBaseTranslation');
+            this.render();
```
dpobel: this should not be there

Resources/public/css/theme/views/languageselectionbox.css
```diff
@@ -0,0 +1,59 @@
...
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+.is-languageselectionbox-hidden .ez-languageselectionbox-container {
+    display: -webkit-box;
+    display: -webkit-flex;
+    display: -ms-flexbox;
+    display: flex;
+
+    background: rgba(0, 0, 0, 0);
```
dpobel: `#000` is much easier to read

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,402 @@
...
+    /**
+     * The language selection box view.
+     *
+     * @namespace eZ
+     * @class LanguageSelectionBoxView
+     * @constructor
+     * @extends eZ.TemplateBasedView
+     */
+    Y.eZ.LanguageSelectionBoxView = Y.Base.create('languageSelectionBoxView', Y.eZ.TemplateBasedView, [], {
+        BASE_TRANSLATION_VISIBLE_CLASS: 'is-visible',
```
dpobel: please use vars like `CONFIRM` instead

dpobel: and `is-visible` and `is-selected` are a bit too generic should be something like `is-base-translation-visible` and `is-translation-selected`.

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,402 @@
...
+        /**
+         * Sets the state for base translation checkbox
+         *
+         * @method _uiBaseTranslationState
+         * @protected
+         * @param {String} baseTranslation
+         */
+        _uiBaseTranslationState: function (baseTranslation) {
+            var c = this.get('container'),
+                toggledContainer = c.one('.ez-languageselectionbox-existingtranslations-container'),
```
dpobel: you should set that class on the view container.

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,402 @@
...
+                title: this.get('title'),
+                newTranslations: this.get('newTranslations'),
+                existingTranslations: this.get('existingTranslations'),
+                canBaseTranslation: this.get('canBaseTranslation'),
+                baseTranslation: this.get('baseTranslation'),
+                selectedNewTranslation: this.get('selectedNewTranslation'),
+                selectedBaseTranslation: this.get('selectedBaseTranslation'),
+            }));
+
+            container.one('.ez-base-translation-checkbox').on('change', function (e) {
```
dpobel: this should be here, why don't you use the `events` property for that ?

Resources/public/js/views/services/ez-languageselectionboxviewservice.js
```diff
@@ -0,0 +1,55 @@
...
+     * @namespace eZ
+     * @class LanguageSelectionBoxViewService
+     * @constructor
+     * @extends eZ.TemplateBasedView
+     */
+    Y.eZ.LanguageSelectionBoxViewService = Y.Base.create('languageSelectionBoxViewService', Y.eZ.ViewService, [], {
+        getViewParameters: function () {
+            var config = this.get('config');
+
+            config.availableTranslations = this.get('availableTranslations');
```
dpobel: by doing that you are modifying an object which is not *yours*, you should first *clone* (with `Y.merge`) config.

Resources/public/templates/languageselectionbox.hbt
```diff
@@ -0,0 +1,32 @@
...
+
+<ul class="ez-languageselectionbox-languages ez-languageselectionbox-newtranslations">
+    {{#each newTranslations}}
+        <li class="ez-translation ez-new-translation" data-languagecode="{{ this }}">{{ this }}</li>
+    {{/each}}
+</ul>
+
+{{#if canBaseTranslation}}
+    <div class="ez-languageselectionbox-basetranslation">
+        <label>
```
dpobel: add the `for` attribute and the corresponding `id` on the input

Resources/public/templates/languageselectionbox.hbt
```diff
@@ -0,0 +1,32 @@
...
+            <input type="checkbox" value="1" class="ez-base-translation-checkbox"
+                name="baseTranslation" {{#if baseTranslation}}checked="checked"{{/if}}>
+            Base this translation on an existing translation
+        </label>
+    </div>
+
+    <div class="ez-languageselectionbox-existingtranslations-container {{#if baseTranslation}}is-visible{{/if}}">
+        <h2 class="ez-languageselectionbox-title">Select a base language:</h2>
```
dpobel: use `h3` to respect the hierarchy of headings

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,402 @@
...
+            });
+            this._syncEventHandler('confirm');
+            this._syncEventHandler('cancel');
+            this.after('activeChange', function () {
+                if ( this.get('active') ) {
+                    this._uiUpdateTitle();
+                }
+            });
+            this._publishEvents();
+            this.on(['existingTranslationsChange','canBaseTranslationChange'], function (e) {
```
dpobel: > 1. There are some view attributes used in template which may change. For example when state is being reset (when languageSelectionBox is closed) I render the view again. I wanted do the same, when attributes related to languages lists are changing: https://github.com/ezsystems/PlatformUIBundle/pull/280/files?diff=unified#diff-7add4a5b132be8c1d7fa15ac84ff9c8eR117
but it crashes - this line fails and return undefined instead of collection of available languages (currently set in LanguageSelectionBoxViewService).


for `canBaseTranslation` instead of trying to rerender completely the view when it changes, I would just add or remove a class on the container to show or hide the corresponding UI part.

for `existingTranslations`, this one should indeed trigger a render and I fail to see why it would not work.

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,402 @@
...
+            var container = this.get('container'),
+                that = this;
+
+            container.setHTML(this.template({
+                title: this.get('title'),
+                newTranslations: this.get('newTranslations'),
+                existingTranslations: this.get('existingTranslations'),
+                canBaseTranslation: this.get('canBaseTranslation'),
+                baseTranslation: this.get('baseTranslation'),
+                selectedNewTranslation: this.get('selectedNewTranslation'),
```
dpobel: > 2. I'm passing to the template attributes: selectedNewTranslation and selectedBaseTranslation which are being set by clicking on the translation and add/remove class from translations <li>s. What I'm wodering is if I should set in template the class on <li> element that highlights selected element as probably in the future there will be need of seting those attributes on initiating the view. I wanted to do this but it needs kind of {{#if this=selectedNewStatement}} statement and AFAIK handlebars are not providing such conditions. I've seen that ppl are writing helpers for it, but I'm not sure if it's proper way or even if it's needed.

as you write those attributes are filled by the user action so don't pass them to the template as it's not needed now.

Resources/public/js/views/actions/ez-translateactionview.js
```diff
@@ -112,6 +119,49 @@ YUI.add('ez-translateactionview', function (Y) {
...
+         * @private
+         * @param {EventFacade} e
+         */
+        _newTranslation: function (e) {
+            var languageCode = e.target.get('selectedNewTranslation'),
+                isBaseLanguage = e.target.get('baseTranslation'),
+                data = {
+                    content: this.get('content'),
+                    languageCode: languageCode,
+                    isBaseLanguage: isBaseLanguage
```
dpobel: this property name is very confusing but actually it's not need so it's better to remove it.

Resources/public/js/views/actions/ez-translateactionview.js
```diff
@@ -112,6 +119,49 @@ YUI.add('ez-translateactionview', function (Y) {
...
+         * @method _newTranslation
+         * @private
+         * @param {EventFacade} e
+         */
+        _newTranslation: function (e) {
+            var languageCode = e.target.get('selectedNewTranslation'),
+                isBaseLanguage = e.target.get('baseTranslation'),
+                data = {
+                    content: this.get('content'),
+                    languageCode: languageCode,
```
dpobel: to improve the readability, I would call it to `toLanguageCode`

Resources/public/js/views/actions/ez-translateactionview.js
dpobel: and you should document it

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -208,6 +209,42 @@ YUI.add('ez-locationviewviewservice', function (Y) {
         },
 
         /**
+         * translate event handler, makes the application to navigate to edit content available
+         * in the facade with given language and base language
+         *
+         * @method _translateContent
+         * @protected
+         * @param {EventFacade} e
+         * @param {String} e.languageCode language code of new translation
```
dpobel: this doc should go to the event doc

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -208,6 +209,42 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+         */
+        _translateContent: function (e) {
+            var app = this.get('app'),
+                baseLanguageCode = this.get('languageCode'),
+                routeParams = {
+                        id: e.content.get('id'),
+                        languageCode: e.languageCode,
+                    };
+
+            if (e.baseLanguageCode) {
```
dpobel: > 3 Here I have prepared the `translate` event: https://github.com/ezsystems/PlatformUIBundle/pull/280/files?diff=unified#diff-9b8359562331d6a0f062d74df2f6525bR226
It's called here: https://github.com/ezsystems/PlatformUIBundle/pull/280/files?diff=unified#diff-659f9de6ec2e07d2db4ea7912dc5e01cR151
It's bit complicated but I don't know if it could be simplier. Or maybe my approach is incorrect.

indeed that's too complicated because if the user did not choose a base language code you don't have to use the current language as the base. Basically this can be simplified to:
```js
            var routeParams = {
                    id: e.content.get('id'),
                    languageCode: e.languageCode,
                };

            if (e.baseLanguageCode) {
                routeParams.baseLanguageCode = e.baseLanguageCode;
            }
            this.get('app').navigate(
                app.routeUri('editContent', routeParams)
            );
```

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -208,6 +209,42 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+         * @param {EventFacade} e
+         * @param {String} e.languageCode language code of new translation
+         * @param {Boolean} e.isBaseLanguage defines if new translation is basing on another one or not
+         *                  if it's set to true then languageCode of base language will be set, otherwise
+         *                  base language will be ommited
+         * @param {String} e.baseLanguageCode language code of existing translation on which 
+         *                  new translation will be basing, it it's not set then current
+         *                  languageCode will be used
+         */
+        _translateContent: function (e) {
```
dpobel: > 4. When calling confirmHandler in languageSelectionBoxView, I'm calling the `translate` event which in consequence is redirecting to edit content view.

[...]

> In the edit content view there are few issues I have faced to:
4.1. currently when I pass languageCode of not existing translation (new translation) fields in content edit view are filled with data from main language version - shouldn't they be empty (for now at least when we ommit the baseLanguage)?
4.2. when saving/publishing the new translation I receives 404 from api call with message "Could not find 'Language' with identifier 'ger-DE'", which probably points on that the version for this language wasn't created. Is it bug or maybe something I should develop?

it's because in this case you should use the create content route (and most likely you'll have to tweak its languageCode handling like you did for the edit so that it's not always the main language code)

> 4.3. there should be definition of edit content view changed in router but when I was playing with it a bit it just made me frustrated :) because I wasn't able to set the parameter in URI to be optional. I suppose that for translations should be written new routes with dividing them to creating new translation without basing on any other translation and another one with passing base translation. Is my way of thinking correct or I'm wrong?

indeed you can not have optional parameters in route but you can *duplicate* the route, something like:
```js
                {
                    name: "editContent",
                    path: '/edit/:id/:languageCode',
                    service: Y.eZ.ContentEditViewService,
                    sideViews: {'navigationHub': false, 'discoveryBar': false},
                    view: 'contentEditView',
                },
                {
                    name: "editContentBaseLanguage",
                    path: '/edit/:id/:languageCode/:baseLanguageCode',
                    service: Y.eZ.ContentEditViewService,
                    sideViews: {'navigationHub': false, 'discoveryBar': false},
                    view: 'contentEditView',
                },
```

Resources/public/js/views/ez-languageselectionboxview.js
mhyndle: I have removed it frome here. I have also removed `this.reset()` for  `baseTranslation`, `selectedNewTranslation` and `selectedBaseTranslation` as it shouldn't be done here probably.

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,412 @@
...
+            });
+            this._syncEventHandler('confirm');
+            this._syncEventHandler('cancel');
+            this.after('activeChange', function () {
+                if ( this.get('active') ) {
+                    this._uiUpdateTitle();
+                }
+            });
+            this._publishEvents();
+            this.on('existingTranslationsChange', function (e) {
```
mhyndle: It seems that this part is not working as I wanted... Event is triggered but if call `this.render()` here it crashed inside method `_getNewTranslations` which is getter for `newTranslations` attribute. It seems that at the time when event is triggered, the value of attribute `availableTranslations` from `LanguageSelectionBoxViewService` is not set and `this.get('availableTranslations')` returns `undefined`.

mhyndle:
According to: https://github.com/ezsystems/PlatformUIBundle/pull/280#discussion_r35512573
I'm bit confused, because you mentioned that when creating the translation for content I should go for creating content, but AFAIK for creating the content I need languageCode, contentType and parentId, not existing contentId. Or I missunderstand you ;)
I appreciate any detailed info about this @dpobel

Resources/public/js/views/ez-languageselectionboxview.js
dpobel: it seems it's because availableTranslations is set after the newTranslations attribute. We can of course hack the order of the properties in the return value of `_getViewParameters` but that's a bit dangerous. Instead, I would do a render when is becoming active (so you can remove the `_uiUpdateTitle` thing which was introduce because others side views like that don't need to be rerendered)

dpobel:
you are right, the translation is actually a special case of content edit, so I guess you'll have to change a bit the content edit view service to handle that as well.

Resources/public/js/views/actions/ez-translateactionview.js
```diff
@@ -112,6 +119,57 @@ YUI.add('ez-translateactionview', function (Y) {
...
+            e.preventDefault();
+            this.fire('languageSelect', {
+                config: {
+                    title: "Select a language for your new translation:",
+                    confirmHandler: Y.bind(this._newTranslation, this),
+                    cancelHandler: null,
+                    canBaseTranslation: true,
+                    existingTranslations: this.get('content').get('currentVersion').getTranslationsList()
+                },
+            });
```
dpobel: you should probably hide the expanded area as well

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,418 @@
...
+        },
+
+        /**
+         * Cancel the languageSelectionBox by firing the `cancelLanguageSelect` event
+         *
+         * @protected
+         * @method _cancelLanguageSelection
+         */
+        _cancelLanguageSelection: function () {
+            this.fire('cancelLanguageSelect');
```
dpobel: missing doc on the event

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,418 @@
...
+        },
+
+        /**
+         * Confirm the languageSelectionBox's selection by firing the `languageSelected` event
+         *
+         * @protected
+         * @method _confirmLanguageSelection
+         */
+        _confirmLanguageSelection: function () {
+            this.fire('languageSelected');
```
dpobel: the doc is missing. In addition, the event facade of this event should probably include the selected language and the base language if any.

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,418 @@
...
+            },
+
+            /**
+             * Selected language from new translations list
+             *
+             * @attribute selectedNewTranslation
+             * @type {string|Null}
+             * @default null
+             */
+            selectedNewTranslation: {
```
dpobel: this attribute and `selectedBaseTranslation` should be called `Language` not `Translation` as the language selection box is meant to allow to select a language not only to translate something.

Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -40,15 +44,25 @@ YUI.add('ez-contenteditviewservice', function (Y) {
...
                     version = service.get('version'),
                     content = service.get('content'),
+                    fields = content.get('fields'),
                     resources;
 
+                Y.each(fields, function (field) {
+                    field.languageCode = languageCode;
+                });
+
+                //TODO: if `languageCodeForLoadContent` !== `languageCode` fields should be empty
```
dpobel: any reason to not do it ?

dpobel:
very good :)
For https://github.com/ezsystems/PlatformUIBundle/pull/280#discussion_r37282842, in the case where the translation is not based on any existing language, you'll have to initialize the fields of the version based on the content type like it's done in the content create view service.

Resources/public/js/views/ez-languageselectionboxview.js
mhyndle: Is it really needed? Currently only thing that `languageSelected` event is doing is hiding sideView without using any of attributes. What's more this function `_confirmLanguageSelection` is only default function and it should be overriden by passing own function to `confirmHandler`.

dpobel: actually your comment makes me realize there's a deeper issue in your patch. Basically when the user chooses a language first the `confirm` event is fired and then the `languageSelected` event is fired. The second one is used to hide the side view while the first one is the one you use with the confirm handler. But there should be only one event (`languageSelected`). This one should be triggered when the user uses the confirm button and its event facade should contain the selected language code and the base language code if any. The `confirmHandler` should then be renamed to `languageSelectedHandler`.
The exact same thing is valid for the case where the user uses the cancel button.

I guess you kind of mixed the confirm box event system with the universal discovery event system while the language selection box is actually close to the universal discovery.

Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -40,24 +44,30 @@ YUI.add('ez-contenteditviewservice', function (Y) {
...
+                    loadedContentLanguageCode = Y.Object.values(fields)[0].languageCode,
+                    setEmptyFields = false,
                     resources;
 
-                version.set('fields', content.get('fields'));
-
                 resources = content.get('resources');
 
+                if (!baseLanguageCode && loadedContentLanguageCode !== languageCode) {
+                    setEmptyFields = true;
```
dpobel: this is a bad name, *empty* is kind of wrong or at least inaccurate as the content type might define default value. I would rather call it `setDefaultFieldValue` and so the same kind of change everywhere you use the word *empty*.

Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -66,6 +76,56 @@ YUI.add('ez-contenteditviewservice', function (Y) {
...
+         * Sets fields of edited version.
+         *
+         * @method _setFields
+         * @private
+         * @param {Object} fields
+         * @param {Boolean} setEmptyFields
+         * @param {String} contentTypeId
+         * @param {Function} callback
+         */
+        _setFields: function (fields, setEmptyFields, contentTypeId, callback) {
```
dpobel: `_setVersionFields`

mhyndle:
ready for review, ping @dpobel, @yannickroger , @StephaneDiot

StephaneDiot:
+1 congrats :+1:  !

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,414 @@
...
+             * @attribute canBaseTranslation
+             * @type {Boolean}
+             * @default true
+             */
+            canBaseTranslation: {
+                value: true
+            },
+
+            /**
+             * Defines if new translation will be basing on already existing one
```
dpobel: *will be based on*

Tests/js/alloyeditor/buttons/assets/ez-alloyeditor-button-heading-tests.js
```diff
@@ -0,0 +1,96 @@
+/*
```
dpobel: why did you add this file and ez-alloyeditor-toolbar-appendcontent-tests.js ? they were not added to git on purpose

Resources/public/js/views/services/ez-languageselectionboxviewservice.js
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-languageselectionboxviewservice', function (Y) {
+    "use strict";
+    /**
+     * Provides the language selection box view service class
+     *
+     * @method ez-languageselectionboxviewservice
```
dpobel: not `method`, `module` :)

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -0,0 +1,414 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-languageselectionboxview', function (Y) {
+    "use strict";
+    /**
+     * Provides the language selection box view class
+     *
+     * @method ez-languageselectionboxview
```
dpobel: `module`

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -208,6 +209,32 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+         *
+         * @method _translateContent
+         * @protected
+         * @param {EventFacade} e
+         */
+        _translateContent: function (e) {
+            var app = this.get('app'),
+                routeName = 'editContent',
+                routeParams = {
+                        id: e.content.get('id'),
```
dpobel: CS: should be written:
```
    routeParams = {
        id: e.content.get('id'),
        languageCode: e.toLanguageCode,
    };
```

Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -40,24 +44,23 @@ YUI.add('ez-contenteditviewservice', function (Y) {
...
-                version.set('fields', content.get('fields'));
-
                 resources = content.get('resources');
 
                 tasks = new Y.Parallel();
 
                 service._loadOwner(resources.Owner, tasks.add());
                 service._loadLocation(resources.MainLocation, tasks.add());
-                service._loadContentType(resources.ContentType, tasks.add());
+                service._setVersionFields(tasks.add());
```
dpobel: the code organization here is far from being clear, this needs to be refactored/reorganized. First, you added a `_setVersionFields` which does a lot more then just setting the fields of the version like checking if the content was loaded in the correct language, loading the content type... Then you also repeat the `languageCodeForLoadContent` in `_load` and in `_setVersionFields`.

So to improve a bit the code, I would keep `_loadContentType` here in parallel with the `_loadOwner` and `_loadLocation` and implement the new logic in the `done` callback a few lines below.

The check on the actual language of the content if we have a baseLanguage should be the first thing done in the `_loadContent` callback (it's an edge case and will happen only if the user tweak the URI manually). And please introduce a method to do that so that we have a self-explanatory name to understand the actual tasks done by the code and a doc block explaining why we need that.

After having done that, `_setVersionFields` should be way smaller and easier to understand and will be able to see if how we can maybe improve its code as well.

Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -66,6 +69,85 @@ YUI.add('ez-contenteditviewservice', function (Y) {
...
+                version = this.get('version'),
+                fields = content.get('fields'),
+                languageCode = this.get('languageCode'),
+                baseLanguageCode = this.get('baseLanguageCode'),
+                languageCodeForLoadContent = baseLanguageCode ? baseLanguageCode : languageCode,
+                setDefaultFields = false,
+                languageInTranslationsList;
+
+            languageInTranslationsList = Y.Array.find(
+                                            content.get('currentVersion').getTranslationsList(),
```
dpobel: btw CS issue and instead if ( condition ) { return true } you can directly return the result of the condition, something like:
```js
languageInTranslationsList = Y.Array.find(
    content.get('currentVersion').getTranslationsList(), function (translation) {
        return (translation === languageCodeForLoadContent);
    }
);
```

Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -66,6 +69,85 @@ YUI.add('ez-contenteditviewservice', function (Y) {
...
+                if ( languageInTranslationsList === null) {
+                    setDefaultFields = true;
+                }
+            }
+
+            this._loadContentType(contentType, function () {
+                if (setDefaultFields) {
+                    fields = service._getDefaultFields(service.get('contentType'), service.get('languageCode'));
+                } else {
+                    Y.each(fields, function (field) {
```
dpobel: you are modifying the fields of the content, you should create a new `fields` object containing clones of the field on which you set the language code.

Tests/js/views/assets/ez-languageselectionboxview-tests.js
```diff
@@ -0,0 +1,500 @@
...
+            Assert.isTrue(templateCalled, "The template should have used to render the view");
+        },
+
+        "Test available variables in the template": function () {
+            var origTpl = this.view.template,
+                that = this;
+
+            this.view.template = function (variables) {
+                Assert.isObject(variables, "The template should receive some variables");
+                Assert.areEqual(4, Y.Object.keys(variables).length, "The template should receive 1 variable");
```
dpobel: "The template should receive 4 variables"

Tests/js/views/assets/ez-languageselectionboxview-tests.js
```diff
@@ -0,0 +1,500 @@
...
+        "Should set the selectedBaseLanguageCode attribute": function () {
+            var selector = '.ez-base-translation',
+                element = this.view.get('container').one(selector),
+                attributeName = 'selectedBaseLanguageCode',
+                expectedAttributeValue = element.getAttribute('data-languagecode');
+
+            this._attributeChangeTest(element, attributeName, expectedAttributeValue);
+        },
+
+        "Should set the basTranslation attribute": function () {
```
dpobel: baseTranslation

Tests/js/views/services/assets/ez-contenteditviewservice-tests.js
```diff
@@ -501,4 +844,4 @@ YUI.add('ez-contenteditviewservice-tests', function (Y) {
     Y.Test.Runner.add(eventTest);
     Y.Test.Runner.add(redirectionUrlTest);
     Y.Test.Runner.add(getViewParametersTest);
-}, '', {requires: ['test', 'ez-contenteditviewservice']});
+}, '', {requires: ['test', 'ez-contenteditviewservice', 'array-extras']});
```
dpobel: as far as I can see, there's no call to anything provided by array-extras so that's not needed (since you added it to the dependencies of ez-contenteditviewservice).

Tests/js/views/services/assets/ez-contenteditviewservice-tests.js
```diff
@@ -227,22 +547,29 @@ YUI.add('ez-contenteditviewservice-tests', function (Y) {
             });
             Y.Mock.expect(this.version, {
                 method: 'set',
-                args: ['fields', this.fields]
+                args: ['fields', Y.Mock.Value.Object]
```
dpobel: hum usually it's not enough to test only Mock.Value.*, is this change really needed ?

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -733,6 +739,13 @@ YUI.add('ez-platformuiapp', function (Y) {
                     view: 'studioPlusPresentationView',
                     callbacks: ['open', 'checkUser', 'handleSideViews', 'handleMainView']
                 }, {
+                    name: "translateContent",
```
dpobel: since you are adding a route which is almost the same as editContent, you also need to duplicate the route configuration passed to the application in shell.html.twig (will change that soon anyway...)

Tests/js/alloyeditor/buttons/assets/ez-alloyeditor-button-heading-tests.js
mhyndle: probably I have missed it on committing, however shouldn't these file to be added to .gitgnore as they are generated on `grunt test`?

Resources/public/js/apps/ez-platformuiapp.js
mhyndle: I have added translateContent route config in shell.html.twig: https://github.com/ezsystems/PlatformUIBundle/pull/280/files#diff-def2dab3cc9208048b5a2e96663033a7R75

Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -60,9 +65,123 @@ YUI.add('ez-contenteditviewservice', function (Y) {
...
+         */
+        _getFieldsForEdit: function (content, translationExists) {
+            var baseLanguageCode = this.get('baseLanguageCode'),
+                languageCode = this.get('languageCode'),
+                contentFields = content.get('fields'),
+                setDefaultFields = false,
+                fields;
+
+            if (baseLanguageCode) {
+                if (!translationExists && !content.get('alwaysAvailable')) {
```
dpobel: as I wrote in https://github.com/ezsystems/PlatformUIBundle/pull/280#discussion_r38178912, this check should be the first thing done in the `_loadContent` callback, doing it here is way too late (and you don't have to check the alwaysAvailable flag).

Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -60,9 +65,123 @@ YUI.add('ez-contenteditviewservice', function (Y) {
                 service._loadContentType(resources.ContentType, tasks.add());
 
                 tasks.done(function () {
-                    next(service);
+                    service._setVersionFields(content, translationExists, function () {
```
dpobel: `_setVersionFields` is completely synchronous so it should not have a callback parameter and you can just write:
```js
service._setVersionFields(content, translationExists);
next(service);
```

Tests/js/views/services/assets/ez-contenteditviewservice-tests.js
mhyndle: Simple object for fields is being passed in tests, but in app it seems that it contains whole `__proto__` (which simple object in tests doesn't have), so objects are not exactly the same. I have added an assertion in `run` function to check if important attributes are same.

mhyndle:
ping @dpobel

Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -40,16 +44,25 @@ YUI.add('ez-contenteditviewservice', function (Y) {
...
                     resources;
 
-                version.set('fields', content.get('fields'));
+                translationExists = service._checkIfTranslationExists(content, languageCodeForLoadContent);
+
+                if (baseLanguageCode && !translationExists) {
+                    service._error(
+                        "Could not load the content with id '" + content.get('contentId')
+                        + "' and languageCode '" + baseLanguageCode + "'"
+                    );
```
dpobel: I guess you should add a `return` here as it's useless to try to load the owner, location and content type

mhyndle: `return;` added

mhyndle:
I have removed also reset of `canBaseTranslation` attribute as it shouldn't be done.
ping @dpobel

andrerom:
Not fully voting member, but looks good to me :+1:

dpobel:
+1

PR #281, opened 2015-07-16
Fixed a JS error displaying a content with a filled Relation field

dpobel:
# Description

When displaying a content in PlatformUI after a content having a filled Relation field, the following error is visible in the console:
```
Uncaught TypeError: Cannot read property 'destination' of undefined
```

this is happening because the event to get the related content is fired when the `active` attribute changes instead of being fired when the view becomes active.

# Tests

manual tests + unit tests


andrerom:
+1

yannickroger:
+1

PR #283, opened 2015-07-20
EZP-24579 Change URL pattern to call PlatformUI

StephaneDiot:
https://jira.ez.no/browse/EZP-24579

## Description

Changing the url to call, to load the PlatformUI application. It's now "myDomainName"/ez#/ instead of "myDomainName"/shell#/

### DIFF
--- a/Resources/config/routing.yml
+++ b/Resources/config/routing.yml
@@ -1,6 +1,6 @@
 
-shell:
-    path: /shell
+ez:
+    path: /ez
     defaults:
         _controller: ezsystems.platformui.controller:shellAction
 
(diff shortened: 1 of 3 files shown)


StephaneDiot:
ping @dpobel

Resources/config/routing.yml
```diff
@@ -1,6 +1,6 @@
 
 shell:
```
dpobel: you should change that as well (for consistency)

dpobel:
beside the inline comment, +1

mhyndle:
+1

andrerom:
+1

PR #285, opened 2015-07-21
EZP-24612: Improved support of PSR-2

lolautruche:
> https://jira.ez.no/browse/EZP-24612

Added php-cs-fixer configuration file and fixed CS with new rules.


lolautruche:
Review ping @andrerom @bdunogier @dpobel @glye @yannickroger @pspanja

andrerom:
+1

PR #286, opened 2015-07-21
EZP-24129: Application Config Provider

bdunogier:
> Implements https://jira.ez.no/browse/EZP-24129
> Status: approved

Adds an Aggregator, passed on to the PlatformUIController, that aggregates a bunch of `ApplicationConfigProviders`.

The existing configs were converted to their own implementation of `ApplicationConfigProvider`: `AnonymousUserId`, `CountriesInfo`, `SessionInfo`.

A compiler pass parses the `ezsystems.platformui.application_config_provider` tag, and registers them in the Aggregator. The `key` attribute of the tag is used as the namespace. 

The controller assigns the whole thing to the template as the `parameters` array.

### TODO
- [x] Improve naming, it's terrible right now
- [x] Write tests
- [x] See if the JS code that passes on the config to the app should be updated somehow. With the current code, `shell.html.twig` must be changed to add new config namespaces.
- [x] Add Root Parameter Provider (with assets root and rest root)
- [x] Squash once review is finished

### DIFF
--- a/ApplicationConfig/Aggregator.php
+++ b/ApplicationConfig/Aggregator.php
@@ -1,0 +1,39 @@
+<?php
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+namespace EzSystems\PlatformUIBundle\ApplicationConfig;
+
+/**
+ * Aggregates a set of ApplicationConfig Providers.
+ */
+class Aggregator implements Provider
+{
+    /** @var Provider[] ApplicationConfigProviders, indexed by namespace string*/
+    private $providers = [];
+
+    /**
+     * Adds an array of Provider to the aggregator.
+     * @param \EzSystems\PlatformUIBundle\ApplicationConfig\Provider[] $providers
+     */
+    public function addProviders(array $providers)
+    {
+        $this->providers = array_merge($this->providers, $providers);
+    }
+
+    /**
+     * Aggregates the config from the providers, and returns a hash with the namespace as the key, and the config
+     * as the value.
+     * @return array
+     */
+    public function getConfig()
+    {
+        $config = [];
+        foreach ($this->providers as $key => $provider) {
+            $config[$key] = $provider->getConfig();
+        }
+
+        return $config;
+    }
+}
--- a/ApplicationConfig/Provider.php
+++ b/ApplicationConfig/Provider.php
@@ -1,0 +1,17 @@
+<?php
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+namespace EzSystems\PlatformUIBundle\ApplicationConfig;
+
+/**
+ * Provides parameters as a serializable value.
+ */
+interface Provider
+{
+    /**
+     * @return mixed Anything that is serializable via json_encode()
+     */
+    public function getConfig();
+}
--- a/ApplicationConfig/Providers/Value.php
+++ b/ApplicationConfig/Providers/Value.php
@@ -1,0 +1,29 @@
+<?php
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+namespace EzSystems\PlatformUIBundle\ApplicationConfig\Providers;
+
+use EzSystems\PlatformUIBundle\ApplicationConfig\Provider;
+
+/**
+ * Simple value provider that passes on the value it is given in the constructor.
+ *
+ * Can be used for container config.
+ */
+class Value implements Provider
+{
+    /** @var mixed */
+    private $value;
+
+    public function __construct(array $value)
+    {
+        $this->value = $value;
+    }
+
+    public function getConfig()
+    {
+        return $this->value;
+    }
+}
--- a/Controller/PlatformUIController.php
+++ b/Controller/PlatformUIController.php
@@ -10,50 +10,16 @@
 namespace EzSystems\PlatformUIBundle\Controller;
 
 use eZ\Bundle\EzPublishCoreBundle\Controller;
-use Symfony\Component\HttpFoundation\Session\SessionInterface;
-use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
+use EzSystems\PlatformUIBundle\ApplicationConfig\Provider;
 
 class PlatformUIController extends Controller
 {
-    /**
-     * @var \Symfony\Component\HttpFoundation\Session\SessionInterface
-     */
-    private $session;
+    /** @var Provider */
+    private $configAggregator;
 
-    /**
-     * @var \Symfony\Component\Security\Csrf\CsrfTokenManagerInterface
-     */
-    private $csrfTokenManager;
-
-    /**
-     * @var string
-     */
-    private $csrfTokenIntention;
-
-    /**
-     * The configured anonymous user id.
-     *
-     * @var int
-     */
-    private $anonymousUserId;
-
-    /**
-     * @var array
-     */
-    protected $countriesInfo;
-
-    public function __construct(
-        array $countriesInfo,
-        SessionInterface $session,
-        CsrfTokenManagerInterface $csrfTokenManager,
-        $restIntention = 'rest',
-        $anonymousUserId = 10
-    ) {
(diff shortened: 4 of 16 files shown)


ApplicationConfigProviders/Aggregator.php
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Aggregates configuration from ApplicationConfigProviders
+ */
+class Aggregator
+{
+    /** @var ApplicationConfigProvider[] */
+    private $providers = [];
+
+    public function addProviders( array $providers )
```
dpobel: this method deserves an API doc explaining that `$providers` is supposed to be a hash and that the keys are used in the config returned by `getConfig`

dpobel:
very good :) this is more or less what I had in mind.

The naming is not that bad, maybe we should aim at having the following namespaces/classes:
* EzSystems\PlatformUIBundle\ApplicationConfig\Aggregator
* EzSystems\PlatformUIBundle\ApplicationConfig\Provider (Interface suffix?)
* EzSystems\PlatformUIBundle\ApplicationConfig\Providers\AnonymousUserId
* EzSystems\PlatformUIBundle\ApplicationConfig\Providers\Countries (unsure about the Info suffix on this one)
* EzSystems\PlatformUIBundle\ApplicationConfig\Providers\SessionInfo

for your third TODO item, we should aim at having something like:
```js
var app = new Y.eZ.PlatformUIApp({
    container: '.ez-platformui-app',
    viewContainer: '.ez-view-container',
    // some others required parameters...
    config: {{ parameters|json_encode|raw }},
});
```

that way the logic in shell.html.twig is minimal but this requires some changes in the App code which are out of the scope of this PR (unless you want to give a try :p)

Resources/views/PlatformUI/shell.html.twig
```diff
@@ -54,31 +54,31 @@
                     viewContainer: '.ez-view-container',
                     root: "{{ app.request.attributes.get( 'semanticPathinfo' ) }}",
                     assetRoot: "{{ asset( '/' ) }}",
```
dpobel: maybe `root` and `assetRoot` should also be computed in a config provider ?

ApplicationConfigProviders/Aggregator.php
bdunogier: Hmm, late coding isn't always good. It'd even be missing an interface, actually.

Resources/views/PlatformUI/shell.html.twig
bdunogier: We could do that.

bdunogier:
Do we go for "config" or "parameter" in naming overall ? We even mix them in `shell.html.twig` ;-)

dpobel:
good remark, given the snippet in https://github.com/ezsystems/PlatformUIBundle/pull/286#issuecomment-123601845 indeed, `config` seems better

Controller/PlatformUIController.php
```diff
@@ -63,25 +29,9 @@ public function __construct(
...
-            'sessionInfo' => $sessionInfo,
-            'anonymousUserId' => $this->generateUrl(
-                'ezpublish_rest_loadUser',
-                ['userId' => $this->anonymousUserId]
-            ),
-            'countriesInfo' => $this->countriesInfo,
-        ]);
+        return $this->render(
+            'eZPlatformUIBundle:PlatformUI:shell.html.twig',
+            ['parameters' => $this->parametersAggregator->getConfig()]
```
dpobel: so still `parameters` ?

Resources/views/PlatformUI/shell.html.twig
```diff
@@ -52,33 +52,34 @@
             var app = new Y.eZ.PlatformUIApp({
                     container: '.ez-platformui-app',
                     viewContainer: '.ez-view-container',
-                    root: "{{ app.request.attributes.get( 'semanticPathinfo' ) }}",
-                    assetRoot: "{{ asset( '/' ) }}",
-                    anonymousUserId: "{{ anonymousUserId }}",
+                    config: {{ parameters|json_encode|raw }},
```
dpobel: since the App is not yet able to understand that, I would not add this line in this PR but we'll do that in a follow up issue/PR.

dpobel:
Besides the 2 minor inline comments and the CS issue, +1

dpobel:
ping @yannickroger @andrerom @lolautruche @pspanja

ApplicationConfig/Aggregator.php
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+namespace EzSystems\PlatformUIBundle\ApplicationConfig;
+
+/**
+ * Aggregates a set of ApplicationConfig Providers.
+ */
+class Aggregator
```
lolautruche: Shouldn't it implement the `Provider` interface?

Controller/PlatformUIController.php
```diff
@@ -10,50 +10,16 @@
...
-        CsrfTokenManagerInterface $csrfTokenManager,
-        $restIntention = 'rest',
-        $anonymousUserId = 10
-    ) {
-        $this->session = $session;
-        $this->csrfTokenManager = $csrfTokenManager;
-        $this->csrfTokenIntention = $restIntention;
-        $this->anonymousUserId = $anonymousUserId;
-        $this->countriesInfo = $countriesInfo;
+    public function __construct(Aggregator $parametersAggregator)
```
lolautruche: Assuming you take my first comment into account (aggregator implementing `Provider` interface), you should type hint against the interface instead.

ApplicationConfig/Provider.php
```diff
@@ -0,0 +1,17 @@
...
+
+namespace EzSystems\PlatformUIBundle\ApplicationConfig;
+
+/**
+ * Provides parameters as a serializable value.
+ */
+interface Provider
+{
+    /**
+     * @return array|string|int|\JsonSerializable
```
lolautruche: Why so many possible types? Couldn't we agree on an array?

Resources/views/PlatformUI/shell.html.twig
bdunogier: hmmm, I did not intend to remote them, as a matter of fact :-)

ApplicationConfig/Aggregator.php
bdunogier: That's a good point, it indeed could.

Controller/PlatformUIController.php
bdunogier: Yep, will do :-)

ApplicationConfig/Provider.php
bdunogier: Nah, array won't do (anonymousUserId is a simple string, for instance). But it could be `mixed anything that can be serialized with json_encode()`

Resources/views/PlatformUI/shell.html.twig
bdunogier: Oh, you meant the addition of the variable. Okay, I'll remove it. I've looked into the JS code, and I really can't change that without a bit of training first :-)

ApplicationConfig/Provider.php
lolautruche: :+1:

ApplicationConfig/Providers/Value.php
```diff
@@ -0,0 +1,29 @@
...
+ * Simple value provider that passes on the value it is given in the constructor.
+ *
+ * Can be used for container config.
+ */
+class Value implements Provider
+{
+    /** @var mixed */
+    private $value;
+
+    public function __construct(array $countriesInfo)
```
lolautruche: `$countriesInfo`? Shouldn't this be generic?

lolautruche:
+1

ApplicationConfig/Providers/Value.php
bdunogier: Oups. It was first named CountriesInfo, I made it generic afterwards :)

andrerom:
looks good besides nitpick :)

PR #288, opened 2015-07-23
Fix EZP-24380: Move section form management from PlatformUI to RepositoryForms

lolautruche:
> https://jira.ez.no/browse/EZP-23480


lolautruche:
Now ready for review

@andrerom @bdunogier @dpobel @yannickroger @glye @pspanja

andrerom:
+1, before you rebase you can now merge in #307 if the change is ok for you, I rebased it on top of this one.

Resources/views/Section/view.html.twig
```diff
@@ -38,28 +38,33 @@
...
-                            >
-                        {{ 'section.edit'|trans }}
-                    </a>
+            {% if canEdit %}
+                <a href="{{ path('admin_sectionedit', {'sectionId': section.id}) }}"
+                   class="pure-button ez-button" data-icon="&#xe606;">
+                    {{ 'section.edit'|trans }}
+                </a>
 
+                <a href="{{ path("admin_sectiondelete", {"sectionId": section.id}) }}"
```
dpobel: that should be button to post a form, deleting something with a GET request is bad.

Controller/SectionController.php
```diff
@@ -176,63 +101,40 @@ public function viewAction($sectionId)
...
-                        ['sectionId' => $newSection->id]
-                    )
-                );
-            } catch (UnauthorizedException $e) {
-                return $this->forward('eZPlatformUIBundle:Pjax:accessDenied');
-            } catch (InvalidArgumentException $e) {
-                $this->addAlreadyExistErrorMessage();
-            }
+        try {
+            $section = $this->sectionService->createSection(new SectionCreateStruct([
```
dpobel: by doing that, as soon as the user clicks on "Create a new section", the section is created before he validates the form. Creating something on the server with a GET request is bad (like deleting) and worse, if the user navigates elsewhere or closes his browser, the new section has been created but the uses has absolutely no clue about that.

lolautruche: I know, just like for ContentType... However this is needed to workaround the fact that we have different structs in the API for create and for update.

lolautruche: Alternative would be to implement versioning for sections like @andrerom already suggested, but looks very overkill to me.

Resources/views/Section/list.html.twig
```diff
@@ -49,35 +47,59 @@
...
+                        <td>
+                        {% if canEdit %}
+                            <a href="{{ path('admin_sectionedit', {'sectionId': section.id}) }}" class="pure-button ez-button" data-icon="&#xe606;">{{ 'section.edit'|trans }}</a>
+                        {% else %}
+                            <span class="pure-button ez-button pure-button-disabled" data-icon="&#xe606;">{{ 'section.edit'|trans }}</span>
+                        {% endif %}
+                        </td>
+                        <td>
+                        {% if canEdit %}
+                            <a href="{{ path('admin_sectiondelete', {'sectionId': section.id}) }}"
```
dpobel: button + form here as well

Resources/views/Section/view.html.twig
lolautruche: And doing this by posting a form complicates things a lot :smiley:

bdunogier: Unfortunately I have to agree on this. A pragmatic reason is that doing it via GET would skip the Form Token. And that's kinda very bad.

lolautruche:
All remarks taken into account. Ready for another round of review, especially from master @dpobel :smile:

dpobel:
+1

Yoda
The real master is saying: *Merge this pull request, you can* :)

Joke aside, we'll have to change the redirection after the form submission as described in https://jira.ez.no/browse/EZP-23700 so to avoid having the inconsistency between the app URI and what is actually displayed.

PR #289, opened 2015-07-24
Fix edit content button

mhyndle:
Edit content button was lacking the content passed to attribute.

### DIFF
--- a/Tests/js/views/assets/ez-actionbarview-tests.js
+++ b/Tests/js/views/assets/ez-actionbarview-tests.js
@@ -37,49 +37,46 @@ YUI.add('ez-actionbarview-tests', function (Y) {
             this.view.destroy();
         },
 
-        "Should instantiate createContentActionView with contentType": function () {
-            var createContentActionFound = false;
+        _isActionCreated: function (actionId, testedAttributes) {
+            var actionFound;
 
-            for (var i = 0; i < this.view.get('actionsList').length; i++){
-                if (this.view.get('actionsList')[i].get('actionId') === 'createContent'){
-                    createContentActionFound = true;
+            actionFound = Y.Array.find(this.view.get('actionsList'), function (actionView) {
+                return actionView.get('actionId') === actionId;
+            });
+
+            if (!!actionFound) {
+                Y.Object.each(testedAttributes, function (attrValue, attrName) {
                     Assert.areSame(
-                        this.view.get('contentType'),
-                        this.view.get('actionsList')[i].get('contentType'),
-                        'The contentType should have been set to createContentActionView'
+                        attrValue,
+                        actionFound.get(attrName),
+                        'The ' + attrName + ' should have been set to ' + actionId + 'ActionView'
                     );
-                }
+                });
             }
 
             Assert.isTrue(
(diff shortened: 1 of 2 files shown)


mhyndle:

dpobel:
no unit test ?

mhyndle:
tests added

Tests/js/views/assets/ez-actionbarview-tests.js
```diff
@@ -37,49 +37,60 @@ YUI.add('ez-actionbarview-tests', function (Y) {
             this.view.destroy();
         },
 
-        "Should instantiate createContentActionView with contentType": function () {
-            var createContentActionFound = false;
+        _actionAssertions: function (testedActionId) {
```
dpobel: bad name, method should start with a verb, so here `_isActionCreated` or something like that.

Tests/js/views/assets/ez-actionbarview-tests.js
```diff
@@ -37,49 +37,60 @@ YUI.add('ez-actionbarview-tests', function (Y) {
...
-                    createContentActionFound = true;
-                    Assert.areSame(
-                        this.view.get('contentType'),
-                        this.view.get('actionsList')[i].get('contentType'),
-                        'The contentType should have been set to createContentActionView'
-                    );
+                if (this.view.get('actionsList')[i].get('actionId') === testedActionId){
+                    actionFound = true;
+
+                    switch ( testedActionId ) {
```
dpobel: not generic, instead the assert method should take in second argument the attributes if should check, something like:
```js
this._isActionCreated('translate', {
    location: this.view.get('location'),
    content: this.view.get('content'),
});
```

Tests/js/views/assets/ez-actionbarview-tests.js
```diff
@@ -37,49 +37,49 @@ YUI.add('ez-actionbarview-tests', function (Y) {
...
-                    createContentActionFound = true;
-                    Assert.areSame(
-                        this.view.get('contentType'),
-                        this.view.get('actionsList')[i].get('contentType'),
-                        'The contentType should have been set to createContentActionView'
-                    );
+                if (this.view.get('actionsList')[i].get('actionId') === actionId){
+                    actionFound = true;
+
+                    /*jshint loopfunc: true */
```
mhyndle: I put the directive to ignore loopfunc tests on it as creating external method for just checking the assertion seems too complicated for me.

StephaneDiot:
+1

Tests/js/views/assets/ez-actionbarview-tests.js
dpobel: you can also just find the action view in the loop (btw, Y.Array.find would be a bit nicer than this for loop) and then iterate to compare the attribute.

mhyndle: Oh right, I have separated iteration through actions and iteration through attributes.

mhyndle:
travis test fails, however not on js part ;)

dpobel:
+1

PR #291, opened 2015-07-27
Fix EZP-24581: Creating a new content type failed with error: Failed …

glye:
…to load '/contenttype/create/1'

Solves it by adding unique string to identifier.

https://jira.ez.no/browse/EZP-24581

See also https://github.com/ezsystems/repository-forms/pull/38

### DIFF
--- a/Controller/ContentTypeController.php
+++ b/Controller/ContentTypeController.php
@@ -130,7 +130,7 @@ class ContentTypeController extends Controller
             $contentTypeGroup = $this->contentTypeService->loadContentTypeGroup($contentTypeGroupId);
 
             $contentTypeCreateStruct = new ContentTypeCreateStruct([
-                'identifier' => 'new_content_type',
+                'identifier' => '__new__' . md5(microtime(true)),
                 'mainLanguageCode' => $languageCode,
                 'names' => [$languageCode => 'New ContentType'],
             ]);


Controller/ContentTypeController.php
```diff
@@ -130,7 +130,7 @@ public function createContentTypeAction($contentTypeGroupId, $languageCode = nul
             $contentTypeGroup = $this->contentTypeService->loadContentTypeGroup($contentTypeGroupId);
 
             $contentTypeCreateStruct = new ContentTypeCreateStruct([
-                'identifier' => 'new_content_type',
+                'identifier' => '__new__' . md5(microtime(true)),
```
andrerom: given this is UI maybe it would be enough to use `time()` or at least skip md5 so it is a human readable increasing number as opposed to hash.

glye: It's done this way because that's how @lolautruche did it for section creation. The string is not supposed to be visible in the UI, and the user is supposed to replace it with something more readable anyway (similar to the content type name). It's mainly there to enable quick creation of test types, I guess.

lolautruche: Yeah, this is not meant to appear in UI. This identifier is purely internal, to ensure that it's unique.

lolautruche:
+1

yannickroger:
+1

andrerom:
+1

PR #292, opened 2015-07-28
EZP-24626 As an editor I want to see content type description when editing content

StephaneDiot:
https://jira.ez.no/browse/EZP-24626

## Description

A description of the content type should be shown when passing cursor over the header while editing or creating a content. On 'touch friendly' devices the description should be shown when tapping on header, and should unshow when tapping outside (the same rule has also been added for the technical infos).

## Screencast

http://youtu.be/Y97RexOHWZo
(first part is 'touch friendly' emulated, second is like in a normal browser)

### DIFF
--- a/Resources/public/js/views/ez-contenteditview.js
+++ b/Resources/public/js/views/ez-contenteditview.js
@@ -11,7 +11,7 @@ YUI.add('ez-contenteditview', function (Y) {
      */
     Y.namespace('eZ');
 
-    var IS_SHOWING_TECH = 'is-showing-technicalinfos',
+    var IS_SHOWING_INFOS = 'is-showing-infos',
         CONTENT_SEL = '.ez-main-content',
         ESCAPE_KEY = 27,
         FORM_CONTAINER = '.ez-contenteditformview-container',
@@ -30,7 +30,8 @@ YUI.add('ez-contenteditview', function (Y) {
             '.ez-view-close': {'tap': '_closeView'},
             'header': {
                 'mouseover': '_showDetails',
-                'mouseout': '_hideDetails'
+                'mouseout': '_hideDetails',
+                'tap': '_showDetails',
             },
             '.ez-main-content': {
                 'keyup': '_handleKeyboard'
@@ -156,7 +157,7 @@ YUI.add('ez-contenteditview', function (Y) {
          * @protected
          */
         _showDetails: function () {
-            this.get('container').addClass(IS_SHOWING_TECH);
+            this.get('container').addClass(IS_SHOWING_INFOS);
         },
 
         /**
@@ -166,7 +167,7 @@ YUI.add('ez-contenteditview', function (Y) {
          * @protected
          */
         _hideDetails: function () {
-            this.get('container').removeClass(IS_SHOWING_TECH);
+            this.get('container').removeClass(IS_SHOWING_INFOS);
         },
 
         /**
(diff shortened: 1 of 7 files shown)


StephaneDiot:
ready for review @dpobel @mhyndle @yannickroger

Resources/public/js/views/ez-contenteditview.js
```diff
@@ -53,6 +56,18 @@ YUI.add('ez-contenteditview', function (Y) {
                 }
             });
 
+            this.after('activeChange', function (e) {
+                if ( e.newVal ) {
+                    that._attachedViewEvents.push(
```
dpobel: `this` is perfectly usable here and given the *complexity* of the code, that would deserve an external method

glye:
+0.1, not my domain :)
You can rebase on master to get rid of the ezrobot warning.

Resources/public/templates/contentedit.hbt
```diff
@@ -3,14 +3,17 @@
...
-                <li>Created by {{ owner.name }}</li>
-                {{#if content.contentId}}
-                <li>{{ content.lastModificationDate }}</li>
-                <li>Object ID: {{ content.contentId }}, Node ID: {{ mainLocation.locationId }}</li>
-                {{/if}}
-            </ul>
+            <div class="ez-infos">
+                <ul class="ez-technical-infos">
+                    <li>{{ contentType.names.[eng-GB] }}</li>
+                    <li>Created by {{ owner.name }}</li>
```
glye: Should this be a translateable string?

glye:
+1 apart from question above.

Resources/public/templates/contentedit.hbt
andrerom: This might be because it is ContentInfo object underneath, if so something we would have to fix later. Or @StephaneDiot ?

andrerom:
+1

Resources/public/js/views/ez-contenteditview.js
```diff
@@ -50,13 +51,35 @@ YUI.add('ez-contenteditview', function (Y) {
...
         },
 
         /**
+         * Hide infos (description and technical infos) when clicking outside the header
+         *
+         * @method _removeInfos
+         * @protected
+         * @param {Object} e event facade
+         */
+        _removeInfos: function (e) {
```
dpobel: this is a very bad name (and description in the comment). This method does not remove anything, it sets an event handler on the `clickoutside` event that will hide the infos. So please rename it to something like `_addClickoutsideInfoHandler` and use `_hideDetails` instead of reimplementing it in an anonymous function.

Resources/public/js/views/ez-contenteditview.js
```diff
@@ -50,13 +51,35 @@ YUI.add('ez-contenteditview', function (Y) {
...
+            var container = this.get('container');
+
+            this._attachedViewEvents.push(
+                container.one('header').on(
```
dpobel: use a class instead of the tag name

Resources/public/js/views/ez-contenteditview.js
```diff
@@ -50,13 +51,35 @@ YUI.add('ez-contenteditview', function (Y) {
             this.after('activeChange', function (e) {
                 if ( e.newVal ) {
                     this._setFocus();
+                    this._removeInfos();
                 }
             });
 
+
```
dpobel: not the most useful change ever

StephaneDiot:
updated ! @dpobel

dpobel:
+1

PR #294, opened 2015-08-05
Fix EZP-24685: Role UI (list, view, CRUD)

glye:
https://jira.ez.no/browse/EZP-24685

TODO
- [x] Dashboard entry
- [x] Role list
- [x] Role view
- [x] Role CRUD
- [x] Role assignment view
- [x] Show human readable user/group assignment
- [x] Use different main icon, not the file archive

Requires https://github.com/ezsystems/repository-forms/pull/40 for role editing.

Note: Cancel editing is out of scope now, because it turns out that role versioning is missing. (Cancel does the same as save, for now.) See https://jira.ez.no/browse/EZP-24701

Follow-up:
- https://jira.ez.no/browse/EZP-24699 Enrich Role limitations view with links and human readable names
- https://jira.ez.no/browse/EZP-24700 Role assignment CRUD
- https://jira.ez.no/browse/EZP-24701 Role versioning

Dashboard entry

Role list

Role view


glye:
Proposing https://github.com/ezsystems/ezpublish-kernel/pull/1370 as a way to show human readable limitation values and user/usergroup names.

glye:
Added user and users iconomoon glyphs.

glye:
Added role deleting.

glye:
Fixed issue reported by QA in https://jira.ez.no/browse/EZP-24685, and added notification as a part of the fix.

glye:
Disabled "role draft saved" notification, until we're actually doing that.

Controller/RoleController.php
```diff
@@ -0,0 +1,183 @@
...
+                return $response;
+            }
+
+            return $this->redirect($actionUrl);
+        }
+
+        return $this->render('eZPlatformUIBundle:Role:update_role.html.twig', [
+            'form' => $form->createView(),
+            'action_url' => $actionUrl,
+            'role' => $role,
```
andrerom: Maybe also pass a "is_new" bool flag so template can empty name input and show "New Role" in heading. 
Can detect this with strpos('__new__') === 0 on name.

Form/Processor/RoleFormProcessor.php
```diff
@@ -0,0 +1,116 @@
...
+    public function processRemoveDraft(FormActionEvent $event)
+    {
+        // TODO: This will delete the role completely. Must delete only the draft. Follow-up: EZP-24701
+//        $role = $event->getData()->role;
+//        $this->roleService->deleterole($role);
+//        try {
+//            $this->roleService->deleterole($role);
+//        } catch (UnauthorizedException $e) {
+//            $this->addError('role.notification.unauthorized');
+//        }
```
andrerom: Or, maybe rather here. I guess we can detect here if name contains __new__\^s{32} then delete the role so cancel "works".

Resources/views/Role/update_role.html.twig
```diff
@@ -0,0 +1,36 @@
+{% extends "eZPlatformUIBundle::pjax_admin.html.twig" %}
+{# @var role \eZ\Publish\API\Repository\Values\User\Role #}
+
+{% trans_default_domain "role" %}
+
+{% set edit_title = "role.edit_title"|trans({"%roleName%": role.identifier}) %}
```
andrerom: here use "New Role" (translatable) if `is_new`.

Controller/RoleController.php
andrerom: OR, a `role_name` variable

glye: @andrerom The `role_name` approach is nice except in the role list - hard to solve that in the controller, unless I populate the objects with a custom property.

glye:
I just realised there is no pagination in the role list :( We need that, and BDD for it.

Controller/RoleController.php
andrerom: That was was I was aiming at, if the field comes up empty in interface user will have to fill it to be allowed to save anyway.

glye:
@andrerom Implemented role drafts feature. Note:
- To avoid logic in the template I had to pass two values, `role_name` and `role_input_value`
- Symfony Forms doesn't like that I remove the original `form_widget(form.identifier)` - it puts it back, so there are two fields. I could only get it to work by keeping it but HTML commenting it. Bad hack and I'm sure there must be a better way.

andrerom:
ok, make sure to add story to add identifier as separate field in db so we can cleanup this name/identifier thing a bit, if that is fixed then name can be empty from api side if we allow that.

Controller/RoleController.php
```diff
@@ -0,0 +1,194 @@
...
+            $roleCreateStruct = new RoleCreateStruct([
+                'identifier' => '__new__' . md5(microtime(true)),
+            ]);
+            $role = $this->roleService->createRole($roleCreateStruct);
+        } catch (UnauthorizedException $e) {
+            return $this->forward('eZPlatformUIBundle:Pjax:accessDenied');
+        } catch (InvalidArgumentException $e) {
+            return $this->forward('eZPlatformUIBundle:Pjax:accessDenied'); // TODO: is this correct? Name is already existing, or limitation errors
+        } catch (LimitationValidationException $e) {
+            return $this->forward('eZPlatformUIBundle:Pjax:accessDenied'); // TODO: is this correct? Invalid policy limitation.
```
andrerom: @dpobel Maybe you have some input on these parts?

glye:
Review ping @dpobel @lolautruche @pspanja @andrerom

Controller/RoleController.php
dpobel: as its name suggests, `eZPlatformUIBundle:Pjax:accessDenied` is supposed to be called when the user does not have access to something. So calling it for other cases is not correct. For the others Exception, most of the time, you should add a notification message like https://github.com/ezsystems/PlatformUIBundle/blob/master/Controller/SectionController.php#L287

andrerom: @glye Do you change it in that direction? sync with @dpobel maybe if unclear.

glye: @andrerom Yep, this is fine. Thanks @dpobel

glye:
Added proper notification messages instead of bogus accessDenied.

glye:
Squashed commits.

Controller/RoleController.php
```diff
@@ -0,0 +1,227 @@
...
+            $this->generateUrl('admin_roleList')
+        );
+    }
+
+    /**
+     * Adds an error message to the flashbag.
+     *
+     * @param string $message Translatable error message
+     */
+    private function addErrorMessage($message)
```
andrerom: This should probably be moved out to abstract controller shared between them, and this should also probably allow dynamic input, like details on what specifically went wrong when short versions of that is possible to extract from exception.

andrerom:
+1

Controller/RoleController.php
glye: Ack, followup: https://jira.ez.no/browse/EZP-24729

Form/Processor/RoleFormProcessor.php
```diff
@@ -0,0 +1,117 @@
...
+     * @var NotificationPoolInterface
+     */
+    private $notificationPool;
+
+    /**
+     * @var RoleService
+     */
+    private $roleService;
+
+    public function __construct(RouterInterface $router, NotificationPoolInterface $notificationPool, RoleService $roleService)
```
yannickroger: nitpick: line is too long

yannickroger:
+1

glye:
Fixed long lines.

PR #295, opened 2015-08-05
[Travis] Added Behat testing for PlatformUI

andrerom:
*Just for testing, will need to use correct profile and suite once that has been setup before we merge this.*

- [x] Get the tests to actually run, it is not correct that this is passing
- [x] Update for change to use Selenium
- [x] Update scenario and profile to be relevant tests for PlatformUI
- [ ] ezplatform tests passing: https://github.com/ezsystems/ezplatform/pull/46

### DIFF
--- a/Features/Standard/platformui.feature
+++ b/Features/Standard/platformui.feature
@@ -1,25 +1,25 @@
 Feature: Basic PlatfomrUi interaction tests
 
-    @javascript
+    @common @javascript
     Scenario: As a admin User, I want to login to PlatformUI
         Given I go to PlatformUI app with username "admin" and password "publish"
         Then I should see "Welcome to eZ Platform" title
 
-    @javascript
+    @commom @javascript
     Scenario: As a admin User, I want to minimize the discovery bar
         Given I go to PlatformUI app with username "admin" and password "publish"
         And I click on the navigation zone "Content"
         When I click on the navigation item "Content structure"
         Then I click on the discovery bar button "Minimize"
 
-    @javascript
+    @commom @javascript
     Scenario: As a admin User, I want to minimize the action bar
         Given I go to PlatformUI app with username "admin" and password "publish"
         And I click on the navigation zone "Content"
         When I click on the navigation item "Content structure"
         Then I click on the action bar button "Minimize"
 
-    @javascript
+    @commom @javascript
     Scenario: As a admin User, I want to minimize the action bar
         Given I go to PlatformUI app with username "admin" and password "publish"
         And I click on the navigation zone "Content"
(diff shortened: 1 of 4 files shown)


andrerom:
After 23 runs on travis this is now ready for tests, your turn @ezsystems/qa-team ;)

pcardiga:
Roger :)

bin/travis/prepare_behat.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/sh
+
+# File for setting up system for behat testing, just like done in DemoBundle's .travis.yml
+
+# Install Apache before we add external repo for PHP 5.5 as it will conflict
+sudo apt-get update -q
+sudo apt-get install -q -y --force-yes apache2 libapache2-mod-fastcgi
+
+# Install PHP 5.5, apt-get package is php 5.3 since travis uses ancient Ubuntu
+sudo add-apt-repository -y ppa:ondrej/php5
```
dpobel: you should use (and complete) bin/travis/setupphp.sh

andrerom: I can do that, but then it will install fpm, mysqlnd, ..

andrerom:
Ready for review, @dpobel @yannickroger @StephaneDiot @mhyndle

bin/travis/setupphp.sh
```diff
@@ -2,4 +2,4 @@
 
 add-apt-repository -y ppa:ondrej/php5
 apt-get update -q
-apt-get install -y php5-cli php5-common php5-gd php5-xsl php5-intl
+sudo apt-get install -q -y --force-yes php5-cli php5-common php5-gd php5-xsl php5-intl
```
dpobel: you don't need to add sudo here as the script is already executed with sudo in .travil.yml and in prepare_behat.sh.

dpobel:

bin/travis/setupphp.sh
andrerom: I did not use sudo in .travis.yml for this one, as I try to keep track if our sudo use for possibility to use travis container infrastructure. But I think they will fix that so if you want I can change it back to sudo in the yml file again.

yannickroger:
+1

Features/Standard/platformui.feature
```diff
@@ -1,28 +1,28 @@
 Feature: Basic PlatfomrUi interaction tests
```
pcardiga: Little Nitpick: There is a typo in "PlatfomrUi". Should be "PlatformUi" :)

andrerom: Not part of the PR,PR welcome :ship:

andrerom:
And this is why I "hate"* Travis for functional testing :(


\** when something breaks there is a million things that could be the reason, and no clear way to debug*

andrerom:
Closing as we got it working with ezci.

PR #297, opened 2015-08-13
QA-436 / QA-437 / QA-438 Move, Copy and Remove sentences implementations

miguelcleverti:
Implementations of the Move/Copy/Remove sentences for PlatfomUI BDD.

Jira Issues:
https://jira.ez.no/browse/QA-436
https://jira.ez.no/browse/QA-437
https://jira.ez.no/browse/QA-438

depends on:
ezsystems/behatbundle#28

### DIFF
--- a/Features/Context/SubContext/Authentication.php
+++ b/Features/Context/SubContext/Authentication.php
@@ -51,6 +51,7 @@ trait Authentication
 
     /**
      * @Given I am logged in as admin on PlatformUI
+     * @Given I am logged in as an Editor in PlatformUI
      */
     public function loggedAsAdminPlatformUi()
     {
--- a/Features/Context/PlatformUI.php
+++ b/Features/Context/PlatformUI.php
@@ -93,6 +93,14 @@ class PlatformUI extends Context
     }
 
     /**
+     * @Given I am on :name full view
+     */
+    public function onFullView($name)
+    {
+        $path = $this->getBasicContentManager()->getContentPath($name);
+        $this->openTreePath($path);
+    }
+    /**
      * @Then I see Content :contentName of type :contentType
      */
     public function contentExists($contentName, $contentType)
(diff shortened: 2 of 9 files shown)


miguelcleverti:
@andrerom @pcardiga @micszo

miguelcleverti:
:triumph: ezrobot :triumph:

andrerom:
+1


Side, there is no exception on the loading spin function after x time, it probably should, ref
```
         while ($page->find('css', $loadingSelector) != null) {
             usleep(100 * 1000); // 100ms
         }
```

Features/Context/SubContext/Authentication.php
```diff
@@ -51,6 +51,7 @@ public function goToPlatformUiAndLogIn($username, $password)
 
     /**
      * @Given I am logged in as admin on PlatformUI
+     * @Given I am logged in as an Editor in PlatformUI
      */
     public function loggedAsAdminPlatformUi()
```
joaoinacio: Nitpick / OT: 
Not necessarily admin user, maybe `loggedInAsUserInPlatformUi()` ?
Also IMO since the trait is using vars from the context instance, at least document it?

miguelcleverti:
@andrerom This is a known issue, the waiting mechanism will mostly need to be refactored and it is not really related to this PR in specific

andrerom:
Miguel: ok, ping when Joao is fine with it as well so we have 2x +1, and I'll merge (but the other one first)

Features/Context/SubContext/Authentication.php
miguelcleverti: These involves changes in various places, a dedicated PR would be better

andrerom: I agree, could one of you do a separate PR on this once we have travis up and running for PR sanity?

andrerom:
Ok, thanks everyone :)

PR #300, opened 2015-08-25
Fix EZP-24701: Role versioning

glye:
Makes use of role draft features added in main PR in ezpublish-kernel.

Requirement for https://github.com/ezsystems/ezpublish-kernel/pull/1376
Also see https://github.com/ezsystems/repository-forms/pull/41

https://jira.ez.no/browse/EZP-24701

### DIFF
--- a/Controller/RoleController.php
+++ b/Controller/RoleController.php
@@ -8,6 +8,7 @@
  */
 namespace EzSystems\PlatformUIBundle\Controller;
 
+use eZ\Publish\API\Repository\Exceptions\NotFoundException;
 use eZ\Publish\API\Repository\RoleService;
 use eZ\Publish\Core\MVC\Symfony\Security\Authorization\Attribute;
 use eZ\Publish\Core\Repository\Values\User\RoleCreateStruct;
@@ -116,8 +117,15 @@ class RoleController extends Controller
      */
     public function updateRoleAction(Request $request, $roleId)
     {
-        $role = $this->roleService->loadRole($roleId);
-        $roleData = (new RoleMapper())->mapToFormData($role);
+        try {
+            $roleDraft = $this->roleService->loadRoleDraft($roleId);
+        } catch (NotFoundException $e) {
+            // The draft doesn't exist, let's create one
+            $role = $this->roleService->loadRole($roleId);
+            $roleDraft = $this->roleService->createRoleDraft($role);
+        }
+
+        $roleData = (new RoleMapper())->mapToFormData($roleDraft);
         $form = $this->createForm(new RoleUpdateType(), $roleData);
         $actionUrl = $this->generateUrl('admin_roleUpdate', ['roleId' => $roleId]);
 
@@ -135,10 +143,11 @@ class RoleController extends Controller
             $hasErrors = true;
         }
 
-        // TODO: Just a temporary implementation of draft handling. To be done properly in follow-up: EZP-24701
         $formView = $form->createView();
-        $roleName = $role->identifier;
-        if (preg_match('/^__new__[a-z0-9]{32}$/', $roleName) === 1) {
+
+        // Show empty text input when name is not set, while showing "New role" in the page title
+        $roleName = $roleDraft->identifier;
+        if ($roleData->isNew()) {
             $roleName = 'role.name_new';
             $formView->vars['role_input_value'] = '';
         } else {
@@ -148,7 +157,7 @@ class RoleController extends Controller
         return $this->render('eZPlatformUIBundle:Role:update_role.html.twig', [
             'form' => $formView,
             'action_url' => $actionUrl,
-            'role' => $role,
+            'role_draft' => $roleDraft,
             'role_name' => $roleName,
             'hasErrors' => $hasErrors,
         ]);
--- a/Form/Processor/RoleFormProcessor.php
+++ b/Form/Processor/RoleFormProcessor.php
@@ -10,6 +10,9 @@
  */
 namespace EzSystems\PlatformUIBundle\Form\Processor;
 
+use eZ\Publish\API\Repository\Exceptions\NotFoundException;
+use eZ\Publish\API\Repository\RoleService;
+use eZ\Publish\API\Repository\Values\User\RoleDraft;
 use EzSystems\PlatformUIBundle\Http\FormProcessingDoneResponse;
 use EzSystems\PlatformUIBundle\Notification\NotificationPoolAware;
 use EzSystems\PlatformUIBundle\Notification\NotificationPoolInterface;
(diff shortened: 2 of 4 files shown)


glye:
Rebased on master after @lolautruche changes. Moved logic to repoforms, keeping only notification and response.

glye:
Rebased on master after more changes.

Controller/RoleController.php
```diff
@@ -89,13 +89,13 @@ public function viewRoleAction($roleId)
      */
     public function createRoleAction()
     {
-        if (!$this->isGranted(new Attribute('role', 'update'))) {
+        try {
+            $roleCreateStruct = new RoleCreateStruct(['identifier' => '__new__' . md5(microtime(true))]);
+            $role = $this->roleService->createRoleDraft($roleCreateStruct);
+        } catch (UnauthorizedException $e) {
```
lolautruche: Why do you remove this change? `isGranted()` check was made on purpose, to clean up code (should check role/create though)

Controller/RoleController.php
```diff
@@ -113,14 +113,35 @@ public function updateRoleAction(Request $request, $roleId)
...
+                return $this->redirectToRoute('admin_roleList');
+            }
 
-            return $this->redirectToRoute('admin_roleList');
+            try {
+                $roleDraft = $this->roleService->createRoleDraftByRole($role);
+            } catch (UnauthorizedException $e) {
+                return $this->forward('eZPlatformUIBundle:Pjax:accessDenied');
+            } catch (InvalidArgumentException $e) {
+                $this->notifyError('role.error.has_draft', [], 'role');
```
lolautruche: This should not happen since you already tried to load the role draft. This catch block is useless IMHO

Controller/RoleController.php
```diff
@@ -113,14 +113,35 @@ public function updateRoleAction(Request $request, $roleId)
...
+
+                return $this->redirectToRoute('admin_roleList');
+            } catch (LimitationValidationException $e) {
```
lolautruche: This catch block is also useless since you don't create a role draft with policies.

Controller/RoleController.php
glye: Based on ContentType, as you said :P Ok, I can revert that bit.

Controller/RoleController.php
glye: I do, but the policies are copied from the presumably valid published role, so must still be valid here. OK.

glye:
Feedback fixes.

glye:
Rebased on master. Edit redirects are not working anymore, TODO...

Controller/RoleController.php
```diff
@@ -116,8 +118,23 @@ public function createRoleAction(Request $request)
      */
     public function updateRoleAction(Request $request, $roleId)
     {
-        $role = $this->roleService->loadRole($roleId);
-        $roleData = (new RoleMapper())->mapToFormData($role);
+        try {
+            $roleDraft = $this->roleService->loadRoleDraft($roleId);
+        } catch (NotFoundException $e) {
```
lolautruche: You don't need to catch those any more

Controller/RoleController.php
```diff
@@ -116,8 +118,23 @@ public function createRoleAction(Request $request)
...
+            // The draft doesn't exist, let's create one
+            try {
+                $role = $this->roleService->loadRole($roleId);
+                $roleDraft = $this->roleService->createRoleDraft($role);
+            } catch (UnauthorizedException $e) {
+                return $this->forward('eZPlatformUIBundle:Pjax:accessDenied');
+            } catch (NotFoundException $e) {
```
lolautruche: You don't need to catch those any more (`NotFoundException` / `UnauthorizedException`) as it's now handled by Symfony under the hood. Besides, `eZPlatformUIBundle:Pjax:accessDenied` doesn't exist any more.

Controller/RoleController.php
```diff
@@ -135,9 +152,10 @@ public function updateRoleAction(Request $request, $roleId)
             $hasErrors = true;
         }
 
-        // TODO: Just a temporary implementation of draft handling. To be done properly in follow-up: EZP-24701
         $formView = $form->createView();
-        $roleName = $role->identifier;
+
+        // Show empty text input when name is not set, while showing "New role" in the page title
+        $roleName = $roleDraft->identifier;
         if (preg_match('/^__new__[a-z0-9]{32}$/', $roleName) === 1) {
```
lolautruche: Use `$roleData->isNew()` instead

glye:
Removed unneeded catches, use $roleData->isNew()

Resources/translations/role.en.xlf
```diff
@@ -122,6 +122,10 @@
         <source>role.error.name_exists</source>
         <target>The role name already exists.</target>
       </trans-unit>
+      <trans-unit id="050c49e2a922c585b79243022076154c" resname="role.error.has_draft">
+        <source>role.error.has_draft</source>
```
pspanja: This not used yet or?

glye: Damn. You're right, eagle eyes :)

glye:
Removed unused translation message.

Controller/RoleController.php
```diff
@@ -116,8 +117,15 @@ public function createRoleAction(Request $request)
...
-        $roleData = (new RoleMapper())->mapToFormData($role);
+        try {
+            $roleDraft = $this->roleService->loadRoleDraft($roleId);
+        } catch (NotFoundException $e) {
+            // The draft doesn't exist, let's create one
+            $role = $this->roleService->loadRole($roleId);
+            $roleDraft = $this->roleService->createRoleDraft($role);
+        }
+
+        $roleData = (new RoleMapper())->mapToFormData($roleDraft);
```
pspanja: This should rather be injected or otherwise registered to the form, but AFAICS this is not the case elsewhere as well, so there's probably a plan behind it :)

pspanja:
Looks good, +1.

Controller/RoleController.php
lolautruche: Given what those mappers do, this would really be overkill. Same as for form types, when they are simple (they usually are), it's perfectly fine to instantiate them instead of defining services. Besides there is absolutely no dependency.
Only thing is that it's not extensible, but is it really needed here?

lolautruche: To make things clearer, these mappers are here to avoid using private methods in the controller, which is not a really good practice.

Controller/RoleController.php
```diff
@@ -116,8 +117,15 @@ public function createRoleAction(Request $request)
      */
     public function updateRoleAction(Request $request, $roleId)
```
lolautruche: Maybe we should rename this method to `editRoleAction()`, like for sections ?

lolautruche:
+1

Controller/RoleController.php
pspanja: It is fine with me like this as well, it is simple and I don't really have a ready use-case.

Controller/RoleController.php
glye: @lolautruche Ok, then we should rename the url and everything else too, to be consistent. And contenttype also uses "update", unless you have renamed it recently :)

lolautruche: > And contenttype also uses "update", unless you have renamed it recently :)

Mmm actually I missed that you were duplicating what's done in `createRoleAction()`. You probably need to move the new creation logic to that action and let `NotFound` exception bubble up.

lolautruche: Forget it, I got confused by the draft creation :blush:

PR #301, opened 2015-08-26
Make Application Config extensible

Nattfarinn:
**Related JIRA ticket**: https://jira.ez.no/browse/EZS-204

Description:
---
**Background**: There's feature called *Application Config* in eZ Platform that allows providing custom backend configurations to frontend without need of additional AJAX requests. Long story short: it dumps PHP array as JSON.

**Problem**: Current implementation is not flexible enough to allow anyone to plug in their configuration without modifying Platform UI Bundle, and ability to pass configuration is required by eZ Studio team. Currently every Provider config is dumped to it's own (hardcoded in twig template) JSON key on frontend side, and there's no good and reliable way to extend it.

**Solution**: Without breaking backward compatibility, I've modified `ApplicationConfigProviderPass` a little. Now it makes use of `ezsystems.bundle_application_config_provider` tag to identify custom services and dumps their configuration to `extensions` key on frontend side (namespaced by new `Provider` method `getCategoryName`).

Example:
---

*CustomProvider.php*:
```php
class CustomProvider implements Provider
{
    public function getCategoryName()
    {
        return 'ezexample';
    }

    public function getConfig()
    {
        return [ 'result' => 'It works!' ];
    }
}
```
*services.yml*:
```yaml
ezexample.test_provider:
    class: CustomProvider
    tags:
        -
            name: ezsystems.bundle_application_config_provider
            key: 'example-config'
```

*Output [site source]*:
```javascript
/* (...) */
var app = new Y.eZ.PlatformUIApp({
container: '.ez-platformui-app',
    viewContainer: "(...)",
    root: "(...)",
    assetRoot: "(...)",
    /* (..) */
    extensions: {    // Here's example output
        'ezexample': {
            'example-config': {
                'result': 'It works!'
            }
        }
    }
});
/* (...) */
```

### DIFF
--- a/ApplicationConfig/Aggregator.php
+++ b/ApplicationConfig/Aggregator.php
@@ -9,6 +9,8 @@ namespace EzSystems\PlatformUIBundle\ApplicationConfig;
  */
 class Aggregator implements Provider
 {
+    const CATEGORY_NAME = 'ezsystems';
+
     /** @var Provider[] ApplicationConfigProviders, indexed by namespace string*/
     private $providers = [];
 
@@ -34,10 +36,11 @@ class Aggregator implements Provider
     }
 
     /**
-     * @inheritdoc
+     * {@inheritdoc}
      */
     public function getCategoryName()
     {
+        return self::CATEGORY_NAME;
     }
 
     /**
@@ -47,14 +50,20 @@ class Aggregator implements Provider
      */
     public function getConfig()
     {
-        $config = [];
+        $category = $this->getCategoryName();
+
+        $config = [$category => []];
+
         foreach ($this->providers as $key => $provider) {
             $config[$key] = $provider->getConfig();
         }
 
-        $config['bundles'] = [];
-        foreach ($this->bundleProviders as $bundleKey => $provider) {
-            $config['bundles'][$bundleKey] = $provider->getConfig();
+        foreach ($this->bundleProviders as $key => $provider) {
+            $providerCategory = $provider->getCategoryName();
+            if (!isset($config[$category][$providerCategory])) {
+                $config[$category][$providerCategory] = [];
+            }
+            $config[$category][$providerCategory][$key] = $provider->getConfig();
         }
 
         return $config;
--- a/ApplicationConfig/Aggregator.php
+++ b/ApplicationConfig/Aggregator.php
@@ -12,6 +12,9 @@ class Aggregator implements Provider
     /** @var Provider[] ApplicationConfigProviders, indexed by namespace string*/
     private $providers = [];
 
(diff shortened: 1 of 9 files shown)


lserwatka:
ping @andrerom @dpobel @bdunogier @michalpipa

dpobel:
I'm -1, this make the config more complicated while we have different plan.

> Problem: Current implementation is not flexible enough to allow anyone to plug in their configuration without modifying Platform UI Bundle, and ability to pass configuration is required by eZ Studio team. Currently every Provider config is dumped to it's own (hardcoded in twig template) JSON key on frontend side, and there's no good and reliable way to extend it.

Actually, it's because the feature is not completely finished. [EZP-24129](https://jira.ez.no/browse/EZP-24129) where the server side configuration was added has a followup: [EZP-24654](https://jira.ez.no/browse/EZP-24654). Basically in EZP-24654, we'll change the way the Application receives the configuration, it will receive a *big* `config` object making it possible to inject anything needed by extension bundle like eZStudio. As a bonus, this will allow to move most of the logic from the template to JavaScript and then to test that logic correctly.

lserwatka:
@dpobel Do you have any ETA when this can be finished? I would like to get rid of some things from StudioUI and make it right with the configuration.

dpobel:
@lserwatka well the issue has been prioritized but it has not been started yet, it's currently assigned to @StephaneDiot but he is off those days (he's back at the end of the week) and has several already started issues. So maybe, we can start working on that in 2 weeks.

lserwatka:
@dpobel great thanks, @Nattfarinn can help @StephaneDiotthe on the backend side if needed. Until that we will work on other remaining things in Studio.

dpobel:
there's nothing more to do on the server side, the remaining work is only on the JS side.

lserwatka:
@dpobel hmm, in the current state backend code does not aggregate settings from the external bundles, that is why this pull request was created. With this implementation it is very easy to achieve what was described in the https://jira.ez.no/browse/EZP-24654, except JS part of course. Anyway, we will wait until you are done.

dpobel:
As far as I can see in https://github.com/ezsystems/PlatformUIBundle/blob/master/DependencyInjection/Compiler/ApplicationConfigProviderPass.php#L21, if you tag your own application config provider service with `ezsystems.platformui.application_config_provider` it will be added to the provider list and your configuration will be available in the `parameters` twig variable.

Nattfarinn:
You're obviously right, but that's the point of this change: way you proposed is not flexible enough solution. There's no way of using such services without hardcoding every single one to twig template.

But it's fair enough to wait till it will be done in https://jira.ez.no/browse/EZP-24654

dpobel:
Replaced by https://github.com/ezsystems/PlatformUIBundle/pull/349

PR #302, opened 2015-08-26
EZP-24727: Admin forms improvments

andrerom:
https://jira.ez.no/browse/EZP-24727

Is:
- padding to have space around content in admin part
- use pure-forms to align form elements
- some minor cleanups/sync up of code to have consistent view on context info (modified info)

Is no:
- This is not the end for content type editing, there is a much bigger story in store to display it differently, which is why some field types is looking a bit weird still here, as they will be displayed differently anyway.

Depends on https://github.com/ezsystems/repository-forms/pull/42

Screen shoots:
platform_admin_forms1
Bottom part of the form:
platform_admin_forms2

Also ready for mobile ;):
platform_admin_forms3_mobile

Ready for review: @glye @dpobel @yannickroger

### DIFF
--- a/Resources/views/ContentType/update_content_type.html.twig
+++ b/Resources/views/ContentType/update_content_type.html.twig
@@ -1,6 +1,8 @@
 {% extends "eZPlatformUIBundle::pjax_admin.html.twig" %}
 {# @var contentTypeDraft \eZ\Publish\API\Repository\Values\ContentType\ContentTypeDraft #}
 
+{% trans_default_domain "content_type" %}
+
 {% set edit_title = "content_type.edit_title"|trans({"%contentTypeName%": contentTypeName}, "ezrepoforms_content_type") %}
 
 {% block title %}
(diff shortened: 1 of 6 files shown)


glye:
Much +1! So improvement.

Resources/views/ContentType/update_content_type.html.twig
```diff
@@ -25,72 +27,101 @@
 
 {% block content %}
     <section class="ez-serverside-content">
-        Last modified: {{ contentTypeDraft.modificationDate|localizeddate('medium', 'medium', app.request.locale) }}
 
-        {{ form_start(form, {"action": action_url}) }}
+        <div class="context-information">
```
dpobel: we should always prefix the classes with the `ez-` prefix. That said, `context-information` is very generic and not used anywhere in the CSS so it can be omitted (the same goes for `modified` below).

dpobel:

Resources/views/ContentType/update_content_type.html.twig
andrerom: ok

andrerom: Found `ez-technical-infos`, won't apply here now but adding it as it is the closest fit (used in content edit view atm).

PR #303, opened 2015-08-26
Small readme changes

mhyndle:
It seems that `gulp` needs to be installed because it's needed when installing node packages inside alloy editor.
Also added empty line before bash command which sets the command block to be multiline - easier to read.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ have several additional dependencies:
 * Install grunt, yuidoc, bower and grover globally (usually you need to be root):
 
   ```
-  # npm install -g grunt-cli yuidocjs grover bower
+  # npm install -g grunt-cli yuidocjs grover bower gulp
   ```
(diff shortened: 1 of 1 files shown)


sunpietro:
Maybe you should switch to gulp completely, what do you think?

mhyndle:
ping @dpobel

dpobel:
@sunpietro we have more urgent tasks I guess

sunpietro:
@dpobel definitely, but having 2 automation tools in one project looks weird.

dpobel:
@sunpietro for the record, we need gulp only because AlloyEditor uses it and we need to run an AlloyEditor task to generate our own skin. But I don't see the problem in installing gulp in addition and actually given the number of alternatives to grunt and gulp (brocoli, gear.js, lineman, or even the good old make...) I would not be surprise to have to install another one in a near future because libX or componentY uses it...

PR #304, opened 2015-08-26
[Login] Small tweaks to login screen

andrerom:
Just small tweaks here.

Was:
screen shot 2015-08-25 at 10 46 12

Is:
screen shot 2015-08-26 at 14 12 36

Review ping @dpobel @mhyndle @yannickroger

### DIFF
--- a/Resources/public/css/theme/views/loginform.css
+++ b/Resources/public/css/theme/views/loginform.css
@@ -5,14 +5,14 @@
 
 .ez-view-loginformview {
     background: #f5f4f2;
-    border-top: 3px solid #000;
+    border-top: 3px solid rgba(200,200,200,0.33);
 }
 
 .ez-view-loginformview .ez-loginform {
     background: #fff;
-    border: 1px solid #ccc;
-    border-radius: 3px;
-    box-shadow: 0px 1px 2px #CCC;
+    border: 1px solid #ddd;
+    border-radius: 5px;
+    box-shadow: 3px 3px 10px 1px rgba(175,175,175,.6);
 }
 
 .ez-view-loginformview .ez-loginform-link {
@@ -24,8 +24,9 @@
 }
 
 .ez-view-loginformview .ez-loginform-title {
-    font-size: 0.9rem;
+    font-size: 0.95rem;
     color: #444;
+    font-family: sans-serif;
 }
 
 .ez-view-loginformview .ez-loginform-button {


mhyndle:
+1

yannickroger:
+1

Resources/public/css/theme/views/loginform.css
```diff
@@ -24,8 +24,9 @@
 }
 
 .ez-view-loginformview .ez-loginform-title {
-    font-size: 0.9rem;
+    font-size: 0.95rem;
     color: #444;
+    font-family: sans-serif;
```
dpobel: not sure about this one, what's wrong with the font defined for the whole app ?

andrerom: It just looked a bit more crips, but could be my screen. I guess we should get @inakijv to have a look at this eventually. Maybe we should use the Maven Pro font here and another places for instance.

inakijv: Hi! For me it is more a Marketing decision. Nothing beats the readability of Helvetica and its family, including Neue (just a few more maybe); That’s why I understand Damien’s question. But nothing like to visualize a transformation like a change of the primary font and due to we are precisely working for this… I would to try to be as consistent as possible throughout all spaces. Finally, Maven Pro compared to Helvetica Neue it is slightly thinner and curved, as well as increases the amount of space between letters. So maybe, in some cases, we should pay attention also to the font-weight or letter-spacing (specially for buttons, for instance) with Maven Pro.

andrerom: Ok, changed back in b22daf33facb923f1f9792aec29d48f851f3b669, thanks for input/review :)

sunpietro: One thing I would like to mention is that Helvetica is available on Mac only.

andrerom: are you sure? It's one of the web safe fonts afaik, Helvetica Neue however is mac only by default indeed.

sunpietro: it's a good practice to put: `font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;` because, Helvetica is not available on Windows: http://snook.ca/archives/html_and_css/windows-subs-helvetica-arial

dpobel: and this is why it's the font-family style since almost day 1 https://github.com/ezsystems/PlatformUIBundle/commit/ce186bc96783b69b453c65fa863c78e8eb731350#diff-82166add900f62398a221f90e2ddd9c8R46

sunpietro: @dpobel +1

PR #305, opened 2015-08-28
EZP-24731: As an editor, I want to be able to add embed elements in the RichText editor

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24731

# Description

This PR adds an *embed* button in the add content toolbar. Of course that's a first step as there's no way yet to choose the embed content or location nor any of the embed properties. Thanks to the CKEditor widgets plugin, it's already possible to drag the embed around. But because of the use of a custom HTML tag (`ezembed`) putting several embed elements or some text around one will completely mess the editor content (as shown at the end of the screencast) since the browser tries to guess what to do with that. The `ezembed` tag is also empty that why the embed is invisble when viewing the content.

Screencast: https://youtu.be/KovWZ9nbgNg

## Tasks

* [x] Add the widget plugin and its dependencies for CKEditor
* [x] Add an embed CKEditor plugin so that AlloyEditor/CKEditor recognizes `ezembed` element as widgets
* [x] Add the *embed* button to the append toolbar

# Tests

manual + unit tests


dpobel:
ping @StephaneDiot @yannickroger @mhyndle

mhyndle:
+1

andrerom:
Looks good, and yes makes it clear why we should change the format.

PR #308, opened 2015-09-01
Mink stale reference fix

miguelcleverti:
This PR fixes the problem when the mink node element changes in the browser, and the reference in memory becomes stale - causing an exception.


miguelcleverti:
@andrerom @joaoinacio

andrerom:
+1

andrerom:
@miguelcleverti fixing the cs? :)

miguelcleverti:
All fixed and ready :smiley_cat:

PR #309, opened 2015-09-01
Refactor loggedAsAdminPlaformUi function name

miguelcleverti:
Small refactor in function name that reflects it's functionality better

### DIFF
--- a/Features/Context/SubContext/Authentication.php
+++ b/Features/Context/SubContext/Authentication.php
@@ -53,7 +53,7 @@ trait Authentication
      * @Given I am logged in as admin on PlatformUI
      * @Given I am logged in as an Editor in PlatformUI
      */
-    public function loggedAsAdminPlatformUi()
+    public function loggedAsUserPlatformUi()
     {
         $this->goToPlatformUiAndLogIn($this->user, $this->password);
     }
--- a/Features/Context/SubContext/Fields.php
+++ b/Features/Context/SubContext/Fields.php
@@ -23,7 +23,7 @@ trait Fields
         $fieldManager->setFieldContentState(FieldType::CONTENT_TYPE_PUBLISHED);
         $type = $fieldManager->getThisContentTypeName('eng-GB');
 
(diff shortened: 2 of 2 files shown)


miguelcleverti:
@andrerom @joaoinacio

Features/Context/SubContext/Authentication.php
```diff
@@ -53,7 +53,7 @@ public function goToPlatformUiAndLogIn($username, $password)
      * @Given I am logged in as admin on PlatformUI
      * @Given I am logged in as an Editor in PlatformUI
```
andrerom: well, the two sentences does not really fit with with the name of the function now, exactly what did you have in mind here @joaoinacio? Always difficult to get right, but be specific, add context to proposals for changes ;)

joaoinacio: @andrerom: Indeed.

The `$user` and `$password` are for the default user, which is admin. ref: https://github.com/ezsystems/PlatformUIBundle/blob/loggedAsAdminFIx/Features/Context/PlatformUI.php#L40
(this can be modified in configuration).

Naming the function `loggedAsUser...` seems more inline with its implementation (IMHO), but what if the default user is not an Admin? (perhaps document it?)

Regarding the editor, I'm not certain why it's using this function, but seems related to current login issues with non-admin user (ping @miguelcleverti)

In any case, once it's sorted, the sentence could probably use either a pre-existing user or, create a new user with 'Editor' credentials. see https://github.com/ezsystems/BehatBundle/blob/master/Context/EzContext.php#L248

miguelcleverti: Currently it is only possible to login in platformUI with admin user, this is a known issue but I can't seem to find a related Jira issue.
The editor sentence for now is just a place holder using the admin credentials in order for the tests to work

andrerom:
ok, good enough for now then I guess, +1

@joaoinacio ?

joaoinacio:
+1 for now, should be updated anyway after editor login is resolved.

PR #310, opened 2015-09-01
EZP-23768: change language when creating content

mhyndle:
Jira: https://jira.ez.no/browse/EZP-23768

## Description
The goal of this one is to allow user to change language of content when creating one. Also currently it allows to switch between existing translations when editing the content.

This PR also changes a little the languageSelectionBox feature allowing to choose what languages should be included in the list: already existing translations or only new translations.

## Task
- [x] CSS
- [x] Implementation
- [x] Tests

## Screencasts
- Change language when creating content: https://youtu.be/p4VquWGLy9Q
- Change language when editing content: https://youtu.be/BS4o01TtWPs

## Screenshot
d03_edit new blog post

### DIFF
--- a/Resources/public/js/views/ez-contenteditview.js
+++ b/Resources/public/js/views/ez-contenteditview.js
@@ -35,6 +35,9 @@ YUI.add('ez-contenteditview', function (Y) {
             },
             '.ez-main-content': {
                 'keyup': '_handleKeyboard'
+            },
+            '.ez-change-content-language-link': {
+                'tap': '_changeLanguage',
             }
         },
 
@@ -51,6 +54,11 @@ YUI.add('ez-contenteditview', function (Y) {
             this.after('activeChange', function (e) {
                 if ( e.newVal ) {
                     this._setFocus();
+                }
+            });
+            this.on('languageCodeChange', function (e) {
+                if ( this.get('active') ) {
+                    this._setLanguageIndicator(e.newVal);
                 }
             });
 
@@ -101,7 +109,8 @@ YUI.add('ez-contenteditview', function (Y) {
                 version: this.get('version').toJSON(),
                 mainLocation: this.get('mainLocation').toJSON(),
                 contentType: this.get('contentType').toJSON(),
-                owner: this.get('owner').toJSON()
+                owner: this.get('owner').toJSON(),
+                languageCode: this.get('languageCode')
             }));
             if ( this._isTouch() ) {
                 container.addClass('is-using-touch-device');
@@ -210,6 +219,38 @@ YUI.add('ez-contenteditview', function (Y) {
          */
         _isTouch: function () {
             return Y.UA.touchEnabled;
+        },
+
+        /**
+         * Tap event handler on change language button. It fires `changeLanguage` event.
+         *
+         * @method _changeLanguage
+         * @private
+         * @param {EventFacade} e
+         */
+        _changeLanguage: function (e) {
+            e.preventDefault();
+
+            /**
+             * Fired when the change language link was tapped
+             *
+             * @event changeLanguage
+             */
+            this.fire('changeLanguage');
+        },
+
+        /**
+         * Sets language indicator
+         *
+         * @method setLanguageIndicator
+         * @private
+         * @param {String} languageCode
+         */
+        _setLanguageIndicator: function (languageCode) {
+            var c = this.get('container'),
+                languageContainer = c.one('.ez-content-current-language');
+
+            languageContainer.setHTML(languageCode);
         }
     }, {
         ATTRS: {
@@ -313,7 +354,16 @@ YUI.add('ez-contenteditview', function (Y) {
                         version: this.get('version')
                     });
                 }
-            }
+            },
+
+            /**
+             * The language code in which the content is edited.
+             *
+             * @attribute languageCode
+             * @type {String}
+             * @required
+             */
+            languageCode: {},
         }
     });
 });
--- a/Resources/public/js/views/ez-languageselectionboxview.js
+++ b/Resources/public/js/views/ez-languageselectionboxview.js
@@ -14,8 +14,8 @@ YUI.add('ez-languageselectionboxview', function (Y) {
     var LANGUAGE_SELECTED = 'languageSelected',
         CANCEL_LANGUAGE_SELECTION = 'cancelLanguageSelection',
         BASE_TRANSLATION_AVAILABLE_CLASS = 'is-base-translation-allowed',
-        BASE_TRANSLATIONS_LIST_VISIBLE_CLASS = 'is-base-translations-list-visible',
-        TRANSLATION_HIGHLIGHT_CLASS = 'is-translation-selected';
+        BASE_LANGUAGES_LIST_VISIBLE_CLASS = 'is-base-languages-list-visible',
+        TRANSLATION_HIGHLIGHT_CLASS = 'is-language-selected';
 
     /**
      * The language selection box view.
@@ -33,11 +33,11 @@ YUI.add('ez-languageselectionboxview', function (Y) {
             '.ez-languageselectionbox-confirm': {
                 'tap': '_confirmLanguageSelection',
             },
-            '.ez-new-translation': {
-                'tap': '_selectNewTranslation'
+            '.ez-language-element': {
+                'tap': '_selectLanguage'
             },
-            '.ez-base-translation': {
-                'tap': '_selectBaseTranslation'
+            '.ez-base-language': {
+                'tap': '_selectBaseLanguage'
             },
             '.ez-base-translation-checkbox': {
                 'change': '_toggleBaseTranslation'
@@ -57,12 +57,15 @@ YUI.add('ez-languageselectionboxview', function (Y) {
                     this.render();
                 }
             });
+            this.on('translationModeChange', function (e) {
+                this.render();
+            });
             this.on('canBaseTranslationChange', function (e) {
                 this._uiSetBaseTranslationVisibility(e.newVal);
             });
             this.on('selectedLanguageCodeChange', function (e) {
                 if ( this.get('active') ) {
-                    this._uiHighlightTranslation('new', e.newVal);
+                    this._uiHighlightLanguage(e.newVal, false);
                 }
             });
             this.after('selectedLanguageCodeChange', function (e) {
@@ -70,7 +73,7 @@ YUI.add('ez-languageselectionboxview', function (Y) {
             });
             this.on('selectedBaseLanguageCodeChange', function (e) {
                 if ( this.get('active') ) {
-                    this._uiHighlightTranslation('base', e.newVal);
+                    this._uiHighlightLanguage(e.newVal, true);
                 }
             });
             this.on('baseTranslationChange', function (e) {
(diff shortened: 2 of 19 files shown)


mhyndle:
ready for review, ping @dpobel @yannickroger @StephaneDiot

Resources/public/js/views/services/ez-contentcreateviewservice.js
```diff
@@ -22,6 +22,20 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
...
     Y.eZ.ContentCreateViewService = Y.Base.create('contentCreateViewService', Y.eZ.ContentEditViewService, [], {
+        initializer: function () {
+            this.on('*:changeLanguage', this._changeLanguage);
+            this.after('*:languageCodeChange', function (e) {
+                var app = this.get('app'),
+                    activeView = app.get('activeView'),
+                    container = activeView.get('container'),
+                    languageContainer = container.one('.ez-content-current-language');
+
+                if (languageContainer) {
```
dpobel: you modifying the DOM in a view service that's bad because it's the responsibility of a view. Instead, you should add a method for that in the view and call it there.

Resources/public/js/views/services/ez-contentcreateviewservice.js
```diff
@@ -22,6 +22,20 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
      * @extends eZ.ContentEditViewService
      */
     Y.eZ.ContentCreateViewService = Y.Base.create('contentCreateViewService', Y.eZ.ContentEditViewService, [], {
+        initializer: function () {
+            this.on('*:changeLanguage', this._changeLanguage);
+            this.after('*:languageCodeChange', function (e) {
+                var app = this.get('app'),
+                    activeView = app.get('activeView'),
```
dpobel: that's not super robust if for example we move the language list to a sub view of the ContentEditView. So instead of using the app activeView, you should use the target (`e.target` ie the component which fired the event) of the `changeLanguage` event.

Resources/public/js/views/services/ez-contentcreateviewservice.js
```diff
@@ -93,6 +107,48 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
...
+         * @method _changeContentLanguage
+         * @private
+         * @param {EventFacade} e
+         * @param {String} e.selectedLanguageCode language code to which created ontent will be switched
+         */
+        _changeContentLanguage: function (e) {
+            var version = this.get('version'),
+                service = this;
+
+            version.sync('delete', {api: this.get('capi')}, function () {
```
dpobel: even if `sync` is pubic, you are not really supposed to call it directly. You should use [`Model.destroy`](http://yuilibrary.com/yui/docs/api/classes/Model.html#method_destroy) with the `remove` option instead.

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -407,6 +416,19 @@ YUI.add('ez-languageselectionboxview', function (Y) {
...
+            /**
+             * Defines what the list o translations will contain.
+             * If it is set to TRUE then list of languages will contain not existing (new) translations.
+             * Otherwise list of language will contain already existing translations.
+             *
+             * @attribute baseTranslation
+             * @type {Boolean}
+             * @default false
+             */
+            newTranslationsOnList: {
```
dpobel: the name of this attribute is bad, it describes the technical behaviour, not the actual feature.

Actually, the language selection box should have 2 modes:

1. a "normal" mode (default) where we let the user chooses amongst the languages in `existingTranslations` (this attribute should be renamed, see below).
2. a translation mode where we provide the non existing languages

To make things a bit clearer, some attributes should be renamed:
* `existingTranslations` => `referenceLanguageList`
* `availableTranslations` (which is not really defined and documented btw) => `systemLanguageList`
* `newTranslationsOnList` => `translationMode` (false by default).

Some comments should also be updated to use *language* (or *language code*) instead of *translation* when the involved code is applicable in *normal* mode and to remove any reference to *content* since the language selection box will be used for others things than content.

Last detail, in the template and the CSS, you can not use `ez-languageselectionbox-newtranslations` and `ez-new-translation` classes since we are not always displaying potential new translations there.

mhyndle:
@dpobel : I made changes according to your feedback.
There is one thing I don't how to test: https://github.com/ezsystems/PlatformUIBundle/pull/310/files#diff-5101bec0a213658cf7cfd0892ae095b7R334
Is there any way to pass the mock of the view in handler function call?

andrerom:
+1

Looks good, is the screen cast outdated btw? Does it have similar look and feel now as the screenshot?

mhyndle:
@andrerom : the structure and functionality of language selection box is the same but design is differrent from what is on screenshot. Box styling refers to what was already implemented in other components (transitions, colors, buttons).

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -344,6 +361,14 @@ YUI.add('ez-languageselectionboxview', function (Y) {
             },
 
             /**
+             * List of active system languages. List contains language objects.
```
dpobel: You should mentioned somewhere that the language objects are indexed by language code.

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -354,13 +379,13 @@ YUI.add('ez-languageselectionboxview', function (Y) {
             },
 
             /**
-             * Already existing translations of content
+             * List of languages used as reference
```
dpobel: here you should mention that in translation mode, this is the list of existing translations while not in translation mode, this is the list languages the user can pick in.

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -407,6 +432,21 @@ YUI.add('ez-languageselectionboxview', function (Y) {
              */
             selectedBaseLanguageCode: {
                 value: null
+            },
+
+            /**
+             * Defines what the list o languages will contain.
```
dpobel: nope, this is *too technical*. Should something like:

> Translation mode flag. In translation mode, the `referenceLanguageList` attribute holds
> the list of existing translations, so the language selection box will allow to select
> a language the system language list but not yet the reference list. By default, the translation
> mode is disabled, the user will be able to select a language in the `referenceLanguageList` attribute.

Resources/public/js/views/services/ez-contentcreateviewservice.js
```diff
@@ -22,6 +22,10 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
      * @extends eZ.ContentEditViewService
      */
     Y.eZ.ContentCreateViewService = Y.Base.create('contentCreateViewService', Y.eZ.ContentEditViewService, [], {
+        initializer: function () {
+            this.on('*:changeLanguage', this._changeLanguage);
```
dpobel: bad method naming, the method should be named `selectLanguage` so that you directly know what happen when the `changeLanguage` event is fired.

Resources/public/js/views/services/ez-contentcreateviewservice.js
```diff
@@ -93,6 +97,50 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
...
+
+            e.preventDefault();
+            this.fire('languageSelect', {
+                config: {
+                    title: "Change language to:",
+                    languageSelectedHandler: Y.bind(this._changeContentLanguage, this, e.target),
+                    cancelLanguageSelectionHandler: null,
+                    canBaseTranslation: false,
+                    translationMode: true,
+                    referenceLanguageList: [that.get('languageCode')]
```
dpobel: useless *that* :)

Resources/public/js/views/services/ez-contentcreateviewservice.js
```diff
@@ -93,6 +97,50 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
...
+         * @param {EventFacade} e
+         */
+        _changeLanguage: function (e) {
+            var that = this;
+
+            e.preventDefault();
+            this.fire('languageSelect', {
+                config: {
+                    title: "Change language to:",
+                    languageSelectedHandler: Y.bind(this._changeContentLanguage, this, e.target),
```
dpobel: same remark here, the method name is bad because I have to look at the code to have rough idea of it does

Resources/public/js/views/ez-contenteditview.js
```diff
@@ -210,6 +214,37 @@ YUI.add('ez-contenteditview', function (Y) {
...
+            this.fire('changeLanguage');
+        },
+
+        /**
+         * Sets language indicator
+         *
+         * @method setLanguageIndicator
+         * @param {String} languageCode
+         */
+        setLanguageIndicator: function (languageCode) {
```
dpobel: instead of this method, I would rather add a `languageCode` attribute to the content edit view and then update the indicator with the `languageCodeChange` event. That way, we clearly separate the data handling from the UI behaviour.

mhyndle: Oh right, that will probably solve my problem with writing the test for it ;)

Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -28,6 +28,7 @@ YUI.add('ez-contenteditviewservice', function (Y) {
                 this._setLanguageCode();
                 this._setBaseLanguageCode();
             });
+            this.on('*:changeLanguage', this._changeLanguage);
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/310/files#r39494854

Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -331,6 +333,50 @@ YUI.add('ez-contenteditviewservice', function (Y) {
...
+         * @method _changeContentLanguage
+         * @private
+         * @param {EventFacade} e
+         * @param {String} e.selectedLanguageCode language code to which edited ontent will be switched
+         */
+        _changeContentLanguage: function (e) {
+            var app = this.get('app'),
+                service = this;
+
+            app.navigate(
```
dpobel: nitpick: you can directly write:
```js
app.navigateTo('editContent', {
    id: service.get('content').get('id'),
    languageCode: e.selectedLanguageCode,
});
```

Resources/public/js/views/ez-contenteditview.js
dpobel: It will simplify a bit the test since `set` is a Y.View (even Y.Base) method. While with a custom method like this one, you need to dynamically create a custom View class with the method and then have a real instance of it, then you have to make the view service a bubble target of this view and then fire the event from this view so that `e.target` is a *real* view with the expected method.

mhyndle:
changes made according to feedback

Resources/public/js/views/ez-contenteditview.js
```diff
@@ -313,7 +352,16 @@ YUI.add('ez-contenteditview', function (Y) {
...
+            },
+
+            /**
+             * The language code in which the content is edited.
+             *
+             * @attribute languageCode
+             * @type {String}
+             * @required
+             */
+            languageCode: {},
```
dpobel: for consistency this attribute should also be set on the first content edit display before the editor uses the change link (you have to change `ContentEditViewService._getViewParameters` to also return its `languageCode` attribute + add/change the corresponding test). AFAICT, that also means you'll have to change the `languageCodeChange` event handler to check for the `active` attribute before trying to update the DOM.

dpobel:
other than the last inline comment, that's very good :)

Resources/public/js/views/ez-contenteditview.js
mhyndle: `_getViewParameters` is changed already: https://github.com/ezsystems/PlatformUIBundle/pull/310/files#diff-322d0e70fded2e094b6aea22bb14a63cR281
But I forget to test it :) So now there is test for it and `active` attribute check on `languageCodeChange` event handling.

mhyndle:
ready for (final? ;-) ) review

dpobel:
+1

Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -331,6 +333,48 @@ YUI.add('ez-contenteditviewservice', function (Y) {
...
+                    cancelLanguageSelectionHandler: null,
+                    canBaseTranslation: false,
+                    translationMode: false,
+                    referenceLanguageList: that.get('content').get('currentVersion').getTranslationsList()
+                },
+            });
+        },
+
+        /**
+         * Changes language od edited content
```
yannickroger: typo ?

yannickroger:
+1

PR #313, opened 2015-09-03
EZP-24725: As an editor, I want to have a toolbar to configure the heading element

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24725

# Description

Provides the *Heading* toolbar to be able to change the heading level, the alignment and also to remove the heading element.

toolbar_heading

Screencast: https://youtu.be/aEWI8xwMQHw

## Tasks

* [x] Keep AlloyEditor's UI visible after using the add content button
* [x] Configure a toolbar for heading
* [x] Add the align buttons
* [x] Add the remove button

# Tests

manual + unit test

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -24,6 +24,7 @@ module.exports = function(grunt) {
             "./Resources/public/js/services/*.js",
             "./Resources/public/js/helpers/*.js",
             "./Resources/public/js/alloyeditor/toolbars/*.js",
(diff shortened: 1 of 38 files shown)


andrerom:
Looks very nice from the screen cast.
And as far as I can see code looks good as well: +1

Q: "fake" ? :)

mhyndle:
+1

PR #318, opened 2015-09-08
Travis configurations for PlatformUI Behat tests

miguelcleverti:
Changed travis.yml to run PlatformUI behat tests.
Added a `common` tag to the standard feature, for now we should try to run the standard feature and add the rest later


miguelcleverti:
@andrerom @bdunogier @joaoinacio @pcardiga

andrerom:
+1, no need for behat.yml file or has that been added elsewhere?

PR #319, opened 2015-09-08
EZP-23700: As an editor, I want to see an accurate url and title when I submit a form in the admin part of the PlatformUI

dpobel:
JIRA: [EZP-24781](https://jira.ez.no/browse/EZP-24781) (sub-task of [EZP-23700](https://jira.ez.no/browse/EZP-23700))

# Description

Currently, most sever side form processing do an HTTP redirection to make the application display the result after the form submission. (for instance [redirecting to the section list after deleting a section](https://github.com/ezsystems/PlatformUIBundle/blob/master/Controller/SectionController.php#L137)). This kind of works except that the PlatformUI Application has no way detect such redirection (it's done transparently by the browser/XmlHTTPRequest object), as a result the application gets in an inconsistent state, the URI does not match what is actually displayed and this breaks some buttons/links.

To avoid that, the server side processing of the forms should not generate a 30X response but a 205 (*Reset Content*) response with a custom `PJAX-Location` header so that the redirection is done by the application. This patch adds this behaviour to the PlatformUI application, the server side change is to be done in [EZP-24780](https://jira.ez.no/browse/EZP-24780).

# Tests

manual test + unit

### DIFF
--- a/Resources/public/js/views/services/ez-serversideviewservice.js
+++ b/Resources/public/js/views/services/ez-serversideviewservice.js
@@ -11,6 +11,9 @@ YUI.add('ez-serversideviewservice', function (Y) {
      */
     Y.namespace('eZ');
 
+    var PJAX_DONE_REDIRECT = 205,
+        PJAX_LOCATION_HEADER = 'PJAX-Location';
+
     /**
      * The Server Side View Service class. It is meant to be used to load the
      * content of a server side view.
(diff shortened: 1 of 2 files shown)


dpobel:
ready for review @mhyndle @yannickroger @andrerom @StephaneDiot 
@lolautruche you can start on https://jira.ez.no/browse/EZP-24780

lolautruche:
+1

lolautruche:
Forget @ezrobot, I messed up CS again...

mhyndle:
+1

andrerom:
+1

lolautruche:
What about regular links? e.g. when clicking from admin panel sub-menu, or from admin dashboard.
We might also have redirects here, and it currently doesn't work with this change;

PR #321, opened 2015-09-08
Fix EZP-24780: Implement the custom PJAX redirection server side

lolautruche:
> https://jira.ez.no/browse/EZP-24780

Implemented a new specific `PjaxRedirectResponse`, using `HTTP 205`.
Target location lies in `PJAX-Location` header.


lolautruche:
Review ping @dpobel @andrerom @bdunogier @yannickroger @glye @pspanja

lolautruche:
Only works with redirection after submitting a form as #319 seems to be incomplete (regular links management missing).

dpobel:
it's because the initial intend was to do that after a form processing (in EZP-23700 *when submitting a form, when everything is OK, the Pjax response should have the status code 205 with a custom header PJAX-Location where this header would point to the PJAX Uri to load to display the result.*). The idea behind the 205 response is to replace the redirect response meaning "I did something, you can see the result here".

Actually, AFAICS, in other use cases from a PlatformUI point of view, it's safe to have a redirection except in case like https://github.com/ezsystems/PlatformUIBundle/blob/master/Controller/RoleController.php#L73 but there, a  404 response should be generated (the Role was not found after all).

Controller/RoleController.php
```diff
@@ -73,7 +73,7 @@ public function viewRoleAction($roleId)
         } catch (NotFoundException $e) {
             $this->notifyError('role.error.role_not_found', [], 'role');
 
-            return $this->redirect($this->generateUrl('admin_roleList'));
+            return $this->pjaxRedirect($this->generateUrl('admin_roleList'));
```
dpobel: as said in the PR comment, here we should generate a 404.

Http/PjaxRedirectResponse.php
```diff
@@ -0,0 +1,22 @@
...
+ * @version //autogentag//
+ */
+namespace EzSystems\PlatformUIBundle\Http;
+
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Specific redirect response, dedicated to PlatformUI JS app.
+ */
+class PjaxRedirectResponse extends Response
```
dpobel: to clarify a bit the usage of this kind of response, I would rather call it something like PjaxFormProcessingDone (and then also change the name of the methods in the abstract controller class) as the redirection is purely a behaviour of JavaScript application (different layer)

Controller/RoleController.php
lolautruche: OK, but then, as discussed yesterday on Skype, 404 messages should be reported by the JS app as well (see the error notified on line 74).

lolautruche:
OK, remarks taken into account.
However it doesn't work as change done in #309 doesn't *like* URIs with a leading `/` (probably the same with URLs starting with `http`.

Reported error is always `Failed to load '//contenttype/update/2/eng-GB'` (note the double `/`).

andrerom:
+1 *(when it works)*

dpobel:
I'm looking at the slash issue

ApplicationConfig/Providers/RootInfo.php
```diff
@@ -24,12 +24,28 @@ public function __construct(RequestStack $requestStack, AssetsHelper $assetsHelp
...
+     *
+     * @return string
+     */
+    protected function getApiRoot()
+    {
+        $request = $this->requestStack->getMasterRequest();
+        $baseUrl = $request->getBaseUrl();
+        $pathinfo = $request->getPathInfo();
+        $semanticPathinfo = $request->attributes->get('semanticPathinfo');
+        return $baseUrl . substr($pathinfo, 0, strpos($pathinfo, $semanticPathinfo)) . '/';
```
lolautruche: `$baseUrl` can be inlined

ApplicationConfig/Providers/RootInfo.php
```diff
@@ -24,12 +24,28 @@ public function __construct(RequestStack $requestStack, AssetsHelper $assetsHelp
...
+     * prefix to use for all api/AJAX calls
+     *
+     * @return string
+     */
+    protected function getApiRoot()
+    {
+        $request = $this->requestStack->getMasterRequest();
+        $baseUrl = $request->getBaseUrl();
+        $pathinfo = $request->getPathInfo();
+        $semanticPathinfo = $request->attributes->get('semanticPathinfo');
```
lolautruche: Please add `$pathinfo` as 2nd argument to set it as fallback

lolautruche:
OK, still missing a `+1` to merge

Controller/Controller.php
```diff
@@ -26,4 +27,32 @@ public function performAccessChecks()
     {
         $this->denyAccessUnlessGranted('IS_AUTHENTICATED_REMEMBERED');
     }
+
+    /**
+     * Returns a PJAX sepcific redirect response (HTTP 205 with PJAX-Location header).
```
yannickroger: typo

yannickroger:
+1

PR #323, opened 2015-09-09
EZP-24788: Add X-PJAX header to PJAX requests

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24788

# Description

This patch changes the server side view service so that it adds the custom `X-PJAX` header to allow the server side code to discriminate PJAX requests from the others.

# Tests

unit tests

PR #325, opened 2015-09-09
Fix EZP-24790: Format server-side PJAX errors using PlatformUI base template

lolautruche:
> https://jira.ez.no/browse/EZP-24790

Implemented new `ExceptionListener` / `ExceptionController` based on the Symfony ones.
They will only trigger if exceptions are coming from a PJAX request.

Exception style has been kept to enable preview from one's browser debug console (with stacktrace).
Please note that full exception description, including stack trace, is only displayed with debug on (usually in dev environment).

Depends on #323 and https://github.com/ezsystems/ezpublish-kernel/pull/1405


andrerom:
+1

yannickroger:
If I understand correctly, error messages are the ones returned by the kernel and they are not translated anymore.

lolautruche:
Yes indeed, but that should not be a problem. A not found is a not found

Resources/views/Section/not_found.html.twig
```diff
@@ -1,14 +0,0 @@
-{% extends "eZPlatformUIBundle::pjax.html.twig" %}
-
-{% trans_default_domain "section" %}
-
-{% block title %}{{ 'section.not_found.title'|trans }}{% endblock %}
```
yannickroger: these translation keys (title and message) could be removed from: `Resources/translations/section.en.xlf`

yannickroger:
Code looks good. I am just a bit concerned by the loss of internationalization/customization. ping @dpobel @andrerom 
Once that sorted out: +1

lolautruche:
@yannickroger About i18n of exceptions, this is a kernel problem. Repository exceptions are currently not translatable (may you report that on Jira please?)

dpobel:
@yannickroger that's a very valid concern.
Now the question is, how should we handle this problem ?

lolautruche:
@dpobel @yannickroger AFAICT, this can only be handled inside the repository, by making these exceptions translatable...

lolautruche:
OK, PR is now complete. All not found / access checks are left to repository + symfony now.

Still missing a +1 :innocent:

dpobel:
+1 just make sure to log a JIRA issue for the untranslatable messages

lolautruche:
Here it is: https://jira.ez.no/browse/EZP-24793

PR #326, opened 2015-09-10
EZP-24770: As an editor, I want to be able to configure the paragraph element

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24770

# Description

This patch adds the configuration of the `styles` toolbar so that it's possible to tweak paragraphs like it's already the case for heading. While at it, I also change the positioning strategy of the heading toolbar. For both, the toolbar is placed right above its related block element.

Screencast: https://youtu.be/PKtkqvAX-vQ

## Tasks details

* [x] Move the toolbar configurations to external modules and add test
* [x] Implement the toolbar based on the heading one (added in https://github.com/ezsystems/PlatformUIBundle/pull/313)
* [x] Change the position strategy so that the toolbar is placed just above its related block not above the caret.

# Tests

manual + unit tests

### DIFF
--- a/Resources/config/yui.yml
+++ b/Resources/config/yui.yml
@@ -22,9 +22,15 @@ system:
                 ez-alloyeditor-toolbar-config-table:
                     requires: ['ez-alloyeditor']
                     path: %ez_platformui.public_dir%/js/alloyeditor/toolbars/config/table.js
+                ez-alloyeditor-toolbar-config-block-base:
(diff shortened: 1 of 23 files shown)


dpobel:

andrerom:
+1

mhyndle:
+1

andrerom:
@dpobel @bdunogier was this forgotten? or does it need further changes maybe? :/

PR #328, opened 2015-09-11
Fix EZP-24786: ContentType and Role creation must use POST

lolautruche:
> https://jira.ez.no/browse/EZP-24786


lolautruche:
Review ping @andrerom @bdunogier @dpobel @yannickroger @glye

andrerom:
+1

Any conflicts @glye should be aware of?

lolautruche:
@andrerom @glye AFAICT should not cause any problem.

glye:
+1, conflicts should be minimal, if any.

PR #329, opened 2015-09-14
EZP-24787: Catch and display real error messages as notification

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24787

# Description

As of https://github.com/ezsystems/PlatformUIBundle/pull/325, the Symfony error page embeds the notifications tag. This patch makes sure we use that (when available) to notify the user about a loading error instead of the *Failed to load* something message. This is done both when submitting a form and for *regular* page loading.

# Tests

manual + unit tests.


andrerom:
+1

yannickroger:
+1

PR #332, opened 2015-09-16
EZP-24810: notification when user changes language when creating the content

mhyndle:
Jira: https://jira.ez.no/browse/EZP-24810

# Description
Currently user is allowed to change the language when creating the content, but when it is done, he is not notified about that.
This  PR adds notification after changing the language.

### DIFF
--- a/Resources/public/js/views/services/ez-contentcreateviewservice.js
+++ b/Resources/public/js/views/services/ez-contentcreateviewservice.js
@@ -121,7 +121,7 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
         },
 
         /**
-         * Sets language of created content to one given in event facade.
+         * Sets language of created content to one given in event facade. After that notification is fired.
          *
          * @method _setLanguage
          * @private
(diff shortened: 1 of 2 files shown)


mhyndle:
ready for review

Resources/public/js/views/services/ez-contentcreateviewservice.js
```diff
@@ -137,6 +137,15 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
                 service.set('languageCode',e.selectedLanguageCode);
                 service.set('version', new Y.eZ.Version());
                 view.set('languageCode', e.selectedLanguageCode);
+
+                service.fire('notify', {
+                    notification: {
+                        text: 'Language has been changed to ' + e.selectedLanguageCode,
+                        identifier: 'create_content_change_language_to_' + e.selectedLanguageCode,
```
dpobel: nitpick: use `-` instead of `_` to separate words.

Resources/public/js/views/services/ez-contentcreateviewservice.js
```diff
@@ -137,6 +137,15 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
                 service.set('languageCode',e.selectedLanguageCode);
```
dpobel: actually, I realize that doing that in the callback of `destroy` is wrong because if you saved the version the callback is executed asynchronously. So instead, you should write something like:

```js
version.destroy({api: this.get('capi'), remove: true}, function (error) {
    if ( error ) {
        console.warn('Failed to remove the version ' + version.get('versionId'));
    }
});
this.set('version', new Y.eZ.Version());
view.set('languageCode', e.selectedLanguageCode);
this.fire('notify', {
// ...
```

mhyndle:
Changes done according to Damien's feedback

dpobel:
+1

StephaneDiot:
+1

yannickroger:
+1

PR #333, opened 2015-09-16
Fix EZP-24811: JavaScript error when loading the tree for a Location other than the root

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24811

# Description

Regression introduced in https://github.com/ezsystems/PlatformUIBundle/commit/cca477fa6f907b197c89dfe42c4361b7edff1105 / [EZP-24706](https://jira.ez.no/browse/EZP-24706). The path now only contains the Location model since they embed the ContentInfo.

# Tests

manual, this part not unit tested yet because it'll change soon


dpobel:
ping @yannickroger @mhyndle

mhyndle:
+1

andrerom:
+1

PR #335, opened 2015-09-18
Added BDD User login sentences/implementation

miguelcleverti:
Added implementation for user with diferent roles login.
relates to jira issue: https://jira.ez.no/browse/EZP-24753

depends on: ezsystems/BehatBundle#33

### DIFF
--- a/Features/Authentication/UsersLogin.feature
+++ b/Features/Authentication/UsersLogin.feature
@@ -1,0 +1,11 @@
+Feature: Basic user with different roles login tests
+
+    @javascript
+    Scenario: As a Editor User, I want login to PlatformUI
+        Given I am logged in as an 'Editor' in PlatformUI
+        Then I should see "Welcome to eZ Platform" title
+
+    @javascript
+    Scenario: As a Administrator User, I want login to PlatformUI
+        Given I am logged in as an 'Administrator' in PlatformUI
+        Then I should see "Welcome to eZ Platform" title
--- a/Features/Context/SubContext/Authentication.php
+++ b/Features/Context/SubContext/Authentication.php
@@ -51,11 +51,19 @@ trait Authentication
 
     /**
      * @Given I am logged in as admin on PlatformUI
-     * @Given I am logged in as an Editor in PlatformUI
      */
     public function loggedAsUserPlatformUi()
     {
         $this->goToPlatformUiAndLogIn($this->user, $this->password);
+    }
+
+    /**
+     * @Given I am logged in as an :role in PlatformUI
+     */
+    public function loggedAsEditorPlatformUI($role)
+    {
+        $credentials = $this->getCredentialsFor($role);
+        $this->goToPlatformUiAndLogIn($credentials['login'], $credentials['password']);
     }
 
     /**
--- a/Features/CopyMoveDelete/CopyContent.feature
+++ b/Features/CopyMoveDelete/CopyContent.feature
@@ -4,7 +4,7 @@ Feature: Copy content
     I need to be able to copy an object that I am viewing
 
     Background:
-        Given I am logged in as an Editor in PlatformUI
+        Given I am logged in as an 'Editor' in PlatformUI
 
     ##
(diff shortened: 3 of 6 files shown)


miguelcleverti:
@andrerom @joaoinacio @pcardiga

Features/Context/SubContext/Authentication.php
```diff
@@ -59,6 +58,33 @@ public function loggedAsUserPlatformUi()
     }
 
     /**
+     * @Given I am logged in as an :role in PlatformUI
+     */
+    public function loggedAsEditorPlatformUI($role)
+    {
+        switch ($role) {
```
joaoinacio: I would probably add it in a separate function
````php
/**
 * Returns default user group name from role name. 
 */
public function getUserGroupForRolle($role)
```

joaoinacio:
Regardless of my inline comment, EzContext should already provide the means to create and return a user/password for a specific role:
```php
    /**
     * Get credentials for a specific role
     *
     * @uses \EzSystems\BehatBundle\Context\Object\User
     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
     *
     * @param string $role Role intended for testing
     *
     * @return array Associative with 'login' and 'password'
     */
    protected function getCredentialsFor( $roleIdentifier )
```
See https://github.com/ezsystems/BehatBundle/blob/master/Context/EzContext.php#L243

miguelcleverti:
@andrerom made the refactor suggested by @joaoinacio

Features/Context/SubContext/Authentication.php
```diff
@@ -51,7 +51,6 @@ public function goToPlatformUiAndLogIn($username, $password)
 
     /**
      * @Given I am logged in as admin on PlatformUI
-     * @Given I am logged in as an Editor in PlatformUI
```
joaoinacio: +1 after checking that this sentence is no longer being used :)

miguelcleverti: @joaoinacio checked

Features/Authentication/UsersLogin.feature
```diff
@@ -0,0 +1,26 @@
...
+
+    @javascript
+    Scenario: As a Partner User, I want login to PlatformUI
+        Given I am logged in as an 'Partner' in PlatformUI
+        Then I should see "Welcome to eZ Platform" title
+
+    @javascript
+    Scenario: As a Subscriber User, I want login to PlatformUI
+        Given I am logged in as an 'Subscriber' in PlatformUI
+        Then I should see "Welcome to eZ Platform" title
```
andrerom: Where id you get this from? neither Subscriber, Member, Partner or Anonymous should have access to login, so here the spec is wrong, and maybe also impl, but that is a different issue.

miguelcleverti: @andrerom I made these tests myself, to test the Issue because that is what the fix does it makes possible all the roles login.

andrerom: Yes, but then use some tag for specs that are wrong, cause this is a bug, and you should not reflect bugs in specs.

andrerom: Suggested convention, when a spec specifies a bug
tag it with `@bug` and add link to JIRA issue for it, in this case it's all one bug, so the sections here can point to the same bug.

miguelcleverti: @andrerom so we should add the `@bug` tag instead of removing these scenarios? I think we should remove them or add the inverse test "U should not be able to log in" even if they will fail for now.

andrerom: That would be my preference, but then you should split it into separate PR linked to from the issue reported for this bug ;)

miguelcleverti:
@andrerom removed the users that were not suppose to login

andrerom:
+1 ready for merge? Is it passing? :)

miguelcleverti:
@andrerom yes they are passing :smile:

PR #337, opened 2015-09-18
EZP-24824: As an editor, I want to be able to see and browse the subitems of a Location.

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24824

# Description

This patch provides the very first sub-items list. It performs quite well (one HTTP request to build it) thanks to the Location search exposed *via* the REST views. The bug issue right now is that [the list is not sorted](https://jira.ez.no/browse/EZP-24849) because [the REST views do not support that yet](https://jira.ez.no/browse/EZP-24315) (see also https://github.com/ezsystems/ezpublish-kernel/pull/1417).

Screencast: https://www.youtube.com/watch?v=0alRMQZTcmg

## Tasks

* [x] Fix the bug in Location where the id is always undefined
* [x] Add a location search plugin to be able to trigger a Location search
* [x] Implement the asynchronous view to list the sub items
* [x] Handle the case where the content is not a container
* [x] Pagination

# Tests

manual + unit test


dpobel:

andrerom:
+1 but @inakijv needs to start to think about some design tweaks I think :)

dpobel:
in an ideal world, this should come with the specifications, in an ideal world...

andrerom:
:)

Resources/public/css/views/subitemlist.css
```diff
@@ -0,0 +1,71 @@
...
+.ez-view-subitemlistview .ez-subitemlist-content {
+    margin: 1em;
+}
+
+.ez-view-subitemlistview .ez-subitemlist-table {
+    width: 100%;
+}
+
+.ez-view-subitemlistview .ez-subitemlist-navigation {
+    display: flex;
```
dpobel: self note: missing browser prefixes

Resources/public/js/views/ez-subitemlistview.js
```diff
@@ -0,0 +1,290 @@
...
+         * @method _hasPages
+         * @private
+         * @return {Boolean}
+         */
+        _hasPages: function () {
+            return this.get('location').get('childCount') > this.get('limit');
+        },
+
+        /**
+         * Checks whether the user is on the last "page".
```
mhyndle: `is on first "page".` and the same in `@method` doc below

mhyndle:
+1, besides inline comment above

Resources/public/js/views/ez-subitemlistview.js
```diff
@@ -0,0 +1,290 @@
...
+     * Provides the subitem list view.
+     *
+     * @module ez-subitemlistview
+     */
+    Y.namespace('eZ');
+
+    var IS_PAGE_LOADING = 'is-page-loading',
+        IS_DISABLED = 'is-disabled';
+
+    function linkIsDisabled(link) {
```
yannickroger: Should this need JSdoc ?

yannickroger:
+1

Resources/public/js/views/ez-subitemlistview.js
dpobel: it's not exposed in any way (really private) and I guess the name is clear enough

PR #338, opened 2015-09-21
Fix EZP-24453: Implement ContentType removal

lolautruche:
> https://jira.ez.no/browse/EZP-24453

Depends on https://github.com/ezsystems/repository-forms/pull/46 .

Note that it is only possible to remove ContentTypes **NOT** having Content instances (basically following what `ContentTypeService` does).
If one try to remove such ContentType, repository error will be bubbled as bad request (see https://github.com/ezsystems/ezpublish-kernel/pull/1416)


lolautruche:
Review ping @andrerom @bdunogier @dpobel @yannickroger @glye @pspanja

andrerom:
+1

glye:
+1

PR #339, opened 2015-09-22
[EZP-24846] BDD sentences for users that should not be able to login

miguelcleverti:
BDD tests for the issue:
https://jira.ez.no/browse/EZP-24846

### DIFF
--- a/Features/Authentication/UsersNotAuthorized.feature
+++ b/Features/Authentication/UsersNotAuthorized.feature
@@ -1,0 +1,18 @@
+Scenario: Users not able to login to PlatformUI tests
+
+    @javascript
+    Scenario: As a Member User, I can't login to PlatformUI
+        Given I am logged in as an 'Member' in PlatformUI
+        Then I should see "Invalid username or password" text
+
+    @javascript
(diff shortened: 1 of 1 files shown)


miguelcleverti:
@andrerom @pcardiga

andrerom:
+1, but this will need fixes before it can merged.

bdunogier:
Hmm, I might not have phrased it this way. By reading the 3 scenarios, you don't really know who's supposed to be able to login...

And fact is that *we* don't really know that :(

andrerom:
Looking back at this, why is this passing?

miguelcleverti:
@andrerom they are not, this was before the BDD tests were running in CI

bdunogier:
Needs more work. The roles that are used don't exist anymore (at least for some), due to the cleandata/demo changes.

Furthermore, I suggest that we cover the actually expected behaviour:

```
Given I am on to the PlatformUI login page
When I log in as a user who doesn't have permissions to use PlatformUI
Then I get an error message saying "..."
```

bdunogier:
Closing it since most of it must be changed.

PR #341, opened 2015-09-22
EZP-24850: Handle error while publishing a content

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24850

# Description

For whatever reason, the error handling was only added when using the "Save draft" button, not the "Publish" one. This patch adds a basic error handling so that the application does not stay in loading mode forever when a server side error occurs.

Screencast: http://youtu.be/2pxvK-lgKmA

# Tests

manual + unit test


dpobel:

mhyndle:
+1

andrerom:
+1

PR #342, opened 2015-09-23
Changed openTreeNode accessor from private to protected in CommonActions.php

micszo:


miguelcleverti:
+1

micszo:
Ping @dpobel

andrerom:
@micszo DP does not over see so much the Behat test impl, so I'll add a +1  and merge ;)

PR #296, opened 2015-08-11
Added Role UI BDD feature file

pcardiga:
BDD scenarios for Role UI (list, view, CRUD) - https://jira.ez.no/browse/EZP-24685

Todo:
- [x] Implementation

Later follow up:
- Validation of Role limitations view with links and human readable names
- Role assignment
- Role versioning
- Role list pagination

PR #343, opened 2015-09-23
Added Role UI BDD feature file and implementations

miguelcleverti:
RoleUI BDD sentences and implentations cherry-picked and squashed from #296

### DIFF
--- a/Features/Context/PlatformUI.php
+++ b/Features/Context/PlatformUI.php
@@ -24,6 +24,7 @@ class PlatformUI extends Context
     use SubContext\Move;
     use SubContext\Copy;
     use SubContext\Remove;
+    use SubContext\Role;
 
     /**
      * PlatformUI relative URL path.
@@ -78,6 +79,8 @@ class PlatformUI extends Context
         while ($page->find('css', $loadingSelector) != null) {
             usleep(100 * 1000); // 100ms
         }
+        //Temporary solution for race conditions errors with selenium2, TO BE IMPROVED
+        sleep(1);
     }
 
     /**
@@ -277,6 +280,7 @@ class PlatformUI extends Context
     public function __construct($uri, $user = null, $password = null)
     {
         parent::__construct();
+        $this->pageIdentifierMap['roles'] = '/ez#/admin/pjax%2Frole';
         $this->platformUiUri = $uri;
         if ($user != null) {
             $this->user = $user;
--- a/Features/Context/SubContext/CommonActions.php
+++ b/Features/Context/SubContext/CommonActions.php
@@ -164,13 +164,15 @@ trait CommonActions
     /**
      * Finds an HTML element by class and the text value and clicks it.
      *
-     * @param string $text Text value of the element
-     * @param string $selector CSS selector of the element
+     * @param string    $text           Text value of the element
+     * @param string    $selector       CSS selector of the element
+     * @param string    $textSelector   Extra CSS selector for text of the element
+     * @param string    $baseElement    Element in which the search is based
      */
-    protected function clickElementByText($text, $selector, $textSelector = null, $index = 1)
+    protected function clickElementByText($text, $selector, $textSelector = null, $baseElement = null, $index = 1)
     {
         $index + 1; //for selection of equal buttons
-        $element = $this->getElementByText($text, $selector, $textSelector);
+        $element = $this->getElementByText($text, $selector, $textSelector, $baseElement);
         if ($element) {
             $element->click();
         } else {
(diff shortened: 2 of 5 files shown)


miguelcleverti:
@andrerom @pcardiga @joaoinacio

Features/RoleUI/RoleUI.feature
```diff
@@ -0,0 +1,167 @@
...
+  Scenario: Updating a Role with an existing identifier fails validation
+    Given an "Organizers" role exists
+    And I have a "Security" role
+    And I edit the "Organizers" role
+    And I fill in "Name" with "Security"
+    When I click on "Save" button
+    Then I am notified that the name already exists
+
+  @javascript @common
+  Scenario: cancel the update of a role
```
andrerom: _C_ancel :)

Resources/views/Role/list_roles.html.twig
```diff
@@ -30,10 +30,10 @@
...
                     {% for role in roles %}
                         {# @var role \eZ\Publish\API\Repository\Values\User\Role #}
-                        <tr>
-                            <td><a href="{{ path("admin_roleView", {"roleId": role.id}) }}">{{ role.identifier }}</a></td>
-                            <td>{{ role.id }}</td>
-                            <td>
+                        <tr class="ez-role">
+                            <td class="ez-role-name"><a href="{{ path("admin_roleView", {"roleId": role.id}) }}">{{ role.identifier }}</a></td>
+                            <td class="ez-role-id">{{ role.id }}</td>
+                            <td class="ez-role-edit-button">
```
andrerom: Looks ok with me, but maybe @dpobel has a POV on how we should sugar the output to be testable.

Features/RoleUI/RoleUI.feature
```diff
@@ -0,0 +1,167 @@
...
+  @javascript @common
+  Scenario: Validate the existence of expected fields when creating a role
+    Given I am on the Roles page
+    When I create a new role
+    Then I should see a "Name" input field
+    And I should see a "Save" button
+    And I should see a "Cancel" button
+
+  @javascript @common
+  Scenario: create a valid role
```
andrerom: **C**reate, also below :)

andrerom:
Besides nitpicks, and possibly input from DP on how we semantically sugar the html output for testing, +1

miguelcleverti:
@andrerom nitpicks fixed :smile:

andrerom:
passing and ready for merge? :)

miguelcleverti:
@andrerom only the delete is not passing, because it is not possible to delete role yet

miguelcleverti:
@andrerom after all it is possible to delete roles role, will need just some extra implementation so no merge for now

pcardiga:
The possibility of deleting roles was introduced by latest implementation on Role UI

Resources/views/Role/list_roles.html.twig
dpobel: well, a class called `button` on something which is absolutely not a button, that's not good. You can add such class on the `a` but by following the naming convention of PlatformUI, ie `ez-button-role-edit`

Features/Context/SubContext/CommonActions.php
```diff
@@ -184,14 +184,17 @@ protected function clickElementByText($text, $selector, $textSelector = null, $i
      * @param string    $text           Text value of the element
      * @param string    $selector       CSS selector of the element
      * @param string    $textSelector   Extra CSS selector for text of the element
+     * @param string    $baseElement    Element in which the search is based
      * @param int       $iteration      Iteration number, used to control number of executions
      * @return array
      */
     protected function getElementByText($text, $selector, $textSelector = null, $iteration = 3)
```
andrerom: baseElement missing

miguelcleverti: I noticed, strange it's not there :neutral_face:

miguelcleverti:
Implemented the delete sentences, fixed all the issues caused by the role versioning and removed the "button" from the class name. All tests are passing, ready for merge @andrerom

Features/Context/PlatformUI.php
```diff
@@ -78,6 +79,7 @@ protected function waitForLoadings()
         while ($page->find('css', $loadingSelector) != null) {
             usleep(100 * 1000); // 100ms
         }
+        sleep(1);
```
andrerom: I din't quite get why this is needed, it seem to already be matched above, and would another `usleep(100 * 1000)` be enough?

andrerom: either way, need a inline comment here to mention why this is here.

miguelcleverti: I decided to leave this here because this way it does not happen the "random" race condition errors, this is more of a temporary solution while we don't find a better/definitive one, but I forgot to had some comments about it @andrerom

andrerom:
+1 awaiting checks before merge

PR #344, opened 2015-09-23
Fix EZP-24700: Role assignment CRUD

glye:
TODO:
- [x] Assign to user
- [x] Assign to user group
- [x] ~~Update assignment~~ (not really needed)
- [x] Delete assignment
- [x] ~~UDW should filter out anything but users and groups~~ (this isn't currently possible)
- [x] ~~Handle assignment limitations - needs UI mockup. Multiple screens/dialogs.~~ Follow-up: https://jira.ez.no/browse/EZP-24905

https://jira.ez.no/browse/EZP-24700

### DIFF
--- a/Controller/RoleController.php
+++ b/Controller/RoleController.php
@@ -18,6 +18,7 @@ use EzSystems\RepositoryForms\Data\Mapper\PolicyMapper;
 use EzSystems\RepositoryForms\Data\Mapper\RoleMapper;
 use EzSystems\RepositoryForms\Form\ActionDispatcher\ActionDispatcherInterface;
 use EzSystems\RepositoryForms\Form\Type\Role\PolicyDeleteType;
+use EzSystems\RepositoryForms\Form\Type\Role\RoleAssignmentDeleteType;
 use EzSystems\RepositoryForms\Form\Type\Role\RoleCreateType;
 use EzSystems\RepositoryForms\Form\Type\Role\RoleDeleteType;
 use EzSystems\RepositoryForms\Form\Type\Role\RoleUpdateType;
@@ -71,6 +72,7 @@ class RoleController extends Controller
         return $this->render('eZPlatformUIBundle:Role:list_roles.html.twig', [
             'roles' => $this->roleService->loadRoles(),
             'can_edit' => $this->isGranted(new Attribute('role', 'update')),
+            'can_assign' => $this->isGranted(new Attribute('role', 'assign')),
             'can_create' => $this->isGranted(new Attribute('role', 'create')),
             'can_delete' => $this->isGranted(new Attribute('role', 'delete')),
             'create_form' => $createForm->createView(),
@@ -89,6 +91,7 @@ class RoleController extends Controller
         $role = $this->roleService->loadRole($roleId);
         $roleAssignments = $this->roleService->getRoleAssignments($role);
         $deleteForm = $this->createForm(new RoleDeleteType(), ['roleId' => $roleId]);
+
         $deleteFormsByPolicyId = [];
         foreach ($role->getPolicies() as $policy) {
             $policyId = $policy->id;
@@ -98,13 +101,23 @@ class RoleController extends Controller
             )->createView();
         }
 
+        $deleteFormsByAssignment = [];
+        foreach ($roleAssignments as $roleAssignment) {
+            $deleteFormsByAssignment[$roleAssignment->id] = $this->createForm(
+                new RoleAssignmentDeleteType(),
+                ['assignmentId' => $roleAssignment->id]
+            )->createView();
+        }
+
         return $this->render('eZPlatformUIBundle:Role:view_role.html.twig', [
             'role' => $role,
             'deleteFormsByPolicyId' => $deleteFormsByPolicyId,
             'role_assignments' => $roleAssignments,
             'deleteForm' => $deleteForm->createView(),
             'can_edit' => $this->isGranted(new Attribute('role', 'update')),
+            'can_assign' => $this->isGranted(new Attribute('role', 'assign')),
             'can_delete' => $this->isGranted(new Attribute('role', 'delete')),
+            'deleteFormsByAssignment' => $deleteFormsByAssignment,
         ]);
     }
 
@@ -290,5 +303,28 @@ class RoleController extends Controller
         }
 
         return $this->redirectToRouteAfterFormPost('admin_roleView', ['roleId' => $roleId]);
+    }
+
+    /**
+     * Deletes a role assignment.
+     *
+     * @param Request $request
+     * @param int $roleAssignmentId Role assignment ID
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function deleteRoleAssignmentAction(Request $request, $roleAssignmentId)
+    {
+        $roleAssignment = $this->roleService->loadRoleAssignment($roleAssignmentId);
+        $deleteForm = $this->createForm(new RoleAssignmentDeleteType(), ['assignmentId' => $roleAssignmentId]);
+        $deleteForm->handleRequest($request);
+        if ($deleteForm->isValid()) {
+            $this->roleService->removeRoleAssignment($roleAssignment);
+            $this->notify('role.assignment.deleted', [], 'role');
+        } elseif ($deleteForm->isSubmitted()) {
+            $this->notifyError('role.assignment.error.delete', [], 'role');
+        }
+
+        return $this->redirectToRouteAfterFormPost('admin_roleView', ['roleId' => $roleAssignment->role->id]);
     }
 }
--- a/Resources/config/routing_pjax.yml
+++ b/Resources/config/routing_pjax.yml
@@ -148,3 +148,9 @@ admin_policyEdit:
     defaults:
         _controller: ezsystems.platformui.controller.role:editPolicyAction
         policyId: ~
+
+admin_roleAssignmentDelete:
+    path: /role/deleteroleassignment/{roleAssignmentId}
+    methods: [POST]
+    defaults:
+        _controller: ezsystems.platformui.controller.role:deleteRoleAssignmentAction
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -107,6 +107,9 @@ YUI.add('ez-platformuiapp', function (Y) {
             sectionServerSideView: {
                 type: Y.eZ.SectionServerSideView,
             },
+            roleServerSideView: {
+                type: Y.eZ.RoleServerSideView,
+            },
             contentTypeEditServerSideView: {
                 type: Y.eZ.ContentTypeEditServerSideView,
             },
@@ -776,6 +779,15 @@ YUI.add('ez-platformuiapp', function (Y) {
                     sideViews: {'navigationHub': true, 'discoveryBar': false},
                     service: Y.eZ.SectionServerSideViewService,
                     view: "sectionServerSideView",
+                    callbacks: ['open', 'checkUser', 'handleSideViews', 'handleMainView']
+                }, {
+                    name: "adminRole",
+                    regex: /\/admin\/(pjax%2Frole$|pjax%2Frole%2F.*)/,
+                    keys: ['uri'],
+                    path: "/admin/:uri",
+                    sideViews: {'navigationHub': true, 'discoveryBar': false},
+                    service: Y.eZ.RoleServerSideViewService,
+                    view: "roleServerSideView",
                     callbacks: ['open', 'checkUser', 'handleSideViews', 'handleMainView']
                 }, {
                     name: "adminContentTypeEdit",
--- a/Resources/public/js/views/serverside/ez-roleserversideview.js
+++ b/Resources/public/js/views/serverside/ez-roleserversideview.js
@@ -1,0 +1,86 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-roleserversideview', function (Y) {
+    "use strict";
+    /**
+     * Provides the role server side view
+     *
+     * @module ez-roleserversideview
+     */
+    Y.namespace('eZ');
+
+    var events = {
+            '.ez-role-assign-button': {
+                'tap': '_pickSubtree'
+            },
+        };
+
+    /**
+     * The role server side view. It adds the handling of the role assign
+     * button.
+     *
+     * @namespace eZ
+     * @class RoleServerSideView
+     * @constructor
+     * @extends eZ.ServerSideView
+     */
+    Y.eZ.RoleServerSideView = Y.Base.create('roleServerSideView', Y.eZ.ServerSideView, [], {
+        initializer: function () {
+            this.events = Y.merge(this.events, events);
+        },
+
+        /**
+         * tap event handler on the role assign buttons. It launches the
+         * universal discovery widget so that the user can pick some contents.
+         *
+         * @method _pickSubtree
+         * @protected
+         * @param {EventFacade} e
+         */
+        _pickSubtree: function (e) {
+            var button = e.target,
+                unsetLoading = Y.bind(this._uiUnsetAssignRoleLoading, this, button);
+
+            e.preventDefault();
+            this._uiSetAssignRoleLoading(button);
+            this.fire('contentDiscover', {
+                config: {
+                    title: button.getAttribute('data-universaldiscovery-title'),
+                    cancelDiscoverHandler: unsetLoading,
+                    multiple: true,
+                    data: {
+                        roleId: button.getAttribute('data-role-rest-id'),
+                        roleName: button.getAttribute('data-role-name'),
+                        afterUpdateCallback: unsetLoading,
+                    },
+                },
+            });
+        },
+
+        /**
+         * Changes the state of the provided assign button to be *loading* and
+         * to be disabled
+         *
+         * @method _uiSetAssignRoleLoading
+         * @protected
+         * @param {Y.Node} button
+         */
+        _uiSetAssignRoleLoading: function (button) {
+            button.addClass('is-loading').set('disabled', true);
+        },
+
+        /**
+         * Changes the state of the provided assign button to not be *loading*
+         * and to be enabled.
+         *
+         * @method _uiUnsetAssignRoleLoading
+         * @protected
+         * @param {Y.Node} button
+         */
+        _uiUnsetAssignRoleLoading: function (button) {
+            button.removeClass('is-loading').set('disabled', false);
+        },
+    });
+});
--- a/Resources/public/js/views/services/ez-roleserversideviewservice.js
+++ b/Resources/public/js/views/services/ez-roleserversideviewservice.js
@@ -37,6 +37,7 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
         _assignRole: function (e) {
             var data = e.target.get('data'),
                 userService = this.get('capi').getUserService(),
+                countAssigned = 0,
                 service = this;
 
             this._assignRoleNotificationStarted(
@@ -45,22 +46,37 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
 
             userService.loadRole(data.roleId, function (error, response) {
                 var tasks = new Y.Parallel(),
-                    end = tasks.add(function (err, response) {
-                        if (err) {
-                            service._error('The role could not be assigned: ' + err.message);
-                        }
-                    });
+                    role;
 
                 if (error) {
-                    service._error('The role could not be loaded: ' + error.message);
+                    service._loadRoleErrorNotification(
+                        data.roleId, data.roleName, e.selection
+                    );
                     return;
                 }
 
-                service._assignRoleToEachAssignee(e, response.document.Role, end);
+                role = response.document.Role;
+
+                Y.Array.each(e.selection, function (struct) {
+                    var end = tasks.add(function (err, response) {
+                            if (err) {
+                                service._notify(
+                                    'Role has not been assigned to "' + struct.content.get('name') + '": ' + err.message,
+                                    'assign-role-failed-' + struct.content.get('id'),
+                                    'error',
+                                    0
+                                );
+                                return;
+                            }
+                            countAssigned++;
+                        });
+
+                    service._assignRoleToAssignee(struct, role, end);
+                });
 
                 tasks.done(function () {
                     service._assignRoleCallback(
-                        data.roleId, data.roleName, e.selection,
+                        data.roleId, data.roleName, e.selection, countAssigned,
                         Y.bind(data.afterUpdateCallback, service)
                     );
                 });
@@ -68,37 +84,28 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
         },
 
         /**
-         * Assign the role to each user/group
+         * Assign the role to selected content
          *
-         * @method _assignRoleToEachAssignee
+         * @method _assignRoleToAssignee
          * @protected
-         * @param {EventFacade} e
+         * @param {Object} struct the object selected with content discovery
          * @param {Role} role
-         * @param {Function} end Callback function
+         * @param {Function} callback the callback function
          */
-        _assignRoleToEachAssignee: function (e, role, end) {
-            var userService = this.get('capi').getUserService(),
-                service = this,
-                roleAssignInputStruct;
-
-            Y.Array.each(e.selection, function (struct) {
-                var limitation = null,
-                    contentTypeIdentifier = struct.contentType.get('identifier');
-
-                // TODO: EZP-24905 Role assignment limitations
+        _assignRoleToAssignee: function (struct, role, callback) {
+            var // TODO: EZP-24905 Role assignment limitations
                 // limitation will be filled like example in UserService.prototype.newRoleAssignInputStruct()
                 // hrefs in the limitation must be converted to IDs, as expected by the API
+                limitation = null,
+                contentTypeIdentifier = struct.contentType.get('identifier');
 
-                roleAssignInputStruct = userService.newRoleAssignInputStruct(role, limitation);
-
-                if (contentTypeIdentifier === 'user') {
-                    service._assignRoleToUser(roleAssignInputStruct, struct.content, end);
-                } else if (contentTypeIdentifier === 'user_group') {
-                    service._assignRoleToUserGroup(roleAssignInputStruct, struct.content, end);
-                } else {
-                    end({message:'The select content is not a user or group.'});
-                }
-            });
+            if (contentTypeIdentifier === 'user') {
+                this._assignRoleToUser(role, limitation, struct.content, callback);
+            } else if (contentTypeIdentifier === 'user_group') {
+                this._assignRoleToUserGroup(role, limitation, struct.content, callback);
+            } else {
+                callback({message:'Selected content is not a user or group.'});
+            }
         },
 
         /**
@@ -106,18 +113,20 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
          *
          * @method _assignRoleToUser
          * @protected
-         * @param {RoleAssignInputStruct} roleAssignInputStruct
+         * @param {Object} role
+         * @param {Object|null} limitation
          * @param {Content} content
-         * @param {Function} end Callback function
+         * @param {Function} callback the callback function
          */
-        _assignRoleToUser: function (roleAssignInputStruct, content, end) {
+        _assignRoleToUser: function (role, limitation, content, callback) {
             var userService = this.get('capi').getUserService(),
                 discoveryService = this.get('capi').getDiscoveryService(),
+                roleAssignInputStruct= userService.newRoleAssignInputStruct(role, limitation),
                 userIdStr;
 
             discoveryService.getInfoObject('users', function (error, usersInfoObject) {
                 if (error) {
-                    end(error);
+                    callback(error);
                     return;
                 }
 
@@ -126,7 +135,7 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
                 userIdStr = usersInfoObject._href + '/' + content.get('contentId');
 
                 userService.assignRoleToUser(
-                    userIdStr, roleAssignInputStruct, end
+                    userIdStr, roleAssignInputStruct, callback
                 );
             });
         },
@@ -136,19 +145,21 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
          *
          * @method _assignRoleToUserGroup
          * @protected
-         * @param {RoleAssignInputStruct} roleAssignInputStruct
+         * @param {Object} role
+         * @param {Object|null} limitation
          * @param {Content} content
-         * @param {Function} end Callback function
+         * @param {Function} callback the callback function
          */
-        _assignRoleToUserGroup: function (roleAssignInputStruct, content, end) {
+        _assignRoleToUserGroup: function (role, limitation, content, callback) {
             var userService = this.get('capi').getUserService(),
                 discoveryService = this.get('capi').getDiscoveryService(),
+                roleAssignInputStruct= userService.newRoleAssignInputStruct(role, limitation),
                 location,
                 groupIdStr;
 
             discoveryService.getInfoObject('rootUserGroup', function (error, rootUserGroup) {
                 if (error) {
-                    end(error);
+                    callback(error);
                     return;
                 }
 
@@ -161,7 +172,7 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
                 groupIdStr = /.+user\/groups/.exec(rootUserGroup._href) + /(\/\d+)+$/g.exec(location)[0];
 
                 userService.assignRoleToUserGroup(
-                    groupIdStr, roleAssignInputStruct, end
+                    groupIdStr, roleAssignInputStruct, callback
                 );
             });
         },
@@ -174,10 +185,30 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
          * @param {String} roleId the role id
          * @param {String} roleName the role name
          * @param {Array} contents the array of Content items to which role is being assigned to
+         * @param {Integer} countAssigned number of successfully assignments
          * @param {Function} callback the callback to call when other tasks are done
          */
-        _assignRoleCallback: function (roleId, roleName, contents, callback) {
-            this._assignRoleNotificationDone(roleId, roleName, contents);
+        _assignRoleCallback: function (roleId, roleName, contents, countAssigned, callback) {
+            var notificationIdentifier = this._getAssignRoleNotificationIdentifier(
+                    'assign-role', roleId, contents
+                );
+
+            if (countAssigned>0) {
+                this._notify(
+                    '"' + roleName + '" role has been assigned to ' + countAssigned + ' Content items',
+                    notificationIdentifier,
+                    'done',
+                    5
+                );
+            } else {
+                this._notify(
+                    'Role has not been assigned to any Content item',
+                    notificationIdentifier,
+                    'error',
+                    0
+                );
+            }
+
             callback();
         },
 
@@ -195,38 +226,34 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
                     'assign-role', roleId, contents
                 );
 
-            this.fire('notify', {
-                notification: {
-                    identifier: notificationIdentifier,
-                    text: 'Assigning the role "' + roleName + '" to ' + contents.length + ' Content items',
-                    state: 'started',
-                    timeout: 0
-                },
-            });
+            this._notify(
+                'Assigning the role "' + roleName + '" to ' + contents.length + ' Content items',
+                notificationIdentifier,
+                'started',
+                5
+            );
         },
 
         /**
-         * Notification changed to *done* after assigning role to Content items
+         * Notification with *error* state when loadRole fails
          *
-         * @method _assignRoleNotificationDone
+         * @method _loadRoleErrorNotification
          * @protected
          * @param {String} roleId the role id
          * @param {String} roleName the role name
          * @param {Array} contents the array of Content items to which role is being assigned to
          */
-        _assignRoleNotificationDone: function (roleId, roleName, contents) {
+        _loadRoleErrorNotification: function (roleId, roleName, contents) {
             var notificationIdentifier = this._getAssignRoleNotificationIdentifier(
                     'assign-role', roleId, contents
                 );
 
-            this.fire('notify', {
-                notification: {
-                    identifier: notificationIdentifier,
-                    text: 'Role "' + roleName + '" assigned to ' + contents.length + ' Content items',
-                    state: 'done',
-                    timeout: 5
-                },
-            });
+            this._notify(
+                'The role "' + roleName + '" could not be loaded',
+                notificationIdentifier,
+                'error',
+                0
+            );
         },
 
         /**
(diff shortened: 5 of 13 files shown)


Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,153 @@
...
+     * The Role Server Side View Service class.
+     *
+     * @namespace eZ
+     * @class RoleServerSideViewService
+     * @constructor
+     * @extends eZ.ServerSideViewService
+     */
+    Y.eZ.RoleServerSideViewService = Y.Base.create('roleServerSideViewService', Y.eZ.ServerSideViewService, [], {
+        initializer: function () {
+            this.on('*:contentDiscover', function (e) {
```
mhyndle: I suppose that you should create another event for it, for example `assignRole` which trigger will be calling `_assignRole`.
Modifying config for `contentDiscover` event on the fly is also something that I think shouldn't be done.

Resources/public/js/views/serverside/ez-roleserversideview.js
```diff
@@ -0,0 +1,86 @@
...
+            var button = e.target,
+                unsetLoading = Y.bind(this._uiUnsetAssignRoleLoading, this, button);
+
+            e.preventDefault();
+            this._uiSetAssignRoleLoading(button);
+            this.fire('contentDiscover', {
+                config: {
+                    title: button.getAttribute('data-universaldiscovery-title'),
+                    cancelDiscoverHandler: unsetLoading,
+                    multiple: true,
```
mhyndle: `config` should contain here the `contentDiscoveredHandler` function, let's name it `_contentSelected`. This function will be called after selecting the content from UDW, and selected content will be passed to this function. So this function should trigger event in viewService responsible for assigning the role, see https://github.com/ezsystems/PlatformUIBundle/pull/344#discussion_r40299143

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,153 @@
...
+            );
+
+            Y.Array.each(e.selection, function (struct) {
+                var role = userService.loadRole(data.roleId);
+                var roleAssign = userService.newRoleAssignInputStruct(role); // TODO assign with limitation
+                var userGroupId = struct.content.get('id');
+
+                if (userGroupId.indexOf('/user/groups/') > 0) {
+                    userService.assignRoleToUserGroup(
+                        userGroupId, roleAssign, tasks.add()
```
mhyndle: You should pass function as param into `tasks.add()`.
The simplest would be `tasks.add(function () {})`
That should be valid, however I suppose that you should prepare some error handling in the case of role assignment fail.

glye: @mhyndle Tried it, but I get the same error as before, "TypeError: callback is not a function CAPI.js:1389:12"
Found it now in my local CAPI.js, it corresponds to this:
https://github.com/ezsystems/ez-js-rest-client/blob/master/dist/CAPI.js#L1389
Does this tell you anything?

Resources/public/js/views/services/ez-roleserversideviewservice.js
dpobel: actually I guess @glye followed the section view service where it's done like that and that's OK I think.

glye: Yes I did, and I'm glad if that's OK because I have no idea what this is about :)

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,190 @@
...
+
+            this._assignRoleNotificationStarted(
+                data.roleId, data.roleName, e.selection
+            );
+
+            userService.loadRole(data.roleId, function (error, role) {
+                var tasks = new Y.Parallel(),
+                    end = tasks.add(function (err, response) {
+                        if (err) {
+                            service._error('tasks.add() failed: ' + response.status);
```
glye: @mhyndle @dpobel tasks.add is still failing, giving me status 400. Any idea what's wrong?

mhyndle: That's the result of loading `assignRoleToUser` or `assignRoleToUserGroup` which means that probably user/group wasn't found and 404 was returned. That would be my first thought. Please verify if userId/groupId are OK.

glye: In this case the ID is of a user, /api/ezp/v2/user/users/120, which is correct if 120 is supposed to be a content object ID...? CC @dpobel

glye:
I finally got basic role assignment to work. It turns out to be a problem either in the REST parser, or in the data submitted by the UDW - I don't know which of them is correct. The following hack makes it work: https://gist.github.com/glye/8c005203f0a2342b9f11
Ref: https://github.com/ezsystems/ezpublish-kernel/commit/8d29aff21dbb49f95cebd230f1d6b989fd176861#diff-036233eec01cc61c839b7d5266f4916fR41

@emodric Please comment, this is partly your code. Also CC @dpobel @mhyndle

emodric:
@dpobel @mhyndle @glye Where does the `document` element come from? That doesn't sound right. None of the input parsers deal with `document` element from input data.

This is how data should look like (hopefully, as this is couple of years old :)):

https://github.com/emodric/ezpRestScripts/blob/master/json/role_assignments/02-createUserRoleAssignment.sh

glye:
@emodric @dpobel @mhyndle This is how the response looks now:
https://gist.github.com/glye/e2d435814d90d9d7cd1f

emodric:
@glye Is this specific to RoleAssignInput? I didn't work with platform UI so don't know if that request is valid there, but this definitely is not valid from REST standpoint.

glye:
@emodric I'm not sure. (Login is just `{"SessionInput":{"login":"admin","password":"publish"}}`.) Role assignment is the only thing I'm using REST for so far.

emodric:
@glye Your `SessionInput` looks fine and is in line with input parser:

https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/REST/Client/Output/ValueObjectVisitor/SessionInput.php#L35-L39

andrerom:
@glye How you inspected what you have in your `role` variable in your js code here?

glye:
@andrerom At what point? `role` loads normally in `_assignRole` in the serversideviewservice.

glye:
I pushed my assignment-by-limitation code to make sure we're on the same code.
@dpobel @mhyndle Could you please test if you see the same role response as I do, with `document` and all? I'm checking `$request->getContent()` in `eZ\Publish\Core\REST\Server\Controller\Role::assignRoleToUser()`. Action: From role list, "assign to...", and select a user (not a group). Without the kernel hack gist I posted above, this will fail.

(I have tested in 3 different browsers, and the response is the same for all.)

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,208 @@
...
+            var data = e.target.get('data'),
+                userService = this.get('capi').getUserService(),
+                discoveryService = this.get('capi').getDiscoveryService(),
+                service = this;
+
+            this._assignRoleNotificationStarted(
+                data.roleId, data.roleName, e.selection
+            );
+
+            userService.loadRole(data.roleId, function (error, role) {
```
mhyndle: `userService.loadRole` returns response from REST, so the Role is embed inside the response. This line should rather looks like:
```javascript
userService.loadRole(data.roleId, function (error, response) {
  var tasks =  new Y.Parallel(),
      end = tasks.add( /* function */),
      role;

  if (error) {
    /* error handling */
  }

  role = response.document.Role;
  /* iterate through e.selection */
        
});
```

glye: Yep, with this the kernel hack isn't needed anymore. Thanks!

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,208 @@
...
+                                "_media-type": "application/vnd.ez.api.Section+json"
+                            }, {
+                                "_href": "/api/ezp/v2/content/sections/4",
+                                "_media-type": "application/vnd.ez.api.Section+json"
+                            }]
+                        }
+                    };
+                    var roleAssignInputStruct = userService.newRoleAssignInputStruct(role, limitation),
+                        contentTypeIdentifier = struct.contentType.get('identifier'),
+                        contentIdStr = struct.content.get('id');
```
mhyndle: `id` is full ID and we anly need the number, so maybe changing it to: 
```javascript
contentIdStr = struct.content.get('contentId');
```
It will simplify a bit later tasks

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,208 @@
...
+
+                    if (contentTypeIdentifier === 'user') {
+                        // TODO get this outside of loop
+                        discoveryService.getInfoObject('users', function (error, users) {
+                            if (error) {
+                                service._error('getInfoObject for "users" failed');
+                                return;
+                            }
+
+                            var userIdStr = users._href + '/' + /\d+$/.exec(contentIdStr);
```
mhyndle: According to previous comment (https://github.com/ezsystems/PlatformUIBundle/pull/344/files#r40662570) here we don't need to trim the content ID:
```javascript
var userIdStr = users._href + '/' + contentIdStr;
```

glye:
The `document` problem is fixed, and avoided a couple regexps. Thanks @mhyndle & others.
Continuing with assignment limitations implementation.

glye:
Both user and group assignments working now, limitations are todo.

glye:
Assignment limitation by section and subtree working internally, UI is todo.
Fixed display of limited assignments in role view template.

Github isn't updating, wtf...?

glye:
@lolautruche @andrerom Deleting role assignments: There's an API/SPI issue with this:
You can create multiple role assigments to the same user/usergroup, each with different assignment limitations (section/subtree). But when deleting, RoleService provides only `unassignRoleFromUser*(Role, User*)` meaning multiple assigments can be wiped in one go. It's the same in the whole stack - in DoctrineDatabase there is only `removeRole($contentId, $roleId)`, no way to remove a single assignment.

We do not read the `id` from the `ezuser_role` table. The `id` is specific to each assignment, so I think it would solve the problem if we read it, add it to SPI and API, and modify methods accordingly (or rather, add new `*ByAssignmentID()` methods). Are you OK with this?

glye:
Rebased on master, and added role assignment delete form. Assignment deletion works now.

glye:
Fixed @ezrobot javascript complaint.

andrerom:
If you need mockups for assignments with limitations, then maybe move this to review w/o and take that after, given beta1 is tagged end of day tomorrow.

glye:
Assignment limitations moved to follow-up: https://jira.ez.no/browse/EZP-24905

andrerom:
+1 but I think the js gang of four needs to review this :)  /cc @dpobel @yannickroger @mhyndle

glye:
Modified to take roleservice simplification into account (just one `loadRoleAssignment()` method).

glye:
Changed *contents* to *Content items*.

Form/Type/RoleAssignmentDeleteType.php
```diff
@@ -0,0 +1,49 @@
...
+ * @version //autogentag//
+ */
+namespace EzSystems\PlatformUIBundle\Form\Type;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Component\Translation\TranslatorInterface;
+
+class RoleAssignmentDeleteType extends AbstractType
```
lolautruche: This class should go to repository-forms

Form/Type/RoleAssignmentDeleteType.php
```diff
@@ -0,0 +1,49 @@
...
+
+    public function getName()
+    {
+        return 'roleassignmentdelete';
+    }
+
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver->setDefaults(
+            ['data_class' => 'eZ\Publish\API\Repository\Values\User\RoleUpdateStruct']
```
lolautruche: Not necessary

Resources/config/routing_pjax.yml
```diff
@@ -136,3 +136,12 @@ admin_roleDelete:
...
         _controller: ezsystems.platformui.controller.role:deleteRoleAction
+
+admin_roleAssignmentDelete:
+    path: /role/deleteusergrouproleassignment/{roleAssignmentId}/{redirectErrorsTo}
+    methods: [POST]
+    defaults:
+        _controller: ezsystems.platformui.controller.role:deleteRoleAssignmentAction
+        redirectErrorsTo: view
+    requirements:
+        redirectErrorsTo: (list|view)
```
lolautruche: copy/paste leftover ?

glye:
Removed vestigial `RoleAssignmentDeleteType`, this is already in repository-forms.
Removed useless routing requirement.

glye:
Review ping @dpobel @yannickroger @mhyndle & CC @lolautruche @pspanja

Resources/config/routing_pjax.yml
```diff
@@ -136,3 +136,10 @@ admin_roleDelete:
     methods: [POST]
     defaults:
         _controller: ezsystems.platformui.controller.role:deleteRoleAction
+
+admin_roleAssignmentDelete:
+    path: /role/deleteusergrouproleassignment/{roleAssignmentId}/{redirectErrorsTo}
```
pspanja: deleteusergrouproleassignment :)

pspanja:
+1 for the non-JS part :)

glye:
Fixed leftover `deleteusergrouproleassignment` from rushed commit yesterday, thanks Petar :)

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,230 @@
...
+            userService.loadRole(data.roleId, function (error, response) {
+                var role,
+                    tasks = new Y.Parallel(),
+                    end = tasks.add(function (err, response) {
+                        if (err) {
+                            service._error('tasks.add() failed: ' + err.message);
+                            return;
+                        }
+
+                        console.log('end'); // TODO just to let us know that it was picked, error handling should be here
```
mhyndle: Can be removed I think (that was only temporary way to sign that we hit that function).
As there can be many objects selected for assign the role, maybe we could show user bit more information about that how many assigns was successful and also notify user about every failed try. Something like here: https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/services/plugins/ez-locationcreateplugin.js#L83
Depends on specification as I don't know if it was precisely described.

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,230 @@
...
+                            if (limitation._identifier == "Section") {
+                                limitation.values.ref[i]._href = /\d+$/.exec(limitation.values.ref[i]._href)[0];
+                            } else if (limitation._identifier == "Subtree") {
+                                limitation.values.ref[i]._href = /(\/\d+)+$/.exec(limitation.values.ref[i]._href)[0];
+                            }
+                        }
+                    }
+
+                    var roleAssignInputStruct = userService.newRoleAssignInputStruct(role, limitation),
+                        contentTypeIdentifier = struct.contentType.get('identifier');
```
mhyndle: Should be moved to the beginning of that function, just after
`var limitation = null`

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,230 @@
...
+                        for ( var i = 0; i < limitation.values.ref.length; i++ ) {
+                            if (limitation._identifier == "Section") {
+                                limitation.values.ref[i]._href = /\d+$/.exec(limitation.values.ref[i]._href)[0];
+                            } else if (limitation._identifier == "Subtree") {
+                                limitation.values.ref[i]._href = /(\/\d+)+$/.exec(limitation.values.ref[i]._href)[0];
+                            }
+                        }
+                    }
+
+                    var roleAssignInputStruct = userService.newRoleAssignInputStruct(role, limitation),
```
mhyndle: Variable should be declared in the beginning of function as undefined: `var roleAssignInputStruct;` and here it should be only binded: `roleAssignInputStruct = userService.newRoleAssignInputStruct(role, limitation);`

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,230 @@
...
+                    });
+
+                if (error) {
+                    service._error('loadRole failed: ' + error.message);
+                    return;
+                }
+
+                role = response.document.Role;
+
+                Y.Array.each(e.selection, function (struct) {
```
mhyndle: I guess it could be moved to the separate method for make it easier to read.

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,230 @@
...
+                                limitation.values.ref[i]._href = /(\/\d+)+$/.exec(limitation.values.ref[i]._href)[0];
+                            }
+                        }
+                    }
+
+                    var roleAssignInputStruct = userService.newRoleAssignInputStruct(role, limitation),
+                        contentTypeIdentifier = struct.contentType.get('identifier');
+
+                    if (contentTypeIdentifier === 'user') {
+                        discoveryService.getInfoObject('users', function (error, usersInfoObject) {
```
mhyndle: Could be also moved to separate method.

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,230 @@
...
+
+                            // We need a user ID, e.g.: /api/ezp/v2/user/users/10
+                            var userIdStr = usersInfoObject._href + '/' + struct.content.get('contentId');
+
+                            userService.assignRoleToUser(
+                                userIdStr, roleAssignInputStruct, end
+                            );
+                        });
+                    } else if (contentTypeIdentifier === 'user_group') {
+                        discoveryService.getInfoObject('rootUserGroup', function (error, rootUserGroup) {
```
mhyndle: Could be also moved to separate method.

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,230 @@
...
+
+                    if (contentTypeIdentifier === 'user') {
+                        discoveryService.getInfoObject('users', function (error, usersInfoObject) {
+                            if (error) {
+                                service._error('getInfoObject for "users" failed: ' + error.message);
+                                return;
+                            }
+
+                            // We need a user ID, e.g.: /api/ezp/v2/user/users/10
+                            var userIdStr = usersInfoObject._href + '/' + struct.content.get('contentId');
```
mhyndle: Should be declared in the beginning of function as indefined variable: `var userIdStr;` and here the value should be just binded: `userIdStr = usersInfoObject._href + '/' + struct.content.get('contentId');`

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,230 @@
...
+                            if (error) {
+                                service._error('getInfoObject for "rootUserGroup" failed: ' + error.message);
+                                return;
+                            }
+
+                            /* We need to convert content ID to user group ID, e.g.
+                               change this: /api/ezp/v2/content/locations/1/5/14
+                               into this: /api/ezp/v2/user/groups/1/5/14
+                             */
+                            var location = struct.content.get('resources').MainLocation;
```
mhyndle: That and the next line: the same as https://github.com/ezsystems/PlatformUIBundle/pull/344/files#r41520575 (declaration of variable in the beginning of function)

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,230 @@
...
+                    //        "ref": [{
+                    //            "_href": "/api/ezp/v2/content/locations/1/5/14",
+                    //            "_media-type": "application/vnd.ez.api.Location+json"
+                    //        }]
+                    //    }
+                    //};
+
+                    // Convert hrefs to IDs, required by API
+                    if (limitation) {
+                        for ( var i = 0; i < limitation.values.ref.length; i++ ) {
```
mhyndle: As this is an array, maybe `Y.Array.each()` would be better?

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,230 @@
...
+                            if (limitation._identifier == "Section") {
+                                limitation.values.ref[i]._href = /\d+$/.exec(limitation.values.ref[i]._href)[0];
```
mhyndle: For this one, the one below, `userIdStr` few lines below and finally `groupIdStr` - shouldn't those be somehow gathered in other way?
ping @dpobel

Resources/views/Role/list_roles.html.twig
```diff
@@ -34,6 +34,14 @@
                             <td class="ez-role-name"><a href="{{ path("admin_roleView", {"roleId": role.id}) }}">{{ role.identifier }}</a></td>
                             <td class="ez-role-id">{{ role.id }}</td>
                             <td class="ez-role-edit">
+                                <button
+                                        data-universaldiscovery-title="{{ 'role.assign.universaldiscovery.title'|trans({'%roleIdentifier%': role.identifier })|e('html_attr') }}"
```
mhyndle: nitpick: double indent?

Resources/public/js/views/services/ez-roleserversideviewservice.js
glye: @mhyndle This is what Damien recommended - there is no nicer way ATM.

glye:
Performed unspeakably horrid acts of black magic under the guidance of necromantic magus @mhyndle. I remember little of the ordeal, but I'm shocked and appalled to find that the code actually still seems to work.

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,278 @@
...
+                userService.assignRoleToUser(
+                    userIdStr, roleAssignInputStruct, end
+                );
+            });
+        },
+
+        /**
+         * Assign the role to a user group
+         *
+         * @method _assignRoleToUser
```
mhyndle: `_assignRoleToUserGroup`

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,278 @@
...
+                        if (limitation._identifier == "Section") {
+                            refs[i]._href = /\d+$/.exec(ref._href)[0];
+                        } else if (limitation._identifier == "Subtree") {
+                            refs[i]._href = /(\/\d+)+$/.exec(ref._href)[0];
+                        }
+                    });
+                }
+
+                if (contentTypeIdentifier === 'user') {
+                    service._assignRoleToUser(userService.newRoleAssignInputStruct(role, limitation), struct, end);
```
mhyndle: Maybe it would be easier to read if you would pass to the `_assignRoleToUser` method separately `role`, `limitation`, `content` (`struct.content`), `end` and build assignInputStruct inside that method.

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,278 @@
...
+                } else if (contentTypeIdentifier === 'user_group') {
+                    service._assignRoleToUserGroup(userService.newRoleAssignInputStruct(role, limitation), struct, end);
```
mhyndle: The same as above https://github.com/ezsystems/PlatformUIBundle/pull/344/files#r41549301

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,278 @@
...
+        /**
+         * Assign the role to a user
+         *
+         * @method _assignRoleToUser
+         * @protected
+         * @param {RoleAssignInputStruct} roleAssignInputStruct
+         * @param {Content} struct
+         * @param {Function} end
+         */
+        _assignRoleToUser: function (roleAssignInputStruct, struct, end) {
```
mhyndle: According to https://github.com/ezsystems/PlatformUIBundle/pull/344/files#r41549301 I would require as params `role`, `limitation`, `content` and `callback`

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,278 @@
...
+        /**
+         * Assign the role to a user group
+         *
+         * @method _assignRoleToUser
+         * @protected
+         * @param {RoleAssignInputStruct} roleAssignInputStruct
+         * @param {Content} struct
+         * @param {Function} end
+         */
+        _assignRoleToUserGroup: function (roleAssignInputStruct, struct, end) {
```
mhyndle: Same as https://github.com/ezsystems/PlatformUIBundle/pull/344/files#r41549539

Resources/public/js/views/services/ez-roleserversideviewservice.js
glye: Here I disagree. For simplicity and readability I'd rather pass one parameter than two, and create the struct in one area of the code, rather than two. Instead, I went back to how it was before, with the struct assigned to a variable immediately before it is passed to the subfunctions.

Resources/public/js/views/services/ez-roleserversideviewservice.js
glye: See above.

Resources/public/js/views/services/ez-roleserversideviewservice.js
mhyndle: So the name of method can be confusing (or not?).
And again - better would be to pass directly `content` instead of `struct` as `struct` is only used as `struct.content.get('contentId')`

Resources/public/js/views/services/ez-roleserversideviewservice.js
mhyndle: See above ;)

glye:
Passing `struct.content` rather than `struct` to subfunctions, as that's all that's needed.

@mhyndle I don't think there is any confusion/readability issue with passing `roleAssignInputStruct` to these functions, it's clearer as it is. One of the reasons for using such structs in the first place is to avoid passing around lots of variables, isn't it? But I'll change it if another JS gang member agrees with you :)

mhyndle:
I agree that the `roleAssignInputStruct` issue should be verified by someone else, maybe by our JS guru @dpobel ;)

glye:
Rebased on master after policy edit commit.

Controller/RoleController.php
```diff
@@ -98,13 +101,23 @@ public function viewRoleAction($roleId)
             )->createView();
         }
 
+        $deleteFormsByAssignment = [];
+        foreach ($roleAssignments as $roleAssignment) {
+            $deleteFormsByAssignment[$roleAssignment->id] = $this->createForm(
+                new RoleAssignmentDeleteType($this->roleService),
```
lolautruche: You don't have to inject the role service into `RoleAssignmentDeleteType`

Resources/config/routing_pjax.yml
```diff
@@ -148,3 +148,10 @@ admin_policyEdit:
     defaults:
         _controller: ezsystems.platformui.controller.role:editPolicyAction
         policyId: ~
+
+admin_roleAssignmentDelete:
+    path: /role/deleteroleassignment/{roleAssignmentId}/{redirectErrorsTo}
```
lolautruche: Please remove `redirectErrorsTo` (also below in defaults)

Resources/views/Role/view_role.html.twig
```diff
@@ -18,152 +18,186 @@
...
+                                </td>
+                                <td>
+                                    {% set deleteForm = deleteFormsByAssignment[role_assignment.id] %}
+                                    {{ form_start(deleteForm, {"action": path("admin_roleAssignmentDelete", {"roleAssignmentId": role_assignment.id, "redirectErrorsTo": "view"})}) }}
+                                    {{ form_widget(deleteForm.assignmentId) }}
+                                    {{
+                                        form_widget(
+                                            deleteForm.delete,
+                                            {
+                                                "disabled": not can_edit,
```
lolautruche: Shouldn't it be `not can_assign`?

lolautruche:
Besides my latest comments, +1

glye:
Don't inject role service into RoleAssignmentDeleteType
Removed `redirectErrorsTo` from routing entry
Fixed unassign permission check
thanks JV.

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,281 @@
...
+            var userService = this.get('capi').getUserService(),
+                service = this,
+                roleAssignInputStruct;
+
+            Y.Array.each(e.selection, function (struct) {
+                // TODO assign with limitation
+                var limitation = null,
+                    contentTypeIdentifier = struct.contentType.get('identifier');
+
+                //var limitation = { // DUMMY SECTION DATA
```
dpobel: please no commented code

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,281 @@
...
+                //        "ref": [{
+                //            "_href": "/api/ezp/v2/content/locations/1/5/14",
+                //            "_media-type": "application/vnd.ez.api.Location+json"
+                //        }]
+                //    }
+                //};
+
+                // Convert hrefs to IDs, required by API
+                if (limitation) {
+                    Y.Array.each(limitation.values.ref, function (ref, i, refs) {
```
dpobel: since `limitation` is always null right now, this is dead code so you can remove it and this is where you should the todo comment with some explanation and a link to the corresponding JIRA issue.

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,281 @@
...
+                service = this,
+                userIdStr;
+
+            discoveryService.getInfoObject('users', function (error, usersInfoObject) {
+                if (error) {
+                    service._error('getInfoObject for "users" failed: ' + error.message);
+                    return;
+                }
+
+                // We need a user ID, e.g.: /api/ezp/v2/user/users/10
```
dpobel: here as well you should add a TODO with a link to JIRA saying that the UDW should allow to browse the users/user groups only and should then provide the User or User Group instead of the Content, Location and cie.

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,281 @@
...
+                location,
+                groupIdStr;
+
+            discoveryService.getInfoObject('rootUserGroup', function (error, rootUserGroup) {
+                if (error) {
+                    service._error('getInfoObject for "rootUserGroup" failed: ' + error.message);
+                    return;
+                }
+
+                /* We need to convert content ID to user group ID, e.g.
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/344/files#r41731615

glye:
Removed unfinished code for assignment by limitation.

Resources/public/js/views/services/ez-roleserversideviewservice.js
glye: Added.

dpobel:
The tests for RoleServerSideView and RoleServerSideViewService are missing as well.

glye:
Improved phantomjs/grover instructions in README.md
Added JS tests for RoleServerSideView and RoleServerSideViewService

@dpobel: 2 of 3 RoleServerSideViewService tests are failing. Not sure why, or for that matter, why one of them succeeds. Work in progress, let me know if you have any quick tips.
- Should assign the role to the discovered user/group(s): The afterUpdateCallback should have been called
- Should notify about the success of assigning user/group to the role: The notify event should have been fired

glye:
All tests passing locally now, except perhaps a sanity test on yours truly. Thanks @mhyndle 
Review ping @mhyndle @yannickroger @dpobel

README.md
```diff
@@ -79,7 +79,8 @@ have several additional dependencies:
   ```
   # npm install -g grunt-cli yuidocjs grover bower gulp
   ```
-* Install [phantomjs](http://phantomjs.org)
+* Install [phantomjs](http://phantomjs.org) NB: Install the 1.9 branch, 2.0
```
dpobel: I would just phrase like:
```md
Install [phantomjs](http://phantomjs.org) version 1.9.x
```

also this kind of change should go in a separate commit

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,253 @@
...
+                if (error) {
+                    service._error('loadRole failed: ' + error.message);
+                    return;
+                }
+
+                service._assignRoleToEachAssignee(e, response.document.Role, end);
+
+                tasks.done(function () {
+                    service._assignRoleCallback(data.roleId, data.roleName, e.selection,
+                        Y.bind(data.afterUpdateCallback, service));
```
dpobel: nitpick CS: should be rewritten
```js
service._assignRoleCallback(
    data.roleId, data.roleName, e.selection,
    Y.bind(data.afterUpdateCallback, service)
);
```

README.md
glye: @dpobel OK. I'll just push that without a PR, then.

glye:
Moved README change to separate commit https://github.com/ezsystems/PlatformUIBundle/commit/4b9fe8441839edb3fe478626c3b21f546d8097ed
Roleserversideviewservice CS fix

Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
```diff
@@ -0,0 +1,609 @@
...
+                },
+                callback = function () {
+                    callbackCalled = true;
+                    Assert.areSame(
+                        that.service, this,
+                        "The callback should be executed in the service context"
+                    );
+                };
+
+            Mock.expect(content, {
```
dpobel: it seems like you are configuring several expectations the same way in many test methods, I guess the definition of those expectations can be moved to the `setUp` method.

glye: Yes, I know. Tried it, didn't work. I can make another attempt. (effing black magic)

glye:
Consolidated mock expectations as much as I can without collapsing the whole house of cards.
Review ping @mhyndle @yannickroger @dpobel

glye:
Undid a readme change which has sneaked in.

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,255 @@
...
+
+            this._assignRoleNotificationStarted(
+                data.roleId, data.roleName, e.selection
+            );
+
+            userService.loadRole(data.roleId, function (error, response) {
+                var tasks = new Y.Parallel(),
+                    end = tasks.add(function (err, response) {
+                        if (err) {
+                            service._error('tasks.add() failed: ' + err.message);
```
mhyndle: This message will be displayed to the user, and that's rather technical detail. This will be handled when `_assignRoleToEachAssignee` will fail, so maybe message that role wasn't assigned will be better.

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,255 @@
...
+                // hrefs in the limitation must be converted to IDs, as expected by the API
+
+                roleAssignInputStruct = userService.newRoleAssignInputStruct(role, limitation);
+
+                if (contentTypeIdentifier === 'user') {
+                    service._assignRoleToUser(roleAssignInputStruct, struct.content, end);
+                } else if (contentTypeIdentifier === 'user_group') {
+                    service._assignRoleToUserGroup(roleAssignInputStruct, struct.content, end);
+                } else {
+                    service._error('The selected content "' + struct.content.get('name') + '" is not a user or user group.');
```
mhyndle: There is no `end` callback called and I'm wondering if whole process won't fail? I guess that instead of triggering error notification you cald call `end(true);` which will trigger error in `end` function.

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,255 @@
...
+        },
+
+        /**
+         * Assign the role to each user/group
+         *
+         * @method _assignRoleToEachAssignee
+         * @protected
+         * @param {EventFacade} e
+         * @param {Role} role
+         * @param {Function} end
```
mhyndle: nitpick: info about that it's callback function

mhyndle:
Except inline comments, the implementation is OK for me, but for the the `ez-roleserversideviewservice-tests` I think there could more things tested, because you are testing only optimistic scenario right now. Tests should cover error cases, for example: 
- `loadRole()` will fail
- trying to assign role to another content than user or user group (it will be solved by adding possibility of filtering content to UDW, but currently you have got that case covered in code so I guess it should be tested)
- other things that my trigger `_error`

Resources/public/js/views/services/ez-roleserversideviewservice.js
glye: This, and the above change, gives me a "The role was not assigned: undefined" when I try to assign to the wrong content type. Meaning, when I call `end(true)` there is of course no error message to output. I think we should send a more detailed error message somehow.

mhyndle: Yes, we should, but we can do it in `end()` function, maybe calling here something like `end({message: "The selected content...."});` so it will be handled inside the `end` function and error will be showed. I don't know if it's best solution, maybe @dpobel could verify it.
However, maybe it's not something that we should bother? I just feel that the `tasks.done()` won't be proceeded if we won't call `end()` for that case. And if `tasks.done()` won't be proceeded the `_assignRoleCallback` won't be called as well which I guess is not good. What's more it's not covered with tests - see https://github.com/ezsystems/PlatformUIBundle/pull/344#issuecomment-148638678

glye:
Made sure that `end()` is called in error situations, with sensible error messages, instead of direct notifications.
Added doc info about callbacks.

mhyndle:
ready for JS part review - ping @dpobel @yannickroger @StephaneDiot

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -0,0 +1,300 @@
...
+         * @param {Function} callback the callback to call when other tasks are done
+         */
+        _assignRoleCallback: function (roleId, roleName, contents, countAssigned, callback) {
+            var notificationIdentifier = this._getAssignRoleNotificationIdentifier(
+                    'assign-role', roleId, contents
+                );
+
+            if (countAssigned>0) {
+                this._notify(
+                    '"' + roleName + '" role has been assigned to ' + countAssigned + ' Content items',
```
glye: For these messages I have been using "users/groups" rathern than "content items" since you cannot assign a role to an article, for example. A bit clearer, I think. Same goes for the other messages.

glye:
Lots of black magic here, but +1 to the bits I do understand, excluding those I wrote myself (not that I don't like them...)

yannickroger:
+1

glye:
JS part review ping @dpobel

dpobel:
+1

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -778,6 +781,15 @@ YUI.add('ez-platformuiapp', function (Y) {
                     view: "sectionServerSideView",
                     callbacks: ['open', 'checkUser', 'handleSideViews', 'handleMainView']
                 }, {
+                    name: "adminRole",
```
andrerom: Noticed this first now, but this one is now duplicated with the one a bit further below, this is the right one now, the other should be removed.

glye: Ack, thanks.

PR #298, opened 2015-08-17
EZP-24637 field description when editing content

StephaneDiot:
https://jira.ez.no/browse/EZP-24637

## Description

This the new display of field description while editing or creating content. The display is not the same on touch devices (see https://doc.ez.no/display/PR/Content+type+and+field+description+tooltips )

## Screencast

With mouse : http://youtu.be/pjcm_ikCFTo
Touch Device : http://youtu.be/ycJpvjxjlHA

PR #345, opened 2015-09-24
EZP-21821: As an editor, I want to be able to edit content with an User field

dpobel:
JIRA: https://jira.ez.no/browse/EZP-21821

# Description

This patch provides the User edit view which allows to have complete form to edit user. It implements all the needed browser validation (including the login availability check added in https://github.com/ezsystems/ezpublish-kernel/pull/1415 / https://github.com/ezsystems/ez-js-rest-client/pull/65).

But actually, this PR is not be enough to be able to create/edit users because any content with a ezuser has to be saved through the User REST API.

Screencast: http://youtu.be/zodmsr0SP48

## Tasks

* [x] Basic user edit view
* [x] Check login availability
  * [x] Add a method for that in the JS REST client https://github.com/ezsystems/ez-js-rest-client/pull/65

# Tests

manual + unit tests

### DIFF
--- a/Resources/public/js/views/fields/ez-user-editview.js
+++ b/Resources/public/js/views/fields/ez-user-editview.js
@@ -1,0 +1,352 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-user-editview', function (Y) {
+    "use strict";
+    /**
+     * Provides the field edit view for the User (ezuser) fields
+     *
+     * @module ez-user-editview
+     */
+    Y.namespace('eZ');
+
+    var FIELDTYPE_IDENTIFIER = 'ezuser';
+
+    /**
+     * User edit field view
+     *
+     * @namespace eZ
+     * @class UserEditView
+     * @constructor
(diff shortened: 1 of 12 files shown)


dpobel:
ping @andrerom @mhyndle @yannickroger @StephaneDiot

Resources/public/js/views/fields/ez-user-editview.js
```diff
@@ -0,0 +1,446 @@
...
+                this.set('emailErrorStatus', 'The email is not valid');
+            } else if ( emailRequired && !this._getInputFieldValue('email') ) {
+                this.set('emailErrorStatus', 'The email is required');
+            } else {
+                this.set('emailErrorStatus', false);
+            }
+        },
+
+        /**
+         * Validates the password inputs. Both value should be the same.
```
mhyndle: nitpick: `values`

mhyndle:
+1

dpobel:
self note: I need to also apply the same change as in https://github.com/ezsystems/PlatformUIBundle/pull/298 to handle the field definition description in the new way.

yannickroger:
+1

andrerom:
+1

*Future possible followup: yellow feedback when password is weak.*

PR #349, opened 2015-09-28
EZP-24654: Inject the PlatformUI application config in one object

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24654

# Description

This patch changes the way the PlatformUIApp receives its configuration. With it, it now receives a single `config` hash build with the *Application Config Providers* introduced in https://github.com/ezsystems/PlatformUIBundle/pull/286 / [EZP-24129](https://jira.ez.no/browse/EZP-24129).

This has 2 main benefits:

* the logic is moved from the shell.html.twig template to a testable and tested code
* the configuration is completely extensible, ie an external bundle can add its own application provider(s) to inject any configuration and can add its own JavaScript components and/or plugins to consume it (ping @sunpietro).

So now, the view services receive by default the app configuration in the `config` attribute. In view services dedicated to side views (bars, udw, confirm box, ...), the `config` attribute was also used to set some execution parameters (like the title of the confirm box for instance). To avoid mixing everything, the side views have now a `parameters` attribute dedicated to that.

## Tasks

* [x] Inject the `parameters` Twig variables and dispatch its content
* [x] Change the *route config* system to avoid dealing with route names

# Tests

unit tests + manual tests

### DIFF
--- a/Resources/public/js/views/services/ez-notificationhubviewservice.js
+++ b/Resources/public/js/views/services/ez-notificationhubviewservice.js
@@ -19,13 +19,13 @@ YUI.add('ez-notificationhubviewservice', function (Y) {
      * @constructor
      * @extends eZ.TemplateBasedView
      */
-    Y.eZ.NotificationHubViewService = Y.Base.create('notificationHubViewService', Y.eZ.ViewService, [], {
+    Y.eZ.NotificationHubViewService = Y.Base.create('notificationHubViewService', Y.eZ.ViewService, [Y.eZ.SideViewService], {
         initializer: function () {
-            this.after('configChange', function () {
-                var config = this.get('config');
+            this.after('parametersChange', function () {
+                var parameters = this.get('parameters');
 
-                if ( config.notification ) {
-                    this._handleNotification(config.notification);
+                if ( parameters.notification ) {
+                    this._handleNotification(parameters.notification);
                 }
             });
         },
--- a/Resources/public/js/views/services/ez-notificationhubviewservice.js
+++ b/Resources/public/js/views/services/ez-notificationhubviewservice.js
@@ -48,7 +48,7 @@ YUI.add('ez-notificationhubviewservice', function (Y) {
             }
         },
 
-        getViewParameters: function () {
+        _getViewParameters: function () {
             return {
                 notificationList: this.get('notificationList'),
             };
--- a/Tests/js/views/assets/ez-rawcontentview-tests.js
+++ b/Tests/js/views/assets/ez-rawcontentview-tests.js
@@ -3,7 +3,8 @@
  * For full copyright and license information view LICENSE file distributed with this source code.
  */
 YUI.add('ez-rawcontentview-tests', function (Y) {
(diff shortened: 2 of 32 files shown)


dpobel:
ping @yannickroger @mhyndle @andrerom @StephaneDiot

yannickroger:
+1

mhyndle:
+1

andrerom:
@dpobel should maybe @sunpietro get a heads up on this btw.. :)

Resources/public/js/views/services/ez-notificationhubviewservice.js
```diff
@@ -19,13 +19,13 @@ YUI.add('ez-notificationhubviewservice', function (Y) {
      * @constructor
      * @extends eZ.TemplateBasedView
      */
-    Y.eZ.NotificationHubViewService = Y.Base.create('notificationHubViewService', Y.eZ.ViewService, [], {
+    Y.eZ.NotificationHubViewService = Y.Base.create('notificationHubViewService', Y.eZ.ViewService, [Y.eZ.SideViewService], {
         initializer: function () {
-            this.after('configChange', function () {
-                var config = this.get('config');
+            this.after('parametersChange', function () {
+                var parameters = this.get('parameters');
```
sunpietro: Why don't you use `event.newVal` instead? I believe it's more optimal in terms of app performance.

Tests/js/views/assets/ez-rawcontentview-tests.js
```diff
@@ -566,9 +532,54 @@ YUI.add('ez-rawcontentview-tests', function (Y) {
         },
     });
 
+    configFieldViewTest= new Y.Test.Case({
```
sunpietro: Missing space before the test case name

sunpietro:
Generally, I'm not so sure how can I use it in Studio app. Maybe I didn't read the code carefully, but I'm not able to see where can I add additional config from backend on extension initialization to avoid additional AJAX requests.

andrerom:
@sunpietro as far as I read this, it builds on the symfony service container tags approach introduced in https://github.com/ezsystems/PlatformUIBundle/pull/286 / [EZP-24129](https://jira.ez.no/browse/EZP-24129), but now this is made avaiable to the application. This means you can get your team started on setting up services that generates the config you need on startup *(so yes, as anything else, slow code will slow down initial page load, with great power comes great responsibility..)*.

Resources/public/js/views/services/ez-notificationhubviewservice.js
dpobel: because it does not mean the same thing, you might not have the same value for instance if there's a `getter` function on the attribute (and really the perf difference does not worth to bother)

dpobel:
@sunpietro : @andrerom is perfectly right, given the discussion on https://github.com/ezsystems/PlatformUIBundle/pull/301 @Nattfarinn knows how to build Symfony service to enhance the app `config` and then it's up to you to use that in any plugin/component you want.

sunpietro:
@andrerom @dpobel thanks for clarification :)

PR #354, opened 2015-09-30
EZP-24654: Additional fixes

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24654

# Description

The patch in https://github.com/ezsystems/PlatformUIBundle/pull/349 was incomplete / buggy:
* the auth is broken because of a wrong SessionAuthAgent initialization
* the view service did not receive the configuration
* the test on that was wrongly written hiding the bug
* as a result, the field (edit) view did not have the necessary data (the country list for Country, the ckeditor plugin path for Rich edit view).
* the default ckeditor plugin path was also wrong

While at it, I also removed the `config` attribute in some views since it's now defined in the `eZ.View`.

# Tests

manual tests + unit test

### DIFF
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -567,6 +567,7 @@ YUI.add('ez-platformuiapp', function (Y) {
                     request: req,
                     response: res,
                     plugins: Y.eZ.PluginRegistry.getPlugins(ServiceContructor.NAME),
+                    config: this.get('config'),
                 });
 
                 viewInfo.service = route.serviceInstance;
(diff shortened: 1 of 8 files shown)


dpobel:
ping @mhyndle @yannickroger @StephaneDiot @an

andrerom:
+1

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -173,8 +173,10 @@ YUI.add('ez-platformuiapp', function (Y) {
 
             this._set('capi', new Y.eZ.CAPI(
                 this.get('apiRoot').replace(/\/{1,}$/, ''),
-                new Y.eZ.SessionAuthAgent(config.sessionInfo ? config.sessionInfo : {})
-            ));
+                new Y.eZ.SessionAuthAgent(
+                    (config.sessionInfo && config.sessionInfo.isStarted) ? config.sessionInfo : undefined)
+            )
```
mhyndle: cs nitpick: indentation

mhyndle:
+1 except the inline comment

PR #356, opened 2015-09-30
EZP-24814: Add the hide/reveal buttons in the locations tab

yannickroger:
Link: https://jira.ez.no/browse/EZP-24814

## Description
In the locations tab, for every location add a button allowing to hide or reveal it.

## Screenshot
hide

## Tests
Manual and update of unit test

## TODO
- [x] Add the button
- [x] Unit test
- [x] error handling
- [x] complete CSS
- [x] Cleanup and rebase

### DIFF
--- a/Resources/public/css/theme/views/tabs/locations.css
+++ b/Resources/public/css/theme/views/tabs/locations.css
@@ -18,3 +18,26 @@
 .ez-view-locationviewlocationstabview .ez-locations-list-table {
     background: #fff;
 }
+
+.ez-view-locationviewlocationstabview .ez-locations-hidden-button-reveal:before {
+    content: "\e607";
+}
+
+.ez-view-locationviewlocationstabview .ez-locations-hidden-button-hide:before {
+    content: "\e607";
+    opacity: 0.3;
+}
+
+.ez-view-locationviewlocationstabview .is-switching-visibility {
+    color: #888;
+    font-style: italic;
+}
+
+.ez-view-locationviewlocationstabview .is-switching-visibility:before {
+    display: inline-block;
+    -webkit-animation: spin 0.7s infinite linear;
+            animation: spin 0.7s infinite linear;
+    content: "\E61c";
+    padding: 0 0.2em;
+    opacity: 1;
+}
--- a/Resources/public/js/models/ez-locationmodel.js
+++ b/Resources/public/js/models/ez-locationmodel.js
@@ -98,7 +98,51 @@ YUI.add('ez-locationmodel', function (Y) {
          */
         move: function (options, parentLocationId, callback) {
             options.api.getContentService().moveSubtree(this.get('id'), parentLocationId, callback);
-        }
+        },
+
+        /**
+         * Updates the hidden status of the location
+         *
+         * @protected
+         * @method _updatehidden
+         * @param {Object} options the required for the update
+         * @param {Object} options.api (required) the JS REST client instance
+         * @param {String} hidden `true` or `false` value to be set on the hidden attribute
+         * @param {Function} callback a callback executed when the operation is finished
+         */
+        _updateHidden: function (options, hidden, callback) {
+            var locationUpdateStruct = options.api.getContentService().newLocationUpdateStruct();
+
+            locationUpdateStruct.body.LocationUpdate.hidden = hidden;
+            //Remove the 2 line bellow once EZP-24899 is fixed and update unit test
+            locationUpdateStruct.body.LocationUpdate.sortField = this.get('sortField');
+            locationUpdateStruct.body.LocationUpdate.sortOrder = this.get('sortOrder');
+            options.api.getContentService().updateLocation(this.get('id'), locationUpdateStruct, callback);
+        },
+
+        /**
+         * Hides the location
+         *
+         * @method hide
+         * @param {Object} options the required for the update
+         * @param {Object} options.api (required) the JS REST client instance
+         * @param {Function} callback a callback executed when the operation is finished
+         */
+        hide: function (options, callback) {
+            this._updateHidden(options, 'true', callback);
+        },
+
+        /**
+         * Reverse the hidden status of the location
+         *
+         * @method unhide
+         * @param {Object} options the required for the update
+         * @param {Object} options.api (required) the JS REST client instance
+         * @param {Function} callback a callback executed when the operation is finished
+         */
+        unhide: function (options, callback) {
+            this._updateHidden(options, 'false', callback);
+        },
     }, {
         REST_STRUCT_ROOT: "Location",
         ATTRS_REST_MAP: [
--- a/Resources/public/js/views/services/plugins/ez-visibilityswitcherplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-visibilityswitcherplugin.js
@@ -1,0 +1,129 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-visibilityswitcherplugin', function (Y) {
+    "use strict";
+    /**
+     * Provides the visibility switcher plugin
+     *
+     * @module ez-visibilityswitcherplugin
+     */
+    Y.namespace('eZ.Plugin');
+
+    /**
+     * Location visibility switcher plugin. It sets an event handler to update the
+     * visibility of a location
+     *
+     * In order to use it you need to fire the `switchVisibility` event with two parameters:
+     *  - `location` to be modified
+     *  - `callback` to be called once done
+     *
+     * @namespace eZ.Plugin
+     * @class VisibilitySwitcherPlugin
+     * @constructor
+     * @extends eZ.Plugin.ViewServiceBase
+     */
+    Y.eZ.Plugin.VisibilitySwitcherPlugin = Y.Base.create('visibilityswitcherplugin', Y.eZ.Plugin.ViewServiceBase, [], {
+
+        initializer: function () {
+            this.onHostEvent('*:switchVisibility', this._switchVisibility);
+        },
+
+        /**
+         * Switch the `hidden` status of a given location
+         * @protected
+         * @method _switchVisibility
+         * @param {Object} e switchVisibility event facade
+         *      e.location: the location to be modified
+         *      e.callback: function to run once done
+         */
+        _switchVisibility: function (e) {
+            var location = e.location,
+                locationId = location.get('id'),
+                hidden = location.get('hidden'),
+                hideOptions = {api: this.get('host').get('capi')},
+                that = this,
+                notificationIdentifier = 'location-switch-visibility-' + locationId,
+                notificationStartedMessage = "Hiding: " + locationId;
+
+            if (hidden) {
+                notificationStartedMessage = "Revealing: " + locationId;
+            }
+
+            that._notify(
+                notificationStartedMessage,
+                notificationIdentifier,
+                'started',
+                5
+            );
+
+            if (hidden) {
+                location.unhide(hideOptions, function (error) {
+                    if (error) {
+                        that._notify(
+                            "Error while revealing location " + locationId,
+                            notificationIdentifier,
+                            'error',
+                            0
+                        );
+                    } else {
+                        that._notify(
+                            "Location " + locationId + " has been revealed",
+                            notificationIdentifier,
+                            'done',
+                            5
+                        );
+                    }
+                    e.callback(error);
+                });
+            } else {
+                location.hide(hideOptions, function (error) {
+                    if (error) {
+                        that._notify(
+                            "Error while hiding location " + locationId,
+                            notificationIdentifier,
+                            'error',
+                            0
+                        );
+                    } else {
+                        that._notify(
+                            "Location " + locationId + " has been hidden",
+                            notificationIdentifier,
+                            'done',
+                            5
+                        );
+                    }
+                    e.callback(error);
+                });
+            }
+        },
+
+        /**
+         * Fire 'notify' event
+         *
+         * @method _notify
+         * @protected
+         * @param {String} text the text shown during the notification
+         * @param {String} identifier the identifier of the notification
+         * @param {String} state the state of the notification
+         * @param {Integer} timeout the number of second, the notification will be shown
+         */
+        _notify: function (text, identifier, state, timeout) {
+            this.get('host').fire('notify', {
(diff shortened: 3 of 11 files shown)


Resources/public/js/models/ez-locationmodel.js
```diff
@@ -98,7 +98,27 @@ YUI.add('ez-locationmodel', function (Y) {
          */
         move: function (options, parentLocationId, callback) {
             options.api.getContentService().moveSubtree(this.get('id'), parentLocationId, callback);
-        }
+        },
+
+        //TODO comment
+        _updatehidden: function (options, hidden, callback) {
```
crevillo: _updatevisibility maybe?

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -98,7 +98,27 @@ YUI.add('ez-locationmodel', function (Y) {
...
+            locationUpdateStruct.body.LocationUpdate.hidden = hidden;
+            //TODO these should not be mandatory when updating
+            locationUpdateStruct.body.LocationUpdate.sortField = this.get('sortField');
+            locationUpdateStruct.body.LocationUpdate.sortOrder = this.get('sortOrder');
+            options.api.getContentService().updateLocation(this.get('id'), locationUpdateStruct, callback);
+        },
+
+        //TODO comment
+        hide: function (options, callback) {
+            this._updatehidden(options, 'true', callback);
```
crevillo: true as string? can't be just boolean?
sorry, asking from the unknowledge of the whole thing.

yannickroger: the REST API expects strings 'true' and 'false' passing a boolean results in '1' and '0' which it doesn't accept.

Resources/public/js/models/ez-locationmodel.js
yannickroger: we keep it like that to be consistant with naming used in the API. We will of course not use these terms in error messages and such :)

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -98,7 +98,51 @@ YUI.add('ez-locationmodel', function (Y) {
...
+         * Updates the hidden status of the location
+         *
+         * @protected
+         * @method _updatehidden
+         * @param {Object} options the required for the update
+         * @param {Object} options.api (required) the JS REST client instance
+         * @param {String} hidden `true` or `false` value to be set on the hidden attribute
+         * @param {Function} callback a callback executed when the operation is finished
+         */
+        _updatehidden: function (options, hidden, callback) {
```
dpobel: CS: camelcase

Resources/public/js/views/services/plugins/ez-visibilityswitcherplugin.js
```diff
@@ -0,0 +1,129 @@
...
+     * In order to use it you need to fire the `switchVisibility` event with two parameters:
+     *  - `location` to be modified
+     *  - `callback` to be called once done
+     *
+     * @namespace eZ.Plugin
+     * @class VisibilitySwitcher
+     * @constructor
+     * @extends eZ.Plugin.ViewServiceBase
+     */
+    Y.eZ.Plugin.VisibilitySwitcher = Y.Base.create('visibilityswitcherplugin', Y.eZ.Plugin.ViewServiceBase, [], {
```
dpobel: `visibilitySwitcherPlugin`

Resources/public/js/views/services/plugins/ez-visibilityswitcherplugin.js
```diff
@@ -0,0 +1,129 @@
...
+        _switchVisibility: function (e) {
+            var location = e.location,
+                locationId = location.get('id'),
+                hidden = location.get('hidden'),
+                hideOptions = {api: this.get('host').get('capi')},
+                that = this,
+                notificationIdentifier = 'location-' + locationId,
+                notificationStartedMessage = "Hiding: " + locationId;
+
+            if(hidden) {
```
dpobel: CS: space after `if`

Resources/public/js/views/services/plugins/ez-visibilityswitcherplugin.js
```diff
@@ -0,0 +1,129 @@
...
+         *      e.location: the location to be modified
+         *      e.callback: function to run once done
+         */
+        _switchVisibility: function (e) {
+            var location = e.location,
+                locationId = location.get('id'),
+                hidden = location.get('hidden'),
+                hideOptions = {api: this.get('host').get('capi')},
+                that = this,
+                notificationIdentifier = 'location-' + locationId,
```
dpobel: it should be more specific like `'location-switch-visibility-' + locationId`

Resources/public/js/views/services/plugins/ez-visibilityswitcherplugin.js
```diff
@@ -0,0 +1,129 @@
...
+                'started',
+                5
+            );
+
+            if (hidden) {
+                location.unhide(hideOptions, function (error) {
+                    if (error) {
+                        that._notify(
+                            "Error while revealing location " + locationId,
+                            "unhide-error-" + notificationIdentifier,
```
dpobel: you should use the same notification identifier, so that the notification bar related to this operation on the location is updated.

Resources/public/templates/tabs/locations.hbt
```diff
@@ -25,8 +25,17 @@
                                     <a href="{{ path "viewLocation" id=id languageCode=contentInfo.mainLanguageCode }}">{{ pathString }}</a>
                                 </td>
                                 <td class="ez-table-data-childcount">{{ childCount }}</td>
-                                <td class="ez-table-data-visibility">
-                                    {{#if hidden}}Hidden{{else}}{{#if invisible}}Hidden by superior{{else}}Visible{{/if}}{{/if}}
+                                <td class="ez-table-data-visibility pure-g">
```
dpobel: you don't need a grid here, the visibility followed by the button would do the trick.

Resources/public/templates/tabs/locations.hbt
```diff
@@ -25,8 +25,17 @@
...
+                                    <div class="ez-table-data-visibility-text-box pure-u-1-2">
+                                        {{#if hidden}}Hidden{{else}}{{#if invisible}}Hidden by superior{{else}}Visible{{/if}}{{/if}}
+                                    </div>
+                                    <div class="ezpure-u-1-2">
+                                        <button
+                                            data-location-id="{{id}}"
+                                            class="ez-locations-hidden-button pure-button ez-font-icon ez-button">
+                                            {{#if hidden}}reveal{{else}}{{#if invisible}}{{else}}hide{{/if}}{{/if}}
```
dpobel: `Reveal` and `Hide`. And if the location is invisible, the button should also contain `Hide`

dpobel:
In addition to the inline comments, right after clicking on the visibility change button while the operation is done, the button should be disabled and it should be possible to style it (ie it gets a new class) so that the user knows its click was taken into account.

Resources/public/css/theme/views/tabs/locations.css
```diff
@@ -18,3 +18,20 @@
 .ez-view-locationviewlocationstabview .ez-locations-list-table {
     background: #fff;
 }
+
+.ez-view-locationviewlocationstabview .ez-locations-hidden-button {
+    width: 9em;
```
dpobel: why do you set a width ?

Resources/public/css/theme/views/tabs/locations.css
```diff
@@ -18,3 +18,20 @@
...
+
+.ez-view-locationviewlocationstabview .ez-locations-visibility-pending {
+    color: #888;
+    font-style: italic;
+}
+
+.ez-view-locationviewlocationstabview .ez-locations-visibility-pending:before {
+    display: inline-block;
+    -webkit-animation: spin 0.7s infinite linear;
+    animation: spin 0.7s infinite linear;
```
dpobel: cs: indent so that the `:` are aligned

Resources/public/js/views/tabs/ez-locationviewlocationstabview.js
```diff
@@ -98,7 +101,40 @@ YUI.add('ez-locationviewlocationstabview', function (Y) {
...
+         * @method _fireSwitchVisibility
+         * @protected
+         * @param {EventFacade} e
+         */
+        _fireSwitchVisibility: function (e) {
+            var locationId = e.target.getAttribute('data-location-id');
+
+            Y.Array.every(this.get('locations'), function (location) {
+                if(location.get('id') === locationId) {
+                    e.target.set('disabled', true).addClass('ez-locations-visibility-pending');
```
dpobel: `is-switching-visibility`

Resources/public/js/views/tabs/ez-locationviewlocationstabview.js
```diff
@@ -98,7 +101,40 @@ YUI.add('ez-locationviewlocationstabview', function (Y) {
...
+        },
+
+        /**
+         * Fire the `switchVisibility` event
+         *
+         * @method _fireSwitchVisibility
+         * @protected
+         * @param {EventFacade} e
+         */
+        _fireSwitchVisibility: function (e) {
```
dpobel: it does more than just firing the event now, a better name would be  `_switchVisibility`

yannickroger:
Ready for review, ping @dpobel @mhyndle

Tests/js/views/services/plugins/assets/ez-visibilityswitcherplugin-tests.js
```diff
@@ -0,0 +1,179 @@
...
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-visibilityswitcherplugin-tests', function (Y) {
+    var tests,
+        registerTest,
+        Assert = Y.Assert,
+        Mock = Y.Mock;
+
+    tests = new Y.Test.Case({
+        name: "eZ User Load Plugin event tests",
```
mhyndle: incorrect name of test case

Tests/js/views/tabs/ez-locationviewlocationstabview.html
```diff
@@ -40,7 +41,7 @@
         filter: loaderFilter,
         modules: {
             "ez-locationviewlocationstabview": {
-                requires: ['ez-locationviewtabview', 'ez-asynchronousview', 'fake-models'],
+                requires: ['ez-locationviewtabview', 'ez-asynchronousview', 'fake-models', 'array-extras'],
```
mhyndle: I think `array-extras` is not needed, or I missed something?

mhyndle:
In `ez-visibilityswitcherplugin` as far as I can see the callback function is called only if changing visibility will finish successfully, but if it will fail, than callback is not called. So if the button after click is changed to disabled and the visbility change will fail, we will get the error notification but the button will remain disabled.
Maybe calling callback function no matter if the visibility switch fails or not would be the solution. Or create in the view another function for reverting the disabled state and pass it to the `switchVisibility` event, so it can be called if changing visibility fails.

yannickroger:
Feedback by @mhyndle taken into account.

Resources/public/js/views/services/plugins/ez-visibilityswitcherplugin.js
```diff
@@ -0,0 +1,131 @@
...
+            if (hidden) {
+                location.unhide(hideOptions, function (error) {
+                    if (error) {
+                        that._notify(
+                            "Error while revealing location " + locationId,
+                            notificationIdentifier,
+                            'error',
+                            0
+                        );
+                        e.callback(error);
```
mhyndle: It's duplicated in `else` so I guess it can be called just after `if () { ... } else { ... }` so it will be single call.

Resources/public/js/views/services/plugins/ez-visibilityswitcherplugin.js
```diff
@@ -0,0 +1,131 @@
...
+            } else {
+                location.hide(hideOptions, function (error) {
+                    if (error) {
+                        that._notify(
+                            "Error while hiding location " + locationId,
+                            notificationIdentifier,
+                            'error',
+                            0
+                        );
+                        e.callback(error);
```
mhyndle: The same as above https://github.com/ezsystems/PlatformUIBundle/pull/356/files#r41358882

mhyndle:
+1, besides inline comments

yannickroger:
@mhyndle Feedback taken into account.
Also I had a bug in the call to the _refresh method, so I fixed it and added a unit test.

Resources/public/css/theme/views/tabs/locations.css
```diff
@@ -18,3 +18,25 @@
...
+    color: #888;
+    font-style: italic;
+}
+
+.ez-view-locationviewlocationstabview .is-switching-visibility:before {
+    display: inline-block;
+    -webkit-animation: spin 0.7s infinite linear;
+            animation: spin 0.7s infinite linear;
+    content: "\E61c";
+    padding: 0 0.2em;
```
dpobel: you should probably force the opacity to 1 here otherwise the loading icon on the hide button will get 0.3 as opacity.

Tests/js/views/tabs/assets/ez-locationviewlocationstabview-tests.js
```diff
@@ -222,6 +223,153 @@ YUI.add('ez-locationviewlocationstabview-tests', function (Y) {
...
+
+        "Should provide the location to the `switchVisibility` event when tapping on the hide/reveal button": function () {
+            var that = this,
+                location = {},
+                button;
+
+            this.view.render();
+            this.view.set('active', true);
+
+            this.view._refresh = function () {
```
dpobel: monkey patching is bad and not needed here. To check if the view is refreshed, you just have to check that the `loadLocations` event is fired.

Tests/js/views/tabs/assets/ez-locationviewlocationstabview-tests.js
```diff
@@ -222,6 +223,153 @@ YUI.add('ez-locationviewlocationstabview-tests', function (Y) {
...
+
+        "Should reactivate the button after error after tapping on the hide/reveal button": function () {
+            var that = this,
+                location = {},
+                button;
+
+            this.view.render();
+            this.view.set('active', true);
+
+            this.view._refresh = function () {
```
dpobel: same here

yannickroger:
@dpobel done :)

dpobel:
+1

PR #357, opened 2015-10-01
EZP-24815: change main location of content

mhyndle:
Jira: https://jira.ez.no/browse/EZP-24815

## Description
The goal of this PR is to allow user to change main location of content.
On the locations list of the content (Locations tab) the new columns is added named "Main". It contains radio button which is checked if location is main location of content.
Clicking on not-checked radio will open confirm box that will ask user for confirm the main location change. If user confirm then main location is changed to the selected one and the view location list is reloaded.

## Tasks
- [x] implementation
- [x] tests
 - [x] manual tests
 - [x] unit tests

## Notes
Plugin is using `updateContentMetadata` method of ez-js-rest-client's contentService and it seems that I encounter REST API bug, that throws error about not-existing location when the passed location actually exists, bug described here: https://jira.ez.no/browse/EZP-24901
So until mentioned bug is solved, changing main location for content is not possible.

### Related PR for ez-js-rest-client
I have created corresponding PR for this one for ez-js-rest-client: https://github.com/ezsystems/ez-js-rest-client/pull/67
That patch adds `setMainLocation` for `ContentUpdateMetadataStruct` and removes required language param from `ContentService.newContentMetadataUpdateStruct` method.

### DIFF
--- a/Resources/public/js/views/services/plugins/ez-contentsetmainlocationplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-contentsetmainlocationplugin.js
@@ -45,7 +45,8 @@ YUI.add('ez-contentsetmainlocationplugin', function (Y) {
                     title: "Are you sure you want to set this location as main location of content?",
                     confirmHandler: Y.bind(function () {
                         this._setMainLocation(e.locationId, e.afterSetMainLocationCallback);
-                    }, this)
+                    }, this),
+                    cancelHandler: Y.bind(e.cancelSetMainLocationCallback, this)
                 }
             });
         },
@@ -73,6 +74,8 @@ YUI.add('ez-contentsetmainlocationplugin', function (Y) {
             );
 
             content.setMainLocation({api: capi}, locationId, function (error, response) {
+                callback();
+
                 if (error) {
                     that._notify(
                         "Changing main location of '" + content.get('name') + "' failed",
@@ -89,7 +92,6 @@ YUI.add('ez-contentsetmainlocationplugin', function (Y) {
                     'done',
                     5
                 );
-                callback();
             });
 
         },
--- a/Resources/public/js/views/services/plugins/ez-contentsetmainlocationplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-contentsetmainlocationplugin.js
@@ -61,13 +61,9 @@ YUI.add('ez-contentsetmainlocationplugin', function (Y) {
         _setMainLocation: function (locationId, callback) {
             var service = this.get('host'),
                 capi = service.get('capi'),
-                contentService = capi.getContentService(),
                 content = service.get('content'),
                 notificationIdentifier = 'set-main-location-' + content.get('id') + '-' + locationId,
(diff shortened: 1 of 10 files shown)


mhyndle:
Ready for review

Resources/public/js/views/services/plugins/ez-contentsetmainlocationplugin.js
```diff
@@ -0,0 +1,128 @@
...
+         * @param {String} locationId
+         * @param {Function} callback
+         */
+        _setMainLocation: function (locationId, callback) {
+            var service = this.get('host'),
+                capi = service.get('capi'),
+                contentService = capi.getContentService(),
+                content = service.get('content'),
+                notificationIdentifier = 'set-main-location-' + content.get('id') + '-' + locationId,
+                updateStruct = contentService.newContentMetadataUpdateStruct(content.get('mainLanguageCode')),
```
dpobel: mainLanguageCode ? the content metadata update struct does not need that but indeed contentService.newContentMetadataUpdateStruc has this parameter and is documented with it. You should remove this parameter here and also fix the JS REST Client

Resources/public/js/views/services/plugins/ez-contentsetmainlocationplugin.js
```diff
@@ -0,0 +1,128 @@
...
+                that = this;
+
+            updateStruct.body.ContentUpdate.MainLocation = {_href: locationId};
```
dpobel: here as well, the JS REST client should be improved to provide a `setMainLocation` (it already has a `setSection` https://github.com/ezsystems/ez-js-rest-client/blob/master/src/structures/ContentMetadataUpdateStruct.js#L27)

dpobel: and the update itself should be done in the content model

Resources/public/templates/tabs/locations.hbt
```diff
@@ -28,6 +29,10 @@
                                 <td class="ez-table-data-visibility">
                                     {{#if hidden}}Hidden{{else}}{{#if invisible}}Hidden by superior{{else}}Visible{{/if}}{{/if}}
                                 </td>
+                                <td class="ez-table-data-main-location">
+                                    <input type="radio" value="1" class="ez-main-location-radio"
+                                        data-location-id="{{ id }}" {{#if isMainLocation}}checked="checked"{{/if}}>
```
dpobel: nitpick: `data-location-id="{{ id }}"{{#if isMainLocation}} checked{{/if}}>`

andrerom:
+1 with the suggested improvements.

mhyndle:
Last feedback taken into account. There is also new PR for ez-js-rest-client which adds `setMainLocation` method for `ContentUpdateMetadataStruct`: https://github.com/ezsystems/ez-js-rest-client/pull/67

dpobel:
Kind of the same thing as in https://github.com/ezsystems/PlatformUIBundle/pull/356#issuecomment-145511582, while the main location is updated the radio inputs should be disabled

mhyndle:
Last comment taken into account

dpobel:
+1

PR #358, opened 2015-10-01
Admin menu fixes

andrerom:
Small fixes to make sure terminology is same on dashboard as well as in menu, also add the missing Roles menu option.


dpobel:
+1

yannickroger:
+1

PR #360, opened 2015-10-04
Implemented ContentCreate/UserCreate gateway

bdunogier:
When a ContentCreate for a User is POSTed, it is handled by the ContentUserGateway.

It creates the user, using the User API, from the request, and returns the created user.
No controller is used in the process, it all happens in the event.

### DIFF
--- a/Rest/ContentUserGateway.php
+++ b/Rest/ContentUserGateway.php
@@ -1,0 +1,155 @@
+<?php
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\PlatformUIBundle\Rest;
+
+use eZ\Publish\API\Repository\Repository;
+use eZ\Publish\Core\Repository\Values\User\UserCreateStruct;
+use eZ\Publish\Core\REST\Common\Input\Dispatcher;
+use eZ\Publish\Core\REST\Common\Message;
+use eZ\Publish\Core\REST\Server\Values\CreatedContent;
+use eZ\Publish\Core\REST\Server\Values\RestContent;
+use eZ\Publish\Core\REST\Server\Values\RestContentCreateStruct;
+use eZ\Publish\Core\REST\Server\View\AcceptHeaderVisitorDispatcher;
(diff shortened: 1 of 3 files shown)


bdunogier:
Ping @dpobel @andrerom.

Rest/ContentUserGateway.php
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    public function onKernelRequest(GetResponseEvent $event)
+    {
+        $request = $event->getRequest();
+        if (!$request->attributes->get('is_rest_request')) {
+            return;
+        }
+
+        if (!$contentTypeHeaderValue = $request->headers->get('content-type')) {
```
andrerom: we tend to prefer parentheses for these right?
`if (!($contentTypeHeaderValue = $request->headers->get('content-type'))) {`

Rest/ContentUserGateway.php
```diff
@@ -0,0 +1,158 @@
...
+        if (!$message->body) {
+            return;
+        }
+        $result = $this->restInputDispatcher->parse($message);
+        if (!$result instanceof RestContentCreateStruct) {
+            return;
+        }
+
+        // Not a user
+        if (($createData = $this->mapContentCreateToUserCreate($result)) === false) {
```
andrerom: nitpick, but  `$userCreateData`

andrerom:
+1, seems like an ok approach

Rest/ContentUserGateway.php
bdunogier: You're right.

bdunogier:
I have been able to test it, and it "works", to some extent.

Problem A: publish will generate an error: creating will work, but will also publish directly (that's what `UserService::createUser()` does. We could remove the "Publish" button, change the label of "Save" to "Publish", and make it exit... @dpobel ?

Problem B: save first, then publish won't work, as it will try to _update_ the user. I have only implemented support for _create_.

bdunogier:
Agreed with @dpobel that it was "better", and could go in as is. Will be improved when needed.

bdunogier:
ezrobot is drunk, I've updated this hours ago. Merging :-)

PR #361, opened 2015-10-04
Fix preview height, and be relative to viewport

andrerom:
Could not find any tests expecting the X/Y values, so should theoretically pass.

As this feature is completely broken w/o this change, I would argue this is an ok approach for now, and somewhat more reliable then calculating height anyway. Would actually be a benefit if we could rely on % width as well, but that is another topic.

review ping @dpobel


dpobel:
+1

mhyndle:
+1

PR #364, opened 2015-10-06
Fix EZP-24892: The relation tab always displays table even without any relation

yannickroger:
Link: https://jira.ez.no/browse/EZP-24892

## Description
When an object has no relation, an empty table was displayed.

Also, while at it, I fixed a small display bug in a separate commit.

## Screenshot
no_content

## Tests
Manual test

### DIFF
--- a/Resources/public/templates/tabs/relations.hbt
+++ b/Resources/public/templates/tabs/relations.hbt
@@ -10,6 +10,7 @@
     </p>
     {{else}}
     <div class="ez-relations-box-list">
+        {{#if relatedContents}}
         <table class="ez-relations-box-table ez-asynchronousview pure-table pure-table-bordered">
         <thead>
             <tr>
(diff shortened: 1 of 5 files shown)


yannickroger:
ping @dpobel @mhyndle

mhyndle:
+1

Resources/public/templates/tabs/relations.hbt
```diff
@@ -36,6 +37,9 @@
             {{/each}}
         </tbody>
         </table>
+        {{else}}
+            <p>This content has no related content</p>
```
dpobel: this deserves to be styled a bit, I would say it should be in italic and you should also removed the default margin.

dpobel:
@ezrobot what's wrong with you ?

yannickroger:

dpobel:
are you sure @yannickroger ? you did not add a class on the paragraph element for the message, so for sure the text won't be in italic. And please add screenshot, that's a lot easier to see the actual change.

yannickroger:
@dpobel Sorry… I forgot the Italic part of your remark. It's now taken care of.

dpobel:
+1 (I guess Travis will agree and @ezrobot please shut up :))

PR #365, opened 2015-10-07
EZP-21821- BDD for edit view for ezuser fields

miguelcleverti:
BDD test for the fields view of ezusers.
PS: also minor doc comment fix of the name of the files

### DIFF
--- a/Features/Context/PlatformUI.php
+++ b/Features/Context/PlatformUI.php
@@ -25,6 +25,7 @@ class PlatformUI extends Context
     use SubContext\Copy;
     use SubContext\Remove;
     use SubContext\Role;
+    use SubContext\Users;
 
     /**
      * PlatformUI relative URL path.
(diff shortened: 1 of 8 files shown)


miguelcleverti:
@andrerom @joaoinacio @pcardiga @bdunogier review pls
ps: (#360) I don't think ezrobot was drunk ,the line [56 in ContentUserGateway.php] (https://github.com/ezsystems/PlatformUIBundle/blob/master/Rest/ContentUserGateway.php#L56) still looks wrong. Shouldn't it be `if (strtolower($mediaType) !== 'application/vnd.ez.api.contentcreate') {`. I would fix it in this PR but it would be really out of scope IMO

miguelcleverti:
@andrerom @joaoinacio @pcardiga @bdunogier

andrerom:
+1 Passing? :)

miguelcleverti:
@andrerom Yes, passing :+1:

PR #366, opened 2015-10-07
EZP-24916: Platform UI Preview broken

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24916

# Description

This patch fixes the display of the preview. With it, we also handle better handle the preview on a device bigger than the one used by the editor.

Screencast: http://youtu.be/NeOrEjiODWA

# Tests

unit tests + manual tests

### DIFF
--- a/Resources/public/css/theme/views/editpreview.css
+++ b/Resources/public/css/theme/views/editpreview.css
@@ -5,34 +5,61 @@
 
 .ez-view-editpreviewview {
     background: #f3f3f3;
-}
-
-.ez-preview-content .ez-preview-hide {
     color: #555;
 }
 
-.ez-preview-content .ez-loader.is-loading {
-    background: #000;
-    background: rgba(0, 0, 0, 0.8);
+.ez-view-editpreviewview .ez-preview-iframe-wrapper:before {
+    transition: opacity ease 0.3s;
+    opacity: 0;
+
+    position: absolute;
+    display: block;
+    content: "";
+    margin-top: 3em;
+    width: 2em;
+    height: 2em;
+    left: calc(50% - 1em);
+    border: 0.5em solid #999;
+    border-left: 0.5em solid transparent;
+    border-radius: 50%;
 }
 
-.ez-preview-content .preview-device.mobile {
-    width: 322px;
-    height: 478px;
-    padding: 118px 30px 118px 31px;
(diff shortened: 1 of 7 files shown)


andrerom:
you'll also need to set the hight to size of screen right? otherwise it will inherit?

dpobel:
well, you are right we have size issues, if we don't set the height like in the current patch, the ez-editpreviewview-container element will take the height needed to display its content but this can be more than the viewport height (and then the bottom of the preview can not be reached) or this can be less than the viewport (and in this case that's ugly). The thing is I can force this element to be the exact viewport height but this leads to the question I already ask several times, how do we handle the case where the preview is supposed to be wider or higher than the actual viewport size ?

dpobel:
ping @StephaneDiot @yannickroger @mhyndle @andrerom

mhyndle:
+1

Resources/public/css/theme/views/editpreview.css
```diff
@@ -5,34 +5,61 @@
...
     background: #f3f3f3;
+    color: #555;
 }
 
-.ez-preview-content .ez-preview-hide {
-    color: #555;
+.ez-view-editpreviewview .ez-preview-iframe-wrapper:before {
+    transition: opacity ease 0.3s;
+    opacity: 0;
+
```
andrerom: On purpose?

mhyndle: I guess it's for separate the part that will be changed with transition: https://github.com/ezsystems/PlatformUIBundle/pull/366/files#diff-e5d16b0809fbbf1b577e2b9454cadbf7R27

andrerom:
+1 looks nice :)

Resources/public/css/theme/views/editpreview.css
andrerom: Ok

dpobel: yep on purpose, just to show/hide the loading animation with a nice transition on the opacity.

andrerom: It was more the empty new line I was commenting on, but yes :)

dpobel: I had my glasses but I need more coffee it seems :)

PR #368, opened 2015-10-08
EZP-24353 Updating author field to empty does not show empty message

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-24353

## Description

When editing a already filled content with author field to empty the "is empty" message was not dislpayed.
The problem wame from the fact that an empty email is considered as valid ( when the field is not requiered ), and that it was the only check in the _getFieldValue() of the field to check if the fieldValue has a value. So when giving an empty one the value was filled by an object with 3 empty properties (name, email, id), which wasn't considered as an empty field for the view.

Solved by adding an _hasContent() check to the author list in the _getFieldValue()

## Tests

- [x] Unit tested

### DIFF
--- a/Resources/public/js/views/fields/ez-author-editview.js
+++ b/Resources/public/js/views/fields/ez-author-editview.js
@@ -654,14 +654,16 @@ YUI.add('ez-author-editview', function (Y) {
         _getFieldValue: function () {
             var value = [];
 
-            this._authorList.each(function (author) {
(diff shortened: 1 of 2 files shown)


StephaneDiot:
ping @dpobel @mhyndle  @yannickroger

mhyndle:
+1

dpobel:
+1

PR #370, opened 2015-10-09
Fix EZP-24928: app is correctly navigating to parent location after removing content

mhyndle:
Jira: https://jira.ez.no/browse/EZP-24928

## Description
After sending content to trash the app should navigate to the parent location of content which was broken. This PR solves it, and after sending content to trash the app is navigating to the parent location.
The reason of the problem was probably changed structure of the `path` attribute as it now contains eZ.Location objects.

## Tests
manual + unit tests


mhyndle:
ready for review

dpobel:
+1

yannickroger:
+1

PR #371, opened 2015-10-09
POC [Not For Merge] - New Behat waiting mechanism

miguelcleverti:
Branch to test the PlatformUI Behat new waiting mechanism approach with travis.
The code still needs major clean up and refactors.
The idea behind is to wait for the desired element instead of waiting for the page to finish loading everything. Using a `spin()` function, that basically makes the test wait while a certain condition is not met(or a time out is reached), even if an exception is thrown. This causes a problem with negative verification tests, when an element is not supposed to be found this causes the tests to wait the full time out value, but still might be a better approach.

### DIFF
--- a/Features/Context/PlatformUI.php
+++ b/Features/Context/PlatformUI.php
@@ -78,7 +78,7 @@ class PlatformUI extends Context
                 // do nothing
             }
 
-            usleep(100 * 3000); //100ms
+            usleep(1000 * 300); //300ms
         }
 
         $backtrace = debug_backtrace();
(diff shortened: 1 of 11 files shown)


miguelcleverti:
@andrerom ready to run on travis

andrerom:
@miguelcleverti for the first robot cs issue, you'll need to rebase on master I think

bdunogier:
Do you think you could briefly described the idea of the change in the PR's description ?

Features/Context/PlatformUI.php
```diff
@@ -59,28 +61,31 @@ class PlatformUI extends Context
...
-            '.ez-view-treeactionview.is-expanded .ez-view-treeview:not(.is-tree-loaded)',
-            '.ez-view-universaldiscoverybrowseview .ez-view-treeview:not(.is-tree-loaded)',
-            '.is-tree-node-loading',
-            // contenttype menu
-            '.ez-view-createcontentactionview.is-expanded:not(.is-contenttypeselector-loaded)',
-        );
-        $loadingSelector = implode(',', $loadingClasses);
-        while ($page->find('css', $loadingSelector) != null) {
-            usleep(100 * 1000); // 100ms
+        for ($i = 0; $i < $wait; ++$i) {
```
dpobel: why not use `$session::wait()` ?

miguelcleverti:
@bdunogier I updated the description

Features/Context/PlatformUI.php
miguelcleverti: @dpobel the reason we do not use the `$session::wait` it's because AFAIK it does not work with the exceptions, but the main focus of this PR is the functionality if it works then I will improve the code overall, maybe the `$session::wait` might be used I will investigate further after we check how it behaves on travis.

andrerom: This might be useful, it most likely does the exact same thing, just with a javascript check instead of xpath via mink. So might be faster.

@miguelcleverti For this to work, assuming it is still in Behat 3.0 / Mink 1.6. As we don't have jQuery, and should not hard code use of framework anyway, you'll need to use the DOM. Modern DOM supported by all browsers we support is rather straight forward using: getElementById, getElementsByTagName, getElementsByClassName, or querySelectorAll *(for more advance expressions, it can do anything, but the once above is more efficient if they fit)*.

miguelcleverti: @andrerom looks good to me, will improve it, so can we try to run it on travis because the logic should not change much and that's what this POC is about.

andrerom:
Tried this out here, unfortunately not working afaik, is something not pushed maybe?
https://travis-ci.org/ezsystems/ezplatform/jobs/85203095

miguelcleverti:
@andrerom the only reason this should be failing is if the timeout was not long enough, anyway I changed it a bit and made it longer(30s) and also made the time out error more descriptive, could you run the tests again?

andrerom:
> could you run the tests again?

I'm about to leave for Symfony live berlin, so that will not work. You can do it yourself by copying what is done in the ezplatform PR, add composer.json repository info for your repo, and open new PR against master with that to test, then every push to that branch will trigger new test and you will thus not rely on me.

miguelcleverti:
> could you run the tests again?
>> I'm about to leave for Symfony live berlin for the next two days, so that will not work. You can do it yourself by copying what is done in my the ezplatform PR, add composer.json repository info for your repo, and open new PR against master with that to test, then every push to that branch will trigger new test and you will thus not rely on me.

okay will do thanks :smile:

miguelcleverti:
closed in favor of #416

PR #372, opened 2015-10-09
EZP-24716 Selection field validation display is wrong while editing

StephaneDiot:
jira : https://jira.ez.no/browse/EZP-24716

## Description 

Selection field errors display was wrong while editing a content. The problem came from the position of the validate() method, which couldn't covver every use cases.

## Tests

- [x] manual tests

### DIFF
--- a/Resources/public/js/views/fields/ez-selection-editview.js
+++ b/Resources/public/js/views/fields/ez-selection-editview.js
@@ -64,6 +64,7 @@ YUI.add('ez-selection-editview', function (Y) {
                 this._set('values', this._getSelectedTextValues());
             });
             this.after('valuesChange', this._uiSyncSelectionValues);
+            this.after('valuesChange', this._validateAddedRemovedValues);
             this.after('showSelectionUIChange', this._uiShowSelectionList);
         },
 
@@ -228,7 +229,6 @@ YUI.add('ez-selection-editview', function (Y) {
          */
         _removeValue: function (e) {
             this._removeSelection(e.target.getAttribute('data-text'), e.target);
-            this.validate();
         },
 
         destructor: function () {
@@ -326,6 +326,20 @@ YUI.add('ez-selection-editview', function (Y) {
                     node = selectionValues.one('.ez-selection-value[data-text="' + e.removed + '"]');
                 }
                 node.remove(true);
+            }
+        },
+
+        /**
+         * Check if there is a value that has been added or removed with the event facade.
+         * If it is, then call validate().
+         *
+         * @method _validateAddedRemovedValues
+         * @protected
+         * @param {Object} e event facade
+         */
+        _validateAddedRemovedValues: function (e) {
+            if ( e.added || e.removed ) {
+                this.validate();
             }
         },
 
--- a/Tests/js/views/fields/assets/ez-selection-editview-tests.js
+++ b/Tests/js/views/fields/assets/ez-selection-editview-tests.js
@@ -3,7 +3,7 @@
  * For full copyright and license information view LICENSE file distributed with this source code.
  */
 YUI.add('ez-selection-editview-tests', function (Y) {
-    var viewTest, registerTest, getFieldTest, uiFunctionalTest,
+    var viewTest, registerTest, getFieldTest, uiFunctionalTest, validateAddedValuesTest,
         options = [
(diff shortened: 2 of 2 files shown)


StephaneDiot:
ping @dpobel @yannickroger @mhyndle

Resources/public/js/views/fields/ez-selection-editview.js
```diff
@@ -64,6 +64,7 @@ YUI.add('ez-selection-editview', function (Y) {
                 this._set('values', this._getSelectedTextValues());
             });
             this.after('valuesChange', this._uiSyncSelectionValues);
+            this.after('valuesChange', this.checkValueChange);
```
dpobel: the name of the method is bad because if I just read this line, I have absolutely no clue of what it does.

Resources/public/js/views/fields/ez-selection-editview.js
```diff
@@ -330,6 +330,18 @@ YUI.add('ez-selection-editview', function (Y) {
         },
 
         /**
+         * Check if there is a value that has been added or removed with the event facade.
+         * If it is, then call validate().
+         *
+         * @method checkValueChange
+         */
+        checkValueChange: function (e) {
```
dpobel: this should be a protected method (tag in the jsdoc + `_`) and maybe called `_validateAddedRemovedValues`

dpobel:
no test ?

StephaneDiot:
updated @dpobel

Tests/js/views/fields/assets/ez-selection-editview-tests.js
```diff
@@ -575,9 +575,101 @@ YUI.add('ez-selection-editview-tests', function (Y) {
         },
     });
 
+    regressionTest = new Y.Test.Case({
```
mhyndle: Variable for test case should point somehow what is it related to - regression doesn't say much ;)

Tests/js/views/fields/assets/ez-selection-editview-tests.js
```diff
@@ -575,9 +575,101 @@ YUI.add('ez-selection-editview-tests', function (Y) {
...
+        name: "eZ Selection View test for UI behavior",
```
mhyndle: Duplicated test case name

Tests/js/views/fields/assets/ez-selection-editview-tests.js
```diff
@@ -575,9 +575,101 @@ YUI.add('ez-selection-editview-tests', function (Y) {
...
+                contentType: this.contentType
+            });
+        },
+
+        tearDown: function () {
+            this.view.destroy();
+            delete this.view;
+        },
+
+        "EZP-24716: Should validate after adding a selection": function () {
```
mhyndle: Jira issue ID is not needed I think

Tests/js/views/fields/assets/ez-selection-editview-tests.js
```diff
@@ -575,9 +575,101 @@ YUI.add('ez-selection-editview-tests', function (Y) {
...
+            this.view.set('fieldDefinition', this._getFieldDefinition(true, false, options));
+            this.view.render();
+            this.view.set('active', true);
+            this.view.set('showSelectionUI', true);
+
+            this.view.validate();
+            Y.Assert.isTrue(!!this.view.get('errorStatus'), "An error should be detected because field is requiered and empty");
+
+            option = container.one('.ez-selection-options li');
+            option.simulateGesture('tap', function () {
```
mhyndle: If I understand the fix, it should call validate on changing `values` attrib (to be precise on add or remove). So I guess simulating user gestures is not needed here as you should just set the value for `values` attrib, then set it again to changed value and just check if the field was validated properly. Please verify it because I may not understand the mechanism.

dpobel: just did almost the same remark to @StephaneDiot. And actually it depends on if you base your test on the bug or on the patch. Since it's a regression test (so written during a bug fix), I guess this approach is OK if the very small refactoring in the patch (the `validate` call moved from `_removeValue` to the `valuesChange` event handler) is covered somehow.

dpobel:
+1

mhyndle:
+1

PR #373, opened 2015-10-13
Fix EZP-24926: UWD Content Tree becomes unusable with a big tree

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24926

# Description

when the tree is too big, the UDW is completely broken. This was initially [the case only in Firefox](https://jira.ez.no/browse/EZP-24086) but lately the bug also appears in Chrome.

udw_tree

# Tests

manual tests (also works well in IE11)


andrerom:
+1

yannickroger:
+1

PR #374, opened 2015-10-13
Fix EZP-24958: Impossible to update BinaryFile fields

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24958

# Description

It impossible to update BinaryFile fields in PlatformUI. This is happening because the field value of such fields has a `url` property when loading the content but this property should not be in the field value when storing a new file in the field.

# Tests

manual + unit test

### DIFF
--- a/Resources/public/js/views/fields/ez-binaryfile-editview.js
+++ b/Resources/public/js/views/fields/ez-binaryfile-editview.js
@@ -141,6 +141,21 @@ YUI.add('ez-binaryfile-editview', function (Y) {
                 "binaryfile": this.get('file'),
             };
         },
+
+        /**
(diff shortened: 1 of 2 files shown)


bdunogier:
Could be fixed in the REST API's Binary FieldTypeProcessor. `url` is only here for BC, and shouldn't crash the input.

I have a short fix for it if interested;

andrerom:
As it is there for bc, I guess both changes would make sense.

andrerom:
+1

dpobel:
ping @yannickroger @mhyndle

mhyndle:
+1

yannickroger:
+1

PR #375, opened 2015-10-13
Fix EZP-24185: Assigned Content count not updated after assigning a section

yannickroger:
Link: https://jira.ez.no/browse/EZP-24185

## Description
When assigning a content to a section, the content count was not updated. The fix refreshed the whole view (not the whole page) involving only one ajax request.

## Test
Manual and unit tests

### DIFF
--- a/Resources/public/js/views/serverside/ez-sectionserversideview.js
+++ b/Resources/public/js/views/serverside/ez-sectionserversideview.js
@@ -41,6 +41,7 @@ YUI.add('ez-sectionserversideview', function (Y) {
          */
         _pickSubtree: function (e) {
             var button = e.target,
+                refreshView = Y.bind(this._refreshView, this),
                 unsetLoading = Y.bind(this._uiUnsetAssignSectionLoading, this, button);
 
             e.preventDefault();
@@ -53,7 +54,7 @@ YUI.add('ez-sectionserversideview', function (Y) {
                     data: {
                         sectionId: button.getAttribute('data-section-rest-id'),
                         sectionName: button.getAttribute('data-section-name'),
-                        afterUpdateCallback: unsetLoading,
+                        afterUpdateCallback: refreshView,
                     },
                 },
             });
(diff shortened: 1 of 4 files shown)


yannickroger:
Ready for reviews @dpobel @mhyndle @StephaneDiot

mhyndle:
+1

Resources/public/js/views/serverside/ez-sectionserversideview.js
```diff
@@ -81,6 +81,12 @@ YUI.add('ez-sectionserversideview', function (Y) {
          */
         _uiUnsetAssignSectionLoading: function (button) {
             button.removeClass('is-loading').set('disabled', false);
+
+            /**
+             * Fired when the view needs to be refreshed
+             * @event refreshView
+             */
+            this.fire('refreshView');
```
dpobel: this does not match the name of the method `_uiUnsetAssignSectionLoading`.

Resources/public/js/views/services/ez-sectionserversideviewservice.js
```diff
@@ -143,5 +145,18 @@ YUI.add('ez-sectionserversideviewservice', function (Y) {
...
+
+        /**
+         * Refreshes the section view
+         *
+         * @method _refreshView
+         * @protected
+         * @param {EventFacade} e
+         */
+        _refreshView: function (e) {
+            this.load(Y.bind(function () {
```
dpobel: you should probably set the app `loading` attribute to true so that the user has a visual feedback

Resources/public/js/views/serverside/ez-sectionserversideview.js
dpobel: also the event will be fired even when the user clicks on Cancel in the UDW

yannickroger:
@dpobel feedback taken into account.

dpobel:
+1 if Travis agrees :)

PR #376, opened 2015-10-13
Fix EZP-22567: Incorrect display of the too high field edit views in Firefox

dpobel:
JIRA: https://jira.ez.no/browse/EZP-22567

# Description

This is a Pure Grid bug reported in https://github.com/yahoo/pure/issues/436. It comes from the fact that in Firefox Pure Grids fallbacks to an `inline-block` based grid system even though Firefox supports flexbox for a while now. This patch is a workaround so that flexbox is used when available.

More generally this brings [the question of using Pure Grid](https://jira.ez.no/browse/EZP-24963), it seems we only support browsers where we can use flexbox so at some point we should/could get rid of Pure Grids.

# Tests

manual test

### DIFF
--- a/Resources/config/css.yml
+++ b/Resources/config/css.yml
@@ -89,6 +89,7 @@ system:
                 - 'bundles/ezplatformui/css/modules/table-data.css'
                 - 'bundles/ezplatformui/css/modules/breadcrumbs.css'
                 - 'bundles/ezplatformui/css/modules/yui-calendar.css'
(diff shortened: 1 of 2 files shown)


andrerom:
+1 (and also +1 on discussion on dropping it)

dpobel:
ping @mhyndle @yannickroger

mhyndle:
+1

yannickroger:
+1

PR #377, opened 2015-10-13
Fix EZP-24625: impossible to edit the content in another language than main language code

mhyndle:
Jira: https://jira.ez.no/browse/EZP-24625

## Description
When editing the content in another language than main language the app has been notifying about error after trying to publish the content. It was specific for contents that have non-translatable fields, for example default folder's `show_children` attribute.
This PR solves it, checking for each field if it's translatable or not. If yes, languageCode for that field is set to given one, otherwise languageCode is set to mainLanguageCode of the content.

## Tests
manual + unit tests

### DIFF
--- a/Resources/public/js/views/ez-contenteditformview.js
+++ b/Resources/public/js/views/ez-contenteditformview.js
@@ -55,7 +55,8 @@ YUI.add('ez-contenteditformview', function (Y) {
                 fieldDefinitions = contentType.get('fieldDefinitions'),
                 views = [],
                 that = this,
-                config = this.get('config');
+                config = this.get('config'),
+                languageCode = this.get('languageCode');
 
             Y.Object.each(fieldDefinitions, function (def) {
                 var EditView, view,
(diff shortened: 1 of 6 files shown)


mhyndle:

dpobel:
I don't think it's the correct place to do that. You are handling the fields for saving but before the form is generated, that's a bit weird.

To me, this is the responsibility of the field edit views to provide a correct field value so you should implement the same kind of logic in eZ.FieldEditView#getField (that means the field edit views should receive the language code in which the version is edited so it should be passed to content edit form view which will give it to the field edit views)

mhyndle:
Is current patch better? ping @dpobel 
I have manually tested it and it seems that it's working, just wanted to be sure before fixing unit tests as there are lot of them broken

dpobel:
yes much better here

mhyndle:
Ready for review

dpobel:
+1

yannickroger:
+1

PR #380, opened 2015-10-14
Fix EZP-24806: Location view of new articles is broken

mhyndle:
Jira: https://jira.ez.no/browse/EZP-24806

## Description
There is a problem, when we have got for example Article created with legacy BO and we want to view it with PlatformUI. For my case it was because Article created with legacy BO hasn't got star_rating field and field definitions for that contentType is pointing that there is such field.
In general - if viewed (or edited) content object didn't contain field that was defined in content's contentType field definitions, the app was throwing error and app was blocked because of that.
This PR solves it, by checking if edited/previewed content contains a field from contentType fieldDefinitions - if yes, view for this field is built and added to field views. Otherwise app is not trying to create view for that field anymore and instead just pushes warn into console.

## Tests
- [x] manual
- [x] unit tests

### DIFF
--- a/Resources/public/js/views/ez-contenteditformview.js
+++ b/Resources/public/js/views/ez-contenteditformview.js
@@ -58,23 +58,28 @@ YUI.add('ez-contenteditformview', function (Y) {
                 config = this.get('config');
 
             Y.Object.each(fieldDefinitions, function (def) {
-                var EditView, view;
+                var EditView, view,
+                    field = version.getField(def.identifier);
 
-                try {
-                    EditView = Y.eZ.FieldEditView.getFieldEditView(def.fieldType);
+                if (field) {
+                    try {
+                        EditView = Y.eZ.FieldEditView.getFieldEditView(def.fieldType);
 
-                    view = new EditView({
-                        content: content,
-                        version: version,
-                        contentType: contentType,
-                        fieldDefinition: def,
-                        field: version.getField(def.identifier),
-                        config: config,
-                    });
-                    views.push(view);
-                    view.addTarget(that);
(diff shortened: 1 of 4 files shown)


mhyndle:
I can't figure out how to test incosistency between content's fields and field definitions. I hoped, that initializing the raw content view ( https://github.com/ezsystems/PlatformUIBundle/pull/380/files#diff-da10d1c31bb009b0d3ef9e0cc253574bR616 ) will do the job so if I remove the fix and run the test it will fail. It didn't. Then I realized that it won't fail the test, because only thing that is done here, is to set the `content.getField(def.identifier)` to the `field` attribute of View. So if there is no such field, then `undefined` will be passed to the View. So the error is triggered somewhere higher I think and I have no idea how to test it. Only solution I'm thinking of is to check the `this._fieldViews` as it shouldn't contain only views for fields that were present in the content.
ping @dpobel

mhyndle:
ready for review

andrerom:
+1

But isn't this a kernel bug? Sounds a bit like it is side effect of the change we recently did there to make repo more tolerable to inconsistencies, however maybe it should always make sure to set missing fields with content type defaults, so maybe there is a bug that should be reported here even if this can be merged to be safe.

dpobel:
@andrerom I think this has more to do with incomplete/buggy field types. In this case, the culprit was Star Rating which has a weird behaviour through REST (I did not check in a Public API based script). In any case, the added console.warn will help us tracking inconsistencies issues.

dpobel:
+1

andrerom:
ok, good to know, and makes sense, it has not been completed yet.

PR #382, opened 2015-10-15
Fix EZP-24907: Impossible to save a Content with a RichText field in Firefox

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24907

# Description

This patch fixes the issue where it's impossible to create a Content with a RichText field in Firefox. This is happening because of our custom XML namespace. For a strict DOM Parser this default namespace basically means we are handling something which is not an XHTML document and then when creating a paragraph with the DOM API, it gets an `xmlns` attribute that breaks the RichText parser.

# Tests

unit tests + manual tests


yannickroger:
+1

dpobel:
ping @mhyndle @StephaneDiot

mhyndle:
+1

andrerom:
+1

PR #386, opened 2015-10-19
Fix EZP-24777: JS error when changing the heading level after adding a heading

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24777

# Description

This patch makes sure the content we add with the `eZAddContent` command is never empty by adding a `<br>`. This prevents the JS error from happening and this is actually what CKEditor does if you hit Enter to add a paragraph for instance.

# Tests

manual + unit tests


dpobel:

mhyndle:
+1

andrerom:
+1

PR #387, opened 2015-10-19
Fix EZP-24962: differrent email address validation

mhyndle:
Jira: https://jira.ez.no/browse/EZP-24962

## Description
In PlatformUI email address validation relies on built-in input type="email" validation and it's differrent from how it looks like in EmailAddressValidator (see https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/FieldType/Validator/EmailAddressValidator.php#L91).
The result of those differences was that PlatformUI was properly validating email address that was after that not validated through REST API call.
This patch solves it, with checking email adrress validity with the same regular expression test as it is in EmailAddressValidator.

## Tests
manual + unit tests

### DIFF
--- a/Resources/public/js/views/fields/ez-emailaddress-editview.js
+++ b/Resources/public/js/views/fields/ez-emailaddress-editview.js
@@ -12,7 +12,10 @@ YUI.add('ez-emailaddress-editview', function (Y) {
 
     Y.namespace('eZ');
 
-    var FIELDTYPE_IDENTIFIER = 'ezemail';
+    /* jshint -W101 */
+    var FIELDTYPE_IDENTIFIER = 'ezemail',
+        VALIDATION_PATTERN = /^((\"[^\"\f\n\r\t\v\b]+\")|([A-Za-z0-9_\!\#\$\%\&\'\*\+\-\~\/\^\`\|\{\}]+(\.[A-Za-z0-9_\!\#\$\%\&\'\*\+\-\~\/\^\`\|\{\}]+)*))@((\[(((25[0-5])|(2[0-4][0-9])|([0-1]?[0-9]?[0-9]))\.((25[0-5])|(2[0-4][0-9])|([0-1]?[0-9]?[0-9]))\.((25[0-5])|(2[0-4][0-9])|([0-1]?[0-9]?[0-9]))\.((25[0-5])|(2[0-4][0-9])|([0-1]?[0-9]?[0-9])))\])|(((25[0-5])|(2[0-4][0-9])|([0-1]?[0-9]?[0-9]))\.((25[0-5])|(2[0-4][0-9])|([0-1]?[0-9]?[0-9]))\.((25[0-5])|(2[0-4][0-9])|([0-1]?[0-9]?[0-9]))\.((25[0-5])|(2[0-4][0-9])|([0-1]?[0-9]?[0-9])))|((([A-Za-z0-9\-])+\.)+[A-Za-z\-]{2,}))$/;
+    /* jshint +W101 */
 
     /**
      * Email Address edit view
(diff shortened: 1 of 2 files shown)


mhyndle:
ready for review

Resources/public/js/views/fields/ez-emailaddress-editview.js
```diff
@@ -101,6 +101,21 @@ YUI.add('ez-emailaddress-editview', function (Y) {
         _getFieldValue: function () {
             return this.get('container').one('.ez-emailaddress-input-ui input').get('value');
         },
+
+        /**
+         * Returns email address format validity
```
dpobel: I would write something like:
```
Checks email address validity based on the same regexp as
the one used in the EmailAddress FieldType.
Otherwise, the email address edit view could accept email
that will be considered invalid when creating/updating a content.
```

Resources/public/js/views/fields/ez-emailaddress-editview.js
```diff
@@ -101,6 +101,21 @@ YUI.add('ez-emailaddress-editview', function (Y) {
...
+         *
+         * @protected
+         * @method _isValidEmail
+         * @return {Boolean}
+         */
+        _isValidEmail: function () {
+            /* jshint -W101 */
+            var pattern = /^((\"[^\"\f\n\r\t\v\b]+\")|([A-Za-z0-9_\!\#\$\%\&\'\*\+\-\~\/\^\`\|\{\}]+(\.[A-Za-z0-9_\!\#\$\%\&\'\*\+\-\~\/\^\`\|\{\}]+)*))@((\[(((25[0-5])|(2[0-4][0-9])|([0-1]?[0-9]?[0-9]))\.((25[0-5])|(2[0-4][0-9])|([0-1]?[0-9]?[0-9]))\.((25[0-5])|(2[0-4][0-9])|([0-1]?[0-9]?[0-9]))\.((25[0-5])|(2[0-4][0-9])|([0-1]?[0-9]?[0-9])))\])|(((25[0-5])|(2[0-4][0-9])|([0-1]?[0-9]?[0-9]))\.((25[0-5])|(2[0-4][0-9])|([0-1]?[0-9]?[0-9]))\.((25[0-5])|(2[0-4][0-9])|([0-1]?[0-9]?[0-9]))\.((25[0-5])|(2[0-4][0-9])|([0-1]?[0-9]?[0-9])))|((([A-Za-z0-9\-])+\.)+[A-Za-z\-]{2,}))$/;
```
dpobel: you should put that in a constant like `VALIDATION_PATTERN` so that it's no recompiled each time this method is called.

dpobel:
besides the inline comment, +1 even if I don't really like the regex duplication

mhyndle:
Feedback from @dpobel taken into account

andrerom:
+1

yannickroger:
Is there a way to inject this regexp in the platfomUi config at some point ?
Also… where does this regexp come from ? Legacy ? the email unit tested sounds perfectly valid to me. `me@localhost` would be rejected for instance.

Appart from the questions +1 :)

mhyndle:
Regexp comes from ezpublish-kernel EmailAddressValidator (https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/FieldType/Validator/EmailAddressValidator.php#L91) on which validation while publish was failing.
Unit test was covering invalid case (for example me@localhost) because specification of email address format used when running tests (Phantom JS?) is differrent from what webbrowser is doing, so I have just added new validation test to be sure it will cover that case, if for example engine running tests will be updated and will use the same email address format spec as web browser.
About injecting the email address pattern to config I'm not sure if it is really needed as if it changes it will have to be changed also in EmailAddressValidator. Or maybe move that pattern to config and use it in both cases (JS and EmailAddressValidator)?

yannickroger:
yes @mhyndle regarding config that's what I had in mind. When the API provides us with all the configuration on app bootup, provide the regexp along, that would be a way to avoid duplication. But I think it is wayyyy overkill for now. 
I think you are good for merge.

PR #390, opened 2015-10-20
[WIP] Fix EZP-24699: Enrich Role limitations view with links and human readable names

glye:
Work in progress. Open questions:
- Should the enrichment code be moved from the `RoleController` into the various `eZ\Publish\API\Repository\Values\User\Limitation` subclasses, with added API for this? It could make it easier to reuse in the content type edit views.
  - `Limitation::toString()` would be too platformui-specific for the api, perhaps. I'd rather have something like `getReadableLimitationValues()` which, as a parallell to the existing `$limitationValues`, returns an array, each entry being a hash of `href` and `name`.
- It adds translation for limitation identifiers, which allows me to change e.g. `Class` to 'Content type'. Should these be translated, and if so should we make such changes to the legacy identifiers?
  - I see the way @lolautruche has done it for limitation editing, if that's +1-ed I can go the same way.

Screenshot

TODO:
- [x] Human readable site access names
- [ ] Verify that all limitation cases are covered
- [ ] Tests

https://jira.ez.no/browse/EZP-24699

### DIFF
--- a/Controller/RoleController.php
+++ b/Controller/RoleController.php
@@ -9,7 +9,10 @@
 namespace EzSystems\PlatformUIBundle\Controller;
 
 use eZ\Publish\API\Repository\Exceptions\NotFoundException;
+use eZ\Publish\API\Repository\ContentTypeService;
 use eZ\Publish\API\Repository\RoleService;
+use eZ\Publish\API\Repository\SectionService;
+use eZ\Publish\API\Repository\Values\User\Role;
 use eZ\Publish\Core\MVC\Symfony\Security\Authorization\Attribute;
 use eZ\Publish\Core\Repository\Values\User\Policy;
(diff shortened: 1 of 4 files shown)


glye:
API / implementation review ping, see description. @lolautruche @andrerom @bdunogier @yannickroger @dpobel

glye:
Fixed output of siteaccess and owner limitations.

andrerom:
-1 for now, as we need to find a better way to handle this and that is to late for this release *(as we are in feature freeze)*

@glye Please do make the Role Limitations *(Location/Section)* fully readable and linkable, but for now lets only deal with SiteAccess Name after @lolautruche is done with his task.

andrerom:
Closing and moving this to [a later epic](https://jira.ez.no/browse/EZP-25442) as this needs to be done extensible, and should have some better support from kernel as defined in story now.

PR #391, opened 2015-10-20
Fix EZP-24964: Hiding/revealing displayed Location not updating sub items

yannickroger:
Link: https://jira.ez.no/browse/EZP-24964

## Description
When in the locations tabs, when changing the visibility of the location being displayed, it would not update the subitem list where most of items (unless they are already hidden) should now show "Hidden by superior".

## Tests
Manual and unit tests

## TODO
- [x] Model updates attributes
- [x] Location model is the same object in view and list
- [x] Trigger subitem list refresh
- [x] Cleanup and rebase

### DIFF
--- a/Resources/public/js/models/ez-contentmodel.js
+++ b/Resources/public/js/models/ez-contentmodel.js
@@ -216,10 +216,12 @@ YUI.add('ez-contentmodel', function (Y) {
          * @method loadLocations
          * @param {Object} options
          * @param {Object} options.api (required) the JS REST client instance
+         * @param {Object} [options.location] current location. If present it will be used instead of loading it from the API.
          * @param {Function} callback
          */
         loadLocations: function (options, callback) {
-            var api = options.api;
+            var api = options.api,
+                currentLocation = options.location;
 
             api.getContentService().loadLocations(this.get('id'), function (error, response) {
                 var tasks = new Y.Parallel(),
@@ -232,6 +234,12 @@ YUI.add('ez-contentmodel', function (Y) {
                 }
                 Y.Array.each(response.document.LocationList.Location, function (loc) {
                     var location,
+                        end;
+
+                    if (currentLocation && currentLocation.get('id') === loc._href) {
+                        locations.push(currentLocation);
+                    } else {
+                        location = new Y.eZ.Location({id: loc._href});
                         end = tasks.add(function (err) {
                             if ( err ) {
                                 loadError = true;
@@ -240,8 +248,8 @@ YUI.add('ez-contentmodel', function (Y) {
                             locations.push(location);
                         });
 
-                    location = new Y.eZ.Location({id: loc._href});
-                    location.load(options, end);
+                        location.load(options, end);
+                    }
                 });
 
                 tasks.done(function () {
--- a/Resources/public/js/models/ez-locationmodel.js
+++ b/Resources/public/js/models/ez-locationmodel.js
@@ -119,7 +119,14 @@ YUI.add('ez-locationmodel', function (Y) {
             //Remove the 2 line bellow once EZP-24899 is fixed and update unit test
             locationUpdateStruct.body.LocationUpdate.sortField = this.get('sortField');
             locationUpdateStruct.body.LocationUpdate.sortOrder = this.get('sortOrder');
-            options.api.getContentService().updateLocation(this.get('id'), locationUpdateStruct, callback);
+            options.api.getContentService().updateLocation(this.get('id'), locationUpdateStruct,
+                Y.bind(function (error, response) {
+                    if (!error) {
+                        this.set('hidden', (hidden === 'true'));
+                    }
+                    callback(error, response);
+                }, this)
+            );
         },
 
         /**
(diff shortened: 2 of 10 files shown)


Resources/public/js/models/ez-contentmodel.js
```diff
@@ -240,8 +244,13 @@ YUI.add('ez-contentmodel', function (Y) {
                             locations.push(location);
                         });
 
-                    location = new Y.eZ.Location({id: loc._href});
-                    location.load(options, end);
+                    if(currentLocation && currentLocation.get('id') === loc._href) {
+                        location = currentLocation;
+                        end();
```
dpobel: since that's absolutely not asynchronous, in this case we should not create the `end` function and then not call it :)

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -119,7 +119,18 @@ YUI.add('ez-locationmodel', function (Y) {
             //Remove the 2 line bellow once EZP-24899 is fixed and update unit test
             locationUpdateStruct.body.LocationUpdate.sortField = this.get('sortField');
             locationUpdateStruct.body.LocationUpdate.sortOrder = this.get('sortOrder');
-            options.api.getContentService().updateLocation(this.get('id'), locationUpdateStruct, callback);
+            options.api.getContentService().updateLocation(this.get('id'), locationUpdateStruct,
+                Y.bind(function (error, response) {
+                    if (!error) {
+                        if (hidden === 'true') {
```
dpobel: can be replaced by `this.set('hidden', (hidden === 'true'));`

Resources/public/js/views/tabs/ez-locationviewlocationstabview.js
```diff
@@ -296,6 +298,17 @@ YUI.add('ez-locationviewlocationstabview', function (Y) {
             },
 
             /**
+             * The location being displayed
```
dpobel: can be confusing in the context of the "locations tab". Should rather be something like:
*The location being displayed in the location view*.

Resources/public/js/models/ez-contentmodel.js
```diff
@@ -216,10 +216,14 @@ YUI.add('ez-contentmodel', function (Y) {
          * @method loadLocations
          * @param {Object} options
          * @param {Object} options.api (required) the JS REST client instance
+         * @param {Object} options.location (optional) current location. If present it will be
```
dpobel:
```
@param {Object} [options.location] current location. If present it will be
```
also don't indent the following line otherwise, you'll have to change it when changing any details of this one.

yannickroger: done

yannickroger:
Ready for review @dpobel @mhyndle @StephaneDiot

Resources/public/js/models/ez-contentmodel.js
```diff
@@ -232,6 +235,12 @@ YUI.add('ez-contentmodel', function (Y) {
                 }
                 Y.Array.each(response.document.LocationList.Location, function (loc) {
                     var location,
+                        end;
+
+                    if(currentLocation && currentLocation.get('id') === loc._href) {
```
dpobel: nitpick CS: space after `if`

Resources/public/js/models/ez-contentmodel.js
```diff
@@ -216,10 +216,13 @@ YUI.add('ez-contentmodel', function (Y) {
...
          * @param {Object} options
          * @param {Object} options.api (required) the JS REST client instance
+         * @param {Object} [options.location] current location. If present it will be used instead of loading it from the API.
          * @param {Function} callback
          */
         loadLocations: function (options, callback) {
-            var api = options.api;
+            var api = options.api,
+                currentLocation = options.location;
+
```
dpobel: nitpick CS: empty line

Resources/public/js/views/ez-subitemlistview.js
```diff
@@ -43,6 +43,14 @@ YUI.add('ez-subitemlistview', function (Y) {
                 this._uiPageLoading();
                 this._fireLocationSearch();
             });
+
+            this.get('location').on(
+                ['hiddenChange', 'invisibleChange'],
+                Y.bind( function () {
```
dpobel: nitpick CS: no space after `(`

Resources/public/js/views/ez-subitemlistview.js
```diff
@@ -43,6 +43,14 @@ YUI.add('ez-subitemlistview', function (Y) {
...
+                    this._uiPageLoading();
```
dpobel: I think you should add a `_refresh` method that calls `_uiPageLoading` and `_fireLocationSearch`, that we can write:
```js
this.after('offsetChange', this._refresh);
this.get('location').after(['hiddenChange', 'invisibleChange'], Y.bind(this._refresh, this));
```

which is bit better in terms of readability

dpobel:
Besides the inline (nitpicking) comments, that's very good, +1

mhyndle:
+1

PR #393, opened 2015-10-21
EZP-24943: Handle publishing of users

bdunogier:
> Fixes https://jira.ez.no/browse/EZP-24943

The content/user gateway will now respond with a 204 No Content if a PUBLISH call is made on a user (users are published on creation).

This makes the "Publish" button work as expected.


andrerom:
+1

glye:
+1

bdunogier:
Closing in favour of the much more reliable controller override: #401.

PR #394, opened 2015-10-21
EZP-22658: Add Language managment to Platform UI

andrerom:
https://jira.ez.no/browse/EZP-22658
Depends on https://github.com/ezsystems/repository-forms/pull/51

- [x] Double check if nothing more on js side is needed
- [x] Be able to disable language *(allow checkbox to be unselected)*
- [x] Testing

### DIFF
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -808,8 +808,8 @@ YUI.add('ez-platformuiapp', function (Y) {
                     view: "serverSideView",
                     callbacks: ['open', 'checkUser', 'handleSideViews', 'handleMainView']
                 }, {
-                    name: "adminRole",
-                    regex: /\/admin\/(pjax%2Frole.*)/,
+                    name: "adminLanguage",
+                    regex: /\/admin\/(pjax%2Flanguage.*)/,
                     keys: ['uri'],
                     path: "/admin/:uri",
                     sideViews: {'navigationHub': true, 'discoveryBar': false},
--- a/Resources/views/Language/list.html.twig
+++ b/Resources/views/Language/list.html.twig
@@ -1,0 +1,78 @@
+{% extends "eZPlatformUIBundle::pjax_admin.html.twig" %}
+
+{% trans_default_domain "language" %}
+
+{% block choice_row %}
+    {{ form_widget(form) }}
+{% endblock choice_row %}
+
+{% block header_breadcrumbs %}
+    {% set breadcrumb_items = [
+        {link: path('admin_dashboard'), label: 'dashboard.title'|trans({}, 'dashboard')},
+        {link: '', label: 'language.list'|trans({}, 'language')}
+    ] %}
+    {{ parent() }}
+{% endblock %}
+
+{% block header_title %}
+    <h1 class="ez-page-header-name" data-icon="&#xe61a;">{{ 'language.list'|trans }}</h1>
+{% endblock %}
+
+{% block content %}
+    <section class="ez-serverside-content">
+        <div class="ez-table-data is-flexible ez-table-data-container">
+            <table class="pure-table pure-table-striped ez-selection-table" data-selection-buttons=".ez-remove-language-button">
+                <thead>
+                    <tr>
+                        <th>{{ 'language.name'|trans }}</th>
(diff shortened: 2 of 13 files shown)


Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -808,8 +808,8 @@ YUI.add('ez-platformuiapp', function (Y) {
                     view: "serverSideView",
                     callbacks: ['open', 'checkUser', 'handleSideViews', 'handleMainView']
                 }, {
-                    name: "adminRole",
-                    regex: /\/admin\/(pjax%2Frole.*)/,
+                    name: "adminLanguage",
```
dpobel: I guess it's a mistake, we need both routes if you want the roles and the language links in the admin navigation

andrerom: see further up, @glye added a second adminRole above which is now the correct one I think given it now needs own service.

glye: Yes - but I just remembered that @mhyndle talked about needing two services. Mateusz, please verify that this was not something you added :)

mhyndle: Yes, I think we all missed that new route was created with already existing name: https://github.com/ezsystems/PlatformUIBundle/pull/344/files#diff-4e897328f624be930d091d1ce688a3ffR784
So I think that Andre's change is OK.

Resources/views/Language/create.html.twig
```diff
@@ -0,0 +1,31 @@
...
+    <section class="ez-serverside-content">
+        <ul class="ez-simpleform-error">
+            {% for flashMessage in app.session.flashbag.get('error') %}
+                <li class="ez-simpleform-error-item">{{ flashMessage }}</li>
+            {% endfor %}
+        </ul>
+
+        {{ form(form) }}
+    </section>
+{% endblock %}
```
lolautruche: You don't need this template, do you?

Resources/views/Language/list.html.twig
```diff
@@ -0,0 +1,94 @@
...
+
+        <tr class="ez-selection-table-row">
+            <td>
+                {{ form_widget(child, {'disabled': not language.canDelete, 'attr': {'class': 'ez-selection-table-checkbox'} }) }}
+            </td>
+            <td><a href="{{ path( 'admin_languageview', {'languageId': language.id} ) }}">{{ language.name }}</a></td>
+            <td><a href="{{ path( 'admin_languageview', {'languageId': language.id} ) }}">{{ language.identifier }}</a></td>
+            <td class="ez-table-data-id"><a href="{{ path( "admin_languageview", {"languageId": language.id} ) }}">{{ language.id }}</a></td>
+            <td>
+                <a href="{{ path('admin_languageedit', {'languageId': language.id}) }}" class="pure-button ez-button{% if not language.canEdit %} pure-button-disabled{% endif %}" data-icon="&#xe606;">{{ 'language.edit'|trans }}</a>
```
lolautruche: This is a link, so adding `pure-button-disabled` is not enough as this won't prevent the browser to follow the link if one clicks on it. (NB: template for section is wrong too).
See [ContentTypeGroups list template for a good example](https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/views/ContentType/list_content_type_groups.html.twig#L38-L42)

lolautruche:
Didn't you forget to update `routing_pjax.yml` to add the new routes?

glye:
+1, seems kosher to me.

dpobel:
+1

dpobel:
Just realized one thing though, after adding or removing a language, we'll have an inconsistency between [the languages list in the JS application](https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/apps/ez-platformuiapp.js#L948) (injected when the application is bootstrapped *via* [the Languages Application config provider]) and the actual list.

yannickroger:
So maybe we should create a follow up if we decide to merge this one.

andrerom:
@dpobel You are ok with that as a JS team followup?

andrerom:
We'll do that as a followup, @yannickroger can you create a story for it?

dpobel:
For the record, this has been detected by a community member: https://jira.ez.no/browse/EZP-25078

PR #395, opened 2015-10-22
Fix EZP-25015: select "multiple" not correctly serialized in ServerSideView

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25015

# Description

This patch fixes the form serialization done in server side view so that "multiple" select form input are correctly recognized and handled. Before this patch, only the first selected option was included in the data posted to the server.

# Tests

manual + unit tests

### DIFF
--- a/Resources/public/js/views/ez-serversideview.js
+++ b/Resources/public/js/views/ez-serversideview.js
@@ -122,9 +122,19 @@ YUI.add('ez-serversideview', function (Y) {
                             data[name] = field.get('value');
                         }
                         break;
+                    case 'select-multiple':
+                        if ( field.get('selectedIndex') >= 0 ) {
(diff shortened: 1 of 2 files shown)


dpobel:
ping @mhyndle @andrerom @StephaneDiot @yannickroger @lolautruche

lolautruche:
+1

glye:
+1. Any relation to https://jira.ez.no/browse/EZP-25006 and https://jira.ez.no/browse/EZP-24997 ? Not sure if those go through this JS.

dpobel:
@glye maybe, I'll let you test with the fix.

glye:
Damn, it actually seems to fix those issues! (partay)

PR #383, opened 2015-10-15
[WIP] Fix EZP-24905 Role assignment limitations

glye:
Work in progress, based on https://github.com/ezsystems/PlatformUIBundle/pull/344

Not working yet. The gist of `RoleServerSideView` is:
- The "Assign to..." button calls `_pickAssignee()` which is used when assigning a role without limitation; this is working.
- The "Assign to... with subtree limitation" button calls `_pickAssigneeLimitSubtree()`, used for assigning with limit on subtree. This fires a UDW for selecting subtree, whose `contentDiscoveredHandler` calls `__pickAssigneeLimitSubtreeAssign()`. That function should fire a UDW for selecting users/groups to assign to, but this fails silently.
- The "Assign to... with section limitation" button starts a new view with a section select dropdown. The Assign button in this view should use JS to read the selected section ID (currently hardcoded to 3) and pass it in the button attribute `data-role-assignment-section-id`. The button calls `_pickAssigneeLimitSection()` which fires a UDW for selecting users/groups. The UDW opens, but when selecting something it fails with "TypeError: s is undefined oop-min.js:8:1932"

- [x] PHP and templates
- [ ] Rebase once https://jira.ez.no/browse/EZP-24700 is stabilised/merged
- [ ] Fix loading of the second UDW when assigning by subtree limitation
- [ ] Ensure that the "assign by section limitation"-button reads section ID from the dropdown
- [ ] Fix oop-min.js error for section limitation UDW
- [ ] Remove console.log debugging lines

https://jira.ez.no/browse/EZP-24905

PR #397, opened 2015-10-22
Fix EZP-24905 Role assignment limitations

mhyndle:
https://jira.ez.no/browse/EZP-24905

## Description
Not working yet. The gist of `RoleServerSideView` is:
- The "Assign to..." button calls `_pickAssignee()` which is used when assigning a role without limitation; this is working.
- The "Assign to... with subtree limitation" button calls `_pickAssigneeLimitSubtree()`, used for assigning with limit on subtree. This fires a UDW for selecting subtree, whose `contentDiscoveredHandler` calls `__pickAssigneeLimitSubtreeAssign()`. That function should fire a UDW for selecting users/groups to assign to, but this fails silently.
- The "Assign to... with section limitation" button starts a new view with a section select dropdown. The Assign button in this view should use JS to read the selected section ID (currently hardcoded to 3) and pass it in the button attribute `data-role-assignment-section-id`. The button calls `_pickAssigneeLimitSection()` which fires a UDW for selecting users/groups. The UDW opens, but when selecting something it fails with "TypeError: s is undefined oop-min.js:8:1932"

## Tasks
- [x] PHP and templates
- [x] Rebase once https://jira.ez.no/browse/EZP-24700 is stabilised/merged
- [ ] ~~Fix loading of the second UDW when assigning by subtree limitation~~ (moved to subtasks)
- [ ] ~~Ensure that the "assign by section limitation"-button reads section ID from the dropdown~~ (moved to subtasks)
- [ ] ~~Fix oop-min.js error for section limitation UDW~~ (moved to subtasks)
- [ ] ~~Remove console.log debugging lines~~ (moved to subtasks)

## Notes
It's new PR, but it's the same as it was in https://github.com/ezsystems/PlatformUIBundle/pull/383
I have messed that one rebasing it on master, so this PR is the clear one based on master

### DIFF
--- a/Controller/RoleController.php
+++ b/Controller/RoleController.php
@@ -10,6 +10,7 @@ namespace EzSystems\PlatformUIBundle\Controller;
 
 use eZ\Publish\API\Repository\Exceptions\NotFoundException;
 use eZ\Publish\API\Repository\RoleService;
+use eZ\Publish\API\Repository\SectionService;
 use eZ\Publish\Core\MVC\Symfony\Security\Authorization\Attribute;
 use eZ\Publish\Core\Repository\Values\User\Policy;
 use eZ\Publish\Core\Repository\Values\User\PolicyDraft;
@@ -34,6 +35,11 @@ class RoleController extends Controller
     protected $roleService;
 
     /**
+     * @var \eZ\Publish\API\Repository\SectionService
+     */
+    protected $sectionService;
+
+    /**
      * @var ActionDispatcherInterface
      */
     private $roleActionDispatcher;
@@ -50,11 +56,13 @@ class RoleController extends Controller
 
     public function __construct(
         RoleService $roleService,
+        SectionService $sectionService,
         ActionDispatcherInterface $roleActionDispatcher,
         ActionDispatcherInterface $policyActionDispatcher,
         TranslatorInterface $translator
     ) {
         $this->roleService = $roleService;
+        $this->sectionService = $sectionService;
         $this->roleActionDispatcher = $roleActionDispatcher;
         $this->policyActionDispatcher = $policyActionDispatcher;
         $this->translator = $translator;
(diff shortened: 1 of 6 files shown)


Controller/RoleController.php
```diff
@@ -327,4 +335,21 @@ public function deleteRoleAssignmentAction(Request $request, $roleAssignmentId)
...
+
+    /**
+     * Lets you select a section which will be used as limitation when assigning the role.
+     *
+     * @param Request $request
+     * @param int $roleId Role ID
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function assignRoleBySectionAction(Request $request, $roleId)
```
crevillo: assignRoleBySection? I'd go for assingSectionLimitationToRole maybe...

andrerom: It's not new, and I would maybe call it `assignRoleBySectionLimitation`

mhyndle:
It should be probably PR made from `ezsystems/PlatformUIBundle` branch but I have unfortunatelly made it from my fork branch. So I'll ask @glye for help if there will be anything I wouldn't know how to handle it.
For now ping for review @andrerom @lolautruche @bdunogier @yannickroger

andrerom:
+1

Controller/RoleController.php
glye: `assignRoleBySectionLimitationAction` then, to stay in sync with other *Action methods.

crevillo: ok. nevermind. depends on how you read it. as i see you could ADD a section limitation to a Role and you could assign A role TO a group of users. With the Limitation suffix looks better anyway. 

i read it as "assing this role to a section of users". thats what's looked weird to me.

glye: @crevillo I wouldn't say it like that. It's always about "assigning a role to user/group", whether the assignment is limited or not. The limitation is for the assignment, not for the role.

Resources/views/Role/assign_role_by_section.html.twig
```diff
@@ -0,0 +1,51 @@
...
+{% block content %}
+    <section class="ez-serverside-content">
+        <label for="data-role-assignment-section">{{ 'role.choose_section'|trans }}</label>
+        <select id="data-role-assignment-section" class="data-role-assignment-section">
+            {% for section in sections %}
+            {# @var section \eZ\Publish\API\Repository\Values\Content\Section #}
+            <option value="{{ section.id }}">{{ section.name }}</option>
+            {% endfor %}
+        </select>
+        {# TODO: Section ID 3 is hardcoded below, it should be read from the select above. #}
```
glye: For reviewers who don't know, this TODO is part of what's moved to a follow-up (to keep the JS heavy stuff separate)

mhyndle: I have removed button's attribute as it won't be used. Section ID will be taken directly from select.

glye:
Guess I can't plus-one my own work, but let's say I strongly endorse it.

andrerom:
ping @yannickroger / @dpobel

dpobel:
+1 but this can not be merged until when implement the sub-tasks to add the behavior to the new buttons.
See https://jira.ez.no/browse/EZP-24905#comment-183506

andrerom:
Seems to have been replaced by #543.

PR #398, opened 2015-10-22
EZP-24927_Preview_title_empty

StephaneDiot:
Jira : https://jira.ez.no/browse/EZP-24927

## Description

The preview title was empty when displaying a preview. Now it can be correctly display with the associated languageCode.
:warning:  But to see the preview title you need to have saved your draft before ! This is due to the fact that there isn't a version of the content before saving it (we can't use currentVersion neither). https://jira.ez.no/browse/EZP-24931 is related to the same problem.

## Tests

- [x] modified units tests to fit

### DIFF
--- a/Tests/js/views/assets/ez-editpreviewview-tests.js
+++ b/Tests/js/views/assets/ez-editpreviewview-tests.js
@@ -14,13 +14,14 @@ YUI.add('ez-editpreviewview-tests', function (Y) {
         setUp: function () {
             this.contentId = 59;
             this.versionNo = 42;
+            this.previewName = 'Test elvish name';
             this.mockContent = new Y.eZ.Content({
                 contentId: this.contentId,
                 name: "Test name"
             });
             this.mockVersion = new Y.eZ.Version({
                 versionNo: this.versionNo,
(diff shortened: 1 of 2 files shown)


Tests/js/views/assets/ez-editpreviewview-tests.js
```diff
@@ -20,7 +20,7 @@ YUI.add('ez-editpreviewview-tests', function (Y) {
             });
             this.mockVersion = new Y.eZ.Version({
                 versionNo: this.versionNo,
-                names: {value: [{'_languageCode': 'eng-GB', '#text': 'Test name'}]}
+                names: {value: [{'_languageCode': 'eng-GB', '#text': 'Test name'}, {'_languageCode': 'quenya', '#text': 'Test name'}]}
```
dpobel: that's not enough in terms of test. First you should put a different value than the one for eng-GB and then in a test (""Test available variable in template" I guess), you should assert that the template receives the expected name.

dpobel:
and the commit message is wrong, should be something like: 'Fix EZP-24927: Preview title is empty'

dpobel:
+1

yannickroger:
+1

PR #399, opened 2015-10-22
Fix EZP-24990: Create a new role broken PlatformUI

glye:
Since we changed back to the old style of role versioning, drafts of published roles have version = published role id. Unpublished role drafts have version = -1. However, the `updateRoleAction()` expects a published role id, which doesn't exist yet for new drafts.

The fix is to try to load the `$roleId` as if it was a draft id first - if that succeeds, we're done, if it fails, we proceed to load draft by role id as before.

https://jira.ez.no/browse/EZP-24990

### DIFF
--- a/Controller/RoleController.php
+++ b/Controller/RoleController.php
@@ -160,11 +160,17 @@ class RoleController extends Controller
     public function updateRoleAction(Request $request, $roleId)
     {
         try {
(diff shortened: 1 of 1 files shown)


glye:
This is blocking QA of role features. Review ping @lolautruche @andrerom @bdunogier @yannickroger @pspanja et al.

andrerom:
+1

lolautruche:
+1

PR #401, opened 2015-10-22
Rest content/user controller override

bdunogier:
> Status: ready for review
> Implements https://jira.ez.no/browse/EZP-25024
> Replaces #393 and the whole user gateway
> Tests: BDD @ https://github.com/ezsystems/ezpublish-kernel/pull/1486. Also tested on PlatformUI: create, edit and delete all seem to work fine

Adds an override of the kernel's REST Content Controller (by means of a class name override).

This override implements new versions of createContent, updateVersion, publishVersion and deleteContent. For those, if the affected item is a User, it executes the request with the user API, and responds as the Content APi would have.

This allows user management in PlatformUI to use the REST content API transparently, without implementing the whole user API.

### TODO
- [x] Fix publishing

### Known issues
- Removing groups won't correctly remove the users they contain.
  The user content will be removed, but not the user account. That's because `deleteUserGroup` uses `deleteContent`. The Content API doesn't have user external data handling, and doesn't delete the database record in `ezuser`. The user doesn't have any permission anymore, but the record is orphan.


Controller/Rest/ContentController.php
```diff
@@ -0,0 +1,288 @@
...
+ * Override of the REST Content Controller.
+ *
+ * Features:
+ * - Proxies Content calls on User content to the UserService.
+ */
+class ContentController extends Content
+{
+    private $settings = ['user-contenttype-id' => 4];
+
+    public function setSettings(array $settings)
```
andrerom: Guess this is inherited from parent, if not this could be a array of user content type ids.

andrerom:
+1, as for the known issue that will have to be solved separately either by foreign key cascading and/or user field type storage gateway implementation as discussed.

Controller/Rest/ContentController.php
bdunogier: Hmm, no, it's not really inherited. I just like to have a settings array in those layers, it makes it easy to add features without touching sighatures.

The array is a good idea, I'll add it.

dpobel:
+1

PR #402, opened 2015-10-23
EZP-25016: Hide the "Save" button when editing users

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25016

# Description

This patch hides the *Save* button when editing a user because the UserService does not support draft for users. I also realized without a draft of user, we can not have a preview, so I removed the preview buttons as well.

buttons_user

Together with https://github.com/ezsystems/PlatformUIBundle/pull/401, this should offer a correct User content handling in PlatformUI.

## Tasks

* [x] Add method to detect a field type in a content type
* [x] Remove the save draft
* [x] Remove the preview buttons

# Tests

manual test + unit test

### DIFF
--- a/Resources/public/js/models/ez-contenttypemodel.js
+++ b/Resources/public/js/models/ez-contenttypemodel.js
@@ -82,8 +82,21 @@ YUI.add('ez-contenttypemodel', function (Y) {
             });
 
             return fieldGroups;
-        }
+        },
 
+        /**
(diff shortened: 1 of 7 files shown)


dpobel:

yannickroger:
+1 and cool that you are doing some cleanup in the meantime

mhyndle:
+1

PR #403, opened 2015-10-23
Fix EZP-24942: Press <Enter> in repository form raises a notification error

glye:
These affected forms all assume that a submit button has been clicked. This is not the case when you have pressed the enter key instead. In that case I now pass `null` for action, meaning that only the default form action is dispatched.

Note: The forms behave a bit differently because actions are set up differently. When pressing enter, the role and content type edit forms save and reload. The section and content type group forms save and close.

https://jira.ez.no/browse/EZP-24942

Also required: https://github.com/ezsystems/repository-forms/pull/53


andrerom:
+1

pspanja:
+1

PR #406, opened 2015-10-26
Fix EZP-24971: Remaining is-block-focused class in the saved richtext html

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24971

# Description

This patch fixes an issue where the class used to style the focused element in the editor is not correctly removed when the RichText edit view retrieves the editor's content. As visible in the patch, it also contains a workaround for [a bug in the RichText parser not accepting an empty `class` attribute](https://jira.ez.no/browse/EZP-25028)...

# Tests

manual + unit tests


dpobel:

mhyndle:
+1

yannickroger:
+1

PR #407, opened 2015-10-26
EZP-24986: Unable to create a new content using PlatformUI in Safari

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24986

# Description

Before this patch and since https://github.com/ezsystems/PlatformUIBundle/commit/d658524c640454fcdf4648ccdca7b8428730f020 it was impossible to create a content in Safari. This was happening because the Content model has an invalid date as default value for the attribute `publishedDate` and the date setter does not validate its input so an invalid date is put in this attribute and Safari throws when calling `toJSON` on that.

So this patch changes the default value and makes the date setter a bit more robust against its input.

# Tests

unit tests + manual test by @sunpietro (thanks !)

### DIFF
--- a/Resources/public/js/models/extensions/ez-contentinfo-attributes.js
+++ b/Resources/public/js/models/extensions/ez-contentinfo-attributes.js
@@ -97,7 +97,7 @@ YUI.add('ez-contentinfo-attributes', function (Y) {
              */
             publishedDate: {
                 setter: '_setterDate',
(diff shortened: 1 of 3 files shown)


dpobel:

mhyndle:
+1

sunpietro:
+1

yannickroger:
+1

PR #412, opened 2015-10-27
EZP-24989: Restrict the possible selection made through the UDW

mhyndle:
Jira: https://jira.ez.no/browse/EZP-24989

## Description
Currently UDW allows to pick up any readable content. This PR allows developer to provide to `contentDiscover` event `isSelectable` function inside config object. `isSelectable` function should return bool value depending on that if content is selectable or not. `isSelectable` function receives selected content struct:
```javascript
{
  'contentInfo': eZ.ContentInfo,
  'contentType': eZ.ContentType,
  'location': eZ.Location
}
```
So basing on selection developer can decide if selected content is allowed for selection or not, for example if selected content is container.
If function return `FALSE` then content is not selectable.

What's more, following @dpobel's instructions, the `isAlreadySelected` function is now passed to the method view so we can check if content is already selected inside the method view.

## Screencasts
For tests I have injected to the UDW's `contentDiscover` event the `isSelectable` function that allows only contents that are containers to be selectable:
```javascript
this.fire('contentDiscover', {
  config: {
    title: "Select content",
    multiple: true, // or false
    // ...
    isSelectable: function (contentStruct) {
      return contentStruct.contentType.get('isContainer');
    },
  },
});
```
- move content: https://youtu.be/aZlbdAB6TFE - single selection mode
- add location: https://youtu.be/YQp5Z56zdzU - multiple selection mode

## Tasks
- [x] implementation
- [x] fill jsdocs
- [x] tests
 - [x] manual
 - [x] unit tests

### DIFF
--- a/Resources/public/js/views/ez-universaldiscoveryview.js
+++ b/Resources/public/js/views/ez-universaldiscoveryview.js
@@ -40,8 +40,14 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
             this.after('multipleChange', this._uiMultipleMode);
             this.after('visibleMethodChange', this._updateMethods);
             this.after('*:selectContent', function (e) {
-                if ( !this.get('multiple') ) {
-                    this._storeSelection(e.selection);
+                var isSelectable = this.get('isSelectable');
+
+                if (!this.get('multiple')) {
+                    if (e.selection !== null && isSelectable(e.selection)) {
+                        this._storeSelection(e.selection);
+                    } else {
+                        this._resetSelection();
+                    }
                 }
             });
             this.after('*:unselectContent', function (e) {
@@ -52,9 +58,6 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
                     this._uiAnimateSelection(e.target);
                     this._storeSelection(e.selection);
                 }
-            });
-            this.after('universalDiscoverySelectedView:contentStructChange', function (e) {
-                e.target.set('confirmButtonEnabled', !this._isAlreadySelected(e.newVal));
             });
             this.after('selectionChange', function () {
                 this._uiSetConfirmButtonState();
@@ -256,7 +259,8 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
         /**
          * Updates the method views depending on the value so that their
          * `visible` flag is consistent with the `visibleMethod` attribute value
-         * and so that they get the correct `multiple` flag value as well.
+         * and so that they get the correct `multiple` flag value as well. What's more
+         * the `isSelectable` function registered in UDW is passed to the method view.
          *
          * @method _updateMethods
          * @protected
@@ -277,7 +281,8 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
 
                 method.setAttrs({
                     'multiple': this.get('multiple'),
-                    'visible': visible
+                    'visible': visible,
+                    'isSelectable': Y.bind(this.get('isSelectable'), this)
                 });
                 if ( visible ) {
                     this._visibleMethodView = method;
@@ -528,6 +533,7 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
                             bubbleTargets: this,
                             priority: 100,
                             multiple: this.get('multiple'),
+                            isAlreadySelected: Y.bind(this._isAlreadySelected, this)
                         }),
                     ];
                 },
@@ -587,7 +593,22 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
              */
             data: {
                 value: {},
-            }
+            },
+
+            /**
+             * Checks wether the content is selectable. Function can be provided in the config
+             * when firing the `contentDiscover` event so it can check if content is selectable
+             * depending on the context where UDW is triggered.
+             *
+             * @attribute isSelectable
+             * @type {Function}
+             */
+            isSelectable: {
+                validator: Y.Lang.isFunction,
+                value: function (contentStruct) {
+                    return true;
+                }
+            },
         }
     });
 });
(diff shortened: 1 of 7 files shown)


Resources/public/js/views/ez-universaldiscoveryview.js
```diff
@@ -237,9 +237,36 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
          * @protected
          */
         _uiSetConfirmButtonState: function () {
-            var confirmButton = this.get('container').one('.ez-universaldiscovery-confirm');
+            var confirmButton = this.get('container').one('.ez-universaldiscovery-confirm'),
+                isSelectable = this.get('isSelectable');
 
-            confirmButton.set('disabled', !this.get('selection'));
+            if (this.get('multiple') && this.get('selection')) {
```
mhyndle: self note: can be simplified a bit

Resources/public/js/views/ez-universaldiscoveryview.js
```diff
@@ -54,7 +54,7 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
                 }
             });
             this.after('universalDiscoverySelectedView:contentStructChange', function (e) {
-                e.target.set('confirmButtonEnabled', !this._isAlreadySelected(e.newVal));
```
dpobel: I realize it was a mistake to write that. Instead the universal discovery methods should receive a `isAlreadySelected` function (which expects a content struct, something like `Y.bind(this._isAlreadySelected, this)` in the current code) so that each method can check the selection or pass this function to a sub-view which can then check the selection. That way, the universalDiscoverySelectedView can decide on its own the status of the confirm button.
Once this change made, pretty much the same strategy should be applied to `isSelectable`. It should be passed to the methods so that when `multiple` is set to true, the method can check if the content is OK.

Resources/public/js/views/ez-universaldiscoveryview.js
```diff
@@ -237,9 +237,36 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
...
         _uiSetConfirmButtonState: function () {
-            var confirmButton = this.get('container').one('.ez-universaldiscovery-confirm');
+            var confirmButton = this.get('container').one('.ez-universaldiscovery-confirm'),
+                isSelectable = this.get('isSelectable');
 
-            confirmButton.set('disabled', !this.get('selection'));
+            if (this.get('multiple') && this.get('selection')) {
+                confirmButton.set('disabled', false);
+            } else {
+                if (this.get('selection') && isSelectable(this.get('selection'))) {
```
dpobel: actually `isSelectable` should not be called to set the state of the button, it should rather be called to accept or deny the value to put in the `selection` attribute

andrerom:
+1

Resources/public/js/views/universaldiscovery/ez-universaldiscoveryselectedview.js
```diff
@@ -42,6 +43,23 @@ YUI.add('ez-universaldiscoveryselectedview', function (Y) {
         },
 
         /**
+         *
+         * @method _setConfirmButtonEnabled
+         * @protected
+         */
+        _setConfirmButtonEnabled: function (contentStruct) {
```
dpobel: I would call it `_setConfirmButtonState`

Resources/public/js/views/universaldiscovery/ez-universaldiscoveryselectedview.js
```diff
@@ -42,6 +43,23 @@ YUI.add('ez-universaldiscoveryselectedview', function (Y) {
...
+            var isAlreadySelected = Y.bind(this.get('isAlreadySelected'), this),
+                isSelectable = Y.bind(this.get('isSelectable'), this);
```
dpobel: I would not bind those functions to the selected view. If you want to bind those functions to a context, that should the universal discovery view (that's already the case for `isAlreadySelected` and maybe we should do the same for `isSelectable`)

Resources/public/js/views/universaldiscovery/ez-universaldiscoverybrowseview.js
```diff
@@ -67,14 +67,18 @@ YUI.add('ez-universaldiscoverybrowseview', function (Y) {
...
+         * new `isSelectable` function to the selected view.
          *
-         * @method _setSelectedViewConfirmButton
+         * @method _setSelectedViewAttrs
          * @protected
          */
-        _setSelectedViewConfirmButton: function () {
-            this.get('selectedView').set('addConfirmButton', this.get('multiple'));
+        _setSelectedViewAttrs: function () {
+            this.get('selectedView').setAttrs({
```
dpobel: It's a bit weird to set 2 attributes when one changes, ie you should only set `addConfirmButton` when `multiple` changes and only set `isSelectable` when `isSelectable`changes.

mhyndle: I did it this way because `this.get('selectedView')` creates new instance of Y.eZ.UniversalDiscoverySelectedView so splitting it to separate event handlers will lead to losing `addConfirmButton` for selectedView if `isSelectable` changes and losing `isSelectable` for selectedView if `multiple` changes. So maybe there is better way of menaging selectedView?

mhyndle:
ready for review

Resources/public/js/views/ez-universaldiscoveryview.js
```diff
@@ -36,12 +36,20 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
...
         initializer: function () {
+            var that = this;
+
             this.on('changeTab', this._updateVisibleMethod);
             this.after('multipleChange', this._uiMultipleMode);
             this.after('visibleMethodChange', this._updateMethods);
             this.after('*:selectContent', function (e) {
-                if ( !this.get('multiple') ) {
-                    this._storeSelection(e.selection);
+                var isSelectable = that.get('isSelectable');
```
dpobel: `that` not needed I think

dpobel:
besides the last inline comment, +1

PR #413, opened 2015-10-28
EZP-24957 Add a combine setting in yui.yml

michalpipa:
> Jira issue: https://jira.ez.no/browse/EZP-24957
> Status: ready for review

Add `combine` setting to `yui.yml` to enable combo loader. If enabled it sets `combine` option for all modules.


michalpipa:
My question: should this setting be enabled by default?

michalpipa:
ping @dpobel

bdunogier:
Please add a description, even if it is one sentence, that describes what the PR changes. 
Also, as mentioned yesterday, specify the PR's status. I usually go with something like:

```
> Implements http://jira.ez.no/...
> Status: work in progress|ready for review
```

When/if people scan the commits, they will go through the PR first, and it should at the very least look decent.

bdunogier:
That aside, +1 for what it's worth, about the code at least :-)

DependencyInjection/Configuration.php
```diff
@@ -45,6 +45,9 @@ protected function defineYui(NodeBuilder $saNode)
                         ->values(['raw', 'min', 'debug'])
                         ->info("Filter to apply to module urls. This filter will modify the default path for all modules.\nPossible values are 'raw', 'min' or 'debug''")
                     ->end()
+                    ->booleanNode('combine')
+                        ->info('If true, YUI combo loader is used.')
```
bdunogier: I'd say "will" instead of "is", since it is what is used in the one above.

Twig/TwigYuiExtension.php
```diff
@@ -90,6 +90,14 @@ public function yuiConfigLoaderFunction($configObject = '')
             'modules' => [],
         ];
 
+        $combine = $this->configResolver->getParameter('yui.combine', 'ez_platformui');
+
+        if (true === $combine) {
```
bdunogier: I really, really don't like this way to organize ifs... I know some people love it, but we don't use it in Platform (except a dozen ones in API tests), and I'd rather favour consistency.

I know it avoids mistakes, but on the other hand, it prevents you from writing something that won't work, so all in all, the benefit is really limited, and it's completely unnatural when you read it :-)

glye: I just looked it up, and learned something new today :) I don't like this either, but does the CS say anything about it? Whatever we choose, we should have consistency.

michalpipa: Ok, I'll change it. Consistency is important.

But I still love it and I think it's better :-)

dpobel:
I'm ok with the code but no tests (btw, I'm surprise this does not break any existing test)
For the default value: eventually it should be true but right that should false since we don't have a working combo loader yet.

michalpipa:
@dpobel I've modified unit test: https://github.com/ezsystems/PlatformUIBundle/pull/413/files#diff-c3a0f60b883e98ec8add0033eea5dcd1

andrerom:
+1 with the inline comments fixed, lets add the combo loader asap so we can enable this :)

dpobel:
+1 with false by default as long as we don't have the combo loader.

bdunogier:
Can this be merged without waiting for #427 ?

dpobel:
if the default value is set to false yes, otherwise this'll break PlatformUI

andrerom:
@michalpipa ?

michalpipa:
@bdunogier @andrerom I can do it. I will change default value to `false` and set it to `true` in th other PR.

michalpipa:
@andrerom I've set default value to `false` and squashed commits.

PR is ready to merge.

PR #414, opened 2015-10-28
Fix EZP-25040: Wrong mimeType is sent after updating a binary file

yannickroger:
Link: https://jira.ez.no/browse/EZP-25040

## Description
An outdated mime type was provided when updating a File. Since the API is not using this mime type we remove it (instead of updating it).

## Tests
Unit test and manual


yannickroger:
Ready for review ping @dpobel @mhyndle @StephaneDiot

mhyndle:
+1

andrerom:
+1

dpobel:
+1

PR #415, opened 2015-10-29
Fix EZP-23733: The delete button in section view/list should be disabled if the section can not be removed

glye:
> Fixes https://jira.ez.no/browse/EZP-23733
> Also requires https://github.com/ezsystems/ezpublish-kernel/pull/1493
> Status: Ready for review

Uses new method for counting policies using section in limitations.

### DIFF
--- a/Controller/SectionController.php
+++ b/Controller/SectionController.php
@@ -46,11 +46,15 @@ class SectionController extends Controller
     {
         $sectionList = $this->sectionService->loadSections();
         $contentCountBySectionId = [];
+        $deletableSections = [];
         $deleteFormsBySectionId = [];
 
         foreach ($sectionList as $section) {
             $sectionId = $section->id;
             $contentCountBySectionId[$sectionId] = $this->sectionService->countAssignedContents($section);
+            if (!$this->sectionService->isSectionUsed($section)) {
+                $deletableSections[$sectionId] = true;
+            }
             $deleteFormsBySectionId[$sectionId] = $this->createForm(
                 new SectionDeleteType($this->sectionService),
                 ['sectionId' => $sectionId]
@@ -61,7 +65,8 @@ class SectionController extends Controller
             'canEdit' => $this->isGranted(new Attribute('section', 'edit')),
             'canAssign' => $this->isGranted(new Attribute('section', 'assign')),
             'sectionList' => $sectionList,
-            'contentCountBySection' => $contentCountBySectionId,
+            'contentCountBySectionId' => $contentCountBySectionId,
+            'deletableSections' => $deletableSections,
             'deleteFormsBySectionId' => $deleteFormsBySectionId,
         ]);
     }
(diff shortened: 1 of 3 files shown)


Resources/views/Section/list.html.twig
```diff
@@ -88,7 +88,7 @@
                                     form_widget(
                                         deleteForm.delete,
                                         {
-                                            "disabled": not canEdit,
+                                            "disabled": contentCountBySection[section.id] or policyLimitationCountBySection[section.id],
```
lolautruche: I'd better leave the logic to the controller.

Resources/views/Section/view.html.twig
```diff
@@ -59,7 +59,7 @@
                         form_widget(
                             deleteForm.delete,
                             {
-                                "disabled": not canEdit,
+                                "disabled": contentCount or policyLimitationCount,
```
lolautruche: Same as above

glye:
Moved delete check logic to the controller.

glye:
Changed to use `isSectionUsed()` as discussed in other PR.

glye:
Review ping @lolautruche @andrerom @bdunogier @crevillo (platformui part of approved kernel fix)

Controller/SectionController.php
```diff
@@ -46,11 +46,13 @@ public function listAction()
     {
         $sectionList = $this->sectionService->loadSections();
         $contentCountBySectionId = [];
+        $canDeleteBySectionId = [];
```
crevillo: maybe i'd try another variable naming here ```can``` makes me expect a boolean here, but it's an array.
i'm +1 anyway.

Controller/SectionController.php
```diff
@@ -46,11 +46,13 @@ public function listAction()
...
         $deleteFormsBySectionId = [];
 
         foreach ($sectionList as $section) {
             $sectionId = $section->id;
             $contentCountBySectionId[$sectionId] = $this->sectionService->countAssignedContents($section);
+            $canDeleteBySectionId[$sectionId] = !$this->sectionService->isSectionUsed($section);
```
crevillo: also wondering if going this way won't lead us to have too much lines of code here and even too much calls... (actually 3 per loop).
you wouldn't expect have lot of sessions here, but who knows...

glye: The section list doesn't have pagination yet, so with many sections it can be a problem. But that's the case with or without this PR, and it's a separate concern, I think.

crevillo: ok. just thinking for the future. +1 otherwise :)

Controller/SectionController.php
bdunogier: I'd actually agree with Carlos here: we have said that "can" was reserved for permissions. Could you change it to deletable ? Maybe then change it to only contain the ones that are deletable, and check in the template using is defined `{% if deletableSections[section.id] is defined %}`.

glye: @bdunogier Ok, done.

glye:
Renamed `canDelete*` to `deletable*`

glye:
Got +1 from @bdunogier too, merging.

PR #416, opened 2015-10-29
BDD: Refactor wait/timeout detection, traits

joaoinacio:
Refactor implementation of BDD sentences based on new spin/loading wait timeout mechanism.
Ref: https://github.com/ezsystems/ezplatform/pull/57

The new method uses a "spin function" with a short timeout (5s) to attempt to get a valid DOM node from selenium, while a different "wait" function is used to wait while loading content (from an ajax call, for example) - the later one having a much longer timeout (60s).

In addition, some traits are now actual Context classes, with behat suites accordingly.

TODO
- [x] update BehatBundle for dependency.
- [x] update ezplatform test branch
- [ ] test/update BDD on StudioUI
- [x] cleanup/squash commits


miguelcleverti:
@andrerom @bdunogier this PR is ready for merge.

bdunogier:
I really wish you had separated the traits change from the wait mechanism... is it simple enough so that you could do this ?

Could you try to describe how you've approached the traits thing ? We can also do a call so that you explain it directly.

miguelcleverti:
@bdunogier most traits were removed in favor of context classes, the only ones that remained were the ones related to the authentication(logins) and the common actions(simple navigation,  click elements, etc.).

We decided to leave these two as traits because the are needed for all the tests and including everything in the main context file would make it harder to read.

The reason we did both changes in the same PR was because the waiting one was dependent of the restructure. Before the squash they were in different commits, I can still separate them by commits.

miguelcleverti:
re-ping @andrerom @bdunogier. Related to Bertrand question again, the two changed are connected, so separating them now would take a great effort.

andrerom:
I'm lost here, there are BDD spec changes here, but no explanation about why in the PR.

joaoinacio:
Hi André, could you be more specific?

I do see a sentence change/fix: canceling the confirm box for removal of content (not confirming would mean not clicking confirm, but the action explicitly clicks the cancel button).
```
     /**
      * Cancel content removal.
-      * @When I do not confirm the removal
+      * @When I cancel the removal
      */
     public function cancelRemoval()
     {
         $this->clickElementByText('Cancel', '.ez-confirmbox-cancel');
     }
```

Other than that, AFAICT, the editor->admin and "Copy Subtree" feature changes where necessary for tests to pass (platform issue) but IMO these could safely be fixed/updated separately from this PR.

andrerom:
referring to changes to spec, aka .feature files.

andrerom:
but ok is this ready for merge?

miguelcleverti:
@andrerom yes it is

PR #417, opened 2015-10-30
Fix EZP-25002: Unable to add more options to ezselection field type

glye:
> Fixes https://jira.ez.no/browse/EZP-25002
> Also requires https://github.com/ezsystems/repository-forms/pull/56
> Status: Ready for review

Dynamic JS-based solution with separate add/remove buttons.
- [x] The delete button of newly added options doesn't work until you reload (click Apply) because it isn't added to the YUI events
- [x] TypeError: this.get(...) is undefined ez-contenttypeeditserversideview.js:115:30 when using newly created delete button
- [x] Validation: Ensure that there must be at least one option, handled in https://github.com/ezsystems/ezpublish-kernel/pull/1495
- [ ] Ensure that there must be at least one option - in javascript, in addition to the field type validation

IScreenshot

### DIFF
--- a/Resources/public/js/views/serverside/ez-contenttypeeditserversideview.js
+++ b/Resources/public/js/views/serverside/ez-contenttypeeditserversideview.js
@@ -15,6 +15,12 @@ YUI.add('ez-contenttypeeditserversideview', function (Y) {
             '.ez-relation-pick-root-button': {
                 'tap': '_pickRoot'
             },
+            '.ez-selection-add-option-button': {
+                'tap': '_addSelectionOption'
+            },
+            '.ez-selection-delete-option-button': {
+                'tap': '_deleteSelectionOption'
+            },
         };
 
     /**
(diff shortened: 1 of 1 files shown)


glye:
@mhyndle @dpobel @yannickroger JS-gang question: The delete button(s) are hooked to `deleteSelectionOption()` here:
https://github.com/ezsystems/PlatformUIBundle/blob/ezp-25002_add_options_to_ezselection_v2/Resources/public/js/views/serverside/ez-contenttypeeditserversideview.js#L22
...but of course this doesn't work for new buttons which I append to the DOM here:
https://github.com/ezsystems/PlatformUIBundle/blob/ezp-25002_add_options_to_ezselection_v2/Resources/public/js/views/serverside/ez-contenttypeeditserversideview.js#L98
How can I connect them?

glye:
It is working now, thanks a lot @mhyndle. Strange thing is I'm getting this error
TypeError: this.get(...) is undefined ez-contenttypeeditserversideview.js:115:30
https://github.com/ezsystems/PlatformUIBundle/blob/ezp-25002_add_options_to_ezselection_v2/Resources/public/js/views/serverside/ez-contenttypeeditserversideview.js#L115
which is weird because `optionItem.remove()` below it is actually working. How can an `undefined` lead to a properly initialised Node? Black magic.

glye:
Fixed the javascript error by removing the manual event assignment to the delete button. It's automatically assigned. (That was not the case last week, but I guess some other platformui update introduced it. Voodoo!)

Ready for review, ping @mhyndle @dpobel @yannickroger

andrerom:
Visually would be better with just a basket icon to click on instead of a full button or similar, could have same red color as left menu delete option on hover/focus. @dpobel ?

glye:
@andrerom Your proposal sounds good, I just can't find the red trash bin you mention in the left menu (content tree). Anything special needed to enable it? CC @dpobel

andrerom:
@glye Don't think there is a red one, it is a font, so you just need to set (font) color on `:hover` and `:active` *(for touch screens)*.

Resources/public/js/views/serverside/ez-contenttypeeditserversideview.js
```diff
@@ -68,5 +74,43 @@ YUI.add('ez-contenttypeeditserversideview', function (Y) {
...
+         * new option field to the form.
+         *
+         * @method _addSelectionOption
+         * @protected
+         * @param {EventFacade} e
+         */
+        _addSelectionOption: function (e) {
+            var button = e.target,
+                rootOptionsList = this.get('container').one(button.getAttribute('data-relation-root-options-list-id')),
+                prototype = rootOptionsList.getAttribute('data-prototype'),
```
dpobel: `prototype` is very bad name, this word has a strong meaning in the JavaScript world

dpobel:
as written in https://github.com/ezsystems/repository-forms/pull/56, the whole approach to be thought. In addition:
* shouldn't we prevent the selection to not have any options ?
* for long options list, I guess it would be better to have a set of checkboxes one delete button to remove the checked options.

Resources/public/js/views/serverside/ez-contenttypeeditserversideview.js
glye: @dpobel The concept is called `prototype` in Symfony. We can of course call it `formFieldPrototype`, or find another word altogether, but it will still refer to `form.option.vars.prototype` in twig.

glye:
Thanks @dpobel. Indeed we should make sure there is at least one option, or perhaps two (though in the multiple choice scenario, "no choice" is a valid choice, so one would be enough).

Checkboxes make some things easier. But there is also precedent (e.g. by Google) for using a simple "X" to directly remove form fields. The unlabelled red trashbins @andrerom suggested would be the equivalent of that. Either way is fine with me.

dpobel:
Closed in favor of https://github.com/ezsystems/PlatformUIBundle/pull/739

PR #421, opened 2015-11-03
Fix EZP-25051: email address is always required

mhyndle:
Jira: https://jira.ez.no/browse/EZP-25051

## Description
It fixes regression introduced by https://jira.ez.no/browse/EZP-24962 where email address field was always required, no matter if it was defined to be required or not.
This PR changes the validation mechanism to check valid with regexp only if field is not empty.


mhyndle:
Ready for review

dpobel:
no regression test?

mhyndle:
Added the regression test - empty field should be valid if the field is not required.

dpobel:
+1

andrerom:
+1

dpobel:
@mhyndle this one can be merged

StephaneDiot:
+1

PR #423, opened 2015-11-04
[WIP] Fix EZP-25050: BDD for EZP-24569: Automatic FieldDefinition position

glye:
> Implementing https://jira.ez.no/browse/EZP-25050
> Ref: https://jira.ez.no/browse/EZP-24569
> Status: Backlog
- [ ] Scenarios
- [ ] Implementation

### DIFF
--- a/Features/ContentTypeManagement/field_definitions_automatic_ordering.feature
+++ b/Features/ContentTypeManagement/field_definitions_automatic_ordering.feature
@@ -1,0 +1,74 @@
+Feature: Automatic ordering of field definitions when editing a content type
+  In order to manage content types
+  As an Administrator user
+  I need to be able to see field definitions sorted in a sensible, predictable order
+
+  Background:
+    Given I am logged in as admin on PlatformUI
+
+  ##
+  #
+  # Order by position
+  #
+  ##
+  @javascript @common
+  Scenario: Order field types of Article content type by position
+    Given an "Article" content type exists
+    And I am on the "Article" content type view page
+    When I edit the "Article" content type
+    Then I should see the following field definitions:
+      | Title             |
+      | Short title       |
+      | Summary           |
+      | Body              |
+      | Subscriber teaser |
+      | Image             |
+      | Caption (Image)   |
+      | Location          |
+      | Author            |
+      | Publish date      |
+      | Star Rating       |
+      | Tags              |
+
+  @javascript @common
+  Scenario: Change field type position of Article content type
+    Given an "Article" content type exists
+    And I am on the "Article" content type view page
+    When I edit the "Article" content type
+    And I set the position of "Title" to 2
+    And I set the position of "Short title" to 1
+    And I click the "Apply"-button
+    Then I should see the following field definitions:
+      | Short title       |
+      | Title             |
+      | Summary           |
+      | Body              |
+      | Subscriber teaser |
+      | Image             |
+      | Caption (Image)   |
+      | Location          |
+      | Author            |
+      | Publish date      |
+      | Star Rating       |
+      | Tags              |
+
+  @javascript @common
+  Scenario: Change field type position of Article content type, default to order by identifier
+    Given an "Article" content type exists
+    And I am on the "Article" content type view page
+    When I edit the "Article" content type
+    And I set the position of "Image" to 5
+    And I click the "Apply"-button
+    Then I should see the following field definitions:
+      | Short title       |
+      | Title             |
+      | Summary           |
+      | Body              |
+      | Image             |
+      | Subscriber teaser |
+      | Caption (Image)   |
+      | Location          |
+      | Author            |
+      | Publish date      |
+      | Star Rating       |
+      | Tags              |


Features/ContentTypeManagement/field_definitions_automatic_ordering.feature
```diff
@@ -0,0 +1,74 @@
...
+  #
+  # Order by position
+  #
+  ##
+  @javascript @common
+  Scenario: Order field types of Article content type by position
+    Given an "Article" content type exists
+    And I am on the "Article" content type view page
+    When I edit the "Article" content type
+    Then I should see the following field definitions:
```
bdunogier: I think that this list can now be shortened, with `clean` being the install used by Behat.

Features/ContentTypeManagement/field_definitions_automatic_ordering.feature
```diff
@@ -0,0 +1,74 @@
...
+      | Tags              |
+
+  @javascript @common
+  Scenario: Change field type position of Article content type
+    Given an "Article" content type exists
+    And I am on the "Article" content type view page
+    When I edit the "Article" content type
+    And I set the position of "Title" to 2
+    And I set the position of "Short title" to 1
+    And I click the "Apply"-button
```
bdunogier: Extra hyphen.

bdunogier:
It's overall a good start. Got a couple background remarks, inline.

Features/ContentTypeManagement/field_definitions_automatic_ordering.feature
bdunogier: While this sentence is generic, I'm not sure that it really "covers" what we want to verify.

This scenario alone doesn't make that much sense: is it really an expectation that you see the fields in a particular order ? It feels more like an assertion of the system's state, which should probably not be done as a scenario.

Maybe i'd remove this particular scenario, as it doesn't have any real action.

Features/ContentTypeManagement/field_definitions_automatic_ordering.feature
```diff
@@ -0,0 +1,74 @@
...
+      | Image             |
+      | Caption (Image)   |
+      | Location          |
+      | Author            |
+      | Publish date      |
+      | Star Rating       |
+      | Tags              |
+
+  @javascript @common
+  Scenario: Change field type position of Article content type
```
bdunogier: We can probably make that a bit more specific. What about `Reorder field definitions by setting arbirtrary, non overlapping values` ? Maybe a bit long, but the expectations are clear.

Features/ContentTypeManagement/field_definitions_automatic_ordering.feature
```diff
@@ -0,0 +1,74 @@
...
+    Given an "Article" content type exists
+    And I am on the "Article" content type view page
+    When I edit the "Article" content type
+    And I set the position of "Title" to 2
+    And I set the position of "Short title" to 1
+    And I click the "Apply"-button
+    Then I should see the following field definitions:
```
bdunogier: Here we go again.

We should first ensure that the fields we are interested in are ordered as expected in the Given part (it used to be done in the scenario I suggested you remove).

Then we should check that what was done had the effect we expected:

```
Given that "Title" has the position 1
  And "Short title" has the position 2
...
Then "Title" has the position 2
 And "Short title has the position 1
 And the other field definitions were untouched
```

What do you think ?

Features/ContentTypeManagement/field_definitions_automatic_ordering.feature
```diff
@@ -0,0 +1,74 @@
...
+      | Publish date      |
+      | Star Rating       |
+      | Tags              |
+
+  @javascript @common
+  Scenario: Change field type position of Article content type, default to order by identifier
+    Given an "Article" content type exists
+    And I am on the "Article" content type view page
+    When I edit the "Article" content type
+    And I set the position of "Image" to 5
```
bdunogier: I'd abstract this a bit. What about `And I set the position of a field to a position that is already used` ?

Features/ContentTypeManagement/field_definitions_automatic_ordering.feature
```diff
@@ -0,0 +1,74 @@
...
+    And I click the "Apply"-button
+    Then I should see the following field definitions:
```
bdunogier: We should describe what we expect to happen (we don't _really_ expect a list, we scan the list to figure out if our actions had the expected result => implementation).

Following `And I set the same position for two field definitions`, I'd go for:
```
Then both fields are next to each other
 And the first one is at the given position
 And they are alphabetically ordered
```

Maybe a bit verbose, but each sentence should be fairly simple.

bdunogier:
Hey there. Do you need more help, or are you able to move forward with what was said here ?

glye:
@bdunogier As far as I remember, the only thing this one lacks is priority. Will try to kick it into motion again.

bdunogier:
Sent the story back to the backlog then.

I'd almost close the PR for now as well, but I'd rather not lose the branch completely.

glye:
Whooops. Archaeological cleanup closing. Not a priority for me for a long while.

PR #427, opened 2015-11-12
EZP-24960: Create the YUI combo controller

michalpipa:
> JIRA issue: https://jira.ez.no/browse/EZP-24960
> Status: ready for review

Controller for combo loader. Combines JS ans CSS files.

Needs to be merged after #413.


Controller/PlatformUIController.php
```diff
@@ -33,6 +42,28 @@ public function shellAction()
...
     }
 
+    /**
+     * Load JS or CSS assets.
+     *
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function yuiCombineLoaderAction(Request $request)
```
bdunogier: I think I would have created a new controller.

Resources/config/default_settings.yml
```diff
@@ -1,6 +1,8 @@
 parameters:
     ez_platformui.default.yui: {}
     ez_platformui.default.yui.filter: "min"
+    ez_platformui.default.yui.combine: true
+    ez_platformui.default.yui.build_dir: /bundles/ezplatformuiassets/vendors/yui3/build/
```
bdunogier: Hmmm, odd question maybe, but why do we need a siteaccess aware setting here ? PlatformUI itself isn't really siteaccess aware...

Loader/ComboLoader.php
```diff
@@ -0,0 +1,147 @@
...
+ * File containing the ComboLoader class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\PlatformUIBundle\Loader;
+
+use eZ\Publish\Core\Base\Exceptions\NotFoundException;
+
+class ComboLoader implements Loader
```
bdunogier: I'd like a bit of phpdoc about this implementation of the Loader interface, and what it does in particular.

Controller/PlatformUIController.php
michalpipa: I was thinking about it. I decided to leave it here, because shell action has similar purpose (admin UI loading).

But no problem for me, I can move this to separate file.

Resources/config/default_settings.yml
dpobel: well, the thing is the whole JS module settings system was made siteaccess aware so for consistency I guess that's OK

Loader/Loader.php
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * JS and CSS loader.
+ */
+interface Loader
+{
+    /**
+     * @param array $files Array of file names to be loaded.
+     *
+     * @return array Array which elements are JS/CSS content and content type
```
bdunogier: I'd really rather have two functions, one to get the content, and another one to get the type.

I know that the type is detected while iterating over the items, but you could very easily extract this part, and only check the first (or last) one.

dpobel:
I just tested and it works well :) even though it breaks some asynchronous loading done by AlloyEditor/CKEditor. I made a quick and dirty workaround in https://github.com/ezsystems/PlatformUIBundle/commit/658598b81a2bd9f09c5ea095884e678fd4458f96 but while you are fixing the issues mentioned by @bdunogier and working on the last sub-task of the story I'll try to find a cleaner solution to that.

Controller/PlatformUIController.php
bdunogier: Don't bother, we can do that later.

Resources/config/default_settings.yml
bdunogier: Yeah... we've talked about it, and it's indeed better to favour consistency here. But the fact is that overall, I'm expecting bad surprises if somebody relies on the siteaccess awareness of PlatformUI :-)

bdunogier:
Looks good, with the remarks taken into account.

Could you please add at least a unit test for the loader method(s) ? Might be even better to cover this with a BDD scenario, but at least a unit test.

michalpipa:
@dpobel Do we need to check if any file was passed to this controller? If yes, then what should it return in case of empty files array? HTTP 400 status code?

We can also assume that there always will be at least one file passed.

dpobel:
I guess you should check if at least a file is passed and indeed a 400 if no file is provided seems to be the correct status code.

michalpipa:
Rebased because of conflicts. Also added unit test and check if array of files to combine is not empty.

Loader/Loader.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Returns combined JS or CSS files.
+     *
+     * @param array $files File name array
+     *
+     * @return string Combined files
+     *
+     * @throws \eZ\Publish\Core\Base\Exceptions\NotFoundException
+     */
+    public function getContent(array $files);
```
andrerom: Maybe `getFileContent`

Loader/Loader.php
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Returns content type.
+     *
+     * @param array $files File name array
+     *
+     * @return string
+     * @throws \eZ\Publish\Core\Base\Exceptions\NotFoundException
+     */
+    public function getContentType(array $files);
```
andrerom: Like above, mainly to avoid using the domain names from repository, something like `getCombinedFileContentType`, or `getMediaType` or ..

andrerom:
@dpobel  is this supposed to support \@import ?

dpobel:
@andrerom not really but actually we won't use the combo loader for CSS but YUI dynamically load some CSS with it that's why we need to have a minimal support for CSS.

andrerom:
> @andrerom not really but actually we won't use the combo loader for CSS but YUI dynamically load some CSS with it that's why we need to have a minimal support for CSS.

ok, let's make sure to document that then so we don't end up getting feature requests for something we don't intend to use it for *(there are anyway better solutions for this out there)*.

Controller/PlatformUIController.php
```diff
@@ -8,16 +8,27 @@
...
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
 class PlatformUIController extends Controller
 {
     /** @var Provider */
     private $configAggregator;
 
-    public function __construct(Provider $configAggregator)
+    /** @var  Loader */
```
lolautruche: CS: extra WS

Controller/PlatformUIController.php
lolautruche: I wouldn't prefix with `yui` as there is nothing specific to YUI, at least code wise.

Loader/ComboLoader.php
```diff
@@ -0,0 +1,188 @@
...
+        $this->yuiBuildDirectory = $yuiBuildDirectory;
+        $this->webDirectory = $webDirectory;
+    }
+
+    /**
+     * Returns combined JS or CSS files.
+     *
+     * @param array $files File name array
+     *
+     * @return string Combined files
```
andrerom: "The content of the combined files"

Loader/Loader.php
lolautruche: +1

andrerom:
Besides open nitcpicks; +1

Controller/PlatformUIController.php
bdunogier: CS as well: please use the FQN in PhpDoc (phpstorm tip: ctrl+shift+alt+c on the class name, then paste).

michalpipa: Great tip! Thanks.

Loader/ComboLoader.php
```diff
@@ -0,0 +1,188 @@
...
+
+                $content .= $this->loadFile($file, $ext);
+            } elseif ($ext === 'css') {
+                $file = $this->fixUnresolvedPath($file);
+
+                $content .= $this->loadFile($file, $ext);
+            } else {
+                throw new NotFoundException('file type', $file);
+            }
+        }
```
lolautruche: A `switch` would have been more readable IMHO.

Loader/ComboLoader.php
```diff
@@ -0,0 +1,188 @@
...
+     * @param $file
+     * @param $ext
+     *
+     * @return string
+     * @throws \eZ\Publish\Core\Base\Exceptions\NotFoundException
+     */
+    private function loadFile($file, $ext)
+    {
+        $filename = $this->fixFilename($file, $ext);
+        $fileContent = @file_get_contents($filename);
```
lolautruche: Could you please use Filesystem component instead?

Loader/ComboLoader.php
```diff
@@ -0,0 +1,188 @@
...
+
+        $ext = $this->findFileExtension($file);
+
+        if ($ext === 'js') {
+            $type = 'application/javascript';
+        } elseif ($ext === 'css') {
+            $type = 'text/css';
+        } else {
+            throw new NotFoundException('file type', $file);
+        }
```
lolautruche: Switch as well maybe

Loader/ComboLoader.php
```diff
@@ -0,0 +1,188 @@
...
+     *
+     * @return string
+     * @throws \eZ\Publish\Core\Base\Exceptions\InvalidArgumentException
+     * @throws \eZ\Publish\Core\Base\Exceptions\NotFoundException
+     */
+    public function getContentType(array $files)
+    {
+        $this->assertAtLeastOneFile($files);
+
+        $file = key($files);
```
lolautruche: Why `key()`? This will return the index of the current element. Shouldn't you use `reset()` or `current()` instead?

Loader/ComboLoader.php
```diff
@@ -0,0 +1,188 @@
...
+
+            if ($ext === 'js') {
+                if ($this->isYuiFile($file)) {
+                    $file = $this->fixUnresolvedPath($file);
+                }
+
+                $content .= $this->loadFile($file, $ext);
+            } elseif ($ext === 'css') {
+                $file = $this->fixUnresolvedPath($file);
+
```
lolautruche: empty line not needed.

Loader/ComboLoader.php
```diff
@@ -0,0 +1,188 @@
...
+     * Replace underscores by dots.
+     *
+     * @param $filename
+     * @param $ext
+     *
+     * @return string
+     */
+    private function fixFilename($filename, $ext)
+    {
+        return $this->webDirectory . str_replace('_' . $ext, '.' . $ext, $filename);
```
lolautruche: By making `yuiBuildDirectory` relative to `webDirectory`, you wouldn't have to add the webdirectory here :wink:

Loader/ComboLoader.php
michalpipa: Well, this is because `files` array has file names stored in keys. But I could be a nice idea to swap keys and values in the controller and pass them to loader in values.

lolautruche: Yes, it would make this code more obvious.

Loader/ComboLoader.php
```diff
@@ -0,0 +1,188 @@
...
+     * @throws \eZ\Publish\Core\Base\Exceptions\NotFoundException
+     */
+    private function loadFile($file, $ext)
+    {
+        $filename = $this->fixFilename($file, $ext);
+        $fileContent = @file_get_contents($filename);
+
+        if ($fileContent === false) {
+            throw new NotFoundException('file', $file);
+        }
```
lolautruche: Instead of using `@` silent operator, first check if file exists. If not, throw the exception.

Loader/ComboLoader.php
michalpipa: I think that for two choices `if/elseif/else` is still readable, but if you insist I can change it.

lolautruche: I'm not saying it's not readable at all, just that it *would* be *more* readable with a switch.
But this is not critical at all...

Loader/ComboLoader.php
michalpipa: But `yuiBuildDirectory` is relative to `webDirectory` (it's used only when path is not fully resolved). Can you explain more what do you mean?

lolautruche: I mean that in the service definition, you could pass only one parameter. But currently the YUI build directory is a siteaccess aware setting, so it's not possible as is. 
So then, questions are:
* Does this parameter need to be siteaccess aware?
* If not (which is what I think), why not building a single parameter like this: `ez_platformui.default.yui.build_dir: %webroot_dir%/bundles/ezplatformuiassets/vendors/yui3/build/`? This should simplify the code a bit.

michalpipa: @lolautruche `yuiBuildDirectory` is not always used. 

YUI file path is not fully resolved, so we need to prefix it with `webDirectory` and `yuiBuildDirectory`. But custom file path is fully resolved and it needs `webDirectory` prefix only.

Ref: https://github.com/michalpipa/PlatformUIBundle/blob/f4f9c1ab350b48ef0071635f4c4a60ee9f02872d/Loader/ComboLoader.php#L48-L52

lolautruche: ok then

Loader/ComboLoader.php
```diff
@@ -0,0 +1,169 @@
...
+     *
+     * @return string
+     * @throws \eZ\Publish\Core\Base\Exceptions\NotFoundException
+     */
+    private function loadFile($file, $ext)
+    {
+        $filename = $this->fixFilename($file, $ext);
+
+        if (is_readable($filename)) {
+            $fileContent = @file_get_contents($filename);
```
lolautruche: Please remove the `@` silent operator. It's useless here.

michalpipa: Thanks, I forgot to remove it.

Loader/ComboLoader.php
andrerom: But his argument that `ez_platformui.default.yui.build_dir` does not need to be siteaccess aware is still valid.
either you override the yui folder arg or you don't, no value in doing it per siteaccess.

michalpipa: @andrerom We've discussed this: https://github.com/ezsystems/PlatformUIBundle/pull/427#discussion_r44634948

But if needed, I can change this.

Resources/config/default_settings.yml
lolautruche: > well, the thing is the whole JS module settings system was made siteaccess aware so for consistency I guess that's OK

Yes, but here it is completely useless

Controller/PlatformUIController.php
```diff
@@ -33,6 +44,33 @@ public function shellAction()
...
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function combineLoaderAction(Request $request)
+    {
+        $files = array_keys($request->query->all());
+
+        try {
+            $content = $this->loader->getCombinedFileContent($files);
+            $type = $this->loader->getCombinedFileContentType($files);
```
bdunogier: I'd actually get the types first, and the contents after that: if there is an error in the types,

Loader/ComboLoader.php
```diff
@@ -0,0 +1,169 @@
...
+    public function getCombinedFileContent(array $files)
+    {
+        $this->assertAtLeastOneFile($files);
+
+        $content = '';
+
+        foreach ($files as $file) {
+            $ext = $this->findFileExtension($file);
+
+            if ($ext === 'js') {
```
bdunogier: As a sequel to the comment above about getting the type first, then the content.

We could maybe simplify this a bit. Do the type consistency check in `getCombinedFilesContentType` only, and assume in this `getCombinedFileContent` that the list is consistent. We can update this method's description and say that it combines a valid list of files.

This would restrict type consistency checks in one place, and this method would become much simpler.

crevillo: i'd add we could change 'js' and 'css' to constants?

Loader/ComboLoader.php
```diff
@@ -0,0 +1,169 @@
...
+     *
+     * @return string
+     * @throws \eZ\Publish\Core\Base\Exceptions\NotFoundException
+     */
+    private function loadFile($file, $ext)
+    {
+        $filename = $this->fixFilename($file, $ext);
+
+        if (is_readable($filename)) {
+            $fileContent = file_get_contents($filename);
```
crevillo: is the else needed? i mean, can't we ```return file_get_contents($filename)``` and throw without the else?

bdunogier: Good remark. Return asap works well in that case.

Loader/Loader.php
```diff
@@ -0,0 +1,38 @@
...
+     * Returns combined JS or CSS files.
+     *
+     * @param array $files File name array
+     *
+     * @return string The content of the combined files
+     *
+     * @throws \eZ\Publish\Core\Base\Exceptions\InvalidArgumentException
+     * @throws \eZ\Publish\Core\Base\Exceptions\NotFoundException
+     */
+    public function getCombinedFileContent(array $files);
```
bdunogier: Naming: instead of "getCombined", we could go for the "combine" verb: ``combineFilesContent` (there are multiple files).

Resources/config/routing.yml
```diff
@@ -14,3 +14,8 @@ template_yui_module:
 ez_pjax:
     resource: "@eZPlatformUIBundle/Resources/config/routing_pjax.yml"
     prefix:   /pjax
+
+yui_combo_loader:
+    path: /combo
```
bdunogier: Should we go for something a bit less likely to collide ? If a content is named "Combo", we're gonna have issues :-)

Loader/ComboLoader.php
bdunogier: Or we could remove type check from this method, and get rid of the control structure altogether :-)

Resources/config/routing.yml
michalpipa: That is true. I've used route name suggested in Jira issue.

Any suggestions? /cc @dpobel @andrerom

lolautruche: `/_ez_combo_resources`?

bdunogier: Why not.

dpobel: from a performance perspective, it's better if we keep it short, `/_combo` or `/_ezcombo` or whatever :)

bdunogier:
What are the open issues left on this ?

michalpipa:
> What are the open issues left on this ?

No issues left.

lolautruche:
+1

andrerom:
Failing:
>  [eZ Rest Model tests]: Passed: 17 Failed: 1 Total: 18 (ignored 0) (0.223 seconds)

Any insight on this?

dpobel:
weird failure (and weird report), this has nothing to do with this PR, I'm restarting the job.

dpobel:
@michalpipa: can you cherry-pick https://github.com/ezsystems/PlatformUIBundle/commit/658598b81a2bd9f09c5ea095884e678fd4458f96 into this branch so that merging it does not break the RichText editor (I don't have the time to find a better fix).

+1

michalpipa:
@dpobel Rebased PR and cherry-picked 658598b81a2bd9f09c5ea095884e678fd4458f96

andrerom:
+1

PR #428, opened 2015-11-12
Fix EZP-25031: Inconsistent editor state after <Backspace>

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25031

# Description

Before, this patch, if a user uses Backspace in an empty RichText field, the editor is completely messed up. This is happening because the editor is configured on a non editable element containing the root `section` element which is made editable. As a result, the `section` element can be removed.

To avoid that, this patch changes the way the editor is configured so that the `section` becomes the *editor element* so that it can not be removed. As a benefit, this avoids having to deal with the custom namespace like I did in [EZP-24907](https://jira.ez.no/browse/EZP-24907) / https://github.com/ezsystems/PlatformUIBundle/pull/382

# Tests

manual test + unit tests

### DIFF
--- a/Resources/public/js/views/fields/ez-richtext-editview.js
+++ b/Resources/public/js/views/fields/ez-richtext-editview.js
@@ -16,10 +16,7 @@ YUI.add('ez-richtext-editview', function (Y) {
         L = Y.Lang,
         FOCUS_CLASS = 'is-focused',
         EDITOR_FOCUSED_CLASS = 'is-editor-focused',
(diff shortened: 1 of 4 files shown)


dpobel:
ping @yannickroger @mhyndle @StephaneDiot @andrerom

yannickroger:
+1

mhyndle:
+1

andrerom:
+1

PR #430, opened 2015-11-12
EZP-24871: use ancestor criterion for getting breadcrumb

mhyndle:
> status: ready for review

Jira: https://jira.ez.no/browse/EZP-24871

## Description:
Currently breadcrumb in location view is builded by recursively loading parent locations, which is not very efficient.
This PR introduces `loadPath` method in location model which is loading the path using location query and then it is used for setting the `path` attribute in location view service. Thanks to that number of calls to API can be reduced to single one.
Small comparision of number of REST API calls needed to get the path:

Number of REST API calls | previous method | introduced method
----------------------------------- | ---------------------- | -----------------
Depth 0 (Home) | 0 | 1
Depth 1 (/2/14) | 1 | 1
Depth 2 (/2/14/23) | 2 | 1
Depth 5 (/2/14/23/51/62/72) | 5 | 1

## Tasks:
- [x] implementation
- [x] tests
 - [x] manual tests
 - [x] unit tests

### DIFF
--- a/Resources/public/js/models/ez-locationmodel.js
+++ b/Resources/public/js/models/ez-locationmodel.js
@@ -178,6 +178,73 @@ YUI.add('ez-locationmodel', function (Y) {
                 location.reset();
                 callback(error, response);
             });
+        },
+
+        /**
+         * Loads path of the current location. The result is the array containing
+         * eZ.Location objects present on the path sorted by depth.
+         * The array doesn't contain current location.
+         *
+         * @method loadPath
+         * @param {Object} options
+         * @param {Object} options.api the JS REST client instance
+         * @param {Function} callback
+         * @param {false|Error} callback.error
+         * @param {Array} callback.locations the array of Locations
+         */
+        loadPath: function (options, callback) {
+            var locations = [],
+                that = this;
+
+            this._loadAncestors(options, function (error, response) {
+                var hits;
+
+                if (error) {
+                    callback(error);
+                    return;
+                }
+
+                hits = response.document.View.Result.searchHits.searchHit;
+
+                Y.Array.each(hits, function (hit) {
+                    if (hit.value.Location._href !== this.get('id')) {
+                        var location = new Y.eZ.Location();
+
+                        location.setAttrs(location.parse({document: hit.value}));
+                        locations.push(location);
+                    }
+                }, that);
+
+                locations.sort(function (a, b) {
+                    return (a.get('depth') - b.get('depth'));
+                });
+
+                callback(error, locations);
+            });
+        },
+
+        /**
+         * Loads ancestors of the location basing on the `pathString`. The result is the REST API view
+         * containing locations from the path.
+         *
+         * @method _loadAncestors
+         * @protected
+         * @param {Object} options
+         * @param {Object} options.api the JS REST client instance
+         * @param {Function} callback
+         */
+        _loadAncestors: function (options, callback) {
+            var contentService = options.api.getContentService(),
+                query = contentService.newViewCreateStruct('ancestors-' + this.get('locationId'), 'LocationQuery');
+
+            query.body.ViewInput.LocationQuery.Criteria = {
+                AncestorCriterion: this.get('pathString')
+            };
+
+            contentService.createView(
+                query,
+                callback
+            );
         }
     }, {
         REST_STRUCT_ROOT: "Location",
--- a/Resources/public/js/views/services/ez-locationviewviewservice.js
+++ b/Resources/public/js/views/services/ez-locationviewviewservice.js
@@ -273,8 +273,7 @@ YUI.add('ez-locationviewviewservice', function (Y) {
                 request = this.get('request'),
                 service = this,
                 location = this.get('location'), content = this.get('content'),
-                type = this.get('contentType'),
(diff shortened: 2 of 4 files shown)


Resources/public/js/models/ez-locationmodel.js
```diff
@@ -178,6 +178,66 @@ YUI.add('ez-locationmodel', function (Y) {
...
+                }
+
+                hits = response.document.View.Result.searchHits.searchHit;
+
+                Y.Array.each(hits, function (hit) {
+                    var location = new Y.eZ.Location();
+
+                    location.setAttrs(location.parse({document: hit.value}));
+
+                    if (location.get('id') !== that.get('id')) {
```
dpobel: you should check that before parsing a creating the corresponding Location object.

dpobel: nitpicl: and instead of using `that`, you can set the context of Y.Array.each function with its 3rd argument: http://yuilibrary.com/yui/docs/api/classes/Array.html#method_each

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -178,6 +178,66 @@ YUI.add('ez-locationmodel', function (Y) {
...
             });
+        },
+
+        /**
+         * Loads path of the current location. The result is the array containing
+         * eZ.Location objects present on the path. The array doesn't contain current location.
+         *
+         * @param {Object} options
+         * @param {Object} options.api the JS REST client instance
+         * @param {Function} callback
```
dpobel: you should document that the callback receives an error and the result as an array with something like:
```
@param {Function} callback
@param {false|Error} callback.error
@param {Array} callback.locations the array of Locations
```

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -427,7 +359,6 @@ YUI.add('ez-locationviewviewservice', function (Y) {
         _newLocation: function (params) {
```
dpobel: this one can also be removed

mhyndle:
Ready for review

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -307,21 +306,14 @@ YUI.add('ez-locationviewviewservice', function (Y) {
                 });
 
                 endLoadPath = tasks.add();
-                discoveryService.getInfoObject('rootLocation', function (error, response) {
-                    var rootLocationId;
-
+                location.loadPath(loadOptions, function(error, response) {
```
dpobel: nitpick: in this context, `response` is a bad name, the variable should be called `locations` or `path`.

dpobel:
The `path` Location view service attribute has a `getter` function to sort the path by depth. I don't know if it's still needed (I guess it is) but if it is this logic should be moved to the `Location#loadPath` method.

mhyndle:
@dpobel I've moved the sorting mechanism from locationViewViewService to the location model

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -178,6 +178,72 @@ YUI.add('ez-locationmodel', function (Y) {
                 location.reset();
                 callback(error, response);
             });
+        },
+
+        /**
+         * Loads path of the current location. The result is the array containing
+         * eZ.Location objects present on the path. The array doesn't contain current location.
```
dpobel: *eZ.Location objects present on the path sorted by depth.*

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -178,6 +178,72 @@ YUI.add('ez-locationmodel', function (Y) {
...
+                Y.Array.each(hits, function (hit) {
+                    if (hit.value.Location._href !== this.get('id')) {
+                        var location = new Y.eZ.Location();
+
+                        location.setAttrs(location.parse({document: hit.value}));
+                        locations.push(location);
+                    }
+                }, that);
+
+                locations.sort(function (a, b) {
```
dpobel: Note: that's a workaround because the REST views do not support sorting yet (see [EZP-24315](https://jira.ez.no/browse/EZP-24315), so you should create an improvement in JIRA blocked by EZP-24315 about the removal of this code when EZP-24315 is implemented.

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -504,10 +504,132 @@ YUI.add('ez-locationmodel-tests', function (Y) {
...
+            delete this.model;
+        },
+
+        "Should load path for the location": function () {
+            var that = this, i;
+
+            Y.Mock.expect(this.contentService, {
+                method: 'createView',
+                args: [Y.Mock.Value.Object, Y.Mock.Value.Function],
+                run: function (options, callback) {
```
dpobel: `options` should be renamed to `query` and you should assert it contains a query with the ancestor criterion correctly configured.

Resources/public/js/models/ez-locationmodel.js
mhyndle: I have created issue for that: https://jira.ez.no/browse/EZP-25099

dpobel:
+1

yannickroger:
+1

PR #431, opened 2015-11-12
Fix EZP-25091: "always available" flag is always set on content creation

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25091
Requires https://github.com/ezsystems/ez-js-rest-client/pull/68

# Description

This patch makes sure we set the "always available" flag value based on the `defaultAlwaysAvailable` value set in the content type when creating a content item.

# Tests

unit tests + manual tests


dpobel:
ping @mhyndle @yannickroger @StephaneDiot @andrerom

mhyndle:
+1

andrerom:
+1

PR #432, opened 2015-11-16
EZP-25091: Impossible to translate a content without the the always available flag

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24766

# Description

It's impossible to translate a content without the "always available" flag. This issue was "hidden" because of [EZP-25091](https://jira.ez.no/browse/EZP-25091) (see https://github.com/ezsystems/ez-js-rest-client/pull/68 and https://github.com/ezsystems/PlatformUIBundle/pull/431).

So basically, this patch adds the loading of the content info to check if the content exists and then adds the necessary logic so that no matter if the content is always available or not, it can be translated.

# Tests

manual tests + unit tests

### DIFF
--- a/Resources/config/yui.yml
+++ b/Resources/config/yui.yml
@@ -137,6 +137,7 @@ system:
                         - 'parallel'
                         - 'ez-viewservice'
                         - 'ez-contentmodel'
+                        - 'ez-contentinfomodel'
                         - 'ez-locationmodel'
                         - 'ez-contenttypemodel'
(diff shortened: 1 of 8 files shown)


dpobel:
ping @mhyndle @yannickroger @andrerom @StephaneDiot

andrerom:
+1

yannickroger:
+1

mhyndle:
+1

PR #433, opened 2015-11-16
Fix EZP-25159: Implement missing limitation mappers Node and Subtree

glye:
> Implements https://jira.ez.no/browse/EZP-25159
> Status: Ready for review
> Required by https://github.com/ezsystems/repository-forms/pull/64

Javascript methods for the UDW buttons used by the...
- [x] Location
- [x] Subtree

...limitation mappers.

- [x] UDW should be multiple choice

### DIFF
--- a/Resources/public/js/views/serverside/ez-roleserversideview.js
+++ b/Resources/public/js/views/serverside/ez-roleserversideview.js
@@ -15,6 +15,9 @@ YUI.add('ez-roleserversideview', function (Y) {
             '.ez-role-assign-button': {
                 'tap': '_pickSubtree'
             },
+            '.ez-pick-location-limitation-button': {
+                'tap': '_pickLocationLimitation'
+            },
         };
 
     /**
@@ -41,10 +44,10 @@ YUI.add('ez-roleserversideview', function (Y) {
          */
         _pickSubtree: function (e) {
             var button = e.target,
-                unsetLoading = Y.bind(this._uiUnsetAssignRoleLoading, this, button);
+                unsetLoading = Y.bind(this._uiUnsetUDWButtonLoading, this, button);
 
             e.preventDefault();
-            this._uiSetAssignRoleLoading(button);
+            this._uiSetUDWButtonLoading(button);
             this.fire('contentDiscover', {
                 config: {
                     title: button.getAttribute('data-universaldiscovery-title'),
@@ -60,26 +63,131 @@ YUI.add('ez-roleserversideview', function (Y) {
         },
 
         /**
-         * Changes the state of the provided assign button to be *loading* and
-         * to be disabled
+         * tap event handler for policy limitation on location ("Node"). It launches the
+         * universal discovery widget so that the user can pick a location.
          *
-         * @method _uiSetAssignRoleLoading
+         * @method _pickLocationLimitation
+         * @protected
+         * @param {EventFacade} e
+         */
+        _pickLocationLimitation: function (e) {
+            var button = e.target,
+                unsetLoading = Y.bind(this._uiUnsetUDWButtonLoading, this, button);
+
+            e.preventDefault();
+            this._uiSetUDWButtonLoading(button);
+            this.fire('contentDiscover', {
+                config: {
+                    title: button.getAttribute('data-universaldiscovery-title'),
+                    cancelDiscoverHandler: unsetLoading,
+                    multiple: true,
+                    contentDiscoveredHandler: Y.bind(this._setLocationLimitation, this, button),
+                },
+            });
+        },
+
+        /**
+         * Puts picked location id into location limitation input. Input is selected by the selector
+         * provided in the `data-location-input-selector` attribute of the button, for example
+         * <button data-location-input-selector="#id_of_input"></button>
+         *
+         * @method _setLocationLimitation
+         * @protected
+         * @param {Y.Node} button
+         * @param {EventFacade} e
+         */
+        _setLocationLimitation: function (button, e) {
+            this._emptyLocationList(button);
+            this._emptyLocationIdHiddenInput(button);
+
+            Y.Array.each(e.selection, function (struct) {
+                this._addLocationToDisplayList(button, struct.contentInfo);
+                this._addLocationIdToHiddenInput(button, struct.location.get('locationId'));
+            }, this);
+
+            this._uiUnsetUDWButtonLoading(button);
+        },
+
+        /**
+         * Empties the existing list of selected locations.
+         *
+         * @method _emptyLocationList
          * @protected
          * @param {Y.Node} button
          */
-        _uiSetAssignRoleLoading: function (button) {
+        _emptyLocationList: function (button) {
+            var selectedLocationList = this.get('container').one(button.getAttribute('data-selected-location-list-selector'));
+
+            selectedLocationList.empty();
+        },
+
+        /**
+         * Empties the content of the hidden input field for selected locations.
+         *
+         * @method _emptyLocationIdHiddenInput
+         * @protected
+         * @param {Y.Node} button
+         */
+        _emptyLocationIdHiddenInput: function (button) {
+            var locationInput = this.get('container').one(button.getAttribute('data-location-input-selector'));
+
+            locationInput.setAttribute('value', '');
+        },
+
+        /**
+         * Adds a location to the list of selected locations.
+         *
+         * @method _addLocationToDisplayList
+         * @protected
+         * @param {Y.Node} button
+         * @param {eZ.ContentInfo} contentInfo
+         */
+        _addLocationToDisplayList: function (button, contentInfo) {
+            var selectedLocationList = this.get('container').one(button.getAttribute('data-selected-location-list-selector'));
+
+            selectedLocationList.appendChild(Y.Node.create('<li>' + Y.Escape.html(contentInfo.get('name')) + '</li>'));
+        },
+
+        /**
+         * Adds a location to the hidden input field for selected locations.
+         *
+         * @method _addLocationIdToHiddenInput
+         * @protected
+         * @param {Y.Node} button
+         * @param {String} locationId
+         */
+        _addLocationIdToHiddenInput: function (button, locationId) {
+            var locationInput = this.get('container').one(button.getAttribute('data-location-input-selector')),
+                existingLocationsStr = locationInput.getAttribute('value');
+
+            if (existingLocationsStr.length > 0) {
+                locationInput.setAttribute('value', existingLocationsStr.concat(',', locationId));
+            } else {
+                locationInput.setAttribute('value', locationId);
+            }
+        },
+
+        /**
+         * Changes the state of the provided UDW button to be *loading* and
+         * to be disabled
+         *
+         * @method _uiSetUDWButtonLoading
+         * @protected
+         * @param {Y.Node} button
+         */
+        _uiSetUDWButtonLoading: function (button) {
             button.addClass('is-loading').set('disabled', true);
         },
 
         /**
-         * Changes the state of the provided assign button to not be *loading*
+         * Changes the state of the provided UDW button to not be *loading*
          * and to be enabled.
          *
-         * @method _uiUnsetAssignRoleLoading
+         * @method _uiUnsetUDWButtonLoading
          * @protected
          * @param {Y.Node} button
          */
-        _uiUnsetAssignRoleLoading: function (button) {
+        _uiUnsetUDWButtonLoading: function (button) {
             button.removeClass('is-loading').set('disabled', false);
         },
     });
--- a/Resources/public/js/views/services/ez-roleserversideviewservice.js
+++ b/Resources/public/js/views/services/ez-roleserversideviewservice.js
@@ -22,8 +22,24 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
     Y.eZ.RoleServerSideViewService = Y.Base.create('roleServerSideViewService', Y.eZ.ServerSideViewService, [], {
         initializer: function () {
             this.on('*:contentDiscover', function (e) {
-                e.config.contentDiscoveredHandler = Y.bind(this._assignRole, this);
+                // If we're not discovering for a role assignment, the contentDiscoveredHandler is already set
+                if (this._isDiscoveringForAssigningRole(e)) {
+                    e.config.contentDiscoveredHandler = Y.bind(this._assignRole, this);
+                }
             });
+        },
+
+        /**
+         * Return true if we are in the process of assigning a role.
+         *
+         * @method _isDiscoveringForAssigningRole
+         * @protected
+         * @param {EventFacade} e
+         */
+        _isDiscoveringForAssigningRole: function (e) {
+            var data = e.config.data;
+
+            return data !== undefined && data.roleId !== undefined;
         },
 
         /**
--- a/Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
+++ b/Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
@@ -35,8 +35,8 @@ YUI.add('ez-roleserversideviewservice-tests', function (Y) {
             delete this.service;
         },
 
-        "Should add the contentDiscovered handler": function () {
-            var config = {};
+        "Should add the contentDiscovered handler on role assignment": function () {
+            var config = { data: { roleId: 42 } };
 
             this.service.fire('whatever:contentDiscover', {
                 config: config,
@@ -45,6 +45,19 @@ YUI.add('ez-roleserversideviewservice-tests', function (Y) {
             Assert.isFunction(
                 config.contentDiscoveredHandler,
                 "The contentDiscovered should have been added"
+            );
+        },
+
+        "Should not add the contentDiscovered handler on policy editing": function () {
+            var config = {};
+
+            this.service.fire('whatever:contentDiscover', {
+                config: config,
+            });
+
+            Assert.isUndefined(
+                config.contentDiscoveredHandler,
+                "The contentDiscovered should not have been added"
             );
         },
     });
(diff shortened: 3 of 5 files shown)


glye:
Review ping @lolautruche @andrerom @bdunogier @mhyndle @dpobel @yannickroger

Resources/public/js/views/serverside/ez-roleserversideview.js
```diff
@@ -60,26 +63,70 @@ YUI.add('ez-roleserversideview', function (Y) {
...
+         * @param {EventFacade} e
+         */
+        _setLocationLimitation: function (button, e) {
+            var locationInput = this.get('container').one(button.getAttribute('data-location-input-selector')),
+                selectedLocation = this.get('container').one(button.getAttribute('data-selected-location-selector')),
+                unsetLoading = Y.bind(this._uiUnsetUDWButtonLoading, this, button);
+
+            locationInput.setAttribute('value', e.selection.location.get('locationId'));
+            selectedLocation.setHTML(e.selection.contentInfo.get('name'));
+            unsetLoading();
```
dpobel: why not just:
```js
this._uiUnsetUDWButtonLoading(button);
```

glye: Why not, indeed.

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -22,7 +22,9 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
     Y.eZ.RoleServerSideViewService = Y.Base.create('roleServerSideViewService', Y.eZ.ServerSideViewService, [], {
         initializer: function () {
             this.on('*:contentDiscover', function (e) {
-                e.config.contentDiscoveredHandler = Y.bind(this._assignRole, this);
+                if (e.config.contentDiscoveredHandler === undefined) {
```
dpobel: this kind of logic feels wrong. I mean this has a very high *WTF potential* if you come back in a month ;-)

joke aside, you should introduce a method called for instance `_isDiscoveringForAssigningRole` (feel free to suggest a better name :)). Also, instead of checking the `contentDiscoveredHandler`, I would rather check that `e.config.data` contains a `roleId`.

dpobel:
unit tests ?

Resources/public/js/views/services/ez-roleserversideviewservice.js
glye: @dpobel Agree. Sadly, my naming imagination is no better than yours.

glye:
> unit tests ?

:scream:

glye:
Ensured that existing unit tests don't fail. Testing of new code is still **very** limited.

andrerom:
+1

glye:
Review ping @lolautruche @bdunogier @mhyndle @dpobel @yannickroger
JS unit tests are still missing, are you OK with doing that in a follow-up, so we can close this? Or is anyone available to help me with them? It's going to take me a while :confounded:

yannickroger:
-1 on merging without unit test, or do it in a follow up. I can work on the unit test if needed. See with @dpobel

glye:
@yannickroger Thanks. The JS is less incomprehensible today, but I'll ask if I get stuck again.

glye:
Added missing unit tests :pray:

glye:
Travis is happy. Not sure how to interpret what Jenkins/Behat is doing, but I see nothing role related there.
Review ping @lolautruche @bdunogier @mhyndle @dpobel @yannickroger

andrerom:
Behat has issues for other reasons +1

bdunogier:
+1 from @andrerom and... that's it.

I can't really bless this code without destroying it. @mhyndle ? @yannickroger ?

Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
```diff
@@ -47,6 +52,63 @@ YUI.add('ez-roleserversideviewservice-tests', function (Y) {
...
+        setUp: function () {
+            this.service = new Y.eZ.RoleServerSideViewService();
+        },
+
+        tearDown: function () {
+            delete this.service;
+        },
+
+        "Should return false when data is not set": function () {
+            var e = {
```
mhyndle: nitpick: can be minimized to single line

Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
```diff
@@ -47,6 +52,63 @@ YUI.add('ez-roleserversideviewservice-tests', function (Y) {
...
+                config: {
+                },
+            };
+
+            Assert.isFalse(this.service._isDiscoveringForAssigningRole(e), 'The method should return false');
+        },
+
+        "Should return false when roleId is not set": function () {
+            var e = {
```
mhyndle: nitpick: can be minimized to single line

Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
```diff
@@ -47,6 +52,63 @@ YUI.add('ez-roleserversideviewservice-tests', function (Y) {
...
+                    data: {
+                    },
+                },
+            };
+
+            Assert.isFalse(this.service._isDiscoveringForAssigningRole(e), 'The method should return false');
+        },
+
+        "Should return true when roleId is set": function () {
+            var e = {
```
mhyndle: nitpick: can be minimized to single line

Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
```diff
@@ -47,6 +52,63 @@ YUI.add('ez-roleserversideviewservice-tests', function (Y) {
...
+            delete this.service;
+        },
+
+        "Should return false when data is not set": function () {
+            var e = {
+                config: {
+                },
+            };
+
+            Assert.isFalse(this.service._isDiscoveringForAssigningRole(e), 'The method should return false');
```
mhyndle: Test should be constructed bit different I think. You shouldn't call directly private/protected methods. You should rather test if the public-available mechanisms are making correct use of them, but as far as I can see it is tested with `"Should not add the contentDiscovered handler on policy editing"` in `contentDiscoverEventTest` test case. So I guess that whole `isDiscoveringForAssigningRoleTest` test case is not needed at all - can you confirm that?

Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
```diff
@@ -129,7 +191,11 @@ YUI.add('ez-roleserversideviewservice-tests', function (Y) {
                 callback = function () {
                     callbackCalled = true;
                 },
-                config = {},
+                config = {
```
mhyndle: nitpick: can be minimized to single line

Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
```diff
@@ -180,7 +246,11 @@ YUI.add('ez-roleserversideviewservice-tests', function (Y) {
...
```
mhyndle: nitpick: can be minimized to single line

Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
```diff
@@ -303,7 +373,11 @@ YUI.add('ez-roleserversideviewservice-tests', function (Y) {
                 contentType = getMockForJson(contentTypeJson),
                 selection = [{contentInfo: contentInfo, contentType: contentType, location: location}],
                 universalDiscovery = new Mock(),
-                config = {},
+                config = {
```
mhyndle: nitpick: can be minimized to single line

Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
```diff
@@ -394,7 +468,11 @@ YUI.add('ez-roleserversideviewservice-tests', function (Y) {
...
```
mhyndle: nitpick: can be minimized to single line

Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
```diff
@@ -466,7 +544,11 @@ YUI.add('ez-roleserversideviewservice-tests', function (Y) {
                 contentInfo = getMockForJson(contentJson),
                 selection = [{contentInfo: contentInfo}],
                 universalDiscovery = new Mock(),
-                config = {},
+                config = {
```
mhyndle: nitpick: can be minimized to single line

Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
```diff
@@ -549,7 +631,11 @@ YUI.add('ez-roleserversideviewservice-tests', function (Y) {
                 contentType = getMockForJson(contentTypeJson),
                 selection = [{contentInfo: contentInfo, contentType: contentType}],
                 universalDiscovery = new Mock(),
-                config = {},
+                config = {
```
mhyndle: nitpick: can be minimized to single line

glye:
This is now related to https://github.com/ezsystems/repository-forms/pull/63 instead.

Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
glye: Yep, that tests it indirectly.

glye:
Removed unneeded test.
Minimised config data.
ping @mhyndle

mhyndle:
+1, seems OK to me but maybe one more JS check would be valuable

glye:
Review ping @dpobel @yannickroger and all other JS mafiosi.

Resources/public/js/views/serverside/ez-roleserversideview.js
```diff
@@ -41,10 +44,10 @@ YUI.add('ez-roleserversideview', function (Y) {
          */
         _pickSubtree: function (e) {
             var button = e.target,
-                unsetLoading = Y.bind(this._uiUnsetAssignRoleLoading, this, button);
+                unsetLoading = Y.bind(this._uiUnsetUDWButtonLoading, this, button);
```
yannickroger: Renaming this was mandatory ? The old name still makes sense, assigning with UDW doesn't make a big difference.

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -22,11 +22,27 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
     Y.eZ.RoleServerSideViewService = Y.Base.create('roleServerSideViewService', Y.eZ.ServerSideViewService, [], {
         initializer: function () {
             this.on('*:contentDiscover', function (e) {
-                e.config.contentDiscoveredHandler = Y.bind(this._assignRole, this);
+                // If we're not discovering for a role assignment, the contentDiscoveredHandler is already set
+                if (this._isDiscoveringForAssigningRole(e)) {
```
yannickroger: is it possible to add an extra parameter to the event facade  to say if we are doing a role assignment or not to avoid trying to figure it out ?

Resources/public/js/views/serverside/ez-roleserversideview.js
glye: @yannickroger With this PR I am not assigning roles, I am editing policy limitations. So I'd say it's a necessary change.

Resources/public/js/views/services/ez-roleserversideviewservice.js
glye: @yannickroger This is what it does - or rather, it removes a parameter, and checks whether it is absent. If I added a boolean for this I could reduce `return data !== undefined && data.roleId !== undefined;` to `return data.isAssigningRole;` - is it worth it? (Did I understand you correctly?)

glye: I could also remove the function `_isDiscoveringForAssigningRole()`, but then I'd have to argue with @dpobel who wanted it in the first place :)

Resources/public/js/views/serverside/ez-roleserversideview.js
yannickroger: So maybe it should be called something that says that (policy limitations edition thing).

Resources/public/js/views/services/ez-roleserversideviewservice.js
yannickroger: Leave the function.
You can leave this part of the code like that. I understood what you wanted to do.

Resources/public/js/views/serverside/ez-roleserversideview.js
glye: @yannickroger It **is** called what it does: unset the loading state of the UDW button. To clarify, the new usage comes in addition to the old one - it is used both when assigning roles, and when editing policy limitations. But what it **does** is the same in both cases. So I named it after what it does :)

glye:
As mentioned in daily, review ping @bdunogier @andrerom @dpobel @yannickroger

bdunogier:
Still +1 for what it's worth. @dpobel should merge this.

glye:
NB: This PR is back to dev / work in progress because of an omission in the repoforms part which will also affect this one.

glye:
Enabled selection of multiple values per limitation.

glye:
Updated test for multi-selection UDW.

StephaneDiot:
+1

Resources/public/js/views/serverside/ez-roleserversideview.js
```diff
@@ -60,26 +63,86 @@ YUI.add('ez-roleserversideview', function (Y) {
...
+         * Puts picked location id into location limitation input. Input is selected by the selector
+         * provided in the `data-location-input-selector` attribute of the button, for example
+         * <button data-location-input-selector="#id_of_input"></button>
+         *
+         * @method _setLocationLimitation
+         * @protected
+         * @param {Y.Node} button
+         * @param {EventFacade} e
+         */
+        _setLocationLimitation: function (button, e) {
```
dpobel: this whole method is very complicated to understand (I had to test it to get what it actually does) because of bad naming and lack of semantic. You should create methods so that the code is self-explanatory.
For instance, this method code should something like:

```js
this._emptyLocationList();
this._emptyLocationIdHiddenInput();

Y.Array.each(e.selection, function (struct) {
    this._addLocationToDisplayList(struct.contentInfo);
    this._addLocationIdToHiddenInput(struct.location.get('locationId'));
}, this);

this._uiUnsetUDWButtonLoading(button);
```

Resources/public/js/views/serverside/ez-roleserversideview.js
```diff
@@ -60,26 +63,86 @@ YUI.add('ez-roleserversideview', function (Y) {
...
+            var locationInput = this.get('container').one(button.getAttribute('data-location-input-selector')),
+                locationIdArray = [],
+                selectedLocation = this.get('container').one(button.getAttribute('data-selected-location-selector')),
+                selectedLocationList = selectedLocation.get('children').item(0);
```
dpobel: the selector in `data-selected-location-selector` should probably directly target this element, that way the code and the markup are less couple. Also, I would rename `data-selected-location-selector` to `data-selected-location-list-selector`

Resources/public/js/views/serverside/ez-roleserversideview.js
```diff
@@ -60,26 +63,86 @@ YUI.add('ez-roleserversideview', function (Y) {
...
+
+            while(selectedLocationList.firstChild) {
```
dpobel: http://yuilibrary.com/yui/docs/api/classes/Node.html#method_empty ?

Resources/public/js/views/serverside/ez-roleserversideview.js
```diff
@@ -60,26 +63,86 @@ YUI.add('ez-roleserversideview', function (Y) {
...
+                selectedLocationList.removeChild(selectedLocationList.firstChild);
+            }
+
+            Y.Array.each(e.selection, function (struct) {
+                var listItem = Y.config.doc.createElement('li');
```
dpobel: you can use [the Node API](http://yuilibrary.com/yui/docs/api/classes/Node.html) which is bit easier.

sunpietro: @dpobel What's the point in using the wrapper on native methods?
What's more, I don't see any point in using `Y.config.doc` instead of `document`.
What are the pros and cons of using Node API in this case and `Y.config.doc`?

dpobel: @sunpietro notice: I wrote *can* not *should* meaning it's not a strong requirement. The pros of using the Node API are the consistency and (maybe) a nicer API.
For `Y.config.doc` over `document`, this might be useful in tests in some cases when you precisely want to check what is done on the document but here it does not really matter as we can check the end result easily.

sunpietro: @dpobel no problem :) sorry for misunderstanding

glye: @dpobel Ack, done.

Resources/public/js/views/serverside/ez-roleserversideview.js
glye: Nice, thanks.

Resources/public/js/views/serverside/ez-roleserversideview.js
glye: Ok, done.

Resources/public/js/views/serverside/ez-roleserversideview.js
glye: @dpobel Readability is in the eyes of the beholder, but I'm not gonna pick a fight with the JS gang, so OK ;)

glye:
Improved according to @dpobel feedback.

andrerom:
+0.9 *( leaving room for js gang to review this fully ;) )*

dpobel:
+1

PR #435, opened 2015-11-16
EZP-25033: search method for UDW

mhyndle:
> status: ready for review

Jira: https://jira.ez.no/browse/EZP-25033

## Description
This PR introduce new content selection method in UDW. The new method is simple search, which allows to make a full text search using existing loaction search plugin.
Additionally, the locationSearchPlugin was changed, so it now allows to load eZ.Content and eZ.ContentType for each location found by passing to the event `loadContent: true` and `loadContentType: true` parameters. Because of that, the result of locationSearch is array of location struct (previously it was array of eZ.Location models). If `loadContent` and `loadContentType` parameters were both set to `true`, then single location struct looks like:
```
{
  location: eZ.Location,
  content: eZ.Content,
  contentType: eZ.ContentType
}
```
If load* flag is `false`, then location struct doesn't contain given field.
This PR also make use of new result of `locationSearch` where it was used before - in subitems view.

## Screencast
YT: https://youtu.be/rZXzblGt_hM
The first part of scrrencast shows how selecting is proceeded in multiple selection method, the second part shows proceeding with single selection part.
**Edit**
Updated screencast with pagination being presented: https://youtu.be/r395vesr4sY

## Tasks
- [x] implementation
- [x] CSS
- [x] tests

## Notes
~~Search results handle pagination, which you can't see currently as pagination is basing on total number of search results (despite limit from pagination) which is not included in the REST response, so currently the total number of records will not be higher than the limit used in search query. There is issue reported for implement the total count for search result: https://jira.ez.no/browse/EZP-25151
Nevertheless, the pagination is handled also so only thing needed is to provide the total count in location search plugin.~~
As https://jira.ez.no/browse/EZP-25151 is done, pagination is working :D


andrerom:
:fireworks:

Resources/public/js/views/services/plugins/ez-universaldiscoverycontentsearchplugin.js
```diff
@@ -0,0 +1,71 @@
...
+    Y.eZ.Plugin.UniversalDiscoveryContentSearch = Y.Base.create('universalDiscoveryContentSearch', Y.eZ.Plugin.ViewServiceBase, [], {
+        initializer: function () {
+            this.onHostEvent('universalDiscoverySearchView:searchContent', this._search);
+        },
+
+        _search: function (e) {
+            var capi = this.get('host').get('capi'),
+                contentService = capi.getContentService(),
+                searchText = e.searchText,
+                query = contentService.newViewCreateStruct('udw-content-search-' + searchText),
```
dpobel: you should do a Location search instead of a Content search because this way you get both the Location and ContentInfo that are needed for the UDW selection handling. But actually, you should reuse the locationSearchPlugin, it does pretty much the same thing (and already support the pagination).

About the needed objects, you'll also need the content type. I think you should add a `loadContentType` option to the `locationSearch` event parameter so that the plugin also load the content type of each result in addition to searching.

mhyndle:
Ready for review

Resources/public/js/views/services/plugins/ez-locationsearchplugin.js
```diff
@@ -96,6 +125,121 @@ YUI.add('ez-locationsearchplugin', function (Y) {
...
+                    endContentLoad = tasks.add(function (error, response) {
+                        if (error) {
+                            loadContentError = true;
+                            return;
+                        }
+
+                        locationStructArr[i].content = content;
+                    });
+
+                content.load(options, endContentLoad);
```
dpobel: instead of loading each content individually, we can load them in one request by doing a content search like it's done in [the content tree plugin](https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/services/plugins/ez-contenttreeplugin.js#L140). It's too late to do that right now though. Can you create a follow up improvement request in JIRA for that ?

Resources/public/js/views/universaldiscovery/ez-universaldiscoverysearchview.js
```diff
@@ -0,0 +1,500 @@
...
+        _fireLocationSearch: function () {
+            var searchText = this.get('searchText');
+
+            this._uiPageLoading();
+
+            this.fire('locationSearch', {
+                viewName: 'udwsearch-' + searchText,
+                resultAttribute: 'searchResultList',
+                resultTotalCountAttribute: 'searchResultCount',
+                loadContentType: true,
```
dpobel: The UDW now has `loadContent` option as well which is passed to the methods https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/universaldiscovery/ez-universaldiscoverymethodbaseview.js#L107
so you should also set the `loadContent` flag on the event object.

Resources/public/js/views/universaldiscovery/ez-universaldiscoverysearchview.js
```diff
@@ -0,0 +1,500 @@
...
+                searchResultCount: this.get('searchResultCount'),
+                searchResultList: this._convertToJSONList(),
+                isFirst: this._isFirstPage(),
+                isLast: this._isLastPage(),
+                hasPages: this._hasPages(),
+                loadingError: this.get('loadingError'),
+                multiple: this.get('multiple')
+            }));
+
+            this.get('selectedView').set('contentStruct', null);
```
dpobel: this has nothing to do with rendering the view, this should not be done here.

Resources/public/js/views/universaldiscovery/ez-universaldiscoverysearchview.js
```diff
@@ -0,0 +1,500 @@
...
+            container.one('.ez-ud-search-selected').append(
+                this.get('selectedView').render().get('container')
+            );
+            this._uiPageEndLoading();
```
dpobel: that's weird to that here as well. I guess this should be done in a `searchResultListChange` event handler.

Resources/public/js/views/universaldiscovery/ez-universaldiscoverysearchview.js
```diff
@@ -0,0 +1,500 @@
...
+        _search: function (e) {
+            var c = this.get('container'),
+                searchInput = c.one(e.target.getAttribute('data-search-input-selector')),
+                searchText = searchInput.get('value');
+
+            e.preventDefault();
+
+            this.set('searchText', searchText);
+            this.set('offset', 0);
+            this._fireLocationSearch();
```
dpobel: to decouple things a bit, I would do that in a after `searchTextChange` event handler

Resources/public/css/views/universaldiscovery/search.css
```diff
@@ -0,0 +1,135 @@
...
+.ez-view-universaldiscoverysearchview .ez-searchresult-type-header {
+    width: 10em;
+}
+
+.ez-view-universaldiscoverysearchview .ez-searchresult-preview-header {
+    width: 5em;
+}
+
+.ez-view-universaldiscoverysearchview .ez-searchresult-row.is-selected {
+    background: #CCE4FF;
```
dpobel: background and transition should go the theme CSS

Resources/public/js/views/universaldiscovery/ez-universaldiscoverysearchview.js
```diff
@@ -0,0 +1,500 @@
...
+         * @param {EventFacade} e
+         */
+        _selectContent: function (e) {
+            var c = this.get('container'),
+                locationId = e.target.getAttribute('data-location-id'),
+                locationStruct = this._getLocationStructFromResultList(locationId),
+                location = locationStruct.location,
+                contentType = locationStruct.contentType,
+                contentInfo = location.get('contentInfo'),
+                contentStruct = {
```
dpobel: here as well, you'll have to add the content when the UDW is configured to load the contents.

Resources/public/js/views/universaldiscovery/ez-universaldiscoverysearchview.js
```diff
@@ -0,0 +1,500 @@
...
+    /**
+     * Provides the universal discovery search method
+     *
+     * @module ez-universaldiscoverysearchview
+     */
+    Y.namespace('eZ');
+
+    var events = {
+            '.ez-ud-search-button': {
+                'tap': '_search',
```
dpobel: it would be very nice to also trigger a search if the user types "Return" on his keyboard (I did it several times by habits when I tested the PR :))

Resources/public/js/views/services/plugins/ez-locationsearchplugin.js
mhyndle: Follow up created in JIRA: https://jira.ez.no/browse/EZP-25169

Resources/public/js/views/universaldiscovery/ez-universaldiscoverysearchview.js
mhyndle: I have added `keydown` event on on the search text input. I'm wondering if there is possibility to triggerr the event only if enter was pushed without checking it inside event handler?

Resources/public/js/views/universaldiscovery/ez-universaldiscoverysearchview.js
mhyndle: So now `loadContent` attr is passed to the locationSearchPlugin call and right here there is added `content` field to the `contentStruct` object containing Content returned inside `locationStruct.content` by locationSearchPlugin.

Resources/public/js/views/universaldiscovery/ez-universaldiscoverysearchview.js
mhyndle: Changed it to be subscribed to the form `submit` event.

mhyndle:

mhyndle:
not really, one thing left... sorry

mhyndle:
Now all @dpobel are taken into account. I have simplified things a bit by removing `_refresh` method which was useless.

Resources/public/js/views/universaldiscovery/ez-universaldiscoverysearchview.js
mhyndle: I have removed that from `_search` method and subscribed it to `searchTextChange` event.

bdunogier:
Tests are I think failing because of the renaming of `ezpublish` to `app`.

mhyndle:
@bdunogier anything I should do with that or just ignore them? (as I have changed only JS)

Resources/public/js/views/universaldiscovery/ez-universaldiscoverysearchview.js
```diff
@@ -0,0 +1,503 @@
...
+            e.preventDefault();
+
+            if (this.get('loadContent')) {
+                contentStruct.content = locationStruct.content;
+            }
+
+            that._fireSelectContent(contentStruct);
+            that.get('selectedView').set('contentStruct', contentStruct);
+
+            c.all('.ez-searchresult-row').removeClass(IS_SELECTED_ROW_CLASS);
```
dpobel: here as well, you are mixing UI code with selection handling code. This could be done in a `selectContent` event handler for instance.

dpobel:
besides the last inline comment, nice work ! +1

bdunogier:
> @bdunogier anything I should do with that or just ignore them? (as I have changed only JS)

Nope, you don't do anything about it right now. Don't bother :)

Resources/public/js/views/services/plugins/ez-locationsearchplugin.js
```diff
@@ -49,31 +49,55 @@ YUI.add('ez-locationsearchplugin', function (Y) {
...
+
             contentService.createView(query, Y.bind(function (error, result) {
                 var attrs = {'loadingError': true};
 
                 if ( !error ) {
+                    var parsedResult = this._parseSearchResult(result);
+
                     attrs.loadingError = false;
-                    attrs[e.resultAttribute] = this._parseSearchResult(result);
+                    if (e.resultTotalCountAttribute) {
```
yannickroger: Since we are starting to have a few properties in this even, could we document it in the JSdoc ?

mhyndle: JSdoc filled

yannickroger:
+1

PR #437, opened 2015-11-19
EZP-24864: Add breadcrumbs for each location in the Location tab

yannickroger:
Link: https://jira.ez.no/browse/EZP-24864

## Description
Display a breacrumbs instead of the pathstring. See screenshot.

### Before
location_bread_before

### After
location_bread_after

## Tests
Manual and unit tests

## TODO
- [x] Basic feature
- [x] Fix unit test
- [x] Add unit test
- [x] Cleanup and rebase

### DIFF
--- a/Resources/public/js/models/ez-locationmodel.js
+++ b/Resources/public/js/models/ez-locationmodel.js
@@ -184,6 +184,7 @@ YUI.add('ez-locationmodel', function (Y) {
          * Loads path of the current location. The result is the array containing
          * eZ.Location objects present on the path sorted by depth.
          * The array doesn't contain current location.
+         * The result is available in the `path` attribute or in the `callback`.
          *
          * @method loadPath
          * @param {Object} options
@@ -219,6 +220,8 @@ YUI.add('ez-locationmodel', function (Y) {
                     return (a.get('depth') - b.get('depth'));
                 });
 
+                that._set('path', locations);
+
                 callback(error, locations);
             });
         },
@@ -245,7 +248,26 @@ YUI.add('ez-locationmodel', function (Y) {
                 query,
                 callback
             );
-        }
+        },
+
+        /**
+         * Overrides the RestModel implementation to also deal with the `path` attribute
+         *
+         * @method toJSON
+         * @return {Object}
+         */
+        toJSON: function () {
+            var attrs = Y.eZ.RestModel.prototype.toJSON.call(this);
+
+            if (attrs.path) {
+                attrs.path = Y.Array.map(attrs.path, function (value) {
+                    return value.toJSON();
+                });
+            }
+
+            return attrs;
+        },
+
     }, {
         REST_STRUCT_ROOT: "Location",
         ATTRS_REST_MAP: [
@@ -380,6 +402,19 @@ YUI.add('ez-locationmodel', function (Y) {
                     }
                     return contentInfo;
                 }
+            },
+
+            /**
+             * The location's path. By default it is empty. You need to call the loadPath method to set this attribute.
+             *
+             * @attribute path
+             * @default false
+             * @type array
+             * @readOnly
+             */
+            path: {
+                value: false,
+                readOnly: true
             },
         }
     });
--- a/Resources/public/js/views/services/plugins/ez-locationsloadplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-locationsloadplugin.js
@@ -36,6 +36,10 @@ YUI.add('ez-locationsloadplugin', function (Y) {
          * @method _loadLocations
          * @private
          * @param {EventFacade} e loadLocations event facade
+         * @param {eZ.Location} e.location the current location
+         * @param {eZ.Content} e.content the content
+         * @param {Boolean} e.loadPath (optional) will also load the path if set to `true`
+         *
          */
         _loadLocations: function (e) {
             var service = this.get('host'),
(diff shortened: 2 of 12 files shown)


Resources/public/templates/tabs/locations.hbt
```diff
@@ -29,7 +29,16 @@
                                         data-main-location="{{#if isMainLocation}}1{{else}}0{{/if}}">
                                 </td>
                                 <td>
-                                    <a href="{{ path "viewLocation" id=id languageCode=contentInfo.mainLanguageCode }}">{{ pathString }}</a>
+                                    <ul class="ez-breadcrumbs-list">
+                                        {{#each this.path}}
+                                        <li class="ez-breadcrumbs-item">
+                                            <a href="{{ path "viewLocation" id=id languageCode=contentInfo.mainLanguageCode }}">{{contentInfo.name}}</a>
```
crevillo: nitpick: spacess after brackets and contentInfo.name. Also before the closing ones.

Resources/public/templates/tabs/locations.hbt
```diff
@@ -29,7 +29,16 @@
...
+                                        </li>
+                                        {{/each}}
+                                        <li class="ez-breadcrumbs-item">
+                                            <a href="{{ path "viewLocation" id=id languageCode=contentInfo.mainLanguageCode }}">{{contentInfo.name}}</a>
```
crevillo: same as above.

Resources/public/templates/tabs/locations.hbt
```diff
@@ -29,7 +29,16 @@
...
                                 <td>
-                                    <a href="{{ path "viewLocation" id=id languageCode=contentInfo.mainLanguageCode }}">{{ pathString }}</a>
+                                    <ul class="ez-breadcrumbs-list">
+                                        {{#each this.path}}
+                                        <li class="ez-breadcrumbs-item">
+                                            <a href="{{path "viewLocation" id=id languageCode=contentInfo.mainLanguageCode}}">{{contentInfo.name}}</a>
+                                        </li>
+                                        {{/each}}
+                                        <li class="ez-breadcrumbs-item">
+                                            <a href="{{path "viewLocation" id=id languageCode=contentInfo.mainLanguageCode}}">{{contentInfo.name}}</a>
```
lolautruche: Point by @crevillo was not to remove the spaces inside `{{ }}`, but to **add** missing ones around `contentInfo.name`...

crevillo: that's it :)

yannickroger:
We should clearly document this at some point BTW, you can find a bit of both in the code.

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -245,7 +250,31 @@ YUI.add('ez-locationmodel', function (Y) {
...
+         *
+         * @method toJSON
+         * @return {Object}
+         */
+        toJSON: function () {
+            var attrs = Y.eZ.RestModel.prototype.toJSON.call(this),
+                path = [];
+
+            if (attrs.path) {
+                Y.Object.each(attrs.path, function (value, key) {
```
dpobel: when filled, `path` is an array, so a Y.Array method would more suited also since you want to create an array you can write:
```js
attrs.path = Y.Array.map(attrs.path, function (value) {
    return value.toJSON();
});
```

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -381,6 +410,17 @@ YUI.add('ez-locationmodel', function (Y) {
...
+
+            /**
+             * The location's path. By default it is empty. You need to call the loadPath method to set this attribute.
+             *
+             * @attribute path
+             * @default false
+             * @type array
+             */
+            path: {
+                value: false
```
dpobel: I would set it as readOnly, since one is not supposed to change anything on it directly.

Resources/public/js/views/services/plugins/ez-locationsloadplugin.js
```diff
@@ -49,10 +53,21 @@ YUI.add('ez-locationsloadplugin', function (Y) {
...
                     e.target.set('loadingError', true);
                 } else {
-                    e.target.set('locations', locations);
+                    if (e.loadPath) {
+                        Y.Array.each(locations, function (item) {
+                            item.loadPath(options, function (error) {
+                                if (error) {
+                                    e.target.set('loadingError', true);
+                                } else {
+                                    e.target.set('locations', locations);
```
dpobel: that's not gonna work, you need to use Y.Parallel to load the path for all Locations and set the result once everything is loaded.

andrerom:
> See screenshot.

there is no screenshot.

yannickroger:
@andrerom I should change the status of the PR from "work in progress" to "screenshot in progress" :) I am working on it at the moment.

andrerom:
Thanks, looks a bit better :)

This is loaded when opening the location tab, or on every "view" of an content item?

yannickroger:
Ready for review @dpobel @andrerom @mhyndle @StephaneDiot

yannickroger:
@andrerom every view. But asynchronously.

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -381,6 +403,18 @@ YUI.add('ez-locationmodel', function (Y) {
...
                 }
             },
+
+            /**
+             * The location's path. By default it is empty. You need to call the loadPath method to set this attribute.
+             *
+             * @attribute path
+             * @default false
+             * @type array
+             * @readOnly
```
dpobel: that's not enough to set an attribute as read only, see http://yuilibrary.com/yui/docs/attribute/#howtoconfig

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -576,13 +576,7 @@ YUI.add('ez-locationmodel-tests', function (Y) {
...
             });
 
-            this.model.loadPath({
-                api: this.capiMock
-            }, function (error, locations) {
-                Assert.isFalse(
-                    error,
-                    "No error should be detected"
-                );
+            this._assertLocations = function (locations) {
```
dpobel: that'll work but that's a bit weird, define a proper method or put the function in the local variable

dpobel:
besides the 2 comments, +1

yannickroger:
@dpobel 's feedback taken into account.

dpobel:
still +1
@mhyndle @StephaneDiot ?

mhyndle:
+1

PR #438, opened 2015-11-19
Fix EZP-25073: Adding user assignment to role doesn't refresh view

glye:
> Fixes https://jira.ez.no/browse/EZP-25073
> Status: Work in progress

- [x] Add way to specify which content tab should be shown on page load
- [x] Ensure the page is refreshed when a role assignment is added
- [x] Ensure that the assignments tab is shown when a role assignment is added
- [x] Fix breaking tests
- [ ] Fix the same problem for Section view
- [ ] Find a solution that doesn't require the tabId URL parameter?

### DIFF
--- a/Resources/config/routing_pjax.yml
+++ b/Resources/config/routing_pjax.yml
@@ -147,7 +147,7 @@ admin_roleList:
         _controller: ezsystems.platformui.controller.role:listRolesAction
 
 admin_roleView:
-    path: /role/view/{roleId}
+    path: /role/view/{roleId}/{tabId}
     defaults:
         _controller: ezsystems.platformui.controller.role:viewRoleAction
 
--- a/Resources/public/js/views/serverside/ez-roleserversideview.js
+++ b/Resources/public/js/views/serverside/ez-roleserversideview.js
@@ -41,7 +41,8 @@ YUI.add('ez-roleserversideview', function (Y) {
          */
         _pickSubtree: function (e) {
             var button = e.target,
-                unsetLoading = Y.bind(this._uiUnsetAssignRoleLoading, this, button);
+                unsetLoading = Y.bind(this._uiUnsetAssignRoleLoading, this, button),
+                refreshAssignmentsTab = Y.bind(this._refreshAssignmentsTab, this, button);
 
             e.preventDefault();
             this._uiSetAssignRoleLoading(button);
@@ -51,9 +52,10 @@ YUI.add('ez-roleserversideview', function (Y) {
                     cancelDiscoverHandler: unsetLoading,
                     multiple: true,
                     data: {
-                        roleId: button.getAttribute('data-role-rest-id'),
+                        roleId: button.getAttribute('data-role-id'),
+                        roleRestId: button.getAttribute('data-role-rest-id'),
                         roleName: button.getAttribute('data-role-name'),
-                        afterUpdateCallback: unsetLoading,
+                        afterUpdateCallback: refreshAssignmentsTab,
                     },
                 },
             });
@@ -81,6 +83,24 @@ YUI.add('ez-roleserversideview', function (Y) {
          */
         _uiUnsetAssignRoleLoading: function (button) {
             button.removeClass('is-loading').set('disabled', false);
+        },
+
+        /**
+         * Refreshes the role view, showing the assignments tab.
+         *
+         * @method _refreshAssignmentsTab
+         * @protected
+         * @param {Y.Node} button
+         */
+        _refreshAssignmentsTab: function (button) {
+            /**
+             * Fired when the view needs to be refreshed.
+             * @event refreshAssignmentsTab
+             * @param roleId {String} The ID of the role we are showing assignments for
+             */
+            this.fire('refreshAssignmentsTab', {
+                roleId: button.getAttribute('data-role-id')
+            });
         },
     });
 });
--- a/Resources/public/js/views/services/ez-roleserversideviewservice.js
+++ b/Resources/public/js/views/services/ez-roleserversideviewservice.js
@@ -24,6 +24,8 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
             this.on('*:contentDiscover', function (e) {
                 e.config.contentDiscoveredHandler = Y.bind(this._assignRole, this);
             });
+
+            this.on('*:refreshAssignmentsTab', this._refreshAssignmentsTab);
         },
 
         /**
@@ -41,16 +43,16 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
                 service = this;
 
             this._assignRoleNotificationStarted(
-                data.roleId, data.roleName, e.selection
+                data.roleRestId, data.roleName, e.selection
             );
 
-            userService.loadRole(data.roleId, function (error, response) {
+            userService.loadRole(data.roleRestId, function (error, response) {
                 var tasks = new Y.Parallel(),
                     role;
 
                 if (error) {
                     service._loadRoleErrorNotification(
-                        data.roleId, data.roleName, e.selection
+                        data.roleRestId, data.roleName, e.selection
                     );
                     return;
                 }
@@ -76,7 +78,7 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
 
                 tasks.done(function () {
                     service._assignRoleCallback(
-                        data.roleId, data.roleName, e.selection, countAssigned,
+                        data.roleRestId, data.roleName, e.selection, countAssigned,
                         Y.bind(data.afterUpdateCallback, service)
                     );
                 });
@@ -181,15 +183,15 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
          *
          * @method _assignRoleCallback
          * @protected
-         * @param {String} roleId the role id
+         * @param {String} roleRestId the role id
          * @param {String} roleName the role name
          * @param {Array} contents the array of users/groups to which role is being assigned to
          * @param {Integer} countAssigned number of successfully assignments
          * @param {Function} callback the callback to call when other tasks are done
          */
-        _assignRoleCallback: function (roleId, roleName, contents, countAssigned, callback) {
+        _assignRoleCallback: function (roleRestId, roleName, contents, countAssigned, callback) {
             var notificationIdentifier = this._getAssignRoleNotificationIdentifier(
-                    'assign-role', roleId, contents
+                    'assign-role', roleRestId, contents
                 );
 
             if (countAssigned>0) {
@@ -216,13 +218,13 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
          *
          * @method _assignRoleNotificationStarted
          * @protected
-         * @param {String} roleId the role id
+         * @param {String} roleRestId the role id
          * @param {String} roleName the role name
          * @param {Array} contents the array of users/groups to which role is being assigned to
          */
-        _assignRoleNotificationStarted: function (roleId, roleName, contents) {
+        _assignRoleNotificationStarted: function (roleRestId, roleName, contents) {
             var notificationIdentifier = this._getAssignRoleNotificationIdentifier(
-                    'assign-role', roleId, contents
+                    'assign-role', roleRestId, contents
                 );
 
             this._notify(
@@ -238,13 +240,13 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
          *
          * @method _loadRoleErrorNotification
          * @protected
-         * @param {String} roleId the role id
+         * @param {String} roleRestId the role id
          * @param {String} roleName the role name
          * @param {Array} contents the array of users/groups to which role is being assigned to
          */
-        _loadRoleErrorNotification: function (roleId, roleName, contents) {
+        _loadRoleErrorNotification: function (roleRestId, roleName, contents) {
             var notificationIdentifier = this._getAssignRoleNotificationIdentifier(
-                    'assign-role', roleId, contents
+                    'assign-role', roleRestId, contents
                 );
 
             this._notify(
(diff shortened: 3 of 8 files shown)


Resources/views/Role/view_role.html.twig
```diff
@@ -21,7 +21,15 @@
     <section class="ez-tabs ez-serverside-content">
         <ul class="ez-tabs-list">
             <li class="ez-tabs-label is-tab-selected"><a href="#ez-tabs-role-name">{{ role.identifier }}</a></li>
-            <li class="ez-tabs-label"><a href="#ez-tabs-content">{{ 'assignment.user_or_group_using'|trans({'%role%': role.identifier, '%count%': role_assignments|length }) }}</a></li>
+            <li class="ez-tabs-label">
+                <a href="#ez-tabs-content"
+                   {%- if tab_id == 'ez-tabs-content' -%}
+                   id="ez-tabs-activate-on-load"
```
dpobel: why not just move the `is-tab-selected` classes to the tab and the panel that should be shown ?

glye: @dpobel Indeed, that is easier :) Still struggling with the reload after assignment create, which is supposed to reload the page with a different `tabId` parameter. This doesn't seem to do anything:
`this.get('app').navigateTo('admin_roleView', {roleId: data.roleId, tabId: 'ez-tabs-content'});`

glye:
Use `is-tab-selected` to activate assignments tab.
Fixed reload & tab select after making new assignment.
The functionality is OK now, tests are TODO.

glye:
Fixed breaking tests, and added tests for new stuff.

andrerom:
+1 but would be nice to know why behat tests are failing on most PR's now /cc @pcardiga @vidarl

bdunogier:
Ping @mhyndle @yannickroger @StephaneDiot.

Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
```diff
@@ -632,8 +633,61 @@ YUI.add('ez-roleserversideviewservice-tests', function (Y) {
...
+            this.capi = new Mock();
+            this.app = new Mock();
+            this.roleId = 42;
+
+            this.service = new Y.eZ.RoleServerSideViewService({
+                capi: this.capi,
+                app: this.app,
+            });
+
+            Mock.expect(this.service, {
```
mhyndle: `this.service` is not mock object so it shouldn't be proceeded like it.
Besides that `Y.eZ.RoleServerSideViewService` extends `Y.Base` so if it is initiated like above, then `this.service.get('app')` should return the value of attribute passed to the constructor.

Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
```diff
@@ -632,8 +633,61 @@ YUI.add('ez-roleserversideviewservice-tests', function (Y) {
...
+                method: 'get',
+                args: [Mock.Value.String],
+                returns: this.app,
+            });
+            Mock.expect(this.service, {
```
mhyndle: Same think like https://github.com/ezsystems/PlatformUIBundle/pull/438/files#r46044133

Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
```diff
@@ -632,8 +633,61 @@ YUI.add('ez-roleserversideviewservice-tests', function (Y) {
...
+            Mock.expect(this.app, {
+                method: 'set',
+                args: ['loading', true],
+            });
+            Mock.expect(this.app, {
+                method: 'navigateTo',
+                args: ['adminRole', {uri: 'pjax/role/view/' + this.roleId + '/ez-tabs-content'}],
+            });
+
+            this.service._refreshAssignmentsTab(data);
```
mhyndle: You shouldn't call private/protected methods directly. To run it you should call the event that triggers it so probably something like: 
```javascript
this.service.fire('refreshAssignmentsTab', {...});
```

Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
```diff
@@ -632,8 +633,61 @@ YUI.add('ez-roleserversideviewservice-tests', function (Y) {
...
+        "Should refresh the page with tabId parameter added": function () {
+            var data = {
+                    roleId: this.roleId,
+                };
+
+            Mock.expect(this.app, {
+                method: 'set',
+                args: ['loading', true],
+            });
+            Mock.expect(this.app, {
```
mhyndle: It's not checking if `navigateTo` method was called. It only checks whether the parameters passed to the function are those you are expecting, but if the method wasn't called at all then test wouldn't catch that. To check if the method was called you should do:
`Mock.verify(this.app);` at the end of the test
OR
provide `run` attribute  where you will switch the flag `appNavigatedToCalled` to `true` then make assertion for it at the end.

Resources/public/js/views/serverside/ez-roleserversideview.js
```diff
@@ -82,5 +84,22 @@ YUI.add('ez-roleserversideview', function (Y) {
...
+         *
+         * @method _refreshAssignmentsTab
+         * @protected
+         * @param {Y.Node} button
+         */
+        _refreshAssignmentsTab: function (button) {
+            /**
+             * Fired when the view needs to be refreshed
+             * @event refreshAssignmentsTab
+             */
```
mhyndle: Parameter passed to the event should be documented, something like:
```
* @event refreshAssignmentsTab
* @param roleId {String} ...
*/
```

Tests/js/views/services/assets/ez-roleserversideviewservice-tests.js
glye: I'm trying the first approach.

glye:
Removed unneeded expects
Fire event rather than call method directly
Calling `Mock.verify(this.app)` to verify methods called

@mhyndle I'm stuck on `Unexpected error: 'undefined' is not an object (evaluating 'this.get('request').params')`. My guess is that the mocking is incomplete, but none of the role code is fetching the `request`, as far as I can see, and I'm unsure how to mock it, anyway.

glye:
The travis tests have passed (the age of wonders has not). Review ping @mhyndle @yannickroger @StephaneDiot

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -41,16 +45,16 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
                 service = this;
 
             this._assignRoleNotificationStarted(
-                data.roleId, data.roleName, e.selection
+                data.roleRestId, data.roleName, e.selection
```
yannickroger: should we need to rename this now ? it is not part of the bugfix.

glye: @yannickroger I renamed it to clarify, because there was a "roleId" before which actually was a role REST ID, and this fix introduces a role ID which is an actual role ID :) So I thought this is the cleanest way around the conflict, avoiding future confusion.

Resources/config/routing_pjax.yml
```diff
@@ -147,7 +147,7 @@ admin_roleList:
         _controller: ezsystems.platformui.controller.role:listRolesAction
 
 admin_roleView:
-    path: /role/view/{roleId}
+    path: /role/view/{roleId}/{tabId}
```
dpobel: instead of passing `null` in https://github.com/ezsystems/PlatformUIBundle/pull/438/files#diff-84de834a31786fc3fe688bc4535d6bdeR326 or 0 in https://github.com/ezsystems/PlatformUIBundle/pull/438/files#diff-77c6e07c3a54f32d5394f5241e4953b5R34 as `tabId` I think you should make this parameter optional

Resources/public/js/views/serverside/ez-roleserversideview.js
```diff
@@ -82,5 +84,23 @@ YUI.add('ez-roleserversideview', function (Y) {
...
+         * @protected
+         * @param {Y.Node} button
+         */
+        _refreshAssignmentsTab: function (button) {
+            /**
+             * Fired when the view needs to be refreshed.
+             * @event refreshAssignmentsTab
+             * @param roleId {String} The ID of the role we are showing assignments for
+             */
+            this.fire('refreshAssignmentsTab', {
```
dpobel: in https://github.com/ezsystems/PlatformUIBundle/commit/f4b909cdedbc01ea848c6058111a574a4c518799, @yannickroger did pretty much the same for section view, I guess we need to move what he did to the ServerSideViewService and ServerSideView instead of duplicating the code.

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -24,6 +24,10 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
             this.on('*:contentDiscover', function (e) {
                 e.config.contentDiscoveredHandler = Y.bind(this._assignRole, this);
             });
+
+            this.on('*:refreshAssignmentsTab', function (data) {
```
dpobel: nitpick: can be written:
```js
this.on('*:refreshAssignmentsTab', this._refreshAssignmentsTab);
```

Resources/public/js/views/serverside/ez-roleserversideview.js
dpobel: actually if you do that, we'll be able to remove the `tabId` parameter. Basically, when the view will fire the `refreshView` event, it should keep the selected  tab and should reselect it after the refresh is done.

Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -295,5 +299,19 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
...
+         * Refreshes the role view, showing the assignments tab.
+         *
+         * @method _refreshAssignmentsTab
+         * @protected
+         * @param {Object} data
+         */
+        _refreshAssignmentsTab: function (data) {
+            this.get('app').set('loading', true);
+            this.load(Y.bind(function () {
+                this.get('app').navigateTo('adminRole', {uri: 'pjax/role/view/' + data.roleId + '/ez-tabs-content'});
```
dpobel: following https://github.com/ezsystems/PlatformUIBundle/pull/438/files#r46158645, `data.roleId` will be useless so the `roleRestId` renaming becomes useless.

Also for the record, this code does not mean anything. Basically the server side view will do an AJAX request and when it's done you ask the application to navigate somewhere that will again do an AJAX request (almost the same...).

Resources/config/routing_pjax.yml
glye: @dpobel Agree, and I did that at first, but couldn't get it to work. The link was built without the trailing slash when tabId was omitted, meaning that the route did not match. So I added a second route without the tabId parameter - don't remember the details, but that didn't work well either. A working example would be helpful :)

Resources/public/js/views/serverside/ez-roleserversideview.js
glye: @dpobel That's where I got the idea from :) But do you want this now, or in a followup?

glye: > it should keep the selected tab and should reselect it after the refresh is done.

But some magic is required to make this happen, I think. I started with just the refresh, and added the tab handling when it turned out to be needed.

Resources/public/js/views/services/ez-roleserversideviewservice.js
glye: The 2nd paragraph isn't clear to me. Why/how are there two requests? Is the `set('loading', true)` an ajax request?

Resources/config/routing_pjax.yml
dpobel: http://symfony.com/doc/current/book/routing.html#required-and-optional-placeholders

Resources/public/js/views/serverside/ez-roleserversideview.js
dpobel: Please stop calling everything "magic", yeah maybe we need to add some code, that's our job. And no, no follow up here, we should just try to not duplicate code.

Resources/public/js/views/services/ez-roleserversideviewservice.js
dpobel: this.load basically does an AJAX request (the same one as the one needed to generate the page the application is displaying) then once it's done you tell the application to navigate to the 'adminRole' route which will also request the URI `'pjax/role/view/' + data.roleId + '/ez-tabs-content'`.

Resources/config/routing_pjax.yml
glye: @dpobel Thanks, but this is the doc I was reading when doing the first attempt (with optional parameters), so not very helpful. The problem wasn't Symfony, but our code. But I can give it another go.

dpobel: but in any case, this parameter is not necessary since the selected tab will be handle by the JS code (see https://github.com/ezsystems/PlatformUIBundle/pull/438/files#r46159286)

Resources/public/js/views/serverside/ez-roleserversideview.js
glye: @dpobel 
> maybe we need to add some code
Which is what I'm doing here, remember? ;)

> we'll be able to remove the tabId parameter
That's the big question, how? Please advise on how to reload a page and stay on the same tab, without passing a parameter. (The refactoring you suggest shouldn't be a problem.)

Resources/config/routing_pjax.yml
glye: @dpobel The section view code you are referring to does not load the correct tab either. I agree with the reuse of code, but we have to come up with functioning code first.

Resources/public/js/views/services/ez-roleserversideviewservice.js
glye: @dpobel Done, thanks.

Resources/public/js/views/services/ez-roleserversideviewservice.js
glye: @dpobel Fixed, thanks.

glye:
Fixed some feedback issues and rebased on master, verified patch still works. Some suggested changes are still todo.

glye:
Status update:
- The PR as is fixes the bug in Role view: it reloads the view and shows the correct tab.
- Section view is affected by the same problem, that's not fixed yet: it reloads the view but shows the wrong tab
- Reviewers want a fix that doesn't require the `tabId` URL parameter, I haven't found a way to do that yet.
- The role view service test is failing.

glye:
The current failure is irrelevant ("Unable to fetch some archives").

andrerom:
Closing *temporary* and moving this to [a later epic](https://jira.ez.no/browse/EZP-25442) as this needs to be handled after generic refresh system is in place.

PR #442, opened 2015-11-26
EZP-25133: Allow to edit a draft version of a content

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25133
requires: https://github.com/ezsystems/ez-js-rest-client/pull/69

# Description

This patch adds a route to be able to edit a draft version of a content. It also cleans up the way the editor is redirected to be easier to tweak from a PlatformUI extension bundle and it also decouples a bit the publish draft and discard draft plugins from the content edit view service.

# Tests

Manual and unit tests.

### DIFF
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -744,6 +744,13 @@ YUI.add('ez-platformuiapp', function (Y) {
                     view: 'studioPlusPresentationView',
                     callbacks: ['open', 'checkUser', 'handleSideViews', 'handleMainView']
                 }, {
+                    name: "editContentVersion",
+                    path: '/edit/:id/version/:versionId/:languageCode',
+                    service: Y.eZ.ContentEditViewService,
+                    sideViews: {'navigationHub': false, 'discoveryBar': false},
+                    view: 'contentEditView',
+                    callbacks: ['open', 'checkUser', 'handleSideViews', 'handleMainView']
+                }, {
                     name: "translateContent",
                     path: '/edit/:id/:languageCode/:baseLanguageCode',
                     service: Y.eZ.ContentEditViewService,
--- a/Resources/public/js/models/ez-contentmodel.js
+++ b/Resources/public/js/models/ez-contentmodel.js
@@ -171,7 +171,7 @@ YUI.add('ez-contentmodel', function (Y) {
          * @return {Boolean}
          */
         hasTranslation: function (languageCode) {
-            return (this.get('currentVersion').getTranslationsList().indexOf(languageCode) !== -1);
+            return this.get('currentVersion').hasTranslation(languageCode);
         },
 
         /**
--- a/Resources/public/js/models/ez-versionmodel.js
+++ b/Resources/public/js/models/ez-versionmodel.js
@@ -32,6 +32,8 @@ YUI.add('ez-versionmodel', function (Y) {
          * supported
          * @param {Object} options the options for the sync.
          * @param {Object} options.api the JS REST client instance
+         * @param {Object} options.languageCode the language in which the fields
+         * should be loaded
          * @param {Function} callback a callback executed when the operation is finished
          */
         sync: function (action, options, callback) {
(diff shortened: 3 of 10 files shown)


dpobel:
ping @andrerom @bdunogier @sunpietro @lserwatka

sunpietro:
_When editing the first draft of a content, I don't have access to the parent location that was chosen by the editor (the node assignment) so I don't know how/where to redirect the user if he discards the draft or close the view._

@dpobel if I understood correctly, you should be able to pass parent location object from the locationView to content creator view service. Then in the discardDraftPlugin you should the access to it.
Correct me if I'm wrong.

dpobel:
@sunpietro that's wrong, you don't arrive on the `editContentVersion` route from a location view but from a link in your flex workflow email.

I've come to think the way we are doing this feature wrong. PlatformUI should not provide (yet) the edition of version, it should instead let you do that in Studio (which is kind of possible right now even if not super handy).

sunpietro:
@dpobel the flex workflow is the additional functionality to PlatformUI not the StudioUI itself.

lserwatka:
@sunpietro actually it is for both. Please test what @dpobel made and then we will take comments from there.

dpobel:
@lserwatka  @sunpietro it's a feature provided by Studio, try to find the word "flex" in Platform(UI) if you don't believe me. And I find odd to implement half it in PlatformUI.

lserwatka:
@dpobel true, business wise it belongs to eZ Studio of course, but technically it reuses components from PlatformUI as it does not make sense to re-create exactly the same editorial features in eZ Studio that PlatformUI already does.

sunpietro:
@dpobel we're missing the view similar to locationView to see non-editable version of content.

dpobel:
> @dpobel we're missing the view similar to locationView to see non-editable version of content.

@lserwatka told us that this was not needed yesterday. As far as I can tell, that would not change anything to the redirection problem, I would miss the context in which the content was created.

bdunogier:
>  I would miss the context in which the content was created.

Could you elaborate on that please ?

lserwatka:
@dpobel as fare as I know parent location id is stored in the DB when we create draft. We can track it explicitly in the flex-worflow, but it is a bit hackish, since that info is already in then DB.

dpobel:
@lserwatka yes it's in the DB but not exposed through the REST API (I'm not even sure in the Public API).

@bdunogier if the Content item has not been published yet and if the user discards the draft or closes the edit view, we should redirect to the Location that was chosen as a parent Location for the Content but I don't have access to that through the REST API. Having the ability to display a DRAFT would not really help if the user discards the draft, the draft does not exist anymore so we can redirect him to view it...

lserwatka:
@dpobel it is clear, if this is not exposed by API, then we can track it on our side. Let's assume you will get parent location id (context) as param in the URI, so you can redirect to that view

lserwatka:
@dpobel is this something you could use eZ\Publish\Core\Helper\PreviewLocationProvider::loadMainLocation($contentId);
It is fetching location depending on Content status, whether it is draft or publish content.

bdunogier:
So, we have a [PreviewLocationProvider](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/Helper/PreviewLocationProvider.php) that either gets the location from the repository if there's one, or uses an SPI service to get the node assignment if there isn't.

It is not exposed by the Repository's REST API (because it is not part of the Repository's Public API), and there isn't a chance that's gonna change.

The safest way would be to provide PlatformUI with the redirect URI. It's actually what we are doing with user generated content in the frontend/Sf stack.

dpobel:
I updated the patch. So to sum up:

* this is adding a a route to be able to edit version
* we do several checks on the version (it must be a "DRAFT", be created by the logged in user, be a draft of the content which id is also passed, have a translation in the passed language code)
* about the redirections (the tricky part finally...):
   * if the content has a main location, closing the view, discarding the draft or publishing it redirects to that Location
   * if the content does have a main location (it has never been published)
      * after publishing, the user is redirected to the newly created Location
      * if the user closes the view or discards the draft, it is redirected by default to the dashboard

Of course, you can customize the redirections in a plugin for the Content Edit View Service either by setting one of the `*RedirectionUrl` attributes (to an URL or a function to compute an URL when it will be needed) or by changing the event handler of the `closeView`, `discardedDraft` or `publishedDraft` events.

So next, I'll change that code to meet our quality standards (comments, tests, ...) normally before tomorrow evening.

sunpietro:
I like it :)

lserwatka:
Thank you @dpobel Looks great! We will test it.

dpobel:
ready for review, ping @yannickroger @StephaneDiot @mhyndle @andrerom

mhyndle:
+1

yannickroger:
+1

PR #446, opened 2015-12-01
Fix EZP-25161: Impossible to publish a content with a Richtext field with Firefox

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25161

# Description

It's impossible to save a content item with a RichText field in Firefox (nor in IE11). This is happening because of the way we build the XHTML5Edit version of the RichText field and especially because of our usage of a namespace in this format...

To avoid that, we are kind of forced to build the XML by concatenating strings... While at it, I also refactored this part by delegating the editor's content clean up to a set of dedicated components (`EditorContentProcessors`).

# Test

unit tests and manual tests in Chrome, Firefox and IE11.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -29,6 +29,7 @@ module.exports = function(grunt) {
             "./Resources/public/js/alloyeditor/buttons/*.js",
             "./Resources/public/js/alloyeditor/buttons/mixins/*.js",
             "./Resources/public/js/alloyeditor/plugins/*.js",
+            "./Resources/public/js/alloyeditor/processors/*.js",
         ],
         testFiles = [
             "./Tests/js/**/*.js",
--- a/Resources/config/yui.yml
+++ b/Resources/config/yui.yml
@@ -92,6 +92,14 @@ system:
                 ez-alloyeditor-button-mixin-embedimage:
                     requires: ['ez-alloyeditor']
                     path: %ez_platformui.public_dir%/js/alloyeditor/buttons/mixins/embedimage.js
+                ez-editorcontentprocessorbase:
+                    path: %ez_platformui.public_dir%/js/alloyeditor/processors/base.js
+                ez-editorcontentprocessorxhtml5edit:
+                    requires: ['ez-editorcontentprocessorbase']
+                    path: %ez_platformui.public_dir%/js/alloyeditor/processors/xhtml5edit.js
+                ez-editorcontentprocessorremoveids:
+                    requires: ['ez-editorcontentprocessorbase']
+                    path: %ez_platformui.public_dir%/js/alloyeditor/processors/removeids.js
                 ez-platformuiapp:
                     requires:
                         - 'app'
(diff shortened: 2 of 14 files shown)


dpobel:
ping @yannickroger @mhyndle @StephaneDiot @andrerom 

Also, this makes me realize that [EZP-24732](https://jira.ez.no/browse/EZP-24732) is not fixed when we try to set the value of a RichText field through REST. @bdunogier can you have a look ?

bdunogier:
> Also, this makes me realize that EZP-24732 is not fixed when we try to set the value of a RichText field through REST. @bdunogier can you have a look ?

Yep, I'll dig.

mhyndle:
+1

andrerom:
+1 kind of, care to exapand on the namespace issues? And why we use xml parser in the browser? (As opposed to Dom, ..)

dpobel:
@andrerom: sure. Let's take a simple example. If the editor puts something like
```html
<p>I'm a paragraph</p>
```

in the RichText editor, I should generate something like:

```xml
<section xmlns="http://ez.no/namespaces/ezpublish5/xhtml5/edit">
  <p>I'm a paragraph</p>
</section>
```

to the save RichText field content. This is a equivalent to:

```xml
<ez:section xmlns:ez="http://ez.no/namespaces/ezpublish5/xhtml5/edit">
  <ez:p>I'm a paragraph</ez:p>
</ez:section>
```

But to generate that properly (without concatenating strings and pray that results in a valid XML document) I should first parse what the user put in the rich text editor and then recursively import everything in the xhtml5edit namespace.

But actually, having `ez:p` is semantically wrong to me. The format is called xhtml5edit so basically that means we want to use some part of *xhtml5* so instead of defining the namespace `http://ez.no/namespaces/ezpublish5/xhtml5/edit` for everything we should reuse the xhtml namespace when we are using XHTML elements. That could give something like:

```xml
<ez:section xmlns:ez="http://ez.no/namespaces/ezpublish5/xhtml5/edit" xmlns:xhtml="http://www.w3.org/1999/xhtml">
  <xhtml:p>I'm a paragraph</xhtml:p>
</ez:section>
```

which is equivalent to:

```xml
<ez:section xmlns:ez="http://ez.no/namespaces/ezpublish5/xhtml5/edit" xmlns="http://www.w3.org/1999/xhtml">
  <p>I'm a paragraph</p>
</ez:section>
```

or

```xml
<ez:section xmlns="http://ez.no/namespaces/ezpublish5/xhtml5/edit" xmlns:xhtml="http://www.w3.org/1999/xhtml">
  <xhtml:p>I'm a paragraph</xhtml:p>
</ez:section>
```

Semantically, that's much better IMHO (at least the structure reflects the goal of this format) and it would be much easier for me to generate.

One can also argue that `section` is a HTML5 element so this one too should not be in our custom namespace and the xhtml5edit version of the document could then be:

```xml
<section><!-- maybe with  xmlns="http://www.w3.org/1999/xhtml" -->
  <p>I'm a paragraph</p>
</section>
```

For me, it would be the easiest of course but the RichText field was build with the idea of having a mapping between the namespace and the parser.

> And why we use xml parser in the browser? (As opposed to Dom, ..)

not sure to understand the question, we use DOM when parsing some XML

yannickroger:
+1

PR #447, opened 2015-12-01
Change ezpublish to app in README.md file

crevillo:
Following with the ezpublish to app renames. Let me know if you want me to add a jira issue for this or is this could work without it
ping @dpobel @andrerom @bdunogier @yannickroger

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ On an existing installation, here's what you need to do:
   $ composer require ezsystems/platform-ui-bundle:dev-master
   ```
 
-* In `ezpublish/EzPublishKernel.php` add an instance of the following classes to
+* In `app/AppKernel.php` add an instance of the following classes to
   the list of registered bundles:
 
   ```php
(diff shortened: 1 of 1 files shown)


bdunogier:
+1 here.

@dpobel ?

andrerom:
+1

dpobel:
+1

bdunogier:
As pointed out by @dpobel, the instructions here don't make much sense, when you think about it. Given that PlatformUI outside of eZ Platform doesn't make much sense, there's not much point in maintaining these instructions.

I'll merge it for now, but we might wanna refactor those README files so that they all reference ezplatform's.

bdunogier:
Merged to 8f03ad0 (changed the commit message).

crevillo:
thanks @bdunogier. while on this. should we remove "ezpublish" references in eZ Platform provided routing.yml? Lide this one https://github.com/ezsystems/ezplatform/blob/master/app/config/routing.yml#L24.
There's no problem with the way they are defined now, but as we're trying to get ride of ezpublish...

bdunogier:
It's a good idea, @crevillo. But we have a few ones that will be harder to cover, like the name of the root key for the settings, the name of the CoreBundle... :-)

PR #448, opened 2015-12-02
Fix EZP-25130: 404 when previewing default eZ Publish object

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25130

# Description

In a clean install, if the editor edits the eZ Publish root content and tries to preview it, the preview iframe displays a 404 page. This is happening because this object has only one version which number is 6 and the preview is generated for a version with `versionNo` set to 1 because no version has been created yet. This is kind of related to [EZP-24931](https://jira.ez.no/browse/EZP-24931) but in that case, we can detect the issue and generate the preview of the current version instead. This will also allow to easily reuse the preview component while viewing a content (outside of the editing process) (see https://jira.ez.no/browse/EZP-24740)

# Tests

manual + unit tests

### DIFF
--- a/Resources/public/js/views/ez-editpreviewview.js
+++ b/Resources/public/js/views/ez-editpreviewview.js
@@ -28,6 +28,24 @@ YUI.add('ez-editpreviewview', function (Y) {
         },
 
         /**
+         * Returns the version to use to generate the preview. If the version
+         * was not saved yet, we are actually generating the preview to the
+         * current version.
+         * TODO: that can be confusing for the user if he did some changes but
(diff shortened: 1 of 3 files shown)


dpobel:
ping @yannickroger @mhyndle @andrerom @StephaneDiot

andrerom:
+1

mhyndle:
+1 (why it's failing on rest model tests?)

dpobel:
@mhyndle that's a good question, this is randomly happening from time to time with no obvious reason at the moment. I restarted the Travis task

PR #449, opened 2015-12-02
EZP:25173: Add an iframe to ez.no to the dashboard

yannickroger:
Link: https://jira.ez.no/browse/EZP-25173

## Description
This is a feature demanded by the product management, it replaces the default static content of the dashboard by an iframe pointing on a specific page on ez.no.

## Screenshot
dashboard

## Tests
Manual an unit test.


Resources/public/js/views/ez-dashboardview.js
```diff
@@ -28,7 +33,35 @@ YUI.add('ez-dashboardview', function (Y) {
...
+            /**
+             * Stores the iframe Source
+             *
+             * @attribute iframeSource
+             * @type String
+             * @default 'http://ez.no/in-product/eZ-Platform'
+             * @readOnly
+             */
+            iframeSource: {
+                value: 'http://ez.no/in-product/eZ-Platform',
```
andrerom: I guess you got this url from @SylvainGuittard, if not check with him, there was a different url, with possibility of passing version number, suggested in the issue.

andrerom:
Besides question, +1

dpobel:
technically +1

Resources/public/js/views/ez-dashboardview.js
crevillo: nitpick: maybe https:// instead? or ://? wondering if http could rise error to people installing platform with https...

bdunogier: Could be the protocol used by platformui itself. That's easy to do in JS is it not ?

But this can nonetheless be fixed later. I'd like to merge if no -1.

sunpietro:
There's an issue when you resize the browser window. We got this fixed on eZ Studio side.

yannickroger:
@sunpietro We also have the same issue on other iframes (page and performances) present when studio is not installed.

yannickroger:
Issue created: https://jira.ez.no/browse/EZP-25233

PR #451, opened 2015-12-02
Fix EZP-25177: search result persist after UDW reloads

mhyndle:
> status: ready for review

Jira: https://jira.ez.no/browse/EZP-25177

## Description
While investigating the problem I found 2 things that were involved in generating the problem:
- this._fireLocationSearch was triggered on change of 2 attributes: `offset` and `searchText` so there were 2 almost simultaneus calls when new search was started - one with changed searchText (but still with old offset) and just after that the second one with reseted offset. So displayed results can be randomly from first or second search...
To solve this one I have stopped listening on `offsetChange` and call `this._fireLocationSearch` for every page change.
- As there is locationSearch triggered on `searchTextChange` event so when the view is being reset the problems also happens. So just after UDW was being closed the `locationSearch` call was triggered. Instead, the view should be reset. That's what this PR also provides - when the `searchText` is empty, instead of calling `locationSearch`, the view is being reset (`{offset: 0; searchResultCount: 0; searchResultList: []}`).

## Tasks
- [x] implementation
- [x] unit tests

### DIFF
--- a/Resources/public/js/views/universaldiscovery/ez-universaldiscoverysearchview.js
+++ b/Resources/public/js/views/universaldiscovery/ez-universaldiscoverysearchview.js
@@ -51,7 +51,7 @@ YUI.add('ez-universaldiscoverysearchview', function (Y) {
             this.on('selectContent', this._uiSelectContent);
 
             this.after(['multipleChange', 'isSelectableChange'], this._setSelectedViewAttrs);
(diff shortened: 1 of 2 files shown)


mhyndle:
Ready for review

StephaneDiot:
+1

dpobel:
+1

PR #452, opened 2015-12-03
Fix EZP-25175: Setting incorrect pagination offset in subitems view

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-25175

# Description

When a folder had as many children as a multiple of its subitems list limit per page, the subitems list would not show if you try to go to the last page. The reason of that is incorrect calculation for offset for last page which fails in case when total number of results is multiplication of page limit.

# Tests

- Manual tests
- Unit tests


StephaneDiot:
ping @dpobel @yannickroger @mhyndle

mhyndle:
+1 as far as Travis agree

dpobel:
+1 (Travis and ezrobot agree)

PR #454, opened 2015-12-03
EZP-25076: Move notifications to the bottom

mhyndle:
> status: ready for review
Jira: https://jira.ez.no/browse/EZP-25076 and https://jira.ez.no/browse/EZP-25147

## Description
This PR solves the problem with positioning edit content preview view when triggered while displaying notifications by moving notifications to the bottom of the screen.

## Screencast
YT: https://youtu.be/bWgrP-YVN4U
In first part you can see current (improper) behavior. After presenting that I'm switching branch in the background and reloading the app so after that you can see behavior within the patch. What's more you can see example of how the notifications are added to the notifications container after patch (they are added from bottom to top).
On the video you can notice that the Symfony toolbar is overlapping first (the lowest) notification.

## Tasks
- [x] implementation
- [x] unit tests

### DIFF
--- a/Resources/public/css/layout.css
+++ b/Resources/public/css/layout.css
@@ -83,11 +83,11 @@
 }
 
 .ez-notification-container {
-    position: absolute;
-    top: 0;
+    position: fixed;
+    bottom: 0;
     left: 0;
     width: 100%;
-    z-index: 100;
+    z-index: 6000001;
 }
 
 .is-app-loading .ez-view-container:before {
--- a/Resources/public/css/views/notificationhub.css
+++ b/Resources/public/css/views/notificationhub.css
@@ -11,5 +11,4 @@
 .is-navigationhubview-fixed .ez-view-notificationhubview,
 .is-navigation-hidden .ez-view-notificationhubview {
     width: 100%;
-    position: fixed;
 }
--- a/Resources/public/js/views/ez-notificationhubview.js
+++ b/Resources/public/js/views/ez-notificationhubview.js
@@ -41,22 +41,11 @@ YUI.add('ez-notificationhubview', function (Y) {
          */
         _setListEventHandlers: function (list) {
             list.on('add', Y.bind(function (e) {
-                var oldHeight = this._getContainerHeight();
-
                 this._addNotificationView(e.model);
-                this._fireHeightChange(
-                    oldHeight, this._getContainerHeight()
-                );
             }, this));
 
             list.on('remove', Y.bind(function (e) {
-                var oldHeight = this._getContainerHeight(),
-                    newHeight = oldHeight;
-
-                newHeight -= this._removeNotificationView(e.model);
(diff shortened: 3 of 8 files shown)


dpobel:
Can you style the notification to appear on top the Symfony toolbar ?

Resources/public/css/views/notificationhub.css
```diff
@@ -11,5 +11,5 @@
 .is-navigationhubview-fixed .ez-view-notificationhubview,
 .is-navigation-hidden .ez-view-notificationhubview {
     width: 100%;
-    position: fixed;
+    position: relative;
```
dpobel: sure of that ? I mean the notifications should still be fixed, no ?

dpobel:
also you commit message should also refer to https://jira.ez.no/browse/EZP-25076 even if it's to fix EZP-25147

Resources/public/css/views/notificationhub.css
mhyndle: I have removed position attribute at all. `.ez-view-notificationhubview` is inside `.ez-notification-container` which is positioning with fixed so here is no reason to keep it.

mhyndle:
@dpobel : notification container moved above the symfony toolbar and added EZP-25076 to the commit message. I have also tested solution in other browsers - seems to be OK. Will be back with fixed unit tests soon ;)

mhyndle:
Ready for review

Resources/public/css/layout.css
```diff
@@ -83,11 +83,11 @@
...
 
 .ez-notification-container {
-    position: absolute;
-    top: 0;
+    position: fixed;
+    bottom: 0;
     left: 0;
     width: 100%;
-    z-index: 100;
+    z-index: 999999;
```
dpobel: You should put 6000001 so that the notification bar is also on top of the SF toolbar with SF 2.7

Resources/public/js/views/ez-notificationhubview.js
```diff
@@ -41,22 +41,11 @@ YUI.add('ez-notificationhubview', function (Y) {
...
-                this._fireHeightChange(
-                    oldHeight, this._getContainerHeight()
-                );
             }, this));
 
             list.on('remove', Y.bind(function (e) {
-                var oldHeight = this._getContainerHeight(),
-                    newHeight = oldHeight;
-
-                newHeight -= this._removeNotificationView(e.model);
```
dpobel: since we don't need to do the math anymore, I guess you can remove the height computation from `_removeNotificationView` ?

mhyndle: Yep, I have forget about it. `this._removeNotificationView` cleared from height calculations.

dpobel:
+1 (when Travis agrees)

StephaneDiot:
+1

PR #455, opened 2015-12-03
[bdd] Updated .ez.yml to use clean install

bdunogier:
Title says it all.


andrerom:
+1

this might fail, but then it won't be because of this change, so I'll merge it as what we have just can't work anyway.

PR #459, opened 2015-12-04
Fix EZP-24931: Unexpected result in the preview if the draft has not been saved

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24931

# Description

This patch fixes the issue where the preview would give expected and wrong result if the draft was not saved yet. Basically, when the user wants a preview, the generated event is transformed into a save draft event and the callback of this one actually triggers the preview.

Screencast: http://youtu.be/gw7RrHRu1D8

# Tests

manual + unit tests

### DIFF
--- a/Resources/public/js/views/actions/ez-buttonactionview.js
+++ b/Resources/public/js/views/actions/ez-buttonactionview.js
@@ -68,6 +68,17 @@ YUI.add('ez-buttonactionview', function (Y) {
         },
 
         /**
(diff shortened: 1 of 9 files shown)


dpobel:
ping @mhyndle @andrerom @StephaneDiot @yannickroger

mhyndle:
+1

andrerom:
+1

PR #462, opened 2015-12-07
[WIP] Fix EZP-25189: incorrect richtext toolbar rendering in IE11/Edge

mhyndle:
> status: work in progress

Jira: https://jira.ez.no/browse/EZP-25189

## Description
Button in richtext toolbar were cut in IE11/Edge because height of the button was set to fixed 40px. This PR removes that attribute for IE11/Edge.
While testing I have discovered that `:after` element for toolbar container used as indicator pointed on the caret was moved a bit lower than in other browsers so I pulled it a bit higher.
What's morein IE11/Edge buttons had outline effect, so I have removed it also.

## Tests
Manual tests in: IE11, Edge (IE12), Chrome, Firefox, Safari

### DIFF
--- a/Resources/public/css/external/alloy-editor-ez.css
+++ b/Resources/public/css/external/alloy-editor-ez.css
@@ -217,7 +217,6 @@
 
 .ae-ui IE10-PLUS::-ms-reveal,
 .ae-ui [class^=ae-toolbar] {
-  height: 40px;
   box-sizing: content-box;
 }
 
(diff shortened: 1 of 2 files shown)


mhyndle:
Ready for review:
ping: @dpobel @yannickroger @StephaneDiot @andrerom

andrerom:
+1

dpobel:
`alloy-editor-ez.css` is auto-generated (see README.md) based on https://github.com/ezsystems/PlatformUIBundle/tree/master/Resources/sass/alloy so you should not change it directly.
Also for the change 100%/98% seems a bit too magic, I'm pretty sure there's a cleaner way to fix it. Can you also put those kind of improvement in a separate commit ?

dpobel:
Replaced by https://github.com/ezsystems/PlatformUIBundle/pull/466

PR #463, opened 2015-12-07
Fix EZP-25233: Two scrollbar on iframes

yannickroger:
Link: https://jira.ez.no/browse/EZP-25233

## Description
This issue was introduced when the parameter was added to the `uiSetHeight` method.
https://github.com/ezsystems/PlatformUIBundle/commit/15b15eb8605b225afb37cded5257f947011bd20c#diff-b264e14fcac4d2600fa40aab505b968bR39

When called in an event, the parameter was by default an event facade, meaning not a Number as expected. This patch makes sure the method is called with `0` as parameter. I also added missing dependencies (it worked without them because they were called by other modules).

Thanks @dpobel for the big hint on this one.

## Test
Manual test. This part of the code is not unit tested as window resize is not easy to simulate.

### DIFF
--- a/Resources/public/js/views/ez-dashboardview.js
+++ b/Resources/public/js/views/ez-dashboardview.js
@@ -33,7 +33,7 @@ YUI.add('ez-dashboardview', function (Y) {
          */
         render: function () {
             this.get('container').setHTML(this.template());
-            this._attachedViewEvents.push(Y.on("windowresize", Y.bind(this._uiSetHeight, this)));
+            this._attachedViewEvents.push(Y.on("windowresize", Y.bind(this._uiSetHeight, this, 0)));
 
             return this;
         },
--- a/Resources/config/yui.yml
+++ b/Resources/config/yui.yml
@@ -230,7 +230,7 @@ system:
                     type: 'template'
                     path: %ez_platformui.public_dir%/templates/loginform.hbt
                 ez-dashboardview:
-                    requires: ['ez-templatebasedview', 'dashboardview-ez-template', 'ez-height-fit']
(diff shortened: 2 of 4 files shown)


yannickroger:
ping @dpobel @StephaneDiot @mhyndle @sunpietro

mhyndle:
+1

lserwatka:
@sunpietro please check if this change does not have any impact on studio

Resources/public/js/views/ez-dashboardview.js
```diff
@@ -33,7 +33,13 @@ YUI.add('ez-dashboardview', function (Y) {
          */
         render: function () {
             this.get('container').setHTML(this.template());
-            this._attachedViewEvents.push(Y.on("windowresize", Y.bind(this._uiSetHeight, this)));
+            this._attachedViewEvents.push(
+                Y.on("windowresize", Y.bind(
```
dpobel: the indent is weird :)
Too avoid issue, I would write something like:
```js
var updateHeight = function () {
        this._uiSetHeight();
    };

this._attachedViewEvents.push(Y.on("windowresize", Y.bind(updateHeight, this));
```

dpobel:
> This part of the code is not unit tested as window resize is not easy to simulate.

a perfect example of "not tested code is likely to fail" :)

besides the code style remark, +1

yannickroger:
New approach, with smaller diff (hinted by @dpobel) instead of fixing the CS.
ping @mhyndle

dpobel:
+1 (when Travis agrees even if this line is not actually tested)

mhyndle:
+1

PR #465, opened 2015-12-07
Fix EZP-25203: Pressing <Enter> when creating a role saves the role but don't publish it

glye:
> Fixes https://jira.ez.no/browse/EZP-25203
> Also required: https://github.com/ezsystems/repository-forms/pull/65
> Status: Ready for review

Let the default action call the save role action. Let the save role action both save and publish.

### DIFF
--- a/Form/Processor/RoleFormProcessor.php
+++ b/Form/Processor/RoleFormProcessor.php
@@ -54,14 +54,24 @@ class RoleFormProcessor implements EventSubscriberInterface
         ];
     }
 
+    /**
+     * Save the role if no button was clicked (i.e. enter was pressed in the form).
+     *
+     * The role should not be saved if cancel was pressed, and if save was pressed this is done in the save action.
+     *
+     * @param \EzSystems\RepositoryForms\Event\FormActionEvent $event
+     */
     public function processDefaultAction(FormActionEvent $event)
     {
-        $this->notify('role.notification.draft_saved', [], 'role');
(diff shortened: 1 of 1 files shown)


andrerom:
+1

bdunogier:
Not 100% convinced, but I don't have anything better to suggest. Would need to look deeper into it.

+1 if it works...

Form/Processor/RoleFormProcessor.php
```diff
@@ -56,12 +56,19 @@ public static function getSubscribedEvents()
 
     public function processDefaultAction(FormActionEvent $event)
     {
-        $this->notify('role.notification.draft_saved', [], 'role');
+        // Don't update anything if we just want to cancel the draft.
+        // Don't do the default action when saving is specified, as it's the same action and shouldn't be done twice.
+        if (in_array($event->getClickedButton(), ['removeDraft', 'saveRole'])) {
```
dpobel: wouldn't it be more logic to check if no button was clicked ?

glye: @dpobel Something like this? Ok. Also I can move the explaining comment to the docblock.
```php
if ($event->getClickedButton() === null) {
    $this->processSaveRole($event);
}
```

glye:
Changed the default action to save if no button was clicked. (No functionality change, just clearer code.) @dpobel

bdunogier:
+1

PR #466, opened 2015-12-08
Fix EZP-25189: Richtext tollbar container doesn't render correctly on IE11/Edge

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25189

# Description

Replaces https://github.com/ezsystems/PlatformUIBundle/pull/462
This is fixing a display issue in IE11 (and Edge) because of IE10+ specific rule setting a fixed height in the AlloyEditor CSS (we can not really change that in our skin).

While at it, I also fixed a small gap between the toolbar tail and the toolbar itself.

# Tests

manual test in IE11, Chrome and Firefox (currently installing a VM to test with Edge)

### DIFF
--- a/Resources/public/css/alloyeditor/buttons/labeled-button.css
+++ b/Resources/public/css/alloyeditor/buttons/labeled-button.css
@@ -9,6 +9,15 @@
     margin: 0.3em 0.8em;
 }
 
+.ae-ui IE10-PLUS::-ms-reveal,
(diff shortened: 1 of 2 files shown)


dpobel:
ping @mhyndle @andrerom @yannickroger

andrerom:
+1

mhyndle:
+1

dpobel:
Edge seems happy with this patch

PR #468, opened 2015-12-08
Fix EZP-25225: remove trash icon from cancel button in contenttype edit view

mhyndle:
> Jira: https://jira.ez.no/browse/EZP-25225
> status: ready for review

## Description
This PR removes trash icon from the cancel button in the edit content type view.
While fixing that I've checked another serverside views and there are also the same issue, so I have removed the trash icon from cancel button also in following views (all are in admin panel):
- edit content type group
- edit language
- edit role
- edit role's policies
- edit section

### DIFF
--- a/Resources/views/ContentType/edit_content_type_group.html.twig
+++ b/Resources/views/ContentType/edit_content_type_group.html.twig
@@ -39,7 +39,7 @@
 
         <div class="pure-controls">
             <a href="{{ path("admin_contenttypeGroupList") }}"
-               class="pure-button ez-button ez-button-delete ez-font-icon">{{ "content_type.group.cancel"|trans }}</a>
+               class="pure-button ez-button">{{ "content_type.group.cancel"|trans }}</a>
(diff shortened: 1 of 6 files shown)


mhyndle:
Ready for review

glye:
`ez-button-delete` is used on many cancel buttons - role, section, language etc. This will remove the trash icon from all of them, is that the intention?

yannickroger:
Looks like @glye has a point: `git grep ez-button-delete`

dpobel:
@glye is right, we should remove that class from the cancel buttons all over the place. But the class `ez-button-delete` is used in others places where it's perfectly fine, so the change in the CSS file is wrong.

mhyndle:
Yeah of course @glye you are right that I was hastily removing CSS style.
I have also removed trash icon from other views (check PR description)

glye:
+1, though I'm not a gang member :)

andrerom:
+1, this is not js anymore so gang membership not a must :)

dpobel:
+1

PR #469, opened 2015-12-08
 Fix EZP-25209: Create button seems available even if it's not

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25209

# Description

The issue is about styling the buttons in the bar views when they are disabled and especially the "Create" button. While working on that, I also discovered that the *disabled logic* was actually implemented in the template for the create button and not implemented at all in the generic button.

So this PR fixes EZP-25209 (in c5043f7) and makes sure the button action view implements the disabled logic instead of relying on the template.

disabled_button

# Tests

manual tests + unit tests


dpobel:
ping @yannickroger @mhyndle @andrerom

yannickroger:
+1

mhyndle:
+1

andrerom:
+1

PR #473, opened 2015-12-10
Fix EZP-25264: Policy "Edit limitations" button should be disabled when there is nothing to edit

glye:
> Fixes https://jira.ez.no/browse/EZP-25264
> Status: Ready for review

Disable the edit limitations button when the policy cannot have limitations, thus we avoid showing the user an empty edit form.

The general way of doing things is taken from https://jira.ez.no/browse/EZP-23733
The limitation checker line is taken from `editPolicyAction()`.


glye:
Template fix: Check both `can_edit` and `editablePolicies`, both must be true.

andrerom:
+1

bdunogier:
+1

PR #475, opened 2015-12-11
Fix EZP-25229: disable move action button if given location is the location from 1st level of content tree

mhyndle:
> Jira: https://jira.ez.no/browse/EZP-25229
> status: ready for review

## Description
This PR fixes the bug when "Move" action button was not disabled for locations of 1st level of content tree ("Content Structure", "Media Library", "Users"). This patch disables "Move" action button for those locations.

## Tests
manual + unit tests

### DIFF
--- a/Resources/public/js/views/ez-actionbarview.js
+++ b/Resources/public/js/views/ez-actionbarview.js
@@ -45,11 +45,11 @@ YUI.add('ez-actionbarview', function (Y) {
                             priority: 200,
                             content: this.get('content')
                         }),
-                        new Y.eZ.ButtonActionView({
+                        new Y.eZ.MoveContentActionView({
                             actionId: "move",
-                            disabled: false,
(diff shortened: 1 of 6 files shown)


mhyndle:
ready for review

andrerom:
+1

Resources/public/js/views/ez-actionbarview.js
```diff
@@ -45,11 +45,12 @@ YUI.add('ez-actionbarview', function (Y) {
                             priority: 200,
                             content: this.get('content')
                         }),
-                        new Y.eZ.ButtonActionView({
+                        new Y.eZ.MoveContentActionView({
                             actionId: "move",
                             disabled: false,
```
dpobel: you should remove it since it's "automatically set" now

dpobel:

PR #477, opened 2015-12-11
EZP-24841: display breadcrumbs in the related content tab

mhyndle:
> Jira: https://jira.ez.no/browse/EZP-24841
> status: ready for review

## Description
Currently in the related content tab there are being displayed only names of related contents linked with their main location. This PR changes it to display breadcrumbs instead.

## Screenshot
ezp-24841

## Tasks
- [x] implementation
- [x] css
- [x] tests
 - [x] manual tests
 - [x] unit tests

### DIFF
--- a/Resources/config/css.yml
+++ b/Resources/config/css.yml
@@ -164,7 +164,6 @@ system:
                 - 'bundles/ezplatformui/css/theme/modules/selection-filter.css'
                 - 'bundles/ezplatformui/css/theme/modules/button.css'
                 - 'bundles/ezplatformui/css/theme/modules/selection-table.css'
-                - 'bundles/ezplatformui/css/theme/modules/breadcrumbs.css'
                 - 'bundles/ezplatformui/css/theme/modules/yui-calendar.css'
                 - 'bundles/ezplatformui/css/theme/alloyeditor/general.css'
                 - 'bundles/ezplatformui/css/theme/alloyeditor/content.css'
--- a/Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
@@ -28,45 +28,157 @@ YUI.add('ez-objectrelationsloadplugin', function (Y) {
          * Loads the related contents. Once this is done, it sets the contents in
          * the `relatedContents` attribute of the event target.
          *
+         * Since version 1.1, the `loadLocation` and `loadLocationPath` are taken into account
+         * to provide the Location with optionally its path. If `loadLocation` or `loadLocationPath`
+         * is TRUE, then array of content structs is being set in the `relatedContents` attribute of the event target.
+         * Content struct contains `content` (eZ.Content) and `location` (eZ.Location) attributes.
+         *
          * @protected
          * @method _loadObjectRelations
          * @param {Object} e ObjectRelations event facade
+         * @param {Bool} e.loadLocation flag indicating whether to load the Location
+         * @param {Bool} e.loadLocationPath flag indicating whether to load the Location's path. If it is TRUE it
+         * forces to e.loadLocation be also TRUE
+         * @since 1.1
          */
         _loadObjectRelations: function (e) {
-            var Content = this.get('contentModelConstructor'),
-                loadOptions = {api: this.get('host').get('capi')},
-                relatedContentListArray = [],
+            var relatedContentListArray = [],
                 stack = new Y.Parallel(),
                 loadedRelation = {},
                 loadingError = false,
                 contentDestinations = this.get('host').get('content').relations(
                     e.relationType, e.fieldDefinitionIdentifier
-                );
+                ),
+                end = stack.add(function (error, struct) {
+                    if (error) {
+                        e.target.set("loadingError", true);
+                        loadingError = true;
+                    } else {
+                        relatedContentListArray.push(struct);
+                    }
+                });
 
             Y.Array.each(contentDestinations, function (value) {
-                var actualRelatedContent;
-
                 if (!loadedRelation[value.destination]) {
-                    actualRelatedContent = new Content();
                     loadedRelation[value.destination] = true;
 
-                    actualRelatedContent.set('id', value.destination);
-                    actualRelatedContent.load(loadOptions, stack.add(function (error) {
-                        if (error) {
-                            e.target.set("loadingError", true);
-                            loadingError = true;
-                        } else {
-                            relatedContentListArray.push(actualRelatedContent);
-                        }
-                    }));
+                    if (e.loadLocation || e.loadLocationPath) {
+                        this._loadContentStruct(value.destination, e.loadLocation, e.loadLocationPath, end);
+                    } else {
+                        this._loadContent(value.destination, end);
+                    }
                 }
-            });
+            }, this);
 
             stack.done(function () {
                 e.target.setAttrs({
                     relatedContents: relatedContentListArray,
                     loadingError: loadingError,
                 });
+            });
+        },
+
+        /**
+         * Loads content for given content id.
+         *
+         * @method _loadContent
+         * @protected
+         * @param {String} contentId
+         * @param {Function} callback
+         * @param {Bool} callback.error
+         * @param {eZ.Content} callback.content
+         * @since 1.1
+         */
+        _loadContent: function (contentId, callback) {
+            var ContentModel = this.get('contentModelConstructor'),
+                content = new ContentModel(),
+                loadOptions = {api: this.get('host').get('capi')};
+
+            content.set('id', contentId);
+            content.load(loadOptions, function (error) {
+                if (error) {
+                    callback(error);
+                } else {
+                    callback(error, content);
+                }
+            });
+        },
+
+        /**
+         * Loads content struct containing content and main location for given content id.
+         * The contentStruct is being passed to the callback function.
+         *
+         * @method _loadContentStruct
+         * @protected
+         * @param {String} contentId
+         * @param {Bool} loadLocation flag indicating whether to load the Location
+         * @param {Bool} loadLocationPath flag indicating whether to load the Location's path. If it is TRUE it
+         * forces to `loadLocation` be also TRUE
+         * @param {Function} callback
+         * @param {Bool} callback.error
+         * @param {Object} callback.contentStruct content struct with loaded content and location
+         * @param {eZ.Content} callback.contentStruct.content
+         * @param {eZ.Location} callback.contentStruct.location
+         * @since 1.1
+         */
+        _loadContentStruct: function (contentId, loadLocation, loadLocationPath, callback) {
+            var loadLocationContentStruct = Y.bind(this._loadLocationToContentStruct, this),
+                contentStruct;
+
+            this._loadContent(contentId, function (error, content) {
+                if (error) {
+                    callback(error);
+                } else {
+                    contentStruct = {content: content};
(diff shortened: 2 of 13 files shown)


mhyndle:
ready for review

Resources/public/css/views/tabs/relations.css
```diff
@@ -33,3 +33,24 @@
 .ez-view-locationviewrelationstabview .ez-relations-box-list-no-content {
     font-style: italic;
 }
+
+.ez-view-locationviewrelationstabview .ez-breadcrumbs-list {
+    font-size: 90%;
```
dpobel: I think it's more than time to put that CSS code in a dedicated CSS module as we pretty much the same code in different places. Also, I don't see the need for the 90% font size here, I think this was just copy pasted from the Location view breadcrumb but in this context and in the related content tab, this is weird.

Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
```diff
@@ -33,9 +33,8 @@ YUI.add('ez-objectrelationsloadplugin', function (Y) {
          * @param {Object} e ObjectRelations event facade
          */
         _loadObjectRelations: function (e) {
-            var Content = this.get('contentModelConstructor'),
-                loadOptions = {api: this.get('host').get('capi')},
-                relatedContentListArray = [],
+            var that = this,
```
dpobel: nitpick: good resolution idea for the new year, stop writing the ugly `that = this` and bind the function or better here use the third argument of Y.Array.each/Y.Object.each :)

Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
```diff
@@ -64,11 +59,78 @@ YUI.add('ez-objectrelationsloadplugin', function (Y) {
...
+                that = this,
+                contentStruct;
+
+            content.set('id', contentId);
+            content.load(loadOptions, function (error) {
+                if (error) {
+                    callback(error);
+                } else {
+                    contentStruct = {content: content};
+                    that._loadLocationToContentStruct(contentStruct, callback);
```
dpobel: loading the Location and loading the Location's path should be triggered by a flag in the event otherwise, we'll do that everytime while this is not needed for instance when displaying the relation when editing a content.

Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
```diff
@@ -25,7 +25,7 @@ YUI.add('ez-objectrelationsloadplugin', function (Y) {
         },
 
         /**
-         * Loads the related contents. Once this is done, it sets the contents in
+         * Loads the related contents. Once this is done, it sets the content structs in
```
dpobel: that's a BC break. This is also related to this comment https://github.com/ezsystems/PlatformUIBundle/pull/477/files#r47764778
We have 2 options here:
1. we keep the BC and when the event is fired without the `loadLocation` (nor the `loadLocationPath`) flag, we're still sending back only an array of Content models
2. we do this BC break, so we have to document it (I guess we need to create bc document in the repository ?)

Also, in both case, we need to document the new behaviour in the API doc so we have to add the `@since` tag, I guess with the value `2016.02` ?

I would choose 1. even if that's not super consistent in terms of API.
ping @andrerom @bdunogier

Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
```diff
@@ -81,7 +143,17 @@ YUI.add('ez-objectrelationsloadplugin', function (Y) {
...
+            },
+
+            /**
+             * Location constructor
+             *
+             * @attribute locationModelConstructor
+             * @type Y.eZ.Location
+             */
+            locationModelConstructor: {
+                value: Y.eZ.Location
```
dpobel: this should be reflected in `yui.yml`, the Location model is now a dependency of this plugin.

mhyndle: Added `ez-locationmodel` as required for `ez-objectrelationsloadplugin`

Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
andrerom: I'm ok with the suggestion.
Only alternative I see is 1. for 1.0.0 and 2. for 1.1.0.

dpobel: well, we talked about that this morning. And the conclusion is to do the solution 1 and to add `@since 1.1` to be follow semver.

> 2. for 1.1.0.
remark: to follow semver, that would be 2.0 actually

mhyndle:
all feedback taken into account, ready for re-review

Resources/config/css.yml
```diff
@@ -89,6 +89,7 @@ system:
                 - 'bundles/ezplatformui/css/modules/selection-filter.css'
                 - 'bundles/ezplatformui/css/modules/button.css'
                 - 'bundles/ezplatformui/css/modules/table-data.css'
+                - 'bundles/ezplatformui/css/modules/breadcrumbs-list.css'
                 - 'bundles/ezplatformui/css/modules/breadcrumbs.css'
```
dpobel: ah we already had a breadcrumbs.css module ? so I don't get why you are adding breadcrumbs-list.css, aren't all our breadcrumbs based on a list ?

The point here is to avoid having n times the same (or same-ish) CSS for the same thing. So this PR is an opportunity to unify the CSS (and HTML) code of our breadcrumbs to use the same thing and only keep the very specific CSS out of the breadcrumb module.
So you should check the breadcrumb in the LocationView, the ones in the Locations and Related Content tabs and the ones in the admin panel (like when viewing a Section) so that they use the breadcrumbs.css module for most of their style.

dpobel:
on the JS code side, that seems good to me.

mhyndle:
I have made some reorganisation with CSS so currently as a base CSS class for breadcrumbs will be `.ez-breadcrumbs-list`. All uncommon stylings like changing font or padding for separator should be made in particullar views styles.

Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
```diff
@@ -44,23 +51,34 @@ YUI.add('ez-objectrelationsloadplugin', function (Y) {
...
+                                    e.target.set("loadingError", true);
+                                    loadingError = true;
+                                } else {
+                                    relatedContentListArray.push(contentStruct);
+                                }
+                            }
+                        ));
+                    } else {
+                        this._loadContent(value.destination, stack.add(
+                            function (error, content) {
```
dpobel: the very same function is used in both case, you should put it in a local variable.

Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
```diff
@@ -69,6 +87,107 @@ YUI.add('ez-objectrelationsloadplugin', function (Y) {
...
+        /**
+         * Loads content for given content id.
+         *
+         * @method _loadContent
+         * @protected
+         * @param {String} contentId
+         * @param {Function} callback
+         * @param {Bool} callback.error
+         * @param {eZ.Content} callback.content
+         */
```
dpobel: missing `@since` tag

Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
```diff
@@ -69,6 +87,107 @@ YUI.add('ez-objectrelationsloadplugin', function (Y) {
...
+         * @protected
+         * @param {String} contentId
+         * @param {Bool} loadLocation flag indicating whether to load the Location
+         * @param {Bool} loadLocationPath flag indicating whether to load the Location's path. If it is TRUE it
+         * forces to `loadLocation` be also TRUE
+         * @param {Function} callback
+         * @param {Bool} callback.error
+         * @param {Object} callback.contentStruct content struct with loaded content and location
+         * @param {eZ.Content} callback.contentStruct.content
+         * @param {eZ.Location} callback.contentStruct.location
```
dpobel: missing `@since` tag

Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
```diff
@@ -69,6 +87,107 @@ YUI.add('ez-objectrelationsloadplugin', function (Y) {
...
+         * @method _loadLocationToContentStruct
+         * @protected
+         * @param {Object} contentStruct
+         * @param {eZ.Content} contentStruct.content eZ.Content object
+         * @param {Bool} loadLocationPath flag indicating whether to load the Location's path
+         * @param {Function} callback
+         * @param {Bool} callback.error
+         * @param {Object} callback.contentStruct content struct with loaded content and location
+         * @param {eZ.Content} callback.contentStruct.content
+         * @param {eZ.Location} callback.contentStruct.location
```
dpobel: missing `@since` tag

Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
```diff
@@ -81,7 +200,17 @@ YUI.add('ez-objectrelationsloadplugin', function (Y) {
...
             contentModelConstructor: {
                 value: Y.eZ.Content
-            }
+            },
+
+            /**
+             * Location constructor
+             *
+             * @attribute locationModelConstructor
+             * @type Y.eZ.Location
```
dpobel: I would also add hhe `@since` tag here (I hope yuidoc handle it correctly in such context)

Resources/public/templates/tabs/relations.hbt
```diff
@@ -22,7 +22,17 @@
             {{#each relatedContents}}
             <tr>
                 <td>
-                    <a href="{{path "viewLocation" id=mainLocationId languageCode=content.mainLanguageCode}}">{{content.name}}</a>
+                    <ul class="ez-breadcrumbs-list">
+                        {{#each location.path}}
+                        <li class="ez-breadcrumbs-item">
```
StephaneDiot: well that's really niptick, but isn't it missing an indentation level ?

StephaneDiot:
+1

Resources/public/templates/tabs/relations.hbt
mhyndle: That's how indention is done when mixing HTML with handlebars template functions. What's more, HTML generated with that approach will be well-indented. But of course I can change it if it's against used CS. ping @dpobel

Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
mhyndle: `@since 1.1` added

Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
mhyndle: `@since 1.1` added

Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
mhyndle: `@since 1.1` added

mhyndle:
@dpobel feedback taken into account

dpobel:
+1

yannickroger:
+1

PR #479, opened 2015-12-14
Fix EZP-24848: Content Type removal does not take into account Contents in the trash

glye:
> Fixes https://jira.ez.no/browse/EZP-24848
> Also requires https://github.com/ezsystems/ezpublish-kernel/pull/1539
> Status: Ready for review

Checks for existing content, if there is any it disables delete buttons and shows an alt text.

### DIFF
--- a/Controller/ContentTypeController.php
+++ b/Controller/ContentTypeController.php
@@ -126,9 +126,7 @@ class ContentTypeController extends Controller
                 ['contentTypeId' => $contentTypeId]
             )->createView();
 
-            $countQuery = new Query(['filter' => new Query\Criterion\ContentTypeId($contentTypeId), 'limit' => 0]);
-            $contentCount = $this->searchService->findContent($countQuery, [], false)->totalCount;
-            $canDeleteById[$contentTypeId] = $canDelete && $contentCount == 0;
+            $canDeleteById[$contentTypeId] = $canDelete && !$this->contentTypeService->isContentTypeUsed($contentType);
         }
 
         return $this->render('eZPlatformUIBundle:ContentType:view_content_type_group.html.twig', [
@@ -225,7 +223,7 @@ class ContentTypeController extends Controller
             'modifier' => $this->userService->loadUser($contentType->modifierId),
             'delete_form' => $deleteForm->createView(),
             'can_edit' => $this->isGranted(new Attribute('class', 'update')),
-            'can_delete' => ($this->isGranted(new Attribute('class', 'delete')) && $contentCount == 0),
+            'can_delete' => ($this->isGranted(new Attribute('class', 'delete')) && !$this->contentTypeService->isContentTypeUsed($contentType)),
         ]);
     }
 
--- a/Resources/translations/content_type.en.xlf
+++ b/Resources/translations/content_type.en.xlf
@@ -138,6 +138,10 @@
         <source>content_type.notification.deleted</source>
         <target>The Content type "%contentTypeName%" was successfully deleted.</target>
       </trans-unit>
+      <trans-unit id="1d549f54e575ebb7f13e696af6c5bb96">
+        <source>content_type.is_in_use</source>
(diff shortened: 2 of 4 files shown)


glye:
Review ping @andrerom @bdunogier

Resources/translations/content_type.en.xlf
```diff
@@ -138,6 +138,10 @@
         <source>content_type.notification.deleted</source>
         <target>The Content type "%contentTypeName%" was successfully deleted.</target>
       </trans-unit>
+      <trans-unit id="1d549f54e575ebb7f13e696af6c5bb96">
+        <source>content_type.is_in_use</source>
+        <target>The content type cannot be deleted because it has content instances.</target>
```
bdunogier: "content items" instead of "content instances" ?

Controller/ContentTypeController.php
```diff
@@ -37,6 +38,11 @@ class ContentTypeController extends Controller
     private $contentTypeService;
 
     /**
+     * @var ContentService
```
bdunogier: FQN in phpdoc please.

ctrl+shift+c on "SearchService" in PhpStorm, then ctrl+w, then paste :-)

bdunogier:
+1 the nitpicks aside.

glye:
Changed translation message, FQN added, thanks.

bdunogier:
Still missing a +1, ping @andrerom ?

glye:
@bdunogier NB see the "Also requires" kernel dependency, which should be reviewed and merged first.

joaoinacio:
Hi, my 2 cents here:
If I understand it right, it seems the message ("content_type.is_in_use") is the same regardless of the reason, which could make things confusing if there is no delete permission but there are also no content objects of the type.
IMHO, it would make more sense to have a separate message for each case?

glye:
@joaoinacio Seems legit. I see three solutions:
- Separate message for each case (complicates code and template)
- Refuse without stating reason (simple, but bad ux)
- Refuse while stating multiple possible reasons ("The content type cannot be deleted because it has content items, or beause you don't have permission").
I like the last one. Opinions?

joaoinacio:
@glye: agreed, option 3 seems quite reasonable here as the first could complicate things quite a bit, in any case @andrerom should have some feedback on this.

glye:
Moved the code to ContentTypeService, see related PR.
Changed error message to state both possible reasons why you can't delete the CT (see option 3: https://github.com/ezsystems/PlatformUIBundle/pull/479#issuecomment-168641338 )

Controller/ContentTypeController.php
```diff
@@ -225,7 +223,7 @@ public function viewContentTypeAction($contentTypeId, $languageCode = null)
             'modifier' => $this->userService->loadUser($contentType->modifierId),
             'delete_form' => $deleteForm->createView(),
             'can_edit' => $this->isGranted(new Attribute('class', 'update')),
-            'can_delete' => ($this->isGranted(new Attribute('class', 'delete')) && $contentCount == 0),
+            'can_delete' => ($this->isGranted(new Attribute('class', 'delete')) && !$this->contentTypeService->isContentTypeUsed($contentType)),
```
crevillo: another nitpick and not related entirely with this pr but as more high level... doesn't ```class``` sounds a bit legacy? :wink: i mean, maybe we could think in change that to "contentType" too...

glye: Yeah, out of scope here but should probably be changed at some point.

andrerom: Renaming policies is a separate task indeed, for that we might need to add a bc layer to map between old and new policies for instance..

andrerom:
With the now updated kernel PR: +1

glye:
@bdunogier Does your old +1 still apply?

bdunogier:
+1 it is.

PR #480, opened 2015-12-15
EZP-25128: Automatically navigate to 1st item of a zone

yannickroger:
Link: https://jira.ez.no/browse/EZP-25128

## Description
Before this patch. When clicking on a top menu item (like "Content") sub items of this menu would be displayed. With this patch, on top of the existing behavior, the first item is automatically opened.

Also, I fixed a few typos I ran into in separate commits.

## Tests
Manual and automated

### DIFF
--- a/.ez.yml
+++ b/.ez.yml
@@ -3,8 +3,8 @@ v: 0
 # This is a bundle, so we need to specify info for meta repo
 meta:
   repo: "https://github.com/ezsystems/ezplatform.git"
-  branch: "v1.2.0"
-  self_alias: "1.2.x-dev"
+  branch: "master"
+  self_alias: "1.3.x-dev"
 
 # ci
 integrate:
--- a/Resources/public/js/views/services/ez-navigationhubviewservice.js
+++ b/Resources/public/js/views/services/ez-navigationhubviewservice.js
@@ -503,7 +503,7 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
              * object for the navigation item view. This configuration must
              * contain a `title` and an `identifier` properties.
              *
-             * @attribute platformNavigationItems
+             * @attribute adminNavigationItems
              * @type Array
              * @default array containing the items for the admin
              * @readOnly
--- a/Resources/public/js/views/services/ez-navigationhubviewservice.js
+++ b/Resources/public/js/views/services/ez-navigationhubviewservice.js
@@ -280,7 +280,7 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
          * the navigation hub view to check which navigation item view is
          * selected. A matched route object is a clone of the application active
          * route without the service, serviceInstance and callbacks entries and
-         * with an additionnal `parameters` property holding the route
+         * with an additional `parameters` property holding the route
          * parameters and their values.
          *
          * @method _matchedRoute
--- a/Resources/public/js/views/services/ez-navigationhubviewservice.js
+++ b/Resources/public/js/views/services/ez-navigationhubviewservice.js
@@ -23,6 +23,7 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
         initializer: function () {
             this.on('*:logOut', this._logOut);
             this.after('*:navigateTo', this._navigateTo);
+            this.after('*:activeNavigationChange', this._navigateToFirstItemRoute);
         },
 
         /**
@@ -51,6 +52,43 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
             app.logOut(function () {
                 app.navigateTo('loginForm');
             });
+        },
+
+        /**
+         * Returns true if a provided item is a NavigationItemView, false if item is a NavigationItemParameterView
+         *
+         * @method _isNavigationViewItem
+         * @private
+         * @param {Object} item
+         * @return {Boolean}
+         */
+        _isNavigationViewItem: function (item) {
+            return !!item.get;
+        },
+
+        /**
+         * Redirects to the first item of the active zone
+         *
+         * @method _navigateToFirstItemRoute
+         * @protected
+         * @param {EventFacade} e
+         * @param {String} e.newVal new value of `activeNavigation`
+         */
+        _navigateToFirstItemRoute: function (e) {
+            var zone = this.get(e.newVal + 'NavigationItems'),
+                firstItem,
+                route;
+
+            if (zone && zone.length > 0) {
+                firstItem = zone[0];
+
+                if (this._isNavigationViewItem(firstItem)) {
+                    route = firstItem.get('route');
+                } else {
+                    route = firstItem.config.route;
+                }
+                this.get('app').navigateTo(route.name, route.params);
+            }
         },
 
         _load: function (next) {
@@ -331,13 +369,13 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
         removeNavigationItem: function (identifier, zone) {
             var attr = zone + 'NavigationItems';
 
-            this._set(attr, Y.Array.reject(this.get(attr), function (elt) {
-                if (elt.get) {
+            this._set(attr, Y.Array.reject(this.get(attr), Y.bind(function (elt) {
+                if (this._isNavigationViewItem(elt)) {
                     return elt.get('identifier') === identifier;
                 } else {
                     return elt.config.identifier === identifier;
                 }
-            }));
+            },this)));
         },
 
         /**
@@ -355,13 +393,13 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
                 items = items.concat(this.get(zone + 'NavigationItems'));
             }, this);
 
-            return Y.Array.find(items, function (elt) {
-                if (elt.get) {
+            return Y.Array.find(items, Y.bind(function (elt) {
+                if (this._isNavigationViewItem(elt)) {
                     return elt.get('identifier') === identifier;
                 } else {
                     return false;
                 }
-            });
+            }, this));
         }
     }, {
         ATTRS: {
--- a/Resources/public/js/views/services/ez-navigationhubviewservice.js
+++ b/Resources/public/js/views/services/ez-navigationhubviewservice.js
@@ -79,7 +79,7 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
                 firstItem,
                 route;
 
-            if (zone && zone.length > 0) {
+            if (zone && zone.length > 0 && e.prevVal !== null) {
                 firstItem = zone[0];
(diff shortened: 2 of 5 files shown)


yannickroger:
@dpobel @mhyndle ready for review.

Resources/public/js/views/services/ez-navigationhubviewservice.js
```diff
@@ -53,6 +54,28 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
             });
         },
 
+        /**
+         * activeNavigationChange event handler, it redirects to the first item of the clicked zone
+         *
+         * @method _activeNavigationChange
+         */
```
mhyndle: missing param docs

yannickroger: Damn, I often ask people for those params, and I forget them :)

Tests/js/views/services/assets/ez-navigationhubviewservice-tests.js
```diff
@@ -895,6 +896,79 @@ YUI.add('ez-navigationhubviewservice-tests', function (Y) {
...
+                        params: this.routeParams,
+                    }
+                }
+            };
+
+            this.appMock = new Mock();
+
+            Mock.expect(this.appMock, {
+                method: 'navigateTo',
+                args: [Mock.Value.String, Mock.Value.Any],
```
mhyndle: You can simplify it a bit with something like:
`args: [that.routeName, that.routeParams]`
so no need of defining `run` then

Tests/js/views/services/assets/ez-navigationhubviewservice-tests.js
```diff
@@ -895,6 +896,79 @@ YUI.add('ez-navigationhubviewservice-tests', function (Y) {
...
+
+            this.view = new Y.View({
+                'bubbleTargets': this.service,
+                'activeNavigation': ''
+            });
+        },
+
+        tearDown: function () {
+            this.service.destroy();
+            delete this.service;
```
mhyndle: Destroy and delete `this.view` also?

Tests/js/views/services/assets/ez-navigationhubviewservice-tests.js
mhyndle: Edit: just `Mock.verify(this.appMock)`  to verify if `nvigateTo` was called.

Resources/public/js/views/services/ez-navigationhubviewservice.js
```diff
@@ -23,6 +23,7 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
         initializer: function () {
             this.on('*:logOut', this._logOut);
             this.after('*:navigateTo', this._navigateTo);
+            this.after('*:activeNavigationChange', this._activeNavigationChange);
```
dpobel: `_activeNavigationChange` is a very bad name as it gives absolutely no clue of it is doing. Good method names usually start with a verb and tell what the method does, not why/when it is called. So here for instance, the name could be `_navigateToFirstItemRoute`. That way, you can almost read that line as an English sentence.

yannickroger:
@mhyndle 's feedback taken into account.

Resources/public/js/views/services/ez-navigationhubviewservice.js
yannickroger: That used to be its name… then I followed the examples of the ones above, named after the even :( I'll fix it.

Resources/public/js/views/services/ez-navigationhubviewservice.js
```diff
@@ -53,6 +54,31 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
...
+         */
+        _activeNavigationChange: function (e) {
+            var zone = this.get(e.newVal + 'NavigationItems'),
+                firstItem,
+                route;
+
+            if (zone && zone.length > 0) {
+                firstItem = zone[0];
+
+                if (firstItem.config) {
```
dpobel: we already have the same kind of test to detect whether it's a config or if we already have a true navigation item view and this is done by checking the `get` method in https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/services/ez-navigationhubviewservice.js#L343 and https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/services/ez-navigationhubviewservice.js#L319 so I guess it's time to introduce a small private method to do that.

Tests/js/views/services/assets/ez-navigationhubviewservice-tests.js
```diff
@@ -895,6 +896,83 @@ YUI.add('ez-navigationhubviewservice-tests', function (Y) {
...
+            this.view.destroy();
+            delete this.view;
+        },
+
+        "Should not do anything when no item": function () {
+            this.view.set('activeNavigation', this.zoneName);
```
dpobel: while the code appears to be covered by that, this is incomplete. Before this line, you should (re)configure `appMock` to not expect any call to `navigateTo` and then call `Mock.verify` at the end.

yannickroger:

Resources/public/js/views/services/ez-navigationhubviewservice.js
```diff
@@ -53,6 +54,44 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
...
         },
 
+
+        /**
+         * Returns true if a provided item is a NavigationItemView
+         *
+         * @method _isNavigationViewItem
+         * @private
+         * @param {Object} item
+         * @return true if item is a NavigationItemView, false if item is a NavigationItemParameterView
```
dpobel: `@return {Boolean}`

if more than enough :)

Resources/public/js/views/services/ez-navigationhubviewservice.js
```diff
@@ -53,6 +54,44 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
...
+         */
+        _isNavigationViewItem: function (item) {
+            return item.get;
```
dpobel: Nitpick: if you want it to return a boolean it should actually be `return !!item.get;`

Resources/public/js/views/services/ez-navigationhubviewservice.js
```diff
@@ -53,6 +54,44 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
...
+        },
+
+        /**
+         * Redirects to the first item of the clicked zone
```
dpobel: nitpick: s/clicked/active/

dpobel:
besides the 3 nitpicks, +1

yannickroger:
" @dpobel 's feedback strike back " taken into account, ping @mhyndle

dpobel:
more "@dpobel awakens" today ;-)

yannickroger:
I was afraid of having a " Return of the @dpobel "

mhyndle:
just one thing about my comment: https://github.com/ezsystems/PlatformUIBundle/pull/480/files#r47629741
`run` is not doing anything else than making assertions which may be bit simplified with what I have proposed in the comment (or it's not valid?)
+1, besides above

yannickroger:
Sorry @mhyndle I missed it. It's taken care of now.

yannickroger:
Can this be merged with the broken BDD @bdunogier @andrerom @pcardiga or can you can (help me? ) fix it ?
I am still a bit lost with the new CI BDD system.

miguelcleverti:
@yannickroger there seems to be some problem with the roles behat tests, I am looking into it. However it does not seem related to this PR.

bdunogier:
I have executed behat (`./bin/behat --profile=platformui vendor/ezsystems/platform-ui-bundle/Features/Users/users.feature --tags='common'`) on my local instance, and it seeems that something is wrong.

Without the patch, ``Create a valid user` passes.
With it, it fails with a timeout. I have added a bit more data, catching the exception thrown by `spin()` from `fillFieldWithValue`: `Timeout while waiting for loading element '.is-app-loading, .is-app-transitioning, .yui3-app-transitioning'. (Exception)`.

miguelcleverti:
It seems that what is happening is that behat is interacting with the page before the automatic navigation to the 1st item of the zone.
This causes failures because behat expects something, but the page changed meanwhile.
The tests need to be updated to take this into account. @bdunogier @yannickroger

dpobel:
@miguelcleverti your patch (https://github.com/ezsystems/PlatformUIBundle/commit/941d718f2eb79f286394e7f8d338b48841c1ffb6) does not seem enough. Can you have another look please ? (and please do not open a new pull request, work directly on the ezp-25128-content_menu_link branch)

miguelcleverti:
A problem seems to have escaped me before! When accessing directly to a page URL instead of interacting with the UI causes the load of the first item. Ex:
Using the Role url `http://ezp6.dev/ez#/admin/pjax%2Frole` instead navigating through `Administration dashboard` > `Roles` in the UI.
IMO I don't think this should happen only when navigating to a zone should it load the first item.

dpobel:
@miguelcleverti you're perfectly right :) After all it's useful to have automated tests :D
@yannickroger can you have a look ? (I know it's been a long time...)

yannickroger:
This kind of long time is relative when you are used to troubleshooting legacy :)
I will have a look soon.

yannickroger:
Issue fixed in a dedicated commit. ping @dpobel @StephaneDiot

.ez.yml
```diff
@@ -4,7 +4,7 @@ v: 0
 meta:
   repo: "https://github.com/ezsystems/ezplatform.git"
   branch: "master"
-  self_alias: "1.0.x-dev"
+  self_alias: "1.2.x-dev"
```
bdunogier: Against master, it would actually be `1.3.x-dev` (same as `master`).

StephaneDiot:
+1

Resources/public/js/views/ez-navigationhubview.js
```diff
@@ -130,7 +130,10 @@ YUI.add('ez-navigationhubview', function (Y) {
                     activeZone = key;
                 }
             }, this);
-            this.set('activeNavigation', activeZone);
+
+            if (this.get('activeNavigation') !== null) {
```
dpobel: I think this is the wrong place for this condition. Here the `activeNavigation` attribute should always be set. But in the `activeNavigationChange` handler in the NavigationHubViewService, you should check the previous `activeNavigation` value to trigger the first route navigation only if there was a `activeNavigation` before, meaning the user has chosen a zone.

yannickroger:

dpobel:
+1

PR #481, opened 2015-12-15
EZP-25162: display info about empty content tree

mhyndle:
> Jira: https://jira.ez.no/browse/EZP-25162
> status: ready for review

## Description
Currently when there is no content created, when opening the content tree the empty space so user don't know that there is no content. This PR provides short message "The Content Tree is empty" in the place of the tree when the tree is empty.

## Screencast
YT: https://youtu.be/e0dQnY19p0U

## Tests
manual + unit tests

### DIFF
--- a/Resources/public/css/theme/views/tree.css
+++ b/Resources/public/css/theme/views/tree.css
@@ -56,3 +56,18 @@
 .ez-view-treeview .is-tree-node-leaf > .ez-tree-node-toggle {
     visibility: hidden;
 }
+
+.ez-view-treeview .ez-tree-empty-info {
+    margin-top: 40px;
+    font-style: italic;
+    text-align: center;
+    font-size: 120%;
+}
+
+.ez-view-treeview .ez-tree-empty-info:before {
+    display: inline-block;
+    content: "\E617";
+    padding: 0 0.15em;
+    text-align: center;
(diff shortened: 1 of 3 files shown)


mhyndle:
ready for review

yannickroger:
Do we have to use a CSS trick to do that ?
Can't we insert the message in the html content ?

mhyndle:
Well, there are few possible approaches for doing that and I wouldn't call it a trick ;) This way it is pretty easy to be tested and message text is independent from the logic (might be both advantage and disadvantage). What's more this is how are placed loading layers so I thought that it will be best approach.

dpobel:
I'm ok with the approach but the empty message needs some love :) It should basically get a similar style to the loading message (italic, big font, centered vertically and horizontally, ...) but a different icon (I think we already have an "info" icon).
Also the new code is not tested at all.

dpobel:
that said after looking a bit more in detail at the code, instead of adding a class, you should maybe add the paragraph with the text when an empty tree is detected. But this operation should be done in `_renderNodeChildren`.

mhyndle:
feedback taken into account, ready for re-review

dpobel:
+1

yannickroger:
+1

PR #483, opened 2015-12-21
BDD: added wait to navigation zone have time to load 1st item

miguelcleverti:
Fix an issue in the BDD cause by #480.
Clicking navigation zone now loads the first item automatically added a sleep to take this into account.

### DIFF
--- a/Features/Context/SubContext/CommonActions.php
+++ b/Features/Context/SubContext/CommonActions.php
@@ -43,6 +43,10 @@ trait CommonActions
     {
         $this->clickElementByText($zone, '.ez-zone-name');
         $this->waitWhileLoading();
+        // Clicking navigation zone triggers load of first item,
+        // we must wait before interacting with the page (see EZP-25128)
+        // this method sleeps for a default amount (see EzSystems\PlatformUIBundle\Features\Context\PlaformUI)
+        $this->sleep();
     }
 
     /**
--- a/.ez.yml
+++ b/.ez.yml
@@ -4,7 +4,7 @@ v: 0
 meta:
   repo: "https://github.com/ezsystems/ezplatform.git"
   branch: "master"
-  self_alias: "1.0.x-dev"
+  self_alias: "1.2.x-dev"
 
 # ci
 integrate:


joaoinacio:
+1 FWIW (tests pass now)

Features/Context/SubContext/CommonActions.php
```diff
@@ -43,6 +43,9 @@ public function clickNavigationZone($zone)
     {
         $this->clickElementByText($zone, '.ez-zone-name');
         $this->waitWhileLoading();
+        // Clicking navigation zone triggers load of first item,
+        // we must wait before interacting with the page (see EZP-25128)
+        $this->sleep();
```
andrerom: how long sleep is this?

andrerom:
+1 besides readability nitpick

yannickroger:
ping @miguelcleverti can you take care of @andrerom 's feedback and merge ?

bdunogier:
Why not wait for the element we are looking for the be loaded ? It would make more sense than a "sleep" would it ?

miguelcleverti:
@andrerom I updated the doc to provide more information.
@bdunogier we are already waiting for the elements, but it was still causing problems, so the on only way we found was to use the `sleep`. IMO it's not that bad since we do have to wait normally, it takes platformUI a bit to load .

bdunogier:
+1

I'd still prever the tests to pass though :-)

miguelcleverti:
@bdunogier the tests in CI are not running it seems the installation fails due to `platformui-assests-bundle` package not being found

yannickroger:
@andrerom @vidarl Can someone look into this ?

andrerom:
Besides potentially being the composer issues last weeks caused by version bumps, .ez.ci needs to be updated. This is in this repo so responsibility of those bumping versions to maintain.

@vidarl To avoid this self_alias should rather be taken from composer.json, otherwise *(if not set)* assuming you merge PR into target branch on ci before pointing composer to it, and assuming you checkout full repo and not just shallow, let composer figure it out.

@miguelcleverti For now can you try to rebase and apply this change:
```diff
-  self_alias: "1.0.x-dev"
+  self_alias: "1.2.x-dev"
```

miguelcleverti:
@andrerom rebased and changed, but something is still failling

andrerom:
Restarting Travis, it was failing on:
> ✖ [eZ Rest Model tests]: Passed: 17 Failed: 1 Total: 18 (ignored 0) (0.336 seconds)

This seems to have some kind of timing issues or something as it often fails and then passes on next run /cc @dpobel

bdunogier:
Passed. Good to merge ?

yannickroger:
I am waiting for it :smile:

dpobel:
The plan is:

1. to merge https://github.com/ezsystems/ezplatform/pull/95 and https://github.com/ezsystems/PlatformUIBundle/pull/524 so that the BDD tests are executed on Travis
2. to rebase https://github.com/ezsystems/PlatformUIBundle/pull/480 on master so that we can see the failure
3. to cherry-pick the fix here in https://github.com/ezsystems/PlatformUIBundle/pull/480
4. to close this PR and merge https://github.com/ezsystems/PlatformUIBundle/pull/480

dpobel:
Closing as changes have been incorporated in https://github.com/ezsystems/PlatformUIBundle/pull/480

PR #484, opened 2015-12-21
Fix EZP-25296: display language name after changing language when creating content

mhyndle:
> Jira: https://jira.ez.no/browse/EZP-25296
> status: ready for review

## Description
When creating the content with PlatformUI we are able to change the language of content. After changing the language the language indicator in content edit view was containing language code of selected language instead of language name. This PR solves this, so after changing the language the language name is displayed.
What's more this PR fixes some inconsistency with content's version, meaning that now we will have got the same version in the createContentService, contentEditView and it's subviews.

## Tasks
- [x] implementation
- [x] manual tests
- [x] unit tests

### DIFF
--- a/Resources/public/js/views/ez-contenteditformview.js
+++ b/Resources/public/js/views/ez-contenteditformview.js
@@ -99,6 +99,25 @@ YUI.add('ez-contenteditformview', function (Y) {
         },
 
         /**
+         * Sets the `version` and `field` attributes on field edit view instances. The `field`
+         * is taken from given version.
+         *
+         * @method _setVersion
+         * @protected
+         * @param {eZ.Version} version
+         * @since 1.1
+         */
+        _setVersion: function (version) {
+            Y.Array.each(this._fieldEditViews, function (fieldEditView) {
+                var fieldIdentifier = fieldEditView.get('field').fieldDefinitionIdentifier,
+                    field = version.getField(fieldIdentifier);
+
+                fieldEditView.set('version', version);
+                fieldEditView.set('field', field);
+            });
+        },
+
+        /**
          * Renders the form view
          *
          * @method render
@@ -240,7 +259,10 @@ YUI.add('ez-contenteditformview', function (Y) {
              * @required
              */
             version: {
-                writeOnce: "initOnly",
+                value: {},
+                setter: function (val, name) {
+                    this._setVersion(val);
+                }
             },
 
             /**
@@ -250,9 +272,7 @@ YUI.add('ez-contenteditformview', function (Y) {
              * @type {String}
              * @required
              */
-            languageCode: {
-                writeOnce: "initOnly",
-            },
+            languageCode: {},
 
             /**
              * The logged in user
--- a/Resources/public/js/views/ez-contenteditview.js
+++ b/Resources/public/js/views/ez-contenteditview.js
@@ -53,10 +53,14 @@ YUI.add('ez-contenteditview', function (Y) {
                     this._setFocus();
                 }
             });
-            this.on('languageCodeChange', function (e) {
+            this.after('languageCodeChange', function (e) {
+                this.get('formView').set('languageCode', this.get('languageCode'));
                 if ( this.get('active') ) {
-                    this._setLanguageIndicator(e.newVal);
+                    this.render();
                 }
+            });
+            this.after('versionChange', function (e) {
+                this.get('formView').set('version', this.get('version'));
             });
 
             this.on(['*:saveAction', '*:publishAction'], this._handleSavePublish);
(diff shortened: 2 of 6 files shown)


Resources/public/js/views/ez-contenteditformview.js
```diff
@@ -35,6 +35,7 @@ YUI.add('ez-contenteditformview', function (Y) {
 
         initializer: function () {
             this._setFieldEditViews();
+            this.after('versionChange', this._setFieldEditViews);
```
dpobel: in some others similar cases, it's done with a `setter` (for instance in https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/ez-editactionbarview.js#L133). Also by calling `_setFieldEditViews` you are going to recreate the whole field edit views list while you just want to update the version (and the field).

Resources/public/js/views/ez-contenteditview.js
```diff
@@ -53,11 +53,15 @@ YUI.add('ez-contenteditview', function (Y) {
                     this._setFocus();
                 }
             });
-            this.on('languageCodeChange', function (e) {
+            this.after('languageCodeChange', function (e) {
+                this.get('formView').set('languageCode', this.get('languageCode'));
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/484/files#r48230407, use `setter`. you should also update the `actionBar`.

Resources/public/js/views/ez-contenteditview.js
```diff
@@ -53,11 +53,15 @@ YUI.add('ez-contenteditview', function (Y) {
...
                 if ( this.get('active') ) {
-                    this._setLanguageIndicator(e.newVal);
+                    this.render();
                 }
             });
+            this.after('versionChange', function (e) {
+                this.get('formView').set('version', this.get('version'));
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/484/files#r48230523

Resources/public/js/views/ez-contenteditview.js
```diff
@@ -253,23 +257,10 @@ YUI.add('ez-contenteditview', function (Y) {
              * Fired when the change language link was tapped
              *
              * @event changeLanguage
+             * @param fields
```
dpobel: missing some details

Resources/public/js/views/services/ez-contentcreateviewservice.js
```diff
@@ -138,14 +142,23 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
                 }
             });
 
-            this.set('version', new Y.eZ.Version());
-            this.set('languageCode',e.selectedLanguageCode);
-            view.set('languageCode', e.selectedLanguageCode);
+            Y.Object.each(fields, function (field) {
```
dpobel: you are duplicating a part of `_initModels`. Also I've just realized you should update the content as well. So you should probably split somehow split `_initModels` and reuse a part of it here.

Resources/public/js/views/ez-contenteditview.js
mhyndle: As I have discussed it with @dpobel , we are going to leave *Change events as doing this with setter will lead to situation when calling `this.get('formView')` inside of setter method is not possible while at the time doing that, the `this.get('languageCode')` will be called and for that moment it will return `undefined` or not actual value. Rebuilding `formView` must be done AFTER the attr taking part in it will be set.

Resources/public/js/views/ez-contenteditview.js
mhyndle: same as https://github.com/ezsystems/PlatformUIBundle/pull/484/files#r48259688

Resources/public/js/views/services/ez-contentcreateviewservice.js
mhyndle: I have provided method `_setFields` which is setting fields on given content and version and then it sets updated content and version attributes of view service. I haven't refactored it more as fields are bit differrent depending on context - when initializing models fields are populated with default values, on the other hand when changing the language fields are populated with data received in EventFacade.

Resources/public/js/views/ez-contenteditformview.js
```diff
@@ -99,6 +99,24 @@ YUI.add('ez-contenteditformview', function (Y) {
         },
 
         /**
+         * Sets the `version` and `field` attributes on field edit view instances. The `field`
+         * is taken from given version.
+         *
+         * @method _setVersion
+         * @protected
+         * @param {eZ.Version} version
```
dpobel: missing `@since` tag

Resources/public/js/views/services/ez-contentcreateviewservice.js
```diff
@@ -65,12 +65,28 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
...
             this.set('owner', this.get('app').get('user'));
+            callback(this);
+        },
+
+        /**
+         * Sets fields on given content and version. After that `content` and `version` are updated
+         * on the view service.
+         *
+         * @method _setFields
+         * @protected
```
dpobel: missing `@since` tag as well

Resources/public/js/views/ez-contenteditformview.js
mhyndle: `@since 1.1` added

Resources/public/js/views/services/ez-contentcreateviewservice.js
mhyndle: `@since 1.1` added

mhyndle:
ready for review

dpobel:
Replaced by https://github.com/ezsystems/PlatformUIBundle/pull/490

PR #489, opened 2015-12-23
EZP-25266: Removing isLoginAvailable check when editing an already existing user

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-25266

## Description

When editing an already existing user there was a check to see if the login was available. Of course this check was always matching the current user edited and throwing an error "this login is already taken"

## Tests

- Manual tests
- Unit tests


StephaneDiot:
ping @dpobel @yannickroger

dpobel:
If I understand the fix correctly, even when the login is read only, we will still issue a REST API call to check it but the result of this call is just hidden. I think that's not the correct fix or at least it's not an efficient one.

StephaneDiot:
updated @dpobel

dpobel:
+1

yannickroger:
+1

PR #492, opened 2016-01-15
EZP-25169: Optimize the content loading in the location search plugin

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-25169

## Description

Instead of loadind each content, one after one, they are now loaded all at the same time and then dispatched in the locationStruct.

## Tests

- Unit tested
- Manually tested

### DIFF
--- a/Resources/public/js/views/services/plugins/ez-locationsearchplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-locationsearchplugin.js
@@ -78,6 +78,7 @@ YUI.add('ez-locationsearchplugin', function (Y) {
 
                 if (e.loadContentType || e.loadContent) {
                     this._loadResources(
+                        e.viewName,
                         e.loadContentType,
                         e.loadContent,
                         attrs[e.resultAttribute],
@@ -134,68 +135,6 @@ YUI.add('ez-locationsearchplugin', function (Y) {
             return location;
         },
 
-
-
-
-        /**
-         * Loads the Content for each tree node representing the children of
-         * `levelLocation`.
-         *
-         * @method _loadContents
-         * @param {eZ.Location} levelLocation
-         * @param {Object} data
-         * @param {Function} callback
-         * @protected
-         */
-        _loadContents: function (levelLocation, data, callback) {
-            var contentService = this.get('host').get('capi').getContentService(),
-                contents = {},
-                query;
-
-            query = contentService.newViewCreateStruct('children_content' + levelLocation.get('locationId'), 'ContentQuery');
-            query.body.ViewInput.ContentQuery.Criteria = {
-                "ParentLocationIdCriterion": levelLocation.get('locationId'),
-            };
-
-            contentService.createView(query, function (err, response) {
-                if ( err ) {
-                    callback(err);
-                    return;
-                }
-                Y.Array.each(response.document.View.Result.searchHits.searchHit, function (hit) {
-                    var content = new Y.eZ.Content({id: hit.value.Content._href});
-
-                    content.loadFromHash(hit.value.Content);
-                    contents[content.get('id')] = content;
-                });
-
-                Y.Object.each(data, function (struct, key) {
-                    data[key].content = contents[struct.contentInfo.get('id')];
-                });
-                callback();
-            });
-        },
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
         /**
          * Loads resources for the given array of location structs. Depending on given
          * `loadContentType` and `loadContent` bool parameters it loads Content and ContentType
@@ -208,7 +147,7 @@ YUI.add('ez-locationsearchplugin', function (Y) {
          * @param {Array} locationStructArr
          * @param {Function} callback
          */
-        _loadResources: function (loadContentType, loadContent, locationStructArr, callback) {
+        _loadResources: function (viewName, loadContentType, loadContent, locationStructArr, callback) {
             var tasks = new Y.Parallel(),
                 loadResourcesError = false;
 
@@ -231,7 +170,7 @@ YUI.add('ez-locationsearchplugin', function (Y) {
                         }
                     });
 
-                this._loadContentForLocations(locationStructArr, endContentLoad);
+                this._loadContentForLocations(viewName, locationStructArr, endContentLoad);
             }
 
             tasks.done(function () {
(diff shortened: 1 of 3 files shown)


StephaneDiot:
ready for first review @dpobel @yannickroger

Resources/public/js/views/ez-subitemlistview.js
```diff
@@ -222,6 +222,7 @@ YUI.add('ez-subitemlistview', function (Y) {
             this.fire('locationSearch', {
                 viewName: 'subitemlist-' + locationId,
                 resultAttribute: 'subitems',
+                loadContent: false,
```
dpobel: it's the default value, so you should enforce it here

Resources/public/js/views/services/plugins/ez-locationsearchplugin.js
```diff
@@ -222,31 +223,41 @@ YUI.add('ez-locationsearchplugin', function (Y) {
...
-                        if (error) {
-                            loadContentError = true;
-                            return;
-                        }
-
-                        locationStructArr[i].content = content;
-                    });
-
-                content.load(options, endContentLoad);
+                contentIdCriterions =
```
dpobel: you should have a look at http://yuilibrary.com/yui/docs/api/classes/Array.html#method_reduce which would be bit more elegant

Resources/public/js/views/services/plugins/ez-locationsearchplugin.js
```diff
@@ -222,31 +223,41 @@ YUI.add('ez-locationsearchplugin', function (Y) {
          * @param {Array|Null} locationStructArr
          * @param {Function} callback
          */
-        _loadContentForLocations: function (locationStructArr, callback) {
-            var tasks = new Y.Parallel(),
-                loadContentError = false;
+        _loadContentForLocations: function (viewName, locationStructArr, callback) {
+            var contentService = this.get('host').get('capi').getContentService(),
+                contentIdCriterions = '',
```
dpobel: that's a bad name, I would just call it `contentIds`

Resources/public/js/views/services/plugins/ez-locationsearchplugin.js
```diff
@@ -222,31 +223,41 @@ YUI.add('ez-locationsearchplugin', function (Y) {
...
-            tasks.done(function () {
-                callback(loadContentError, locationStructArr);
+            query = contentService.newViewCreateStruct('contents-loading-'+ viewName, 'ContentQuery');
+            query.body.ViewInput.ContentQuery.Criteria = {
+                "ContentIdCriterion": contentIdCriterions
+            };
+
+            contentService.createView(query, function (err, response) {
+                if (err) {
+                    loadContentError = true;
```
dpobel: this boolean is not needed. In case of error, you should just call `callback(err);`

Resources/public/js/views/services/plugins/ez-locationsearchplugin.js
```diff
@@ -222,31 +223,41 @@ YUI.add('ez-locationsearchplugin', function (Y) {
...
+                Y.Array.each(response.document.View.Result.searchHits.searchHit, function (hit, i) {
+                    var Content = that.get('contentModelConstructor'),
+                        content;
+
+                    content = new Content();
+                    content.loadFromHash(hit.value.Content);
+                    locationStructArr[i].content = content;
+                });
             });
+            callback(loadContentError, locationStructArr);
```
dpobel: this call should be in the `createView` callback, otherwise you will execute the callback before the Content items are actually loaded and added to `locationStructArr`.

Tests/js/views/services/plugins/assets/ez-locationsearchplugin-tests.js
```diff
@@ -231,14 +252,35 @@ YUI.add('ez-locationsearchplugin-tests', function (Y) {
...
                 method: 'getContentService',
                 returns: this.contentService,
             });
             Mock.expect(this.contentService, {
                 method: 'newViewCreateStruct',
-                args: [this.viewName, 'LocationQuery'],
-                returns: this.query,
+                args: [Mock.Value.String, Mock.Value.String],
+                run: function (viewName, queryType) {
+                    if (viewName == that.viewName && queryType == "LocationQuery") {
```
dpobel: I would just use the `queryType` to return the location query or the content query. The view name are more implementation details here.

Tests/js/views/services/plugins/assets/ez-locationsearchplugin-tests.js
```diff
@@ -253,9 +295,13 @@ YUI.add('ez-locationsearchplugin-tests', function (Y) {
             });
             Mock.expect(this.contentService, {
                 method: 'createView',
-                args: [this.query, Mock.Value.Function],
+                args: [Mock.Value.Object, Mock.Value.Function],
                 run: function (query, cb) {
-                    cb(false, that.response);
+                    if (query == that.locationQuery) {
```
dpobel: since you are comparing two objects, you should use `===`. Also because of that, none the condition should be evaluated to true so I'm a bit suspicious on that code and the test.

Tests/js/views/services/plugins/assets/ez-locationsearchplugin-tests.js
```diff
@@ -274,12 +321,13 @@ YUI.add('ez-locationsearchplugin-tests', function (Y) {
...
+            delete this.locationQuery;
         },
 
         "Should load Content and ContentType into location struct": function () {
             var contentInfo = this.contentInfo,
-                response = this.response,
+                response = this.locationResponse,
                 resultAttr = 'whateverAttr';
 
             Y.eZ.Content = Y.Model;
```
dpobel: since you are now injecting a custom Content constructor, I guess this line can/should be remove.

Tests/js/views/services/plugins/assets/ez-locationsearchplugin-tests.js
```diff
@@ -335,7 +388,20 @@ YUI.add('ez-locationsearchplugin-tests', function (Y) {
...
+                run: function (query, cb) {
+                    if (query == that.locationQuery) {
+                        cb(false, that.locationResponse);
+                    } else if (query == that.contentQuery) {
+                        cb(true, that.contentResponse);
+                    }
+                }
+            });
 
             Y.eZ.Content = function () {};
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/492/files#r49862676 for those 2 lines.

StephaneDiot:
updated @dpobel

dpobel:
Closed in favor of https://github.com/ezsystems/PlatformUIBundle/issues/494

PR #493, opened 2016-01-15
EZP-25135: As an editor, I want to be able to choose the image variation of images in the RichText editor

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25135

# Description

This patch brings a drop down list in the style toolbar for image elements in the RichText editor. This list allows to choose the image alias to use for a given image.

Screencast: https://www.youtube.com/watch?v=G0rDBCYoh2c

In this PR, I also rename the *location search plugin* to *search plugin* since this plugin is now also able to search for Content items. To avoid a BC break, the location search plugin is still there, but if someone uses it, a deprecation warning will be written to the console (see https://github.com/ezsystems/PlatformUIBundle/pull/493/files#diff-c969c75ae71109346a69fbefbc10232fR16).

## Tasks

* [x] Improve the Embed Widget API
* [x] Pass the image variations config to the editor
* [x] Adds the image variations drop down list to the image style toolbar
* [x] Update the corresponding widget with the chosen variations

# Tests

manual + unit tests


dpobel:
ping @yannickroger @andrerom @bdunogier

Resources/config/yui.yml
```diff
@@ -57,11 +57,14 @@ system:
                     requires: ['ez-alloyeditor']
                     path: %ez_platformui.public_dir%/js/alloyeditor/buttons/paragraph.js
                 ez-alloyeditor-button-image:
-                    requires: ['ez-alloyeditor', 'ez-alloyeditor-button-mixin-embeddiscovercontent', 'ez-alloyeditor-button-mixin-embedimage']
+                    requires: ['ez-alloyeditor', 'ez-alloyeditor-button-mixin-embeddiscovercontent', 'ez-alloyeditor-button-mixin-embedimage', 'ez-alloyeditor-button-mixin-widgetbutton']
```
yannickroger: nitpick: you could change the format of the list so it doesn't reach the line limit (same below)

Resources/public/js/alloyeditor/buttons/imagevariation.jsx
```diff
@@ -0,0 +1,96 @@
...
+         *
+         * @method _updateImage
+         * @param {Object} e
+         */
+        _updateImage: function (e) {
+            var widget = this._getWidget(),
+                newVariation = e.target.value,
+                contentId = widget.getHref().replace('ezcontent://', '');
+
+            this.props.editor.get('nativeEditor').fire('contentSearch', {
```
yannickroger: Don't we add a bit of JS doc when we fire an event?

Resources/public/js/views/services/plugins/ez-locationsearchplugin.js
```diff
@@ -5,280 +5,18 @@
...
-             * @type {Function}
-             * @default Y.eZ.Content
-             */
-            contentModelConstructor: {
-                value: Y.eZ.Content
-            }
-        },
-    });
+    console.warn('[DEPRECATED] ez-locationsearchplugin module and Y.eZ.Plugin.LocationSearch are deprecated');
+    console.warn('[DEPRECATED] ez-locationsearchplugin module and Y.eZ.Plugin.LocationSearch will be removed in PlatformUI 2.0');
```
yannickroger: Should this also be added to some kind of upgrade doc ?

yannickroger:
+1 with feedback/nitpick/questions taken into account

Resources/public/js/views/services/plugins/ez-locationsearchplugin.js
andrerom: Would probably be more then sufficient to add info in one of these on what to use instead.

andrerom:
+1

Resources/public/js/views/services/plugins/ez-locationsearchplugin.js
dpobel: it's already the case in line 18 + in the doc of the module in the beginning of the file. But I agree with @yannickroger , when we'll effectively remove the module, we'll have to document that.

PR #496, opened 2016-01-19
[composer] bumped branch alias to 1.1.x-dev

bdunogier:


andrerom:
+1, but we should have a 1.0 branch\* for composer before we merge this, otherwise we breaks `~1.0@dev`, and same changes would be needed on the changed requirements.

\* On this repo this implies need to fork also

bdunogier:
I can't really configure satis and update all the branches just for that. Let's skip this for now.

PR #497, opened 2016-01-20
EZP-23932: Show appropriate version information

andrerom:
~~Also move over version info from the other data source to not duplicate this, means first tab is now called Bundles. @bdunogier only alternative I see regarding the interface is simply deprecate it and rely on the class instead as long as we know we will have to change this over time.*~~~

https://jira.ez.no/browse/EZP-23932

Updated: Simple approach here for 16.02, while @glye works on a more proper approach for 16.04 in separate bundle.

*Bundles tab has been renamed from hard coded "eZ Platform" and if composer info is available will give version of platform ui and kernel from it:*
screen shot 2016-02-23 at 17 11 07

*New tab listing all packages, most important info is added, including link if defined:*
screen shot 2016-02-23 at 17 11 14

### DIFF
--- a/Controller/SystemInfoController.php
+++ b/Controller/SystemInfoController.php
@@ -19,9 +19,15 @@ class SystemInfoController extends Controller
      */
     protected $systemInfoHelper;
 
-    public function __construct(SystemInfoHelperInterface $systemInfoHelper)
+    /**
+     * @var string
+     */
+    private $installDir;
+
+    public function __construct(SystemInfoHelperInterface $systemInfoHelper, $installDir)
     {
         $this->systemInfoHelper = $systemInfoHelper;
+        $this->installDir = $installDir;
     }
 
     public function performAccessChecks()
@@ -40,6 +46,7 @@ class SystemInfoController extends Controller
         return $this->render('eZPlatformUIBundle:SystemInfo:info.html.twig', [
             'ezplatformInfo' => $this->systemInfoHelper->getEzPlatformInfo(),
             'systemInfo' => $this->systemInfoHelper->getSystemInfo(),
+            'composerInfo' => $this->getComposerInfo(),
         ]);
     }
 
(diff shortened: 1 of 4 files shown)


bdunogier:
> @bdunogier only alternative I see regarding the interface is simply deprecate it and rely on the class instead as long as we know we will have to change this over time.

Well, "only" alternative with this architecture.

What about something _a bit_ more ambitions:
- deprecate `SystemInfoHelperInterface` and the implementation
- add an `InfoProvider` interface, with one method (`getInformation()`)
- Implement one provider per info type (`SystemInfoProvider`, `BundlesInfoProvider`, `PackagesInfoProvider`)
- Add a simple service tag + registry for those, and use it in the controller

It's a bit more work, but we are open for additions with less risks of BC breaks.

andrerom:
Sounds plausable.
But be aware I kind of intend this to be extracted into SupportBundle, so ns will change. Could you check with @dpobel if we have everything we need on extensibility to extract it already now?

bdunogier:
>  Could you check with UI team if we have everything we need on menu extensibility to extract this already now?

I will.

bdunogier:
Re-ping @dpobel

dpobel:
> we need on menu extensibility to extract this

not sure to follow that, which menu are you talking about ? are you talking about the tabs ?

andrerom:
> not sure to follow that, which menu are you talking about ? are you talking about the tabs ?

sub menu of admin panel *(so to be able to ass _System info_ from bundle)*

dpobel:
ah yes it's extensible, see https://doc.ez.no/display/TECHDOC/Configure+the+navigation or http://dpobel.github.io/slides-ez/extending-platformui-ezconference-2015/#/17 :p

andrerom:
@dpobel updated as discussed.

review ping @yannickroger @glye

glye:
+1, agree it doesn't need to be more fancy now since it will all change in the new bundle(s) anyhow.

Controller/SystemInfoController.php
```diff
@@ -56,4 +63,32 @@ public function phpinfoAction()
...
+
+        $packages = [];
+        $lockData = json_decode(file_get_contents($this->installDir . 'composer.lock'), true);
+        foreach ($lockData['packages'] as $packageData) {
+            $packages[$packageData['name']] = [
+                'version' => $packageData['version'],
+                'time' => $packageData['time'],
+                'homepage' => isset($packageData['homepage']) ? $packageData['homepage'] : '',
+                'reference' => $packageData['source']['reference'],
+
```
yannickroger: nitpick: empty line

Controller/SystemInfoController.php
```diff
@@ -56,4 +63,32 @@ public function phpinfoAction()
...
+    /**
+     * Temporary function for getting composer package info into templates.
+     *
+     * @return array
+     */
+    private function getComposerInfo()
+    {
+        if (!file_exists($this->installDir . 'composer.lock')) {
+            return array();
```
yannickroger: nitpick: can use short array notation ?

yannickroger:
Could you also add the issue number in the commit message.

+1 after nitpick taken into account.

andrerom:
> Could you also add the issue number in the commit message.
> .. nitpicks taken into account.

fixed and fixed

Controller/SystemInfoController.php
```diff
@@ -56,4 +63,31 @@ public function phpinfoAction()
 
         return $response;
     }
+
+    /**
+     * Temporary function for getting composer package info into templates.
```
dpobel: maybe should precise a bit what you mean by *Temporary*

dpobel:
besides the inline comment +1

Controller/SystemInfoController.php
andrerom: "fixed", as method is private I don't really need to mention this, we can change it at any point.

PR #501, opened 2016-01-25
Fix EZP-25375: Timestamps 0 are not handled properly

yannickroger:
Link: https://jira.ez.no/browse/EZP-25375

## Description
This PR fixes storing and displaying Time and DateAndTime field type with an empty time stamp.

### Storing
When adding a Time with the value `00:00` it generates a timestamp with the value 0.

The code what checking if the value was set with a:
```javascript
// Please don't be 0 or I'll do something stupid :)
if (timestamp) {
    //update
} else {
    return null;
}
```
So `null` was returned for a valid timestamp (`0`)

With this patch I found another issue in the kernel when updating a Time with 00:00 https://github.com/ezsystems/ezpublish-kernel/pull/1565

### Displaying
In order to decide if a field is empty or not I added an overriden `isFieldEmpty` in the views and created an equivalent method on the editviews (made the code more readable to me having a dedicated method).

## Tests
Manual and regression unit tests

### DIFF
--- a/Resources/public/js/views/fields/ez-dateandtime-view.js
+++ b/Resources/public/js/views/fields/ez-dateandtime-view.js
@@ -74,16 +74,23 @@ YUI.add('ez-dateandtime-view', function (Y) {
          * Returns a Date object from the field value
          *
          * @method _getDateObject
-         * @return {Date}
+         * @return {Date|undefined}
          */
         _getDateObject: function () {
             var field = this.get('field');
 
-            if ( field && field.fieldValue ) {
-                return new Date(field.fieldValue.timestamp * 1000);
+            if (this._isFieldEmpty()) {
+                return undefined;
             }
-            return undefined;
-        }
+
+            return new Date(field.fieldValue.timestamp * 1000);
+        },
+
+        _isFieldEmpty: function () {
+            var field = this.get('field');
+
+            return !field || !field.fieldValue || field.fieldValue.timestamp === null || isNaN(field.fieldValue.timestamp);
+        },
     });
 
     Y.eZ.FieldView.registerFieldView('ezdatetime', Y.eZ.DateAndTimeView);
--- a/Resources/public/js/views/fields/ez-time-editview.js
+++ b/Resources/public/js/views/fields/ez-time-editview.js
@@ -116,7 +116,7 @@ YUI.add('ez-time-editview', function (Y) {
                 field = this.get('field'),
                 time = '';
 
-            if ( field && field.fieldValue ) {
+            if (!this._isFieldEmpty()) {
                 if (!this.get('supportsTimeInput') && !def.fieldSettings.useSeconds) {
                     time = Y.Date.format(new Date(this._getUtcTimeStamp(field.fieldValue * 1000)), {format:"%R"});
                 } else {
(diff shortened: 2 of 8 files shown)


andrerom:
+1

dpobel:
+1

Tests/js/views/fields/assets/ez-dateandtime-editview-tests.js
```diff
@@ -913,6 +913,35 @@ YUI.add('ez-dateandtime-editview-tests', function (Y) {
...
+            expectedTimeValue: '00:00',
+
+            _setNewValue: function () {
+                var c = this.view.get('container');
+                c.one('.ez-dateandtime-date-input-ui input').set('value', this.expectedDateValue);
+                c.one('.ez-dateandtime-time-input-ui input').set('value', this.expectedTimeValue);
+            },
+
+            _assertCorrectFieldValue: function (fieldValue, msg) {
+                var expected = 526521600, // '1986-09-08' '10:10' in UTC
```
yannickroger: Typo is comment

yannickroger:
@dpobel @andrerom @StephaneDiot I changed the scope of this PR to also include displaying 0 timestamps. It needs another round of reviews.

Resources/public/js/views/fields/ez-dateandtime-view.js
```diff
@@ -79,11 +79,22 @@ YUI.add('ez-dateandtime-view', function (Y) {
         _getDateObject: function () {
             var field = this.get('field');
 
-            if ( field && field.fieldValue ) {
+            if (this._isFieldEmpty()) {
+                return undefined;
+            } else {
```
dpobel: nitpick: no need for else
nitpick2: while at it you can update the doc of the method to indicate this method can return `undefined`

Resources/public/js/views/fields/ez-dateandtime-view.js
```diff
@@ -79,11 +79,22 @@ YUI.add('ez-dateandtime-view', function (Y) {
...
                 return new Date(field.fieldValue.timestamp * 1000);
             }
-            return undefined;
-        }
+        },
+
+        _isFieldEmpty: function () {
+            var field = this.get('field');
+
+            if (field && field.fieldValue && field.fieldValue.timestamp !== null ) {
```
dpobel: can't you simplify that like in https://github.com/ezsystems/PlatformUIBundle/pull/501/files#diff-87487438348f321146e7acd2eefba122R311 ?

Resources/public/js/views/fields/ez-time-editview.js
```diff
@@ -168,10 +168,12 @@ YUI.add('ez-time-editview', function (Y) {
             var valueOfInput;
 
             valueOfInput = this._getInputNode().get('valueAsNumber');
-            if (valueOfInput) {
+
+            if (isNaN(valueOfInput)) {
+                return null;
+            } else {
```
dpobel: nitpick: no need for else

Resources/public/js/views/fields/ez-time-view.js
```diff
@@ -89,12 +89,21 @@ YUI.add('ez-time-view', function (Y) {
          * @return {Date}
          */
         _getDateObject: function () {
+            if (this._isFieldEmpty()) {
+                return undefined;
+            } else {
```
dpobel: nitpick: no need for else as well :)

Resources/public/js/views/fields/ez-time-view.js
```diff
@@ -89,12 +89,21 @@ YUI.add('ez-time-view', function (Y) {
...
+                return new Date(this.get('field').fieldValue * 1000);
+            }
+        },
+
+        _isFieldEmpty: function () {
             var field = this.get('field');
 
-            if ( field && field.fieldValue ) {
-                return new Date(field.fieldValue * 1000);
+            if (field && field.fieldValue !== null) {
```
dpobel: same question as https://github.com/ezsystems/PlatformUIBundle/pull/501/files#r52433136

dpobel:
besides the inline comments, still +1

StephaneDiot:
+1

andrerom:
this should be 1.1 scope right?

PR #499, opened 2016-01-22
Upgrade to AlloyEditor 0.7

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25235

# Description

Well, it's just the upgrade to AlloyEditor 0.7. Besides some changes in the CSS, this only requires fixes in tests to avoid [React deprecation warnings](https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#two-packages-react-and-react-dom) and [CKEditor warnings](http://docs.ckeditor.com/#!/guide/dev_errors).

## Tasks

* [x] Update bower.json
* [x] Update PlatformUIAssetsBundle (see https://github.com/ezsystems/PlatformUIAssetsBundle/pull/11)
* [x] Regenerate the skin
* [x] Fixed the deprecation warning
* [x] Inspect/Fixed warnings like `[CKEDITOR] Error code: editor-incorrect-destroy`
* [x] Apply same changes in tests introduced in https://github.com/ezsystems/PlatformUIBundle/pull/493

PR #507, opened 2016-02-02
Make sure to not use deprecated React methods

dpobel:
Deprecated calls forgotten in https://github.com/ezsystems/PlatformUIBundle/pull/499


dpobel:
Merging as Travis CI agrees

PR #508, opened 2016-02-04
EZP-25009: Role assignment with section limitation

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-25009

## Description

The goal is to be able to assign a role to a user or user group with a section limitation.

:warning: the redirection is not implemented after the role is assigned. Seems it will wait for https://jira.ez.no/browse/EZP-25442

## Tests

- Manual tests
- Unit tests


StephaneDiot:
ready for early review @dpobel

PR #511, opened 2016-02-12
Switched to TravisCI container based infrastructure

dpobel:
Come on TravisCI, tell me what you think :)


dpobel:

dpobel:
ping @andrerom as well :)

yannickroger:
+1

.travis.yml
```diff
@@ -1,15 +1,16 @@
-language: node_js
-node_js:
-    - "0.10"
+language: php
+php:
+    - "5.6"
```
joaoinacio: Q: any reason for php 5.6 when platform specifies 5.5|5.7 as requirement?

joaoinacio: Q: any reason for php 5.6 when platform specifies 5.5|5.7 as requirement?

dpobel: (5.7 ? I guess you mean 7 :))

Well, actually before this PR, we've been testing only with PHP 5.5 (provided by `ppa:ondrej/php5`). Given that we have 2 code bases to test, to not run the JS tests 3 times, we either have to choose a PHP version or I have to change the JS test be executed only when a specific PHP version is configured but I find that a bit weird to be honest.

joaoinacio: > (5.7 ? I guess you mean 7 :))

Right :)

Thanks for the info, just curious if it was a typo or not, as it looked weird to me as well.

andrerom: `~5.5|~7.0` means 5.x starting 5.5.0 and 7.x starting 7.0.x, so it should mean 5.5, 5.6 and 7.0 at the moment.
Make sure you have an up-to-date composer, and if this is not working as intended, propose composer.json fixes that solves this. *(Aka 5.6 is supposed to be supported)*

dpobel:
ping @joaoinacio @andrerom @bdunogier

andrerom:
+1, followup on updating node.js ?

dpobel:
@andrerom Follow up: BDD tests on Travis CI, see https://github.com/ezsystems/PlatformUIBundle/pull/524

PR #512, opened 2016-02-15
EZP-25249: Restore items from trash

yannickroger:
Link: https://jira.ez.no/browse/EZP-25249

## Description
Allow to restore selected items from trash.

Screencast: https://drive.google.com/file/d/0B5amepf6EkXbeG1ObktCQkxtSlk/view?usp=sharing

trash

## Tests
Manual and unit test

### DIFF
--- a/Resources/public/js/views/ez-trashview.js
+++ b/Resources/public/js/views/ez-trashview.js
@@ -21,9 +21,16 @@ YUI.add('ez-trashview', function (Y) {
      * @constructor
      * @extends eZ.TemplateBasedView
      */
-    Y.eZ.TrashView = Y.Base.create('trashView', Y.eZ.TemplateBasedView, [], {
+    Y.eZ.TrashView = Y.Base.create('trashView', Y.eZ.TemplateBasedView, [Y.eZ.SelectionTable], {
+        events: {
+            '.ez-trashitem-box': {
+                'change': '_updateTrashBarButtons'
+            },
+        },
+
         initializer: function () {
             this.on('*:minimizeTrashBarAction', this._handleMinimizeTrashBar);
+            this.on('*:restoreTrashItemsAction', this._restoreTrashItems);
         },
 
         /**
@@ -86,6 +93,57 @@ YUI.add('ez-trashview', function (Y) {
                     'contentType': trashItem.contentType.toJSON(),
                 };
             });
+        },
+
+        /**
+         * Restores the selected trash items
+         *
+         * @private
+         * @method _restoreTrashItems
+         */
+        _restoreTrashItems: function () {
+            var selectedTrashItems = this.get('container').all('.ez-trashitem-box:checked'),
+                trashItems = [];
+
+            selectedTrashItems.each(function (selectedTrashItem) {
+                Y.Array.some(this.get('trashItems'), function (trashItem) {
+                    if (selectedTrashItem.getAttribute('value') === trashItem.item.get('id')) {
+                        trashItems.push(trashItem.item);
+                        return true;
+                    }
+                });
+            }, this);
+
+            /**
+             * Fired to restore the selected items
+             * @event restoreItems
+             */
+            this.fire('restoreItems', {trashItems: trashItems});
+        },
+
+        /**
+         * Updates the TrashBar buttons
+         *
+         * @private
+         * @method _updateTrashBarButtons
+         */
+        _updateTrashBarButtons: function () {
+            this._updateDisableTrashBarButton(
+                'restoreTrashItems',
+                this.get('container').one('.ez-trashitem-box:checked') === null
+            );
+        },
+
+        /**
+         * Update the disable status on a given button of the trashBar
+         *
+         * @private
+         * @method _updateDisableTrashBarButton
+         * @param {String} actionId of the button
+         * @param {Boolean} isDisabled
+         */
+        _updateDisableTrashBarButton: function (actionId, isDisabled) {
+            this.get('trashBar').getAction(actionId).set('disabled', isDisabled);
         },
 
         destructor: function () {
--- a/Resources/public/js/views/services/ez-trashviewservice.js
+++ b/Resources/public/js/views/services/ez-trashviewservice.js
@@ -25,6 +25,7 @@ YUI.add('ez-trashviewservice', function (Y) {
 
         initializer: function () {
             this.on('*:emptyTrashAction', this._emptyTrashConfirmBox);
+            this.on('*:restoreItems', this._restoreItems);
         },
 
         /**
@@ -115,9 +116,8 @@ YUI.add('ez-trashviewservice', function (Y) {
 
             if (!service._isParentInTrash(itemHash.ParentLocation._href, inTrashList)) {
                 parentLocation.load(loadOptions, tasks.add(function (error) {
-                    if (error) {
-                        service._error("Failed to load parent location with REST API");
-                    } else {
+                    // Do nothing if there is an error. The parent might have been deleted/restored
+                    if (!error) {
                         parentLocation.loadPath(loadOptions, tasks.add(function (error) {
                             if (error) {
                                 service._error("Failed to load parent location path with REST API");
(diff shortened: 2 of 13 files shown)


Resources/public/js/views/ez-trashview.js
```diff
@@ -88,6 +89,27 @@ YUI.add('ez-trashview', function (Y) {
...
         },
 
+        /**
+         * Restores the selected trash items
+         *
+         * @private
+         * @method _restoreTrashItems
+         */
+        _restoreTrashItems: function (e) {
+            var selectedTrashItems = Y.all('.ez-trashitem-box:checked'),
```
dpobel: don't use `Y.all`, a view is only responsible what happens in its own container so here you should write `this.get('container')..all('.ez-trashitem-box:checked');

Resources/public/js/views/services/ez-trashviewservice.js
```diff
@@ -223,6 +223,55 @@ YUI.add('ez-trashviewservice', function (Y) {
...
+         * @return {Boolean}
+         */
+        _restoreItems: function (e) {
+            var contentService = this.get('capi').getContentService(),
+                tasks = new Y.Parallel(),
+                service = this,
+                notificationNamePrefix = "restoreTrashItem-";
+
+            Y.each(e.trashItemsId, tasks.add(function (trashItemId) {
+                contentService.recover(trashItemId, function (error) {
```
dpobel: it would be cleaner if the trash item model have a `restore` method. That also means the `restoreItems` should provide the trash items models, not only the ids.

Resources/public/js/views/ez-trashview.js
```diff
@@ -21,9 +21,17 @@ YUI.add('ez-trashview', function (Y) {
...
+        events: {
+            '.ez-trashitem-box': {
+                'change': '_itemSelectionChange'
+            },
+        },
+
         initializer: function () {
             this.on('*:minimizeTrashBarAction', this._handleMinimizeTrashBar);
+            this.on('*:restoreTrashItemsAction', this._restoreTrashItems);
+            this.after('activeChange', this._itemSelectionChange);
```
dpobel: weird :) to avoid that, the restore button action view should be disabled by default.

Resources/public/js/views/ez-trashview.js
```diff
@@ -79,13 +87,69 @@ YUI.add('ez-trashview', function (Y) {
...
+            this.fire('restoreItems', {trashItems: trashItems});
+        },
+
+        /**
+         * Called when the selection of trashItems is modified
+         *
+         * @private
+         * @method _itemSelectionChange
+         */
+        _itemSelectionChange: function () {
```
dpobel: bad method name

Resources/public/js/views/ez-trashview.js
```diff
@@ -79,13 +87,69 @@ YUI.add('ez-trashview', function (Y) {
...
+         * Restores the selected trash items
+         *
+         * @private
+         * @method _restoreTrashItems
+         */
+        _restoreTrashItems: function (e) {
+            var selectedTrashItems = this.get('container').all('.ez-trashitem-box:checked'),
+                trashItems = [];
+
+            Y.each(selectedTrashItems, function (trashItem) {
```
dpobel: `selectedTrashItems` is a NodeList, to iterate on it you have [a `each` method](http://yuilibrary.com/yui/docs/api/classes/NodeList.html#method_each).

Resources/public/js/views/ez-trashview.js
```diff
@@ -79,13 +87,69 @@ YUI.add('ez-trashview', function (Y) {
...
+        /**
+         * Called when the selection of trashItems is modified
+         *
+         * @private
+         * @method _itemSelectionChange
+         */
+        _itemSelectionChange: function () {
+            this._updateDisableTrashBarButton(
+                'restoreTrashItems',
+                Y.one('.ez-trashitem-box:checked') === null
```
dpobel: `this.get('container').one('.ez-trashitem-box:checked');`

Resources/public/js/views/ez-trashview.js
```diff
@@ -79,13 +87,69 @@ YUI.add('ez-trashview', function (Y) {
...
+        /**
+         * Update the disable status on a given button of the trashBar
+         *
+         * @private
+         * @method _updateDisableTrashBarButton
+         * @param {String} actionId of the button
+         * @param {Boolean} isDisabled
+         */
+        _updateDisableTrashBarButton: function (actionId, isDisabled) {
+            Y.Array.some(this.get('trashBar').get('actionsList'), function (button) {
```
dpobel: you can be simplified with:
```js
this.get('trashBar').getAction(actionId).set('disabled', isDisabled);
```

Resources/public/js/views/services/ez-trashviewservice.js
```diff
@@ -223,6 +225,54 @@ YUI.add('ez-trashviewservice', function (Y) {
...
+         * @param {Object} e restoreItems event facade
+         * @param {Array} e.trashItems  List of trashItems to be restored
+         */
+        _restoreItems: function (e) {
+            var loadOptions = {api: this.get('capi')},
+                tasks = new Y.Parallel(),
+                service = this,
+                notificationNamePrefix = "restoreTrashItem-";
+
+            Y.each(e.trashItems, tasks.add(function (trashItem) {
```
dpobel: `tasks.add` should not be called here but on the `restore` callback

yannickroger:
Ready for review @dpobel @StephaneDiot

Tests/js/models/assets/ez-trashitemmodel-tests.js
```diff
@@ -204,8 +205,48 @@ YUI.add('ez-trashitemmodel-tests', function (Y) {
...
+        },
+
+        tearDown: function () {
+            this.model.destroy();
+            delete this.model;
+        },
+
+        "Should call the api on restore": function () {
+            this.model.set('id', this.id);
+            this.model.restore(this.options, function () {});
```
dpobel: this test is incomplete. You should check that the model restore callback parameter is passed to the content service recover method (ie the contentService mock should be mocked here)

Tests/js/views/services/assets/ez-trashviewservice-tests.js
```diff
@@ -942,9 +957,141 @@ YUI.add('ez-trashviewservice-tests', function (Y) {
...
+        },
+
+        tearDown: function () {
+            this.service.destroy();
+            delete this.service;
+        },
+
+        _createTrashItem: function (id, error) {
+            return {
+                restore: function (loadOptions, callback) {
```
dpobel: a Mock would be more suited here.
Also a test where you assert that `location#restore` is called is missing.

dpobel:
can you please record a screencast to see the full app behaviour for that feature ?

Resources/public/js/views/ez-trashview.js
```diff
@@ -88,6 +95,57 @@ YUI.add('ez-trashview', function (Y) {
             });
         },
 
+        /**
+         * Restores the selected trash items
+         *
+         * @private
```
StephaneDiot: missing event facade doc

StephaneDiot: infact, do you need the event facade?

Resources/public/js/views/services/ez-trashviewservice.js
```diff
@@ -223,6 +223,54 @@ YUI.add('ez-trashviewservice', function (Y) {
...
+         * @protected
+         * @param {Object} e restoreItems event facade
+         * @param {Array} e.trashItems  List of trashItems to be restored
+         */
+        _restoreItems: function (e) {
+            var loadOptions = {api: this.get('capi')},
+                tasks = new Y.Parallel(),
+                service = this,
+                notificationNamePrefix = "restoreTrashItem-";
+
```
StephaneDiot: that's maybe a bit of niptick but i think you could have even fired a started notification here in addition of the ones that are in your restore() callback

StephaneDiot:
+1

Resources/public/js/views/services/ez-trashviewservice.js
dpobel: that's far from being a nitpick, this raises an interesting question on the notifications because one notification per restored content might be a bit too much. Actually, there should be one notification before starting to restore any content which is updated in the `done` callback. And in addition to that, a notification in case of error but with a different id.

yannickroger: I pinged PM about it in the Jira issue.
We still need to have a single notification if only one item is selected as we will later display the name and a link to the content. We could have a dedicated notification behavior if multiple items are selected.

Resources/public/js/views/services/ez-trashviewservice.js
```diff
@@ -223,6 +223,60 @@ YUI.add('ez-trashviewservice', function (Y) {
...
+         * @method _restoreItems
+         * @protected
+         * @param {Object} e restoreItems event facade
+         * @param {Array} e.trashItems  List of trashItems to be restored
+         */
+        _restoreItems: function (e) {
+            var loadOptions = {api: this.get('capi')},
+                tasks = new Y.Parallel(),
+                service = this,
+                notificationName = "restoreTrashItems";
```
dpobel: should be named `notificationIdentifier` (name != identifier)

Resources/public/js/views/services/ez-trashviewservice.js
```diff
@@ -223,6 +223,60 @@ YUI.add('ez-trashviewservice', function (Y) {
...
+
+            Y.each(e.trashItems, function (trashItem) {
+                trashItem.restore(loadOptions, tasks.add(function (error) {
+                    var trashItemId = trashItem.get('id');
+
+                    if (error) {
+                        service._notify(
+                            "An error occurred while restoring trash item with Id: " + trashItemId
+                            + ". For the moment item's whose original location has been deleted can not be restored",
+                            notificationName,
```
dpobel: You should use a different identifier for this case

Resources/public/js/views/services/ez-trashviewservice.js
```diff
@@ -223,6 +223,60 @@ YUI.add('ez-trashviewservice', function (Y) {
...
+                            0
+                        );
+                    }
+                }));
+            });
+
+            tasks.done(function () {
+
+                service._notify(
+                    'Trash item(s) have been restored',
```
dpobel: I would include the number of restored items (Should be called "Content items" as well) and this notification should be thrown only if at least one item was successfully restored.

Resources/public/js/views/services/ez-trashviewservice.js
```diff
@@ -223,6 +223,66 @@ YUI.add('ez-trashviewservice', function (Y) {
...
+
+            Y.each(e.trashItems, function (trashItem) {
+                trashItem.restore(loadOptions, tasks.add(function (error) {
+                    var trashItemId = trashItem.get('id');
+
+                    if (error) {
+                        service._notify(
+                            "An error occurred while restoring trash item with Id: " + trashItemId
+                            + ". For the moment item's whose original location has been deleted can not be restored",
+                            notificationIdentifier + "-error",
```
dpobel: you should also add the trash item id in the identifier otherwise there will be only one notification even if several items were not restored.

yannickroger:
@dpobel feedback taken into account

dpobel:
+1 when Travis agrees

PR #513, opened 2016-02-17
EZP-25446: Add the user profile in the navigation hub

dew326:
Jira: https://jira.ez.no/browse/EZP-25446

## Description
The user profile in navigation hub should contain the user name, role, image and the user menu. Menu should provide ability to add new items manually or using view plugins. 

## TODO
- [x] JS Unit Tests


dew326:
ping @dpobel @sunpietro @lserwatka

Resources/public/js/views/ez-usermenuitemview.js
```diff
@@ -0,0 +1,56 @@
...
+             */
+            title: {},
+
+            /**
+             * The suffix for CSS class of the view
+             *
+             * @attribute CSSclassSuffix
+             * @type {String}
+             */
+            CSSclassSuffix: {}
```
sunpietro: `cssClassSuffix`

Resources/public/js/views/services/ez-navigationhubviewservice.js
```diff
@@ -53,6 +54,84 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
...
+            .then(function(rolePath) {
+                userService.loadRole(rolePath, Y.bind(function (error, response) {
+                    try {
+                        userInfo.role = response.document.Role.identifier;
+                        event.target.set('userInfo', userInfo);
+                    } catch (e) {
+                        this._handleError({
+                            message: 'Cannot get user data',
+                            identifier:  'fetch-user-info-' + Date.now()
+                        });
```
sunpietro: `}, e);`

dpobel:
small remark before reviewing, for this kind of feature a screencast (or some screenshots) helps a lot the review.

Resources/public/js/views/ez-navigationhubview.js
```diff
@@ -244,15 +241,29 @@ YUI.add('ez-navigationhubview', function (Y) {
...
         /**
+         * Renders user profile to the view
+         *
+         * @method _renderUserProfile
+         * @protected
+         */
+        _renderUserProfile: function () {
+            var profileView = this.get('userProfileView');
+
+            profileView.addTarget(this);
```
dpobel: this should be done as soon as soon the user profile view is instantiated. Basically, the `userProfileView` attribute should be defined with something like:

```js
            userProfileView: {
                valueFn: function () {
                    return new Y.eZ.UserProfileView({
                        bubbleTargets: this,
                    });
                }
            },
```

Resources/public/js/views/ez-navigationhubview.js
```diff
@@ -244,15 +241,29 @@ YUI.add('ez-navigationhubview', function (Y) {
...
+            this.get('container').append(profileView.render().get('container'));
+            profileView.set('active', true);
```
dpobel: that's not needed normally thanks to https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/ez-view.js#L53

(and kind of wrong, `active` means *attached to the DOM* which is not the case yet when this code is executed)

Resources/public/js/views/services/ez-navigationhubviewservice.js
```diff
@@ -53,6 +54,84 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
...
+         *
+         * @method _fetchUserInfo
+         * @param event {Object} event facade
+         */
+        _fetchUserInfo: function (event) {
+            var user = this.get('app').get('user'),
+                userService = this.get('capi').getUserService(),
+                userInfo = {},
+                promise = new Y.Promise(function (resolve, reject) {
+                    userService.loadUser(user.get('id'), Y.bind(function (error, response) {
```
dpobel: this part is useless, the currently logged in user is already loaded and available in the `user` attribute of the application. (`this.get('app').get('user');` from here).

dpobel: well actually, I've realized you already use it on this line, so why loading it again ?

Resources/public/js/views/ez-userprofileview.js
```diff
@@ -0,0 +1,130 @@
...
+        *
+        * @method render
+        * @return {eZ.UserProfileView} the view itself
+        */
+        render: function () {
+            this.get('container')
+                .setHTML(this.template())
+                .addClass(CLASS_USER_PROFILE_HIDDEN);
+
+            this.fire('getUserInfo');
```
dpobel: it's weird to fire an event like that and in `render`. Normally the data you need to render the view should be provided by default without the need to ask anything. So basically, the navigation hub view service load method should more or less do what is currently done when handling this event.

Resources/public/js/views/ez-userprofileview.js
```diff
@@ -0,0 +1,130 @@
...
+
+        /**
+        * Renders the user menu view
+        *
+        * @method _renderUserMenu
+        */
+        _renderUserMenu: function () {
+            var view = this.get('userMenuView');
+
+            view.addTarget(this);
```
dpobel: same as above on the view being added a bubble target and on the `active` flag

Resources/public/js/views/services/ez-navigationhubviewservice.js
```diff
@@ -53,6 +54,84 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
...
+                        }
+                    }, this));
+                });
+
+            promise.then(function(userId) {
+                return new Y.Promise(function (resolve, reject) {
+                    userService.getRoleAssignmentsForUser(userId, Y.bind(function (error, response) {
+                        try {
+                            if (response.document.RoleAssignmentList && response.document.RoleAssignmentList.RoleAssignment[0]) {
+                                resolve(response.document.RoleAssignmentList.RoleAssignment[0].Role._href);
```
dpobel: actually getting the right role is a lot more complicated than that and even not really possible, because:
1. it's common to have several roles, and it's impossible to guess which one we should choose to display its name so that makes sense
1. the role is usually assigned to a group not to a user directly so in addition, so the same logic should be applied to groups
1. a user can be in several groups...

so IMHO, to decrease the complexity and avoid bugs we can't fix, we should skip that part and not display the user role(s) for now. If we really want that feature and we want it reliable, then we'll have to make changes server side.

@rolandbenedetti @inakijv @jince any opinion on that ?

Resources/public/js/views/ez-userprofileview.js
```diff
@@ -0,0 +1,130 @@
...
+         */
+        _setUserInfo: function () {
+            var user = this.get('userInfo'),
+                container = this.get('container'),
+                fields = user.userData.Version.Fields.field,
+                imageUrl,
+                imageAltText;
+
+                fields.forEach(function (field) {
+                    if (field.fieldDefinitionIdentifier === 'image' && field.fieldValue) {
```
dpobel: there are several issues in this method:

1. it's a bad practice to base some logic on the `fieldDefinitionIdentifier` because a user is a Content item with a Content Type that might be different from what you expect. To do that in a reliable way, you should get the field definition identifier of an `ezimage` field def. from the Content Type with [`getFieldDefinitionIdentifiers`](http://ezsystems.github.io/platformui-javascript-api/classes/eZ.ContentType.html#method_getFieldDefinitionIdentifiers) (of course it's also possible for the user content type to not have any `ezimage` field.)
1. to properly get the field and its value, you should use the Content model
1. to display the user's avatar, you use the original image variation which can be quite big while it's displayed in a very small way. Also the img is styled to be a square (why in em ?), while the image might not have such ratio. So instead, you use a dedicated image variation to have a square generated from the image

so for these reasons, the code should be organized in a different way to have less logic here and more spread over the models and the view services. So basically, here is the changes you should do:

1. change the user model to add a `loadAvatar` method. Internally, this method will first load the corresponding Content Type, check if it has an `ezimage` field then load the corresponding Content item and retrieve the `ezimage` field:
  ```js
  user.loadAvatar({api: capiInstance}, function (err, field) {
      // field can be undefined if the user Content Type does not have any ezimage field def.
      // otherwise it's the field value of the first ezimage field.
  });
  ```
2. in the navigation hub view service, from the field, you should [load a specific image variation](http://ezsystems.github.io/javascript-rest-client/classes/ContentService.html#method_loadImageVariation) defined in image.yml which is then pass to the view in `getViewParameters`

Resources/public/js/views/ez-userprofileview.js
```diff
@@ -0,0 +1,130 @@
...
+                        imageUrl = field.fieldValue.uri;
+                        imageAltText = field.fieldValue.fileName;
```
dpobel: the original filename is most likely a bad alternative text, this should rather be something like *User's avatar image*

Resources/public/js/views/ez-userprofileview.js
dpobel: and as we were saying yesterday in the demo, when the user does not have an image, you should display a generic avatar

crevillo: Apart for @dpobel, i'd probably make this a bit more configurable. Or at least i won't hardcode ```image``` there. Someone may want to remove add another one for any reason. Some others may just delete this attribute...

Resources/public/js/views/ez-userprofileview.js
```diff
@@ -0,0 +1,130 @@
...
+        },
+
+        /**
+         * Toggle the user menu. 
+         *
+         * @method _toggleUserMenu
+         * @protected
+         */
+        _toggleUserMenu: function () {
+            this.get('userMenuView').fire('toggleUserMenu');
```
dpobel: in this case, since you have direct access to the instance of the object on which you want to act, it would be better to directly call a public method instead of using an event. The events are usually used to communicate with others components which are *far away* and higher in the hierarchy of components (view to parent views to view services to app).

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -0,0 +1,225 @@
...
+        },
+
+        /**
+         * Toggles the visibility of a user menu.
+         *
+         * @method _toggleUserMenu
+         * @protected
+         */
+        _toggleUserMenu: function () {
+            var methodName = this.get('displayed') ? 'addClass' : 'removeClass',
```
dpobel: this should be separated in 2 steps:

1. `toggleUserMenu` should set the `displayed` attribute value based on the existing one
1. a `displayedChange` event handler should reflect the state in the markup. Also usually, it's easier to set the class on the view container, this avoids to have various selectors in the code, it's easier to unit test and this view the opportunity the style anything in the view (not only the visibility of a specific part).

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -0,0 +1,225 @@
...
+         * Renders the user menu view
+         *
+         * @method render
+         * @return {eZ.UserMenuView} the view itself
+         */
+        render: function () {
+            var container = this.get('container');
+
+            container.setHTML(this.template(this.getAttrs()));
+            container.on('clickoutside', this._hideMenuOnClickOutside, this);
```
dpobel: you should store the subscription object (the return value of `on`) to make sure the subscription is detached when the view is destroyed.

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -0,0 +1,225 @@
...
+        },
+
+        /**
+         * Checks if clicked outside the user profile container and hides a user menu if it is visible.
+         *
+         * @method _hideMenuOnClickOutside
+         * @protected
+         */
+        _hideMenuOnClickOutside: function (event) {
+            if (!event.target.hasClass(CLASS_USER_PROFILE) && event.target.ancestor(SELECTOR_USER_PROFILE) === null && this.get('displayed')) {
```
dpobel: testing `displayed` attribute should be enough.

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -0,0 +1,225 @@
...
+        /**
+         * Renders menu items views
+         *
+         * @method _renderMenuItems
+         * @protected
+         */
+        _renderMenuItems: function () {
+            var menuItemsContainer = this.get('container').one(SELECTOR_USER_MENU),
+                logOutView = this.get('logOutItemView'),
+                additionalItems = this.get('additionalItems');
```
dpobel: I think `eZ.UserMenuView` should be simplified, IMHO it's a bit complicated partly because you do pretty much the same thing as for the navigation items while the constraints are a different. Also the way you handle the logout is not correct because the `eZ.UserMenuView` and the view for the logout link are too much coupled, `eZ.UserMenuView` should have no knowledge of what is generated by its sub-views.

So here is how this could be done:

1. I would put all the items view in one attribute (`items`) as the logout item is not really different from the others.
1. Sorting by an attribute before rendering is a good idea but we tend to use `priority` over `order` (ie big values first) and the base view class `eZ.UserMenuItemView` should have such attribute with a default value.
1. to ease the whole process, I would make sure the `eZ.UserMenuView` container is an `<ul>`, that way you can directly add the sub-views to the container.
1. the trick where you describe the sub-views with a plain object with the `Constructor` property is not necessary here, so you should remove that option, the `items` attribute will handle only view instances.
1. as explained above, the `cssClassSuffix` system tends to couple a bit the components together. For the items in the user menu, one use case is to fire an event when the user clicks on the items like for logout. So I think you should implement a `eZ.UserMenuItemFireEventView` (feel free to propose a better name :)) which extends `eZ.UserMenuItemView` and expects in addition to the title an event name that is triggered when the user clicks on it.

Basically, you'll end up having the `items` attribute defined like:

```js
items: {
    readOnly: true,
    valueFn: function () {
        return [
            new Y.eZ.UserMenuItemFireEventView({
                bubbleTargets: this,
                title: "Logout",
                eventName: 'logout',
            }),
        ];
    },
}
```
for this to work, you'll have to make the `active` flag is correctly updated on the items. That means you have to implement an `activeChange` event handler so that the items receive the correct value when the active flag changes.

Resources/config/yui.yml
```diff
@@ -1112,3 +1112,34 @@ system:
                 ez-app-extension-admin:
                     requires: ['ez-serversideview', 'ez-serversideviewservice', 'ez-app-extension']
                     path: %ez_platformui.public_dir%/js/apps/extensions/ez-app-extension-admin.js
+                ez-userprofileview:
```
dpobel: `yui.yml` is kind of sorted. Given this is about the navigation hub and you only add new views, those module definitions should be close the navigation hub view definition in this file.

Resources/public/css/views/userprofile.css
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+.ez-view-userprofileview {
+    position: absolute;
+    top: 5px;
+    right: 0;
+}
+
+.is-user-profile-hidden {
+    display: none !important;
```
dpobel: please never use `!important`

Resources/public/css/views/userprofile.css
```diff
@@ -0,0 +1,58 @@
...
+}
+
+.ez-view-userprofileview .ez-user-profile-image {
+    width: 2em;
+    height: 2em;
+    margin-left: 0.5em;
+    -webkit-user-select: none;
+    -moz-user-select: none;
+    -ms-user-select: none;
+    user-select: none;
```
dpobel: for readability, please align the `:`

Resources/public/css/theme/views/userprofile.css
```diff
@@ -0,0 +1,68 @@
...
+}
+
+.ez-view-userprofileview .ez-user-profile {
+    cursor: pointer;
+    line-height: 2em;
+    text-align: right;
+}
+
+.ez-view-userprofileview .ez-user-profile:after {
+    content: "";
```
dpobel: since we already have an icon for the arrow and that it's already used in different places, I would use it there as well for consistency.

dpobel:
I'm kind of done on the view for now. I'm not super fan of the current style of the menu but we can take that later. And of course, I'm looking forward the unit tests but most of my comments will make then easier to write.

Resources/public/js/views/services/ez-navigationhubviewservice.js
jince: I'm good with your recommendation, Damien. It's a nice-to-have feature, but not essential.

rolandbenedetti: Fine by me too.

Resources/public/js/views/ez-userprofileview.js
dpobel: well, actually I've just realized the user response already contains the current version of the Content item of the user and after thinking a bit about it, I think the change to the User model to get the avatar should be done differently. Basically, the user model should get a new `avatar` attribute than we'll give access to the field value of the image field (if there's one). To do that, here are a needed changes:

1. add an `avatar` attribute to the user model
1. when loading the user, we should load the corresponding the content type directly, this will result in something like (not tested):
   ```js
sync: function (action, options, callback) {
    var api = options.api;

    if ( action === 'read' ) {
        api.getUserService().loadUser(this.get('id'), function (err, response) {
            var contentType;
            if ( err ) {
                return callback(err, response);
            }
            contentType = new Y.eZ.ContentType({id: response.document.User.ContentType._href});
            contentType.load({api: api}, function (error) {
                response.document._contentType = contentType;
                callback(error, response);
            });
        });

    ```
1. then in a custom `_parseStruct` method, from the Content type you can get the image field definition identifier and then get the corresponding field value in the version available in the REST User response, something like:
    ```js
        _parseStruct: function (struct, responseDoc) {
            var attrs, imageField,
                imageIdentifiers = responseDoc._contentType.getFieldDefinitionIdentifiers('ezimage');

            attrs = this.constructor.superclass._parseStruct.call(this, struct);
            imageField = struct.Version.Fields.field.filter(function (field) {
                return field.fieldDefinitionIdentifier == imageIdentifiers[0];
            });
            attrs.avatar = imageField.length ? imageField[0].fieldValue : null;
            return attrs;
        },
     ```

that way, once a user is loaded, the avatar field value is directly available in the `avatar` attribute.

dew326:
ping @dpobel 
- comments applied
- JS Unit Tests added

Resources/config/yui.yml
```diff
@@ -434,6 +434,41 @@ system:
...
+                        - 'userprofileview-ez-template'
+                    path: %ez_platformui.public_dir%/js/views/ez-userprofileview.js
+                userprofileview-ez-template:
+                    type: 'template'
+                    path: %ez_platformui.public_dir%/templates/userprofile.hbt
+                ez-usermenuview:
+                    requires:
+                        - 'event-tap'
+                        - 'event-outside'
+                        - 'template-micro'
```
dpobel: not needed I think

Resources/config/yui.yml
```diff
@@ -434,6 +434,41 @@ system:
...
+                        - 'ez-usermenuitemview'
```
dpobel: this dependency as well is not needed

Resources/public/js/models/ez-usermodel.js
```diff
@@ -36,17 +36,46 @@ YUI.add('ez-usermodel', function (Y) {
...
+                        callback(error, response);
+                    });
+                });
             } else {
                 callback("Only read operation is supported at the moment");
             }
+        },
+
+        /**
+         * Set the avatar to the user item.
```
dpobel: not exactly, more something like:
```
Override the default implementation to find the avatar (if any) in the user Content item.
```

also the `@protected` tag is missing.

Resources/public/js/models/ez-usermodel.js
```diff
@@ -36,17 +36,46 @@ YUI.add('ez-usermodel', function (Y) {
...
+                return field.fieldDefinitionIdentifier == imageIdentifiers[0];
+            });
+            attrs.avatar = imageField.length ? imageField[0].fieldValue : null;
+            return attrs;
         }
     }, {
         REST_STRUCT_ROOT: "User",
         ATTRS_REST_MAP: [
-            'email', 'login', 'enabled', 'name',
+            'email', 'login', 'enabled', 'name', 'avatar',
```
dpobel: you should not add `avatar` there. the `ATTRS_REST_MAP` list the attributes that can filled by simply reading the REST response. `avatar` is not in that category since you build its value in your custom `_parseStruct` method.

Resources/public/js/models/ez-usermodel.js
```diff
@@ -104,6 +133,17 @@ YUI.add('ez-usermodel', function (Y) {
              */
             remoteId: {
                 value: ''
+            },
+
+            /**
+             * The user's avatar
+             *
+             * @attribute avatar
+             * @default {}
```
dpobel: I think you should set it to `null` (both here and in `value` below) to be consistent when the default value in `_parseStruct` and the *empty value* of the Image field. This will also ease the check to detect if there's an avatar or not.

Resources/public/js/views/ez-navigationhubview.js
```diff
@@ -244,15 +241,28 @@ YUI.add('ez-navigationhubview', function (Y) {
...
         /**
+         * Renders user profile to the view
+         *
+         * @method _renderUserProfile
+         * @protected
+         */
+        _renderUserProfile: function () {
+            var profileView = this.get('userProfileView');
+
+            profileView.set('user', this.get('user'));
```
dpobel: this should be in a `userChange` event handler.

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -0,0 +1,170 @@
...
+    /**
+     * The User Menu view
+     *
+     * @namespace eZ
+     * @class UserMenuView
+     * @constructor
+     * @extends eZ.TemplateBasedView
+     */
+    Y.eZ.UserMenuView = Y.Base.create('userMenuView', Y.View, [], {
+        _clickOutsideUserMenuHandler: null,
```
dpobel: you should not declare the property here because it ends up in the prototype and then it is shared by all the instances (ok there's normally only one instance of `UserMenuView` but still).

So you should initialize the `_clickOutsideUserMenuHandler` property in the `initializer` method. And it should be documented (like http://ezsystems.github.io/platformui-javascript-api/files/Resources_public_js_apps_ez-platformuiapp.js.html#l131 for instance).

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * The User Menu view
+     *
+     * @namespace eZ
+     * @class UserMenuView
+     * @constructor
+     * @extends eZ.TemplateBasedView
+     */
+    Y.eZ.UserMenuView = Y.Base.create('userMenuView', Y.View, [], {
```
dpobel: it should extend `Y.eZ.View` to get the `active` attribute and the ability to write plugin for it.

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -0,0 +1,170 @@
...
+        },
+
+        /**
+         * Renders the user menu view
+         *
+         * @method render
+         * @return {eZ.UserMenuView} the view itself
+         */
+        render: function () {
+            this.get('container').addClass(CLASS_HIDDEN).addClass(CLASS_USER_MENU);
```
dpobel: I don't really see the point of the class `ez-user-menu` (`CLASS_USER_MENU`) since we already have `ez-view-usermenuview` on the container.

Also you should not add `is-user-menu-hidden` (`CLASS_HIDDEN`) here because the `displayed` attribute might be `true` and also because we should leave this responsibility to the `displayedChange` handler. So to clean up this, you should:

1. add this class to the `containerTemplate` in `initializer`
1. set the `displayed` attribute as read only so that the `toggleDisplayed` method is the only way to change it (that also means you have to use `_set` in that method).

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -0,0 +1,170 @@
...
+        },
+
+        /**
+         * Toggles the visibility of a user menu.
+         *
+         * @method _toggleUserMenu
+         * @protected
+         * @param event {Object} event facade
+         */
+        _toggleUserMenu: function (event) {
```
dpobel: to help readability, this kind of method is usually prefixed with `_ui` (so `_uiToggleUserMenu`) to reflect the fact it acts on the UI.

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -0,0 +1,170 @@
...
+                items = this.get('items');
+
+            items.sort(function (a, b) {
+                return b.get('priority') - a.get('priority');
+            });
+
+            items.forEach(Y.bind(function (view) {
+                view.addTarget(this);
+                menuItemsContainer.append(view.render().get('container'));
+                view.set('active', true);
```
dpobel: as I wrote in a previous comment, you should not set the `active` flag while rendering. Instead, the UserMenuView should have a `activeChange` handler that forwards the `active` value to the items.

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -0,0 +1,170 @@
...
+        _renderMenuItems: function () {
+            var menuItemsContainer = this.get('container'),
+                items = this.get('items');
+
+            items.sort(function (a, b) {
+                return b.get('priority') - a.get('priority');
+            });
+
+            items.forEach(Y.bind(function (view) {
+                view.addTarget(this);
```
dpobel: it's not really the place to do that. Instead, I would introduce a `addMenuItem` method that is responsible for adding a menu item to the UserMenuView and while doing that would also call `addTarget`.

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -0,0 +1,170 @@
...
+            },
+
+            /**
+             * Additional list of menu items
+             *
+             * @attribute additionalItems
+             * @type Array
+             * @default []
+             */
+            additionalItems: {
```
dpobel: this can be removed now.

Resources/public/js/views/ez-usermenuitemview.js
```diff
@@ -0,0 +1,52 @@
...
+        ATTRS: {
+            /**
+             * Stores the item title
+             *
+             * @attribute title
+             * @type {String}
+             */
+            title: {},
+
+            priority: {
```
dpobel: missing attribute doc

Resources/public/js/views/ez-userprofileview.js
```diff
@@ -0,0 +1,134 @@
...
+            },
+
+            /**
+             * Contains the default users avatar image
+             *
+             * @attribute defaultAvatarImage
+             * @type {String}
+             */
+            defaultAvatarImage: {
+                value:  'data:image/png;base64, iVBORw0KGgoAAAANSUhEUgAAACgAAAAoCAYAAACM/rhtAAAAGXRFWHRTb2Z0' +
```
dpobel: good idea :)

Resources/public/js/views/ez-usermenuitemview.js
```diff
@@ -0,0 +1,52 @@
...
+        },
+
+        /**
+        * Renders the item view
+        *
+        * @method render
+        * @return {eZ.UserMenuItemView} the view itself
+        */
+        render: function () {
+            this.get('container').setHTML(this.template(this.getAttrs()));
```
dpobel: I would rather only pass the `title` to the template

Resources/public/js/views/ez-usermenuitemfireeventview.js
```diff
@@ -0,0 +1,76 @@
...
+        },
+
+        /**
+        * Renders the item view
+        *
+        * @method render
+        * @return {eZ.UserMenuItemFireEventView} the view itself
+        */
+        render: function () {
+            this.get('container').setHTML(this.template(this.getAttrs()));
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/513/files#r53772287

Resources/public/js/views/ez-usermenuitemfireeventview.js
```diff
@@ -0,0 +1,76 @@
...
+        },
+
+        /**
+         * Tap event handler on the logout link
+         *
+         * @method _logOut
+         * @protected
+         * @param event {Object} tap event facade
+         */
+        _logOut: function (event) {
```
dpobel: it should not be called `_logOut` as it's much more generic now. `_fireEvent` maybe ? Also the js doc should be updated.

Resources/public/js/views/ez-userprofileview.js
```diff
@@ -0,0 +1,134 @@
...
+        *
+        * @method render
+        * @return {eZ.UserProfileView} the view itself
+        */
+        render: function () {
+            this.get('container')
+                .setHTML(this.template({
+                    user: this.get('user').toJSON(),
+                    avatarAltText: this.get('avatarAltText'),
+                    defaultAvatarImage: this.get('defaultAvatarImage')
```
dpobel: instead of passing `defaultAvatarImage`, it would be better to pass an `avatar` variable and to move the condition about the user's avatar here (so that the template becomes dumb)

Resources/public/templates/usermenuitemfireevent.hbt
```diff
@@ -0,0 +1 @@
+<a class="ez-user-menu-item">{{ title }}</a>
```
dpobel: I would add a `href="#"` to reflect the fact it's a *real* link.

Resources/public/js/views/ez-userprofileview.js
dpobel: actually, that's a bit more complicated than that. I've just realized you are still displaying the original image variation (which can be big while the image is super small).
So, you should a dedicated image variation definition in `image_variations.yml` and load this variation from the user's `avatar` attribute (if filled) in the navigation hub view service. This variation should then be passed to the navigation hub view which pass it to the user profile view.

Tests/js/views/ez-navigationhubview.html
```diff
@@ -83,9 +103,25 @@
         filter: loaderFilter,
         modules: {
             "ez-navigationhubview": {
-                requires: ['ez-templatebasedview', 'ez-navigationitemview', 'event-tap', 'node-screen', 'node-style', 'event-outside', 'ez-height-change'],
+                requires: ['ez-templatebasedview', 'ez-navigationitemview', 'event-tap', 'node-screen', 'node-style', 'event-outside', 'ez-height-change', 'ez-userprofileview'],
                 fullpath: "../../../Resources/public/js/views/ez-navigationhubview.js"
             },
+            "ez-userprofileview": {
```
dpobel: you should try to limit the dependencies as much as possible.
Actually, this is one of the reason why the sub-views are stored in attributes with a `valueFn` (like `userProfileView` in the case of the navigation hub view). Because of this, in the test to remove the need of `ez-userprofileview`, you can write something like:

```js
setUp: function () {
    this.userProfileView = new Y.View();
    this.view = new Y.eZ.NavigationHubView({
        userProfileView: this.userProfileView,
        // rest of the needed config...
    });
}
```

it's also possible to use a mock instead but in the case of a View that's usually a bit painful. Also, depending on how the sub-view is used, it's also common to dynamically create a fake implementation of the sub-view, this gives something like:

```js
setUp: function () {
    var UserProfileView = Y.Base.create('userProfileView', Y.View, [], {
            /* some needed methods */
        }, {/* even some ATTRS */});
    this.userProfileView = new UserProfileView();
    this.view = new Y.eZ.NavigationHubView({
        userProfileView: this.userProfileView,
        // rest of the needed config...
    });
}
```

Tests/js/views/ez-usermenuitemfireeventview.html
```diff
@@ -0,0 +1,63 @@
...
+</head>
+<body>
+<li class="container ez-view-usermenuitemfireeventview"></li>
+
+
+<script type="text/x-handlebars-template" id="usermenuitemfireeventview-ez-template">
+    <a class="ez-user-menu-item">{{ title }}</a>
+</script>
+
+<script type="text/x-handlebars-template" id="usermenuitemview-ez-template">
```
dpobel: not really needed I think

Tests/js/views/assets/ez-usermenuitemfireeventview-tests.js
```diff
@@ -0,0 +1,66 @@
...
+                container: '.container'
+            });
+        },
+
+        tearDown: function () {
+            this.view.destroy();
+        },
+
+        "Should render the user menu item": function () {
+            var container = this.view.render().get('container');
```
dpobel: usually we test that the template was used and received the correct variables (example https://github.com/ezsystems/PlatformUIBundle/blob/master/Tests/js/views/assets/ez-confirmboxview-tests.js#L26)

Tests/js/views/assets/ez-usermenuitemview-tests.js
```diff
@@ -0,0 +1,31 @@
...
+                container: '.container'
+            });
+        },
+
+        tearDown: function () {
+            this.view.destroy();
+        },
+
+        "Should render the user menu item": function () {
+            var container = this.view.render().get('container');
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/513/files#r53778249

Tests/js/views/ez-usermenuitemview.html
```diff
@@ -0,0 +1,59 @@
...
+
+    YUI({
+        coverage: ['ez-usermenuitemview'],
+        filter: loaderFilter,
+        modules: {
+            "ez-usermenuitemview": {
+                requires: ['ez-templatebasedview'],
+                fullpath: "../../../Resources/public/js/views/ez-usermenuitemview.js"
+            },
+            "ez-usermenuitemfireeventview": {
```
dpobel: useless (and not loaded as far as I can tell)

Tests/js/views/ez-usermenuview.html
```diff
@@ -0,0 +1,66 @@
...
+        modules: {
+            "ez-usermenuview": {
+                requires: ['event-tap', 'event-outside', 'template-micro', 'ez-usermenuitemview', 'ez-usermenuitemfireeventview'],
+                fullpath: "../../../Resources/public/js/views/ez-usermenuview.js"
+            },
+            "ez-usermenuitemview": {
+                requires: ['ez-templatebasedview'],
+                fullpath: "../../../Resources/public/js/views/ez-usermenuitemview.js"
+            },
+            "ez-usermenuitemfireeventview": {
```
dpobel: same remark as in https://github.com/ezsystems/PlatformUIBundle/pull/513/files#r53777165
This one should not be loaded and the logout item should be replaced in the test as it's not the component being tested.

for `ez-usermenuitemview`, it depends how you'll test ez-usermenuview. If you inject some mocks as its `items` then it should be removed as well but if you choose to inject *real* item views, then it is needed and should be used.

Tests/js/views/ez-userprofileview.html
```diff
@@ -0,0 +1,80 @@
...
+
+    YUI({
+        coverage: ['ez-userprofileview'],
+        filter: loaderFilter,
+        modules: {
+            "ez-userprofileview": {
+                requires: ['ez-templatebasedview', 'event-tap', 'ez-usermenuview'],
+                fullpath: "../../../Resources/public/js/views/ez-userprofileview.js"
+            },
+            "ez-usermenuview": {
```
dpobel: same remark as in https://github.com/ezsystems/PlatformUIBundle/pull/513/files#r53777165, you don't need `ez-usermenuview` to test the UserProfileView as long as you inject a view (or a mock) with a `toggleDisplayed` method.

dew326:
ping @dpobel 
done

Resources/config/image_variations.yml
```diff
@@ -9,3 +9,7 @@ system:
                 reference: reference
                 filters:
                     - {name: geometry/scalewidthdownonly, params: [300]}
+            platformui_profileview:
+                reference: reference
+                filters:
+                    - {name: geometry/scalewidthdownonly, params: [40]}
```
dpobel: you are displaying the avatar as a square, so the variation should generate a square. I think we should get that with something like (not tested):
```yml
- {name: geometry/scaledownonly, params: [40, 40]}
- {name: geometry/crop, params: [40, 40, 0, 0]}
```

Resources/public/css/views/userprofile.css
```diff
@@ -0,0 +1,54 @@
...
+.ez-view-userprofileview .ez-user-profile .ez-user-profile-info {
+    display: inline-block;
+}
+
+.ez-view-userprofileview .ez-user-profile .ez-user-profile-name {
+    margin: 0;
+}
+
+.ez-view-userprofileview .ez-user-profile-image {
+    width: 2em;
```
dpobel: since it's a bitmap image, I think its size should be set in pixels and to the chosen value in image_variations.yml.

Resources/public/js/views/ez-usermenuitemfireeventview.js
```diff
@@ -0,0 +1,76 @@
...
+        }
+    }, {
+        ATTRS: {
+            /**
+             * Stores the item title
+             *
+             * @attribute title
+             * @type {String}
+             */
+            title: {},
```
dpobel: nitpick: you don't have to redefine this attribute since it's already defined in the parent class.

Resources/public/js/views/ez-usermenuview.js
dpobel: the condition seems really too complicated but maybe I'm mistaken ?

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -0,0 +1,187 @@
...
+         * Sets active flag to the item view
+         *
+         * @method _setActiveViews
+         * @protected
+         */
+        _setActiveViews: function () {
+            var items = this.get('items');
+
+            items.forEach(function (item) {
+                item.set('active', true);
```
dpobel: not true but the `active` value being set.

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -0,0 +1,187 @@
...
+         * Adds item to the menu
+         *
+         * @method _addMenuItem
+         * @protected
+         * @param view {Object}
+         */
+        _addMenuItem: function (view) {
+            var menuItemsContainer = this.get('container');
+
+            view.addTarget(this);
```
dpobel: it seems like I was not clear enough in https://github.com/ezsystems/PlatformUIBundle/pull/513#discussion_r53769841 :(

Calling `addTarget` at render time is not correct because you might expect some events from child view before the view is rendered and this is absolutely not related to the render logic.

So my initial comment was about separating `addTarget` from the rendering logic. Basically, `addTarget` should be called when a view is added an item of the `UserMenuView`. To do that, in my initial comment, I was suggesting that you add a public method called `addMenuItem` that would take a view into parameter and add it to the `items` attribute and after doing that would call `addTarget`.
Then when rendering `UserMenuView` you would just iterate over the items, render them and add the container to the UserMenuView's container.

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -0,0 +1,187 @@
...
+     * The User Menu view
+     *
+     * @namespace eZ
+     * @class UserMenuView
+     * @constructor
+     * @extends Y.eZ.View
+     */
+    Y.eZ.UserMenuView = Y.Base.create('userMenuView', Y.eZ.View, [], {
+        initializer: function () {
+            this.containerTemplate = '<ul class="ez-view-usermenuview is-user-menu-hidden"/>';
```
dpobel: nitpick: you can use CLASS_HIDDEN defined a few lines above

Resources/public/js/views/ez-userprofileview.js
```diff
@@ -0,0 +1,136 @@
...
+        _toggleUserMenu: function () {
+            this.get('userMenuView').toggleDisplayed();
+        }
+    }, {
+        ATTRS: {
+            /**
+             * Contains data about current user
+             *
+             * @attribute user
+             * @type {Object}
```
dpobel: nitpick: `{eZ.User}`

Resources/public/js/views/ez-usermenuview.js
dew326: I need this condition because when there is only `this.get('displayed')` and you click on the user profile to show user menu it's set the displayed to true and then set displayed to false by the `clickoutside` event.

Tests/js/models/assets/ez-usermodel-tests.js
```diff
@@ -28,9 +30,21 @@ YUI.add('ez-usermodel-tests', function (Y) {
...
+                    "Version":  {
+                        "Fields": {
+                            "field": []
+                        }
+                    }
+                },
+                "_contentType": this.contentType
             };
+            Y.Mock.expect(this.contentType, {
+                method: this.getFieldDefinitionIdentifiers,
```
dpobel: I don't really see the point in putting the method name in a property of the test object. You can directly put the literal string here.

Tests/js/models/assets/ez-usermodel-tests.js
```diff
@@ -14,7 +14,9 @@ YUI.add('ez-usermodel-tests', function (Y) {
             this.serviceMock = new Y.Mock();
             this.serviceLoad = 'loadUser';
             this.rootProperty = "User";
-            this.parsedAttributeNumber = Y.eZ.User.ATTRS_REST_MAP.length + 1; // links
+            this.getFieldDefinitionIdentifiers = 'getFieldDefinitionIdentifiers';
+            this.contentType = new Y.Mock();
+            this.parsedAttributeNumber = Y.eZ.User.ATTRS_REST_MAP.length + 2; // links
```
dpobel: you should update the comment to `// links + avatar` since it's not "+ 1" anymore

Tests/js/models/assets/ez-usermodel-tests.js
```diff
@@ -40,11 +54,37 @@ YUI.add('ez-usermodel-tests', function (Y) {
...
+
+            Y.Mock.expect(this.capiMock, {
+                method: this.capiGetService,
+                returns: this.serviceMock
+            });
+            Y.Mock.expect(this.serviceMock, {
+                method: this.serviceLoad,
+                args: [
+                    modelId,
+                    Y.Mock.Value.Function
```
dpobel: since you don't call this function, the entire `loadUser` callback is not covered. This mock shoud be configured to run that function something like:
```
Mock.expect(this.serviceMock, {
   method: this.serviceLoad,
   args: [modelId, Mock.Value.Function],
   run: Y.bind(function (id, callback) {
       callback(false, this.loadResponse);
   }, this)
});
```
this will have several consequences:
1. the configured `loadResponse` should have a `User.ContentType._href` value
1. you should dynamically create a fake version of eZ.ContentType
1. to get full coverage the mock should also be configured in another test to execute the callback with a truthy value as the error

Tests/js/models/assets/ez-usermodel-tests.js
```diff
@@ -28,9 +30,21 @@ YUI.add('ez-usermodel-tests', function (Y) {
                     "login": "admin",
                     "email": "dp@ez.no",
                     "enabled": "true",
-                    "name": "Administrator User"
-                }
+                    "name": "Administrator User",
+                    "Version":  {
+                        "Fields": {
+                            "field": []
```
dpobel: since this array is always empty, the filter function in `_parseStruct` is not covered and the alternative right after it is half covered as well.

Tests/js/views/assets/ez-navigationhubview-tests.js
```diff
@@ -3,7 +3,7 @@
  * For full copyright and license information view LICENSE file distributed with this source code.
  */
 YUI.add('ez-navigationhubview-tests', function (Y) {
-    var viewTest, eventTest, logOutTest,
+    var viewTest, eventTest,
```
dpobel: the `userChange` event handler appears as covered but it's not really tested and the `userAvatarChange` is not covered at all.

Tests/js/views/assets/ez-usermenuitemfireeventview-tests.js
```diff
@@ -0,0 +1,91 @@
...
+                container: '.container',
+                eventName: 'logOut'
+            });
+        },
+
+        tearDown: function () {
+            this.view.destroy();
+        },
+
+        "Should fire event the user": function () {
```
dpobel: nitpick: `Should fire the configured event`

Tests/js/views/assets/ez-usermenuview-tests.js
```diff
@@ -0,0 +1,74 @@
...
+
+            Y.Assert.areSame(
+                1,
+                container.get('children').size(),
+                'The user menu should have one child'
+            );
+        },
+    });
+
+    eventTest = new Y.Test.Case({
```
dpobel: you are not testing any event here but the `displayed` attribute and the consequence on the dom so this test should be called something like `displayTest` and its name should be updated.

Tests/js/views/assets/ez-usermenuview-tests.js
```diff
@@ -0,0 +1,74 @@
...
+        tearDown: function () {
+            this.view.destroy();
+            delete this.view;
+        },
+
+        "Should show the user menu": function () {
+            var view = this.view,
+                container = view.render().get('container');
+
+            view._set('displayed', true);
```
dpobel: you should use `toggleDisplayed` since it's the public API to manipulate `displayed`. Then you have to assert both the `displayed` value and the class on the container.

Tests/js/views/assets/ez-usermenuview-tests.js
```diff
@@ -0,0 +1,74 @@
...
+
+            Y.Assert.isFalse(container.hasClass(CLASS_HIDDEN), 'The user menu should be visible');
+        },
+
+        "Should hide the user menu": function () {
+            var view = this.view,
+                container = view.render().get('container');
+
+            view._set('displayed', true);
+            view._set('displayed', false);
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/513/files#r53945725

Tests/js/views/assets/ez-usermenuview-tests.js
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-usermenuview-tests', function (Y) {
+    var renderTest, eventTest,
```
dpobel: after adding the `addMenuItem` method, you'll also have to add a dedicated test case for it (like for any public method). A test is missing for the clickoutside event handler (you can add a method in the test about `displayed`).
You should also add a test for the valueFn of the `items` attribute. To avoid adding too many dependencies, in the setUp of this test you can write something like:
```js
Y.eZ.UserMenuItemFireEventView = Y.View;
```
and then you just have to check that the view in `items` receives the correct title and eventName and that the UserMenuView is an event target of the UserMenuItemFireEventView.

Tests/js/views/assets/ez-usermenuview-tests.js
```diff
@@ -0,0 +1,74 @@
...
+        CLASS_HIDDEN = 'is-user-menu-hidden';
+
+    renderTest = new Y.Test.Case({
+        name: "eZ User Menu render test",
+
+        setUp: function () {
+            this.userMenuItemFireEventView = new Y.View();
+            this.view = new Y.eZ.UserMenuView({
+                items: [this.userMenuItemFireEventView]
```
dpobel: you should put at least 2 views and set on them a priority so that you can check that the views are sorted by priority.

Tests/js/views/assets/ez-userprofileview-tests.js
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-userprofileview-tests', function (Y) {
+    var renderTest;
```
dpobel: the case where the view receives a valid `userAvatar` object is not covered. A test on the tap event is also missing and like in https://github.com/ezsystems/PlatformUIBundle/pull/513/files#r53947305, the `valueFn` of the `userMenuView` is not tested.

Tests/js/views/ez-navigationhubview.html
```diff
@@ -15,8 +15,11 @@
 item
 </script>
 
+<script type="text/x-handlebars-template" id="itemtest-ez-template">
```
dpobel: not needed, no ?

Tests/js/models/assets/ez-usermodel-tests.js
```diff
@@ -40,11 +54,37 @@ YUI.add('ez-usermodel-tests', function (Y) {
...
+            Y.Mock.verify(this.capiMock);
+            Y.Mock.verify(this.serviceMock);
+        },
     }));
 
     Y.Test.Runner.setName("eZ User Model tests");
     Y.Test.Runner.add(modelTest);
 
-}, '', {requires: ['test', 'model-tests', 'ez-usermodel', 'ez-restmodel']});
+}, '', {requires: ['test', 'model-tests', 'ez-usermodel', 'ez-restmodel', 'model']});
```
dpobel: not needed, no ?

Tests/js/views/ez-usermenuview.html
```diff
@@ -0,0 +1,58 @@
...
+    } else {
+        loaderFilter = filter;
+    }
+
+    YUI({
+        coverage: ['ez-usermenuview'],
+        filter: loaderFilter,
+        modules: {
+            "ez-usermenuview": {
+                requires: ['event-tap', 'event-outside', 'template-micro', 'ez-usermenuitemview'],
```
dpobel: `template-micro` and `ez-usermenuitemview` should not be there but `ez-view` should be added. That also means, `ez-usermenuitemview` and `ez-templatebasedview` modules can be removed below.

Tests/js/views/services/assets/ez-navigationhubviewservice-tests.js
```diff
@@ -686,6 +699,40 @@ YUI.add('ez-navigationhubviewservice-tests', function (Y) {
...
+                args: ['avatar'],
+                returns: this.avatar,
+            });
+
+            Y.Mock.expect(this.getContentServiceMock, {
+                method: 'loadImageVariation',
+                args: [Mock.Value.String, Y.Mock.Value.Function],
+                run: function (avatarPath, callback) {
+                    Y.Assert.areSame(expectedAvatarUrl, avatarPath, 'Should pass the correct url');
+                    callback(false, response);
```
dpobel: the error case is not covered

Tests/js/views/services/assets/ez-navigationhubviewservice-tests.js
```diff
@@ -686,6 +699,40 @@ YUI.add('ez-navigationhubviewservice-tests', function (Y) {
...
+
+            Mock.expect(this.app, {
+                method: 'get',
+                args: ['user'],
+                returns: this.user,
+            });
+            Mock.expect(this.user, {
+                method: 'get',
+                args: ['avatar'],
+                returns: this.avatar,
```
dpobel: the case where there's no avatar in the user is not covered as well

Resources/public/js/views/ez-usermenuview.js
dpobel: hum weird, that should not happen, let me check the details

dew326:
ping @dpobel 
done

Resources/public/js/views/ez-usermenuview.js
dpobel: ok this is happening because the `tap` event used in eZ.UserProfileView to show the menu is not prevented, so with:

```js
        _toggleUserMenu: function (e) {
            e.preventDefault();
            this.get('userMenuView').toggleDisplayed();
        }
```

you can simplified the condition.

dew326: it doesn't help. 
Condition still needed

dew326:
ping @dpobel 
do you see something else to improve?

Resources/public/js/views/ez-usermenuview.js
dpobel: indeed you're right, I don't know why this worked for me this morning, maybe missing coffee for the PC or myself :)

Anyway, after digging a bit more in the code, the way the toggle display responsibility is shared is not super clear. On one hand, UserProfileMenu calls a public method on the UserMenuView to toggle its visibility and on the other hand UserMenuView tries to manage the click outside to hide itself.
I would move the click outside handling to the UserProfileMenu and even simplify it. So basically, UserProfileView would always subscribe to the clickoutside event on its container and when it happens, it would call a `hide` method on the UserMenuView.

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -0,0 +1,193 @@
...
+            this.containerTemplate = '<ul class="ez-view-usermenuview ' + CLASS_HIDDEN + '"/>';
+
+            this.on('displayedChange', this._uiToggleUserMenu);
+            this.on('activeChange', this._setActiveViews);
+            this.after('displayedChange', this._setClickOutsideUserMenuHandler);
+
+            /**
+             * The click outside user menu event handler
+             *
+             * @method _clickOutsideUserMenuHandler
```
dpobel: it's a property (`@property`) not a method and it's should be marked as protected.

Resources/public/js/views/ez-usermenuitemfireeventview.js
```diff
@@ -0,0 +1,68 @@
...
+     */
+    Y.namespace('eZ');
+
+    /**
+     * The user menu item fire event view. 
+     *
+     * @namespace eZ
+     * @class UserMenuItemFireEventView
+     * @constructor
+     * @extends Y.eZ.UserMenuItemView
```
dpobel: I think the `Y.` should be omitted so that yuidoc finds the reference and generates a link.

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -0,0 +1,193 @@
...
+         *
+         * @method addMenuItem
+         * @public
+         * @param view {Object}
+         */
+        addMenuItem: function (view) {
+            var items = this.get('items');
+
+            items.push(view);
+            this.set('items', items);
```
dpobel: nitpick: actually you don't have to do that, the array is updated by the previous line and this array is referenced in the `items` attribute.

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -0,0 +1,193 @@
...
+                    ];
+                }
+            },
+
+            /**
+             * Whether the user menu is displayed or not
+             *
+             * @attribute displayed
+             * @type Boolean
+             * @default false
```
dpobel: missing `@readOnly` tag

Tests/js/models/assets/ez-usermodel-tests.js
```diff
@@ -28,9 +29,56 @@ YUI.add('ez-usermodel-tests', function (Y) {
...
             };
+
+            Y.eZ.ContentType = Y.Model;
+
+            Y.Mock.expect(this.contentType, {
+                method: 'getFieldDefinitionIdentifiers',
+                args: [
+                    'ezimage'
+                ],
+                returns: 'image'
```
dpobel: weird this works, `getFieldDefinitionIdentifiers` is supposed to return an array and its return value is used as an array https://github.com/ezsystems/PlatformUIBundle/pull/513/files#diff-c1e1c619e4630afebc3e8790954d6118R71

dpobel: most likely related: in `_parseStruct` the case where there's an imageField is not covered.

Tests/js/views/assets/ez-navigationhubview-tests.js
dpobel: the `valueFn` of `userProfileView` attribute is not covered

Tests/js/views/assets/ez-usermenuitemfireeventview-tests.js
```diff
@@ -0,0 +1,92 @@
...
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-usermenuitemfireeventview-tests', function (Y) {
+    var renderTest, eventTest;
+
+    renderTest = new Y.Test.Case({
+        name: "eZ User Menu Item Fire Event render test",
+
+        setUp: function () {
+            this.title = "Welcome to the jungle";
```
dpobel: good choice :D

Tests/js/views/assets/ez-usermenuview-tests.js
```diff
@@ -0,0 +1,157 @@
...
+
+            Y.Assert.areSame(
+                2,
+                container.get('children').size(),
+                'The user menu should have one child'
+            );
+
+            Y.Assert.areSame(
+                this.userMenuItemSecondView.get('priority'),
+                this.view.get('items')[0].get('priority'),
```
dpobel: I would compare the object directly and both

Tests/js/views/assets/ez-usermenuview-tests.js
```diff
@@ -0,0 +1,157 @@
...
+        },
+
+        "Should add item to the user menu items": function () {
+            this.view.addMenuItem(this.userMenuItemSecondView);
+
+            Y.Assert.areSame(
+                2,
+                this.view.get('items').length,
+                'Should add menu item to the items attribute'
+            );
```
dpobel: you should also check the effect of the `addTarget`

Tests/js/views/assets/ez-usermenuview-tests.js
```diff
@@ -0,0 +1,157 @@
...
+        name: "eZ User Menu attribute test",
+
+        setUp: function () {
+            Y.eZ.UserMenuItemFireEventView = Y.View;
+            this.view = new Y.eZ.UserMenuView();
+        },
+
+        tearDown: function () {
+            this.view.destroy();
+            delete this.view;
```
dpobel: you should also delete Y.eZ.UserMenuItemFireEventView

Tests/js/views/assets/ez-usermenuview-tests.js
```diff
@@ -0,0 +1,157 @@
...
+            this.userMenuItemFireEventView = new Y.View();
+            this.view = new Y.eZ.UserMenuView({
+                items: [this.userMenuItemFireEventView],
+                container: '.container'
+            });
+        },
+
+        tearDown: function () {
+            this.view.destroy();
+            delete this.view;
```
dpobel: missing destroy and delete of this.userMenuItemFireEventView

Tests/js/views/assets/ez-usermenuview-tests.js
```diff
@@ -0,0 +1,157 @@
...
+            var view = this.view.render();
+
+            view.toggleDisplayed();
+
+            view.after('displayedChange', function () {
+                Y.Assert.isTrue(view.get('container').hasClass(CLASS_HIDDEN), 'The user menu should be hidden');
+                Y.Assert.isFalse(view.get('displayed'), 'Value of displayed attribute should be false');
+            });
+
+            Y.one(SELECTOR_OUTSIDE).simulate('click');
```
dpobel: the `displayedChange` event handler should set a boolean flag so that you verify that it was actually called. Otherwise, it can not be called and the test will still pass.

Tests/js/views/assets/ez-userprofileview-tests.js
```diff
@@ -0,0 +1,168 @@
...
+
+        tearDown: function () {
+            this.view.destroy();
+            delete this.view;
+        },
+
+        "Should set event target to the view": function () {
+            var item = this.view.get('userMenuView');
+
+            Y.Assert.areSame(this.view, item.getTargets()[0], 'Should set correct event target');
```
dpobel: you should also check that `item` is an instance of `Y.eZ.UserMenuView`

Tests/js/views/assets/ez-userprofileview-tests.js
```diff
@@ -0,0 +1,168 @@
...
+        name: "eZ User Profile attribute test",
+
+        setUp: function () {
+            Y.eZ.UserMenuView = Y.View;
+            this.view = new Y.eZ.UserProfileView();
+        },
+
+        tearDown: function () {
+            this.view.destroy();
+            delete this.view;
```
dpobel: missing delete of `Y.eZ.UserMenuView`

Tests/js/views/assets/ez-userprofileview-tests.js
```diff
@@ -0,0 +1,168 @@
...
+            Y.Assert.areSame(this.view, item.getTargets()[0], 'Should set correct event target');
+        },
+    });
+
+    eventTest = new Y.Test.Case({
+        name: "eZ User Profile Event test",
+
+        setUp: function () {
+            this.userMenuView = new Y.Mock();
+            this.fakeView = new Y.View();
```
dpobel: not used anywhere ?

Tests/js/views/ez-userprofileview.html
```diff
@@ -0,0 +1,60 @@
...
+<body>
+<div class="container"></div>
+<div class="ez-view-usermenuview"><div class="ez-view-usermenuview"></div></div>
+
+<script type="text/x-handlebars-template" id="userprofileview-ez-template">
+    <div class="ez-user-profile">
+        <div class="ez-user-profile-info">
+            <p class="ez-user-profile-name">{{user.name}}</p>
+        </div>
+        <img class="ez-user-profile-image" src="{{#if user.avatar.uri}}{{user.avatar.uri}}{{else}}{{defaultAvatarImage}}{{/if}}" alt="{{avatarAltText}}">
```
dpobel: it's not aligned anymore with the actual template. On template, the good practice is to try to simplify it as much as possible in the test so it generates only what is needed for the view to work.

Tests/js/views/assets/ez-userprofileview-tests.js
dew326: this is used as a return of a mock few lines below

dpobel: oh yeah my mistake sorry

dew326:
ping @dpobel 
done, comments are not hide because the changes are line below

Tests/js/views/assets/ez-usermenuview-tests.js
```diff
@@ -0,0 +1,166 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-usermenuview-tests', function (Y) {
+    var renderTest, displayTest, addItemTest, attrTest,
```
dpobel: `hide` is not tested

Tests/js/models/assets/ez-usermodel-tests.js
```diff
@@ -28,9 +29,57 @@ YUI.add('ez-usermodel-tests', function (Y) {
...
+            this.imageIdentifiers = ['image'];
+
+            Y.eZ.ContentType = Y.Model;
+
+            Y.Mock.expect(this.contentType, {
+                method: 'getFieldDefinitionIdentifiers',
+                args: [
+                    'ezimage'
+                ],
+                returns: this.imageIdentifiers
```
dpobel: so now the case where there's no image field in the User Content Type is not covered. Basically, you need to add a test where `getFieldDefinitionIdentifiers` returns and empty array and you check that `avatar` is null.

dpobel:
besides the last 2 comments on missing tests, +1
but the CSS style definitively not consistent with the rest of PlatformUI, I guess we can change that without delaying the RC (ping @bdunogier @andrerom)

dew326:
ping @dpobel 
done

Tests/js/models/assets/ez-usermodel-tests.js
```diff
@@ -40,8 +89,108 @@ YUI.add('ez-usermodel-tests', function (Y) {
...
+            Y.Mock.verify(this.capiMock);
+            Y.Mock.verify(this.serviceMock);
+        },
+
+        "Should not set avatar attribute with CAPI when user doesn't have avatar": function () {
+            var m = this.model,
+                modelId = "/api/v2/ezp/model/mid",
+                callback = function () { };
+
+            this.imageIdentifiers = [];
```
dpobel: I guess you did not you check the coverage. This is not enough because the `contentType` mock is still configured with the value you set in setUp.

Tests/js/models/assets/ez-usermodel-tests.js
```diff
@@ -40,8 +89,108 @@ YUI.add('ez-usermodel-tests', function (Y) {
...
+                }, this)
+            });
 
+            m.set('id', modelId);
+            m.sync('read', {
+                    api: this.capiMock
+                }, callback
+            );
+
+            Y.Assert.isNull(this.model.get('avatar'), 'Should not set avatar when user doesn`t have one');
```
dpobel: this is not the right place to check that. This check should be done in the callback. Also to make sure the code is correctly executed, the avatar attribute in the model should be initialized to something else since `null` is also the default value of this attribute.

dew326:
ping @dpobel 
I changed the test

dpobel:
+1
ping @StephaneDiot 
(even if the style are not consistent with the rest of the app and the way the clickoutside is handled can be improved. I'll do that today.)

StephaneDiot:
+1 (niptick: a screen cast would have been great to see the changes from Demo)

andrerom:
> niptick: a screen cast would have been great to see the changes from Demo

+1 @dpobel any instructions on that? :)

dpobel:
@andrerom instructions on recording a screencast ? pick whatever tool for your system and do it :) I personally  use https://www.screencastify.com/ which is a Chrome extension but there are plenty of alternatives.

PR #515, opened 2016-02-17
EZP-25482: Fix exception with CSRF protection disabled

emodric:
When CSRF protection is disabled with `framework.csrf_protection.enabled` flag set to false, site crashes with an exception about missing `security.csrf.token_manager` service.

Depends on: https://github.com/ezsystems/ezpublish-kernel/pull/1589


dpobel:
you need to add a test in https://github.com/ezsystems/PlatformUIBundle/blob/master/Tests/ApplicationConfig/Providers/SessionInfoTest.php
other than that, that seems ok to me.

emodric:
@dpobel Done.

dpobel:
+1

andrerom:
+1

dpobel:
@andrerom that's a very good remark, I need to check

emodric:
@andrerom @dpobel In my tests login worked fine.

dpobel:
the actual login should work but did you check with an existing session ?
I think that because of https://github.com/ezsystems/ez-js-rest-client/blob/master/src/authAgents/SessionAuthAgent.js#L81, that won't work, can you try @emodric ?

emodric:
@dpobel I'm not sure I understand. What and how do I need to test? It's not too obvious from JS code.

dpobel:
:)
Just go to PlatformUI, login normally and after you are logged in, refresh the page. In that case, PlatformUI will receive the session info object describing the existing session but without the CSRF token and the SessionAuthAgent will throw an error because it is missing and expects one.

emodric:
Hm... This is what I get now when trying to login:

`Parameter "sessionId" for route "ezpublish_rest_deleteSession" must match "[^/]++" ("" given) to generate a corresponding URL.`

Is this expected?

ApplicationConfig/Providers/SessionInfo.php
```diff
@@ -29,7 +29,7 @@ class SessionInfo implements Provider
 
     public function __construct(
         SessionInterface $session,
-        CsrfTokenManagerInterface $csrfTokenManager,
+        CsrfTokenManagerInterface $csrfTokenManager = null,
```
lolautruche: You cannot have a default value here. If you want to add `null` as default value, you must move the argument at the end.

emodric: Damn it! Slipped through my fingers!

But then it's a breaking change :/

lolautruche: Technically yes, but it should be OK IMHO. And btw the service should be marked `private`

emodric: > Technically yes, but it should be OK IMHO.

Okay, fixed.

> And btw the service should be marked `private`

It already is?

bdunogier:
The consensus reached after discussion with the platform team can be read [on jira](https://jira.ez.no/browse/EZP-25482?focusedCommentId=188896&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-188896).

This pull-request is not sufficient at the moment, as it only fixes an error, but doesn't make the wholee thing work.

What we would like for the moment is that PlatformUI throws a clear and documented exception when the CsrfTokenManager is not set. We can then quietly consider an improvement to the JS rest client to allow it to work without a csrf token (not supported at the moment).

andrerom:
Ok, I guess csrf should only be allowed to be disabled for basic auth mode, but we don't support that from ui atm so not applicable. Better error reporting, in this case especially on logging page is indeed missing piece here: https://jira.ez.no/browse/EZP-25344

PR #518, opened 2016-02-24
EZP-25410: inject site access into CAPI

wizhippo:
This addresses part of EZP-25410.  This will inject the siteaccess into the CAPI that platformui creates.

### DIFF
--- a/ApplicationConfig/Providers/SiteAccess.php
+++ b/ApplicationConfig/Providers/SiteAccess.php
@@ -1,0 +1,31 @@
+<?php
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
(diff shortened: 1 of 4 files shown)


dpobel:
Thanks for the pull request @wizhippo 
This raises an interesting question that has been there for quite some time actually. For now, PlatformUI runs outside any siteaccess and is more bound to a given repository. But that's unclear and I think we should clarify that.

wizhippo:
@dpobel The direction you think will be taken is to have PlatformUI not be siteaccess aware?  That would be too bad in my opinion, especially in multi-site setups.  No longer being able to have customized admins based on site access will be missed:(  That was a great feature of legacy along with being able to specify the root node for the admin based on site access.

andrerom:
@wizhippo Well, what we might move towards is to have a site drop down inside the admin ui, so it will be different.

PR #519, opened 2016-02-26
User menu improvements

dpobel:
Follow up of https://github.com/ezsystems/PlatformUIBundle/pull/513

This PR aims at giving some love to the user menu that is somewhat naked in its current form and feels a bit odd in PlatformUI.

Screencast (in the screencast the menu artificially bigger with several items):

[](http://www.youtube.com/watch?v=0yZUB4oZN9g "Click to play on Youtube.com")


dpobel:
ping @dew326 @sunpietro @StephaneDiot @andrerom

andrerom:
+1

Tests/js/views/assets/ez-userprofileview-tests.js
```diff
@@ -109,6 +110,49 @@ YUI.add('ez-userprofileview-tests', function (Y) {
...
+        tearDown: function () {
+            this.view.destroy();
+            delete this.view;
+        },
+
+        "Should add the menu displayed class": function () {
+            this.view.get('userMenuView').set('displayed', true);
+
+            Assert.isTrue(
+                this.view.get('container').hasClass('is-menu-displayed'),
```
dew326: You use `is-menu-displayed` twice, maybe it's a good idea to define a variable `CLASS_IS_DISPLAYED`? What do you think about it?

StephaneDiot:
+1

Tests/js/views/assets/ez-userprofileview-tests.js
dpobel: done but that's highly debatable :)
if you see this class as just a string and an implementation detail then yes you are perfectly right.
But in this case (ie a test), this class is like a public method (or a public behaviour), this is something that as of the release should stay for BC reason. For instance, you wouldn't put `'container'` or `'get'` (the method name) in variables (even if this test uses a lot those strings).
(in the implementation, we use variables because it can improve compression and minification efficiency and also because it improves the readability in that context)

lserwatka:
+1

dew326:
+1

PR #521, opened 2016-03-02
EZP-25522: User is not able to get variation of his profile image

dew326:
**Jira ticket**: https://jira.ez.no/browse/EZP-25522

Now we do not throw an error when user don't have permission to load image variation.

### DIFF
--- a/Resources/public/js/views/services/ez-navigationhubviewservice.js
+++ b/Resources/public/js/views/services/ez-navigationhubviewservice.js
@@ -64,7 +64,9 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
 
             if (user.get('avatar')) {
                 contentService.loadImageVariation(user.get('avatar').variations.platformui_profileview.href, tasks.add(function (error, response) { //jshint ignore:line
-                    if ( error ) {
+                    if ( error && response.status !== 401 ) {
+                        // 401 error is ignored because the user might not have the right to read
+                        // his/her own avatar see https://jira.ez.no/browse/EZP-25522
                         loadError = true;
                         return;
                     }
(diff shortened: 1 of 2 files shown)


dew326:
ping @dpobel

Resources/public/js/views/services/ez-navigationhubviewservice.js
```diff
@@ -64,7 +64,10 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
 
             if (user.get('avatar')) {
                 contentService.loadImageVariation(user.get('avatar').variations.platformui_profileview.href, tasks.add(function (error, response) { //jshint ignore:line
-                    if ( error ) {
+                    if ( error && error.message.indexOf('401') !== -1 ) {
```
dpobel: that's very bad way of checking the HTTP status code... You can check it in `response.status`.

Resources/public/js/views/services/ez-navigationhubviewservice.js
```diff
@@ -64,7 +64,10 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
...
+                        error = false;
```
dpobel: I don't see the point in modifying `error` which is a callback parameter. I think this can be simplified in something like:

```js
if ( error && response.status !== 401 ) {
    // 401 error is ignored because the user might not have the right to read
    // his/her own avatar see https://jira.ez.no/browse/EZP-25522
    loadError = true;
    return;
}
```

dew326:
ping @dpobel

andrerom:
+1

dpobel:
+1

PR #522, opened 2016-03-03
EZP-25503: The user profile menu stays after opening Notifications form

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-25503

- [x] hide user menu when menu item has been clicked
- [x] JS Unit Tests updated

### DIFF
--- a/Resources/public/js/views/ez-usermenuitemfireeventview.js
+++ b/Resources/public/js/views/ez-usermenuitemfireeventview.js
@@ -28,6 +28,8 @@ YUI.add('ez-usermenuitemfireeventview', function (Y) {
 
         initializer: function () {
             this.containerTemplate = '<li class="' + this._generateViewClassName(this._getName()) + '"/>';
+
+            this.on('addedToUserMenu', this._addUserMenuHideOnEvent, this);
         },
 
         /**
@@ -39,6 +41,17 @@ YUI.add('ez-usermenuitemfireeventview', function (Y) {
         render: function () {
             this.get('container').setHTML(this.template({title: this.get('title')}));
             return this;
+        },
+
+        /**
+         * Adds the event listener to hide the user menu
+         *
+         * @method _addUserMenuHideOnEvent
+         * @protected
+         * @param event {Object} event facade
+         */
+        _addUserMenuHideOnEvent: function (event) {
+            this.on(this.get('eventName'), Y.bind(event.userMenu.hide, event.userMenu));
         },
 
         /**
--- a/Resources/public/js/views/ez-usermenuview.js
+++ b/Resources/public/js/views/ez-usermenuview.js
@@ -113,7 +113,7 @@ YUI.add('ez-usermenuview', function (Y) {
         },
 
         /**
-         * Adds item to the menu
+         * Adds item to the menu and fire event when it's done
          *
          * @method addMenuItem
          * @public
@@ -125,6 +125,16 @@ YUI.add('ez-usermenuview', function (Y) {
             items.push(view);
 
             view.addTarget(this);
+
+            /**
+             * Fired when the view is added to the user menu
+             *
+             * @event addedToUserMenu
+             * @param userMenu {Object} the user menu view
+             */
+            view.fire('addedToUserMenu', {
+                userMenu: this,
+            });
         },
     }, {
         ATTRS: {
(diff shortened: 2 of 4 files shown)


dew326:
ping @dpobel

Resources/public/js/views/ez-usermenuview.js
```diff
@@ -124,6 +124,10 @@ YUI.add('ez-usermenuview', function (Y) {
 
             items.push(view);
 
+            view.fire('addedToUserMenu', {
```
dpobel: I realize this should be done after the `view.addTarget` so that others components can react to that event. Also the event should be documented (example http://ezsystems.github.io/platformui-javascript-api/files/Resources_public_js_views_universaldiscovery_ez-universaldiscoverybrowseview.js.html#l122) and you should also explain in the method description that this event is fired as well.

Resources/public/js/views/ez-usermenuitemfireeventview.js
```diff
@@ -42,6 +44,27 @@ YUI.add('ez-usermenuitemfireeventview', function (Y) {
...
 
         /**
+         * Adds the event listener
+         *
+         * @method _addEventListener
+         * @protected
+         * @param event {Object} event facade
+         */
+        _addEventListener: function (event) {
+            this.on(this.get('eventName'), this._hideUserMenu, event.userMenu);
```
dpobel: that's a very confusing way of writing that `event.userMenu.hide` should be called. I would write something like:
```js
this.on(this.get('eventName'), Y.bind(event.userMenu.hide, event.userMenu));
```

Resources/public/js/views/ez-usermenuitemfireeventview.js
```diff
@@ -42,6 +44,27 @@ YUI.add('ez-usermenuitemfireeventview', function (Y) {
         },
 
         /**
+         * Adds the event listener
+         *
+         * @method _addEventListener
+         * @protected
+         * @param event {Object} event facade
+         */
+        _addEventListener: function (event) {
```
dpobel: that's not a very descriptive name. maybe `_addUserMenuHideOnEvent` or something like that.

Tests/js/views/assets/ez-usermenuitemfireeventview-tests.js
```diff
@@ -84,6 +84,31 @@ YUI.add('ez-usermenuitemfireeventview-tests', function (Y) {
...
+                run: function () {
+                    isMethodCalled = true;
+                }
+            });
+
+            this.view.fire('addedToUserMenu', {userMenu: userMenuView});
+
+            this.view.on('logOut', function () {
+                that.resume(function () {
+                    Y.Assert.isTrue(isMethodCalled, 'Should call the `hide` method');
```
dpobel: instead of using a boolean variable, you should use `Y.Mock.verify(userMenuView)`

Tests/js/views/assets/ez-usermenuview-tests.js
```diff
@@ -81,18 +81,26 @@ YUI.add('ez-usermenuview-tests', function (Y) {
...
+                isAddedEventFired = false;
 
             view.on('*:' + eventName, function () {
                 isEventFired = true;
             });
 
+            userMenuView.on('addedToUserMenu', function (event) {
+                isAddedEventFired = true;
+
+                Y.Assert.areSame(view, event.userMenu, 'Should send correct parameter');
```
dpobel: The error message is not precise enough. imagine if the test is failing in 6 months, what does *correct parameter* mean? The message should be something like *The user menu view should be available in the event parameters*.

dew326:
ping @dpobel

Tests/js/views/assets/ez-usermenuitemfireeventview-tests.js
```diff
@@ -84,6 +84,31 @@ YUI.add('ez-usermenuitemfireeventview-tests', function (Y) {
...
+
+        "Should hide the user menu": function () {
+            var container = this.view.render().get('container'),
+                that = this,
+                userMenuView = new Y.Mock();
+
+            Y.Mock.expect(userMenuView, {
+                method: 'hide',
+                args: [Y.Mock.Value.Object],
+                run: function (event) {
```
dpobel: nitpick: is that necessary ? I think you can remove the `run` part.

dew326:
ping @dpobel

dpobel:
+1
ping @StephaneDiot

StephaneDiot:
+1

PR #523, opened 2016-03-03
Fixed: avoid random failures in the REST model tests

dpobel:
Trying to avoid random failures in REST model tests. Those failures can happen if the 2 compared date objects were not created in the same milliseconds. With this patch, the test will fail only if the dates were not created in the same second, this can still happen but that is 1000 times less likely :)


glye:
Sounds good, +1. I have gotten off-by-one-second test failures before, but in other tests (like content publishing).

dpobel:

andrerom:
+1

PR #502, opened 2016-01-28
Js translation bundle + French Translations

yed30:
Link: https://jira.ez.no/browse/EZP-25428

These modifications use the JsTranslationBundle (cf. https://github.com/willdurand/BazingaJsTranslationBundle)

**Installation:**
The the bundle is included in the PlatformUIBundle  composer.json.
Update the composer and then add activate the bundle in your Kernel:
`new Bazinga\Bundle\JsTranslationBundle\BazingaJsTranslationBundle(),`

By default, all domains are dumped. Add the translations domains in the config file (cf http://symfony.com/doc/current/components/translation/introduction.html#using-message-domains):
```
bazinga_js_translation:
    active_domains:
        - editorial
```
Right now, for the eZPlatformUIBundle only the _editorial_ translation domain is loaded in the main template shell.html.twig.

To generate (dump) the translations in the web/js/translations folder of your application, run:
`php app/console bazinga:js-translation:dump`

**How to use:**
To add a translation in a js file, just use the Translator object, which implements the Symfony2 TranslatorInterface and provides the same trans() and transChoice() methods:

`Translator.trans('key', {}, 'DOMAIN_NAME');`
// the translated message or undefined

`Translator.transChoice('key', 1, {}, 'DOMAIN_NAME');`
// the translated message or undefined

cf. https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/Resources/doc/index.md

Please let me know any question or issue.
Below screen shots with the French translations (The part in the YUI templates are still not translated ;) )
ezjstranslation3
ezjstranslation2
ezjstranslation1

PR #525, opened 2016-03-07
EZP-25428: JS Translation infrastructure

dpobel:
Replaces https://github.com/ezsystems/PlatformUIBundle/pull/502 (ping @yed30)

## Tasks

* [x] Add a test to Y.eZ.Translator component
* [x] Define and use a convention on the translations ids
* [x] Open a PR against ezplatform to instantiate BazingaJsTranslationBundle in the kernel
- [x] Define how we handle translations in templates
- [x] See how we could use `php app/console translation:update` in handlebars templates
- [ ] ~~Check for new stuff to translate since rebase~~ https://jira.ez.no/browse/EZP-26430
- [x] Bump bazinga to 2.6 https://github.com/ezsystems/ezplatform/pull/116
- [x] move french translation in the main folder (no fr subfolder) so it can be updated by the command line
- [x] add all the translation keys in templates ?
- [ ] ~~extractor doesn't detect when 2 translation keys on the same line~~ https://jira.ez.no/browse/EZP-26432


andrerom:
Meta repo PR added, this can be rebased and tested using:
```diff
-git clone --depth 1 --single-branch --branch master https://github.com/ezsystems/ezplatform.git
+git clone --depth 1 --single-branch --branch bazinga https://github.com/ezsystems/ezplatform.git
```
in `bin/travis/prepare_behat.sh`

Resources/translations/editorial.en.xlf
```diff
@@ -0,0 +1,103 @@
...
+      </trans-unit>
+      <trans-unit id="88490010ebf2eef359f946e13db60ffa" resname="discoverybar.minimize">
+        <source>discoverybar.minimize</source>
+        <target>Minimize</target>
+      </trans-unit>
+      <trans-unit id="fc340d93afc3c54ebcd212b246ea4a81" resname="discoverybar.contenttree">
+        <source>discoverybar.contenttree</source>
+        <target>Content tree</target>
+      </trans-unit>
+      <trans-unit id="discoverybar.trash" resname="discoverybar.trash">
```
andrerom: Any reason why this one is not using a md5 id?

Resources/views/PlatformUI/shell.html.twig
```diff
@@ -45,7 +45,17 @@
         <p>{{ 'javascript.disabled.required'|trans }}</p>
         <p><strong>{{ 'javascript.enable'|trans }}</strong></p>
     </noscript>
+    {% javascripts
+        'bundles/bazingajstranslation/js/translator.min.js'
+        'js/translations/config.js'
```
andrerom:
```
  [RuntimeException]                                                                                    
  The source file "/home/travis/build/ezplatform/app/../web/js/translations/config.js" does not exist. 
```

Should this be here?

andrerom: https://github.com/willdurand/BazingaJsTranslationBundle/issues/165

Resources/views/PlatformUI/shell.html.twig
```diff
@@ -45,7 +45,17 @@
...
+    %}
+        <script src="{{ asset_url }}"></script>
+    {% endjavascripts %}
+    {# We only load the translations for the eZPlatformUIBundle JS files #}
+    <script src="{{ url('bazinga_jstranslation_js', { 'domain': 'editorial' }) }}"></script>
```
andrerom: This loads all translations afaik, but depending on how we let users select UI languages in the future we should probably limit it to not load all. see https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/Resources/doc/index.md#locales

andrerom:
Status: Works locally now. Currently no clue why behat fails on Travis.
*Note: It seems we are doing something wrong on translation files, we only give them locale names like `en` and `fr`, however the locales of the system is normally full form `en_GB` and `fr_FR`, which is why I had to somewhat hack it in 7669dd8. Also `bazinga_jstranslation_js` does not load locale fallback so had to specify that in url as well, hard coded for now.*

Resources/public/js/apps/plugins/ez-registerlanguagehelpersplugin.js
```diff
@@ -58,6 +66,26 @@ YUI.add('ez-registerlanguagehelpersplugin', function (Y) {
...
+         * The `translate` helper expects a message that can be found in a domain (See Symfony translation documentation).
+         * A list of variables to be inserted into the translated message can be provided in the `variables` attribute
+         * in the shape of a JSON hash, here is a example:
+         * {"name": "John Doe", "age": "42"}
+         *
+         * @method _registerTranslate
+         * @protected
+         */
+        _registerTranslate: function () {
+            Y.Handlebars.registerHelper('translate', function (message, domain, variables) {
```
dpobel: the order of the function parameters is a bit weird, this should be `message, variables, domain` where `variables` is optional.
That way, you can write this like:

```js
Y.Handlebars.registerHelper('translate', function (message, variables, domain) {
    if ( !domain ) {
        domain = variables;
        variables = {};
    } else {
        variables = JSON.parse(variables);
    }
    return Y.eZ.Translator.trans(message, variables, domain);
});
```

yannickroger:
You can start reviewing @dpobel @StephaneDiot  @bdunogier @andrerom (also php)

sunpietro:
How are these translation files generated, like this one: `editorial.en.xlf`?

yannickroger:
This way: `php app/console translation:update --dump-messages --force --output-format=xlf --no-backup en eZPlatformUIBundle
`

A documentation will be provided.

bin/travis/prepare_behat.sh
```diff
@@ -10,7 +10,7 @@ export BRANCH_BUILD_DIR=$TRAVIS_BUILD_DIR TRAVIS_BUILD_DIR="$HOME/build/ezplatfo
 cd "$HOME/build"
 
 # Checkout meta repo, change the branch and/or remote to use a different ezpublish branch/distro
-git clone --depth 1 --single-branch --branch master https://github.com/ezsystems/ezplatform.git
+git clone --depth 1 --single-branch --branch bazinga https://github.com/ezsystems/ezplatform.git
```
bdunogier: Let's make sure we don't merge this please :)

dpobel:
Status: this is know (almost) ready to be merged in master. We just need to wait for https://github.com/ezsystems/ezplatform/pull/116 I think.

About @andrerom's comment:

> Note: It seems we are doing something wrong on translation files, we only give them locale names like en and fr, however the locales of the system is normally full form en_GB and fr_FR, which is why I had to somewhat hack it in 7669dd8. Also bazinga_jstranslation_js does not load locale fallback so had to specify that in url as well, hard coded for now.

I changed the template to use the full locale by default (ie `en_GB`) and to fallback on the configured locale in parameters.yml (`en` by default). That's a bit cleaner but this has several issues. First, the default locale comes from the matching siteaccess while PlatformUI is now supposed to not be tied to a siteaccess. Second we should probably implement several fallbacks for instance, if I have a site in `fr_CA`, I'd probably like to first fallback to `fr_FR` (or `fr`) and then maybe to `en`.

andrerom:
> Second we should probably implement several fallbacks for instance, if I have a site in fr_CA, I'd probably like to first fallback to fr_FR (or fr) and then maybe to en.

Something like one of these options pherhaps:
A. can the fallback for now reflect siteaccess language fallbacks?
B. To prepare for UI having drop down to select language, it should be selected language _(eg `fr_FR`)_ and then it's compressed form _(eg `fr`)_ and then our default system language _(eg `en`)_.

dpobel:
@andrerom A can easily be done, it's just a matter of building the `locales` parameter (see https://github.com/ezsystems/PlatformUIBundle/pull/525/files#diff-def2dab3cc9208048b5a2e96663033a7R61) according to a setting or anything (but maybe at some point, we should try to use the configured language in the browser)

andrerom:
> (but maybe at some point, we should try to use the configured language in the browser)

unless user has selected something yes.

PR #524, opened 2016-03-04
Run Bdd tests on Travis CI

dpobel:
Come on Travis!

requires https://github.com/ezsystems/ezplatform/pull/95 to have a router for the behat environment and more independent shell scripts to initialize the environment.

PR #526, opened 2016-03-08
EZP-25546: Pointer events block scrolling after loading is finished in Google Chrome

sunpietro:
https://jira.ez.no/browse/EZP-25546

Explicitly set `pointer-events` to `all` when the app is opened but in loading state.

### DIFF
--- a/Resources/public/css/layout.css
+++ b/Resources/public/css/layout.css
@@ -23,6 +23,7 @@
 
 .ez-platformui-app p {
     line-height: 1.4em;
+    pointer-events: auto;
 }
 
 .is-app-open {


sunpietro:
ping @dpobel

andrerom:
+1  (bug fix 1.2.x branch)

lserwatka:
@sunpietro with this patch we don't have to fix anything on the studio side, correct?

sunpietro:
@lserwatka yes, this patch fixes the blocker EZS-605 in eZ Studio.

sunpietro:
I found that, my proposal doesn't fix the issue on Safari. To fix scrolling issue there, I have to remove the `pointer-events: none;` property completely.

sunpietro:
Created a different approach only for Studio app: ezsystems/StudioUIBundle#462

lserwatka:
@andrerom @dpobel We have to make a decision soon. Either we fix it here or in StudioUI. What I understood, that fixing here (with proposed patch) is not enough for Safari browser. I'm more in favor to have a fix in StudioUI and create 1.2.1 tag there (fix for master and 1.2.x).

dpobel:
so I understand correctly, this patch fixes the issue in Chrome but not in Safari ? Did you try with `pointer-events: auto;` instead of `pointer-events: all;` ?

dpobel:
(according to https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events it seems `all` is dedicated to SVG element)

sunpietro:
@dpobel yes, I did
we could either remove the `pointer-events: none;` or override it like I did in the PR for Studio

lserwatka:
I can see one risk of not fixing it here. Other plugin devs might experience same issues.

dpobel:
I meant testing in this patch `pointer-events: auto;` (instead `of all`).

Resources/public/css/layout.css
```diff
@@ -29,6 +29,7 @@
     opacity: 1;
     min-height: 100%;
     overflow: auto;
+    pointer-events: all;
```
dpobel: I would put it on `.ez-platformui-app`

sunpietro: Well, where is it placed doesn't really matter. The `is-app-open` CSS class is added as soon as the Loading application screen disappears.

We can keep the `.is-app-loading` selector alone without `.is-app-open` at the beggining of the selector.
I did it this way to show, that `.is-app-loading` class is added when the app is opened but switching active views.

To sumarize, I'd keep it here.

dpobel: > Well, where is it placed doesn't really matter. The is-app-open CSS class is added as soon as the Loading application screen disappears.

for you maybe, for me (as a maintainer of PlatformUI) it matters. The opposite of 'is-app-loading' is not having 'is-app-loading', not 'is-app-open' which is a different concept.

> We can keep the .is-app-loading selector alone without .is-app-open at the beggining of the selector.

so please do NOT add unneeded changes. A bug fix is a fix with the minimum amount of change.

> I did it this way to show, that .is-app-loading class is added when the app is opened but switching active views.

which is an implementation detail irrelevant to the problem. *open* and *loading* are 2 different concepts which are not really related.

> To sumarize, I'd keep it here as it is.

I prefer to not answer to that...

sunpietro:
ping @dpobel

dpobel:
+1 (when Travis agrees)

sunpietro:
@dpobel behat is failing for some reason

dpobel:
@sunpietro let's ignore CI tests, I have absolutely no clue of what's going on there. Since [yesterday](https://github.com/ezsystems/PlatformUIBundle/pull/524), Travis runs the BDD tests, and there everything is up to date we can see what really happens.

dpobel:
@sunpietro also failing Travis but it seems completely unrelated to your patch. I'm restarting the job.

@andrerom @bdunogier any clue on that https://gist.github.com/dpobel/bdfe9769e41d20117511#file-travis-log-L1633 ?

bdunogier:
> @andrerom @bdunogier any clue on that https://gist.github.com/dpobel/bdfe9769e41d20117511#file-travis-log-L1633 ?

Really not, gotta admit... new one to me :(

lserwatka:
@dpobel @sunpietro is this something we can merge for 1.2.1 tag?

dpobel:

StephaneDiot:
+1

miguelcleverti:
@bdunogier yes the object manager will be removed

dpobel:
Replaced by https://github.com/ezsystems/PlatformUIBundle/pull/530

PR #529, opened 2016-03-15
[travis] Configured for 1.2 branch

bdunogier:
- Enable travis testing on 1.2
- Pull ezplatform 1.2.0 when testing on behat.


andrerom:
There is also self alias in [.ez.yml](https://github.com/ezsystems/PlatformUIBundle/blob/1.2_travis/.ez.yml#L7)

Which is maybe wrongly bumped in #480 for master to 1.2.

andrerom:
+1

PR #530, opened 2016-03-17
EZP-25546: Pointer events block scrolling after loading is finished in Google Chrome

dpobel:
Replaces https://github.com/ezsystems/PlatformUIBundle/pull/526


dpobel:
Travis failure is not related to the patch (memory limit issue with composer install in PHP 5.6).

PR #532, opened 2016-03-18
System info view integration

bdunogier:
> Proof of concept
> http://jira.ez.no/browse/EZP-25579

Integrates the view api being added to https://github.com/ezsystems/ez-support-tools/pull/8.
Customizes `SystemInfo` objects rendering so that they're formatted for pjax.

### TODO
- [ ] Add ezsystems/ez-support-tools to the requirements
- [ ] Remove old system info parts?

### DIFF
--- a/Resources/views/ez-support-tools/info/php.html.twig
+++ b/Resources/views/ez-support-tools/info/php.html.twig
@@ -1,0 +1,46 @@
+{% extends "eZPlatformUIBundle::pjax_admin.html.twig" %}
+
+{% trans_default_domain "systeminfo" %}
+
+{% block header_breadcrumbs %}
+    {% set breadcrumb_items = [
+    {link: path('admin_dashboard'), label: 'dashboard.title'|trans({}, 'dashboard')},
+    {link: '', label: 'system.information'|trans({}, 'systeminfo') },
+    ] %}
+
+    {{ parent() }}
+{% endblock %}
+
+{% block header_title %}
+    <h1 class="ez-page-header-name" data-icon="&#xe617">{{ 'system.information'|trans }}</h1>
+{% endblock %}
+
+{% block content %}
+    <section class="ez-tabs ez-serverside-content">
+        <ul class="ez-tabs-list">
+            <li class="ez-tabs-label is-tab-selected"><a href="#ez-tabs-platform">eZ Platform</a></li>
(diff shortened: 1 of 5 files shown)


Resources/views/ez-support-tools/info/php.html.twig
```diff
@@ -0,0 +1,46 @@
...
+    {% set breadcrumb_items = [
+    {link: path('admin_dashboard'), label: 'dashboard.title'|trans({}, 'dashboard')},
+    {link: '', label: 'system.information'|trans({}, 'systeminfo') },
+    ] %}
+
+    {{ parent() }}
+{% endblock %}
+
+{% block header_title %}
+    <h1 class="ez-page-header-name" data-icon="&#xe617">{{ 'system.information'|trans }}</h1>
```
crevillo: what is this ```&#xe617``` for?

bdunogier: I have no idea. I'm  just copy/pasting :-)

dpobel: it's the entity for the information icon (btw it should end with a `;`  but it seems to work still)

dpobel:
nice :)

glye:
Seems kosher.

Resources/views/ez-support-tools/info/php.html.twig
```diff
@@ -0,0 +1,46 @@
...
+{% endblock %}
+
+{% block content %}
+    <section class="ez-tabs ez-serverside-content">
+        <ul class="ez-tabs-list">
+            <li class="ez-tabs-label is-tab-selected"><a href="#ez-tabs-platform">eZ Platform</a></li>
+            <li class="ez-tabs-label"><a href="#ez-tabs-server">{{ 'server'|trans }}</a></li>
+        </ul>
+        <div class="ez-tabs-panels">
+            <div class="ez-tabs-panel" id="ez-tabs-server">
```
glye: The tab management stuff should go into a shared template, right? And then this template should only render the content of the tab. (The way I've been using the tabs feature before, I had to render all the tabs in the same document, and then JS decides which one to show.)

Though we may want to wait with perfecting these templates until we have the "infocollector registry" in place, and also the "report" feature (gather info from multiple collectors into a single page).

PR #534, opened 2016-03-26
EZP-25626: Add React.js to PlatformUI app and reuse it in Alloy Editor

sunpietro:
https://jira.ez.no/browse/EZP-25626

Added React.js directly into main app template file: `shell.html.twig`.

### DIFF
--- a/Resources/config/yui.yml
+++ b/Resources/config/yui.yml
@@ -9,8 +9,16 @@ system:
                 ez-capi:
                     requires: ['capi']
                     path: %ez_platformui.public_dir%/js/external/ez-capi.js
+                react:
+                    path: %ez_platformui.external_assets_public_dir%/vendors/react/react.js
+                react-dom:
+                    path: %ez_platformui.external_assets_public_dir%/vendors/react/react-dom.js
+                ez-react:
+                    requires: ['react', 'react-dom']
+                    path: %ez_platformui.public_dir%/js/external/ez-react.js
                 alloyeditor:
-                    path: %ez_platformui.external_assets_public_dir%/vendors/alloy-editor/dist/alloy-editor/alloy-editor-all.js
+                    requires: ['ez-react']
+                    path: %ez_platformui.external_assets_public_dir%/vendors/alloy-editor/dist/alloy-editor/alloy-editor-no-react.js
                 ez-alloyeditor:
                     requires: ['alloyeditor']
                     path: %ez_platformui.public_dir%/js/external/ez-alloyeditor.js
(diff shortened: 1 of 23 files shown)


sunpietro:
ping @dpobel 
 I found that adding React.js by creating YUI3 module makes AlloyEditor unstable. Sometimes the editor files are loaded before React.js sometimes they are loaded loaded after.

sunpietro:
It's related to: ezsystems/PlatformUIAssetsBundle#12

sunpietro:
ping @dpobel

sunpietro:
@dpobel I don't know why behat tests are failing.

Resources/public/js/external/ez-alloyeditor.js
```diff
@@ -9,6 +9,4 @@ YUI.add('ez-alloyeditor', function (Y) {
     Y.namespace('eZ');
 
     Y.eZ.AlloyEditor = AlloyEditor;
-    Y.eZ.React = Y.eZ.AlloyEditor.React;
```
dpobel: removing those lines is BC break, so we need to keep the definition of `Y.eZ.React` and `Y.eZ.ReactDOM` somehow.

dpobel:
> I found that adding React.js by creating YUI3 module makes AlloyEditor unstable. Sometimes the editor files are loaded before React.js; sometimes they are loaded loaded after it.

I guess you added a `react` module in yui.yml pointing to `bundles/ezplatformuiassets/vendors/react/react.js` and you added this module as a dependency to `alloyeditor` or similar approach. And indeed, this might not work because by default the YUI Loader loads the file asynchronously and then for files which content is not wrapped in a `YUI.add`, the code might not be executed in the expected order (in your case alloy-editor-no-react.js could be evaluated before react.js is fully loaded).

Maybe to avoid that, you could try to adopt a similar strategy as for the REST Client or AlloyEditor itself. Basically, you'll add a `react` module to yui.yml pointing to `bundles/ezplatformuiassets/vendors/react/react.js` and then you'll also add a `ez-react` module which requires `react` (and defines Y.eZ.React and Y.eZ.ReactDOM see https://github.com/ezsystems/PlatformUIBundle/pull/534/files#r57878901). Then `ez-react` should be added as a dependency of `alloyeditor`. Then if a component needs React alone, then it should depend on `ez-react`.

Resources/public/js/external/ez-alloyeditor.js
andrerom: but we can deprecate and remove in v2 if needed

sunpietro:
ping @dpobel

PR #535, opened 2016-03-28
Move browser methods into PlatformUI Context

miguelcleverti:
This PR moves some of the methods from the [behatbundle](https://github.com/ezsystems/BehatBundle) Contexts into the PlatformUI Contexts.
The main goal is to decrease the dependencies of the Behat BDD tests between the repository, this way the code should be easier to manage and maintain.


miguelcleverti:
ping @bdunogier

Features/Context/SubContext/CommonActions.php
```diff
@@ -12,6 +12,91 @@
 trait CommonActions
 {
     /**
+     * @Given I clicked on/at (the) :link link
+     * @When  I click on/at (the) :link link
+     *
+     * Click a link with text ':link'
+     */
+    public function iClickAtLink($link)
```
bdunogier: Could we please take advantage of this PR to remove this  "atLink" ? You don't click at something, but "on" something. We can keep the alternative in the sentences.

This is also valid for other methods in the same file.

Features/Context/Override/CommonActions.php
```diff
@@ -1,79 +0,0 @@
...
-/**
- * File containing Overriden Functions for context PlatformUI.
- *
- * @copyright Copyright (C) eZ Systems AS. All rights reserved.
- * @license For full copyright and license information view LICENSE file distributed with this source code.
- * @version //autogentag//
- */
-namespace EzSystems\PlatformUIBundle\Features\Context\Override;
-
-trait CommonActions
```
bdunogier: Don't remove, deprecate (at least phpdoc, but I'd also add `trigger_error` calls. Example: `trigger_error(E_USER_DEPRECATED, "The common actions trait is deprecated, use XXX instead")`.

Features/Context/SubContext/CommonActions.php
```diff
@@ -12,6 +12,91 @@
 trait CommonActions
```
bdunogier: Why do we keep using the trait ?

Features/Context/SubContext/CommonActions.php
bdunogier: (can be a follow-up. I need to stick to my own advice)

Features/Context/Override/CommonActions.php
miguelcleverti: This trait only had overriden method that were moved into the `Features/Context/SubContext/CommonActions.php` it causes no breaks

Features/Context/SubContext/CommonActions.php
miguelcleverti: we devided to not remov this one because it was really simple, just a collection of basic sentences. The alternative is to put all this methods inside PlatformUI Context class...

bdunogier:
+1, but Travis needs to pass.

Opened https://github.com/ezsystems/PlatformUIBundle/pull/539 that might fix the Travis errors on master.

Once it passes, rebase this PR, and if it passes, let's merge.

miguelcleverti:
@bdunogier something failed but it does not seem related to Behat...

Features/Context/SubContext/CommonActions.php
bdunogier: I'd really prefer if we could aim at keeping those methods in dedicated contexts, in order to sandbox things. But not now.

bdunogier:
@miguelcleverti now that memory limit errors are fixed on travis, can you rebase on the latest master and force push this ?

PR #536, opened 2016-03-30
say hello to 2016

crevillo:
ping @dpobel @yannickroger 

just a change in the copyright year. let me know if you need a issue in the tracker for it.


crevillo:
build is failing, but i don't think the fail is related with this change...

dpobel:
@crevillo you are right

+1

PR #537, opened 2016-04-01
Fix EZP-25642: Change "Send to Trash Button" to "Delete" button for users

crevillo:
Fixes: https://jira.ez.no/browse/EZP-25642

> Status: work-in-progress

I will need some guide to finish this properly. This tries to delete content insteand of sending content to trash when the content is user type. i have some questions about the code i've written specially for the the testing part. 

Is that the right way to check this? I mean, i'm duplicating lot of code there. Is there any way to extend tests here like we do with phpunit? 

Second. Do we need another icon for deleting or the trash one is enough?

##### tasks
- [x] Add delete or sendToTrash Button depending on content Type
- [x] Create content delete action
- [ ] Add icon for delete action

##### tests
manual and unit tests > in progress
 

ping @bdunogier @dpobel @StephaneDiot @yannickroger

### DIFF
--- a/Resources/public/js/views/services/ez-locationviewviewservice.js
+++ b/Resources/public/js/views/services/ez-locationviewviewservice.js
@@ -25,6 +25,7 @@ YUI.add('ez-locationviewviewservice', function (Y) {
         initializer: function () {
             this.on('*:editAction', this._editContent);
             this.on('*:sendToTrashAction', this._sendContentToTrashConfirmBox);
+            this.on('*:deleteContentAction', this._confirmDeleteContent);
             this.on('*:moveAction', this._selectLocation);
             this.on('*:translateContent', this._translateContent);
             this.on('*:sortUpdate', this._updateSorting);
@@ -96,6 +97,24 @@ YUI.add('ez-locationviewviewservice', function (Y) {
         },
 
         /**
+         * `deleteContentAction` event handler,
+         * it asks confirmation to the user before delete the content item.
+         *
+         * @method _confirmDeleteContent
+         * @protected
+         * @param {Object} e event facade of the deleteAction event
+         */
+        _confirmDeleteContent: function (e) {
+            e.preventDefault();
+            this.fire('confirmBoxOpen', {
+                config: {
+                    title: "Are you sure you want to delete this content?",
+                    confirmHandler: Y.bind(this._deleteContent, this),
+                },
+            });
+        },
+
+        /**
          * moveAction event handler, launch the universal discovery widget
          * to choose a location to move the content
          *
@@ -140,6 +159,27 @@ YUI.add('ez-locationviewviewservice', function (Y) {
         },
 
         /**
+         * Deletes the the content item, triggering loading parent location and notifications
+         *
+         * @method _deleteContent
+         * @protected
+         */
+        _deleteContent: function () {
+            var path = this.get('path'),
+                content = this.get('content'),
+                parentLocation = path[path.length - 1];
+
+            this._notify(
+                'Deleting "' + content.get('name') + '"',
+                'delete-' + content.get('id'),
+                'started',
+                0
+            );
+
+            content.delete({api: this.get('capi')}, Y.bind(this._afterDeleteCallback, this, parentLocation, content));
+        },
+
+        /**
          * Send to trash callback triggering notifications and making app to navigate to parent location
          *
          * @method _afterSendToTrashCallback
@@ -178,6 +218,50 @@ YUI.add('ez-locationviewviewservice', function (Y) {
              * @param {eZ.Location} location
              */
             this.fire('sentToTrash', {location: location});
+
+            app.navigateTo('viewLocation', {
+                id: parentLocation.get('id'),
+                languageCode: content.get('mainLanguageCode')
+            });
+        },
+
+        /**
+         * Delete content item callback triggering notifications and making app to navigate to parent location
+         *
+         * @method _afterDeleteCallback
+         * @protected
+         * @param {eZ.Location} parentLocation the parent location to which app will navigate to
+         * @param {eZ.Content} content the content item that has been deleted
+         * @param {Boolean} error
+         */
+        _afterDeleteCallback: function (parentLocation, content, error) {
+            var app = this.get('app'),
+                contentName = content.get('name');
+
+            if (error) {
+                this._notify(
+                    'An error occurred when deleting "' + contentName + '"',
+                    'delete-' + content.get('id'),
+                    'error',
+                    0
+                );
+                return;
+            }
+
+            this._notify(
+                '"' + contentName + '" was deleted',
+                'delete-' + content.get('id'),
+                'done',
+                5
+            );
+
+            /**
+             * Fired when the content has been deleted
+             *
+             * @event deletedContent
+             * @param {eZ.Content} content
+             */
+            this.fire('deletedContent', {content: content});
 
             app.navigateTo('viewLocation', {
                 id: parentLocation.get('id'),
--- a/Tests/js/models/assets/ez-contentmodel-tests.js
+++ b/Tests/js/models/assets/ez-contentmodel-tests.js
@@ -3,7 +3,7 @@
  * For full copyright and license information view LICENSE file distributed with this source code.
  */
 YUI.add('ez-contentmodel-tests', function (Y) {
-    var modelTest, relationsTest, createContent, loadResponse, copyTest,
+    var modelTest, relationsTest, createContent, deleteContent, loadResponse, copyTest,
         loadLocationsTest, addLocationTest, setMainLocationTest, hasTranslationTest,
         Assert = Y.Assert,
         Mock = Y.Mock;
@@ -639,6 +639,68 @@ YUI.add('ez-contentmodel-tests', function (Y) {
         },
     });
 
+    deleteContent = new Y.Test.Case({
+        name: "eZ Content Model delete content tests",
+
+        setUp: function() {
+            this.model = new Y.eZ.Content();
+            this.contentId = 'Pele';
+
+            this.capi = new Mock();
+            this.contentService = new Mock();
+
+            Y.Mock.expect(this.capi, {
+                method: 'getContentService',
+                returns: this.contentService
+            });
+
+            Y.Mock.expect(this.model, {
+                method: 'get',
+                args: ['id'],
+                returns: this.contentId
+            });
+        },
+
+        tearDown: function () {
+            this.model.destroy();
+            delete this.model;
+            delete this.capi;
+            delete this.contentService;
+        },
+
+        "Should delete the content": function () {
+            var options = {api: this.capi},
+                callback = function () {},
+                content = this.model;
+
+            Y.Mock.expect(this.contentService, {
+                method: 'deleteContent',
+                args: [this.contentId, callback],
+            });
+
+            content.delete(options, callback);
+        },
+
+        "Should handle the error while deleting a content": function () {
+            var content = this.model,
+                that = this,
+                err = new Error();
+
+            Y.Mock.expect(this.contentService, {
+                method: 'deleteContent',
+                args: [this.contentId, Y.Mock.Value.Function],
+                run: function (struct, callback) {
+                    callback(err, that.createResponse);
+                }
+            });
+            content.delete({
+                api: this.capi
+            }, function (error, response) {
+                Assert.areSame(err, error, "The CAPI error should be provided");
+            });
+        },
+    });
+
     loadLocationsTest = new Y.Test.Case({
         name: "eZ Content Model load locations tests",
 
@@ -998,6 +1060,7 @@ YUI.add('ez-contentmodel-tests', function (Y) {
     Y.Test.Runner.add(modelTest);
     Y.Test.Runner.add(relationsTest);
     Y.Test.Runner.add(createContent);
+    Y.Test.Runner.add(deleteContent);
     Y.Test.Runner.add(copyTest);
     Y.Test.Runner.add(loadLocationsTest);
     Y.Test.Runner.add(addLocationTest);
--- a/Tests/js/views/assets/ez-actionbarview-tests.js
+++ b/Tests/js/views/assets/ez-actionbarview-tests.js
@@ -4,7 +4,8 @@
  */
 YUI.add('ez-actionbarview-tests', function (Y) {
     var Assert = Y.Assert,
-        viewTest;
+        Mock = Y.Mock,
+        viewTest, userViewTest;
 
     viewTest = new Y.Test.Case({
         name: "eZ Action Bar View test",
@@ -12,7 +13,7 @@ YUI.add('ez-actionbarview-tests', function (Y) {
         setUp: function () {
             this.location = {};
             this.content = {};
-            this.contentType = {};
+            this.contentTypeMock = new Mock();
 
             Y.eZ.CreateContentActionView = Y.Base.create('createContentActionView', Y.eZ.ButtonActionView, [Y.eZ.Expandable], {}, {
                 contentTypeGroups: {},
@@ -27,16 +28,23 @@ YUI.add('ez-actionbarview-tests', function (Y) {
                 location: {}
             });
 
+            Mock.expect(this.contentTypeMock, {
+                method: 'hasFieldType',
+                args: ['ezuser'],
+                returns: false
+            });
+
             this.view = new Y.eZ.ActionBarView({
                 location: this.location,
                 content: this.content,
-                contentType: this.contentType
+                contentType: this.contentTypeMock
             });
         },
 
         tearDown: function () {
             delete Y.eZ.CreateContentActionView;
             delete Y.eZ.TranslateActionView;
+            delete Y.eZ.MoveContentActionView;
             this.view.destroy();
         },
 
@@ -60,6 +68,19 @@ YUI.add('ez-actionbarview-tests', function (Y) {
             Assert.isTrue(
                 !!actionFound,
                 'The ActionBarView should contain ' + actionId + ' action'
+            );
+        },
+
+        _isNotActionCreated: function (actionId, testedAttributes) {
+            var actionFound;
+
+            actionFound = Y.Array.find(this.view.get('actionsList'), function (actionView) {
+                return actionView.get('actionId') === actionId;
+            });
+
+            Assert.isFalse(
+                !!actionFound,
+                'The ActionBarView should not contain ' + actionId + ' action'
             );
         },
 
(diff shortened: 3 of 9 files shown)


crevillo:
added a new test to check delete action when the action bar is for a user. Let's see what travis has to say now...

bdunogier:
Also fails on master, with the same error. It needs to be fixed (remove memlimit), but it's not because of your code.

bdunogier:
Could you please edit the PR's title ? Work-in-progress doesn't have to be in the title, it is enough to add it in the description's head.

crevillo:
isn't it in the description head now?

Tests/js/models/assets/ez-contentmodel-tests.js
```diff
@@ -639,6 +639,72 @@ YUI.add('ez-contentmodel-tests', function (Y) {
...
+            delete this.contentService;
+        },
+
+        "Should delete the content": function () {
+            var options = {api: this.capi},
+                content = this.model;
+
+            Y.Mock.expect(this.contentService, {
+                method: 'deleteContent',
+                args: [this.contentId, Y.Mock.Value.Function],
```
dpobel: you don't need 2 tests for this method. You only have to check that the delete `callback` is passed to the content service `deleteContent` method. you can write something like:

```js
var options = {api: this.capi},
    callback = function () {},
    content = this.model;

Y.Mock.expect(this.contentService, {
    method: 'deleteContent',
    args: [this.contentId, callback],
});
content.delete(options, callback);
```

Tests/js/views/assets/ez-actionbarview-tests.js
```diff
@@ -87,8 +107,126 @@ YUI.add('ez-actionbarview-tests', function (Y) {
...
+        },
+
+        "Should instantiate deleteActionView": function () {
+            this._isActionCreated('delete', {});
+        },
     });
 
     Y.Test.Runner.setName("eZ Action Bar View tests");
-    Y.Test.Runner.add(viewTest);
+    Y.Test.Runner.add(viewTest, userViewTest);
```
dpobel: `add` only accepts one parameter so the `userViewTest` is completely ignored, you should add `Y.Test.Runner.add(userViewTest);` instead

Tests/js/views/assets/ez-actionbarview-tests.js
```diff
@@ -87,8 +107,126 @@ YUI.add('ez-actionbarview-tests', function (Y) {
...
+            this.view = new Y.eZ.ActionBarView({
+                location: this.location,
+                content: this.content,
+                contentType: this.contentTypeMock
+            });
+        },
+
+        tearDown: function () {
+            delete Y.eZ.CreateContentActionView;
+            delete Y.eZ.TranslateActionView;
```
dpobel: missing `delete Y.eZ.MoveContentActionView`

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -96,6 +97,26 @@ YUI.add('ez-locationviewviewservice', function (Y) {
         },
 
         /**
+         * `deleteAction` event handler,
+         * it asks confirmation to the user before delete the location.
```
dpobel: `it asks confirmation to the user before deleting the content item.`

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -96,6 +97,26 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+         *
+         * @method _deleteConfirmBox
+         * @protected
+         * @param {Object} e event facade of the deleteAction event
+         */
+        _deleteContentConfirmBox: function (e) {
```
dpobel: I would call it `_confirmDeleteContent`

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -96,6 +97,26 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+            e.preventDefault();
+            this.fire('confirmBoxOpen', {
+                config: {
+                    title: "Are you sure you want to delete this content?",
+                    confirmHandler: Y.bind(function () {
```
dpobel: nitpick: since the `_delete` parameter is not used, you can even write:
```js
confirmHandler: Y.bind(this._delete, this),
```

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -140,6 +161,30 @@ YUI.add('ez-locationviewviewservice', function (Y) {
         },
 
         /**
+         * Delete the location, triggering loading parent location and notifications
+         *
+         * @method _delete
+         * @protected
+         */
+        _delete: function () {
```
dpobel: I would call it `_deleteContent`

Tests/js/views/assets/ez-actionbarview-tests.js
crevillo: @dpobel: Ok. But this was a copy paste from the other test.should we fix it in the other too?

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -140,6 +161,30 @@ YUI.add('ez-locationviewviewservice', function (Y) {
         },
 
         /**
+         * Delete the location, triggering loading parent location and notifications
```
dpobel: `Delete the content item`

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -140,6 +161,30 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+         *
+         * @method _delete
+         * @protected
+         */
+        _delete: function () {
+            var that = this,
+                location = this.get('location'),
```
dpobel: `that` is not needed.
The `location` and `locationId` are not really needed as well, you can use the content id in the notification identifier.

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -186,6 +231,52 @@ YUI.add('ez-locationviewviewservice', function (Y) {
         },
 
         /**
+         * Send to trash callback triggering notifications and making app to navigate to parent location
```
dpobel: `Delete content item callback`

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -186,6 +231,52 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+         *
+         * @method _afterDeleteCallback
+         * @protected
+         * @param {eZ.Location} parentLocation the parent location to which app will navigate to
+         * @param {eZ.Content} content the content to be trashed
```
dpobel: `@param {eZ.Content} content the content item that has been deleted`

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -186,6 +231,52 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+                    'An error occurred when deleting "' + contentName,
+                    'delete-' + locationId,
+                    'error',
+                    0
+                );
+                return;
+            }
+
+            this._notify(
+                '"' + contentName + '" deleted',
```
dpobel: `'"' + contentName + '" was deleted',`

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -140,6 +161,30 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+        _delete: function () {
+            var that = this,
+                location = this.get('location'),
+                locationId = location.get('id'),
+                path = this.get('path'),
+                content = this.get('content'),
+                parentLocation = path[path.length - 1];
+
+            this._notify(
+                'Deleting "' + content.get('name'),
```
dpobel: missing ` + '"'` at the end of the string

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -186,6 +231,52 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+         */
+        _afterDeleteCallback: function (parentLocation, content, error) {
+            var app = this.get('app'),
+                location = this.get('location'),
+                locationId = location.get('id'),
+                contentName = content.get('name');
+
+            if (error) {
+                this._notify(
+                    'An error occurred when deleting "' + contentName,
```
dpobel: missing ` + '"'` at the end of the string

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -186,6 +231,52 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+                5
+            );
+
+            /**
+             * Fired when the content deleted
+             *
+             * @event sentToTrash
+             * @param {eZ.Location} location
+             */
+            this.fire('delete', {location: location});
```
dpobel: that's a bit too generic. I would call it `deleteContent` and I would also add the `content` to the event parameter. (also the `@event` in the doc is wrong)

dpobel:
nice start @crevillo :)

> Is that the right way to check this? I mean, i'm duplicating lot of code there. Is there any way to extend tests here like we do with phpunit? 

I guess you are talking about `_isNotActionCreated` and `_isActionCreated` in action bar view tests ? In such case, you can write something like:

```js
var viewTest,
    // others vars...
    isNotActionCreated = function () {
        // code here
    };

viewTest = new Y.Test.Case({
   name: "eZ Action Bar View test",
   setUp: function() { /* code */ },

   _isNotActionCreated: isNotActionCreated,

   // test using _isNotActionCreated
});
```

> Second. Do we need another icon for deleting or the trash one is enough?

I guess we can keep the current one for now, that's really a detail.

More important:

* the code you added to `eZ.LocationViewViewService` should be tested
* The `deleteContent` event is actually useful to make sure the tree does not display the content that was deleted so you should add it in https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/apps/plugins/ez-updatetreeplugin.js#L26 (with the corresponding test)

crevillo:
Thanks @dpobel, for the double check i was talking about how to check an action bar and a specific bar for users. Wondering if one test could extend another or something.

Thanka for the feedback. I Will take care of it

Tests/js/views/assets/ez-actionbarview-tests.js
dpobel: yes

Resources/public/js/views/services/ez-locationviewviewservice.js
crevillo: @dpobel: so, should i strip the `ConfirmBox` part?

crevillo:
> The deleteContent event is actually useful to make sure the tree does not display the content that was deleted so you should add it in

the name of the event should be deleteContent or delete*d* content? asking because in the already defined ones we have `movedContent` or `publishedDraft`

crevillo:
@dpobel: Applied your comments and squashed everything in one commit. Let me know if I forgot something.

crevillo:
Build failing, but again looks because composer problems. 
re-ping @dpobel @StephaneDiot @yannickroger

dpobel:
you can ignore those failures since they are completely unrelated to your change and @bdunogier fixed that lately anyway in master and 1.2 branches.
(I'm reviewing your changes)

Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
```diff
@@ -1368,10 +1368,520 @@ YUI.add('ez-locationviewviewservice-tests', function (Y) {
...
+                }
+            });
+
+            Mock.expect(parentLocation, {
+                method: 'get',
+                args: ['id'],
+                returns: 'alan-shearer'
+            });
+
+            Mock.expect(this.location, {
```
dpobel: that's not needed but I guess this is a left over of another test

Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
```diff
@@ -1368,10 +1368,520 @@ YUI.add('ez-locationviewviewservice-tests', function (Y) {
...
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/537/files#r58863220

dpobel: and same following tests

crevillo: i think i've copied all over the place... but don't you need it to get the `parentLocationId` and navigate to it?

Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
```diff
@@ -1368,10 +1368,520 @@ YUI.add('ez-locationviewviewservice-tests', function (Y) {
...
+                parentLocation = new Mock(),
+                that = this,
+                contentName = 'pierlugi-collina',
+                contentId = 'aContentId',
+                eventFired = false;
+
+            this.service.set('path', [{location: parentLocation}]);
+            this.service.set('content', content);
+
+            Mock.expect(content, {
```
dpobel: this mock configuration and the one for `parentLocation` are the same and repeated in every test, you should put that in one or 2 dedicated methods.

Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
dpobel: this is configuring the Location mock to have a `deleteContent` method which does not exist in the original Location Model so this method never called anywhere. If you remove that, the test still passes.
Note: the `parentLocationId` is not retrieved anywhere, the view service already has the path to the Location being displayed, so the parent Location is taken from it and that's the point of the line `this.service.set('path', [{location: parentLocation}]);` a few lines above.

dpobel:
besides the last comment, that's good to me.
+1
ping @yannickroger

crevillo:
@dpobel, if i remove 
```js
 Mock.expect(parentLocation, {
     method: 'get',
     args: ['id'],
     returns: 'alan-shearer'
});
```
i get following errors running the tests...
```
✖ [eZ Location View View Service tests]: Passed: 34 Failed: 3 Total: 37 (ignored 0) (0.527 seconds)
    Should notify user about success of deleting the content item
       Unexpected error: 'undefined' is not a function (evaluating 'parentLocation.get('id')')
    Should fire deletedContent event after deleting a content item
       Unexpected error: 'undefined' is not a function (evaluating 'parentLocation.get('id')')
    Should navigate the app to view of parent location
       Unexpected error: 'undefined' is not a function (evaluating 'parentLocation.get('id')')
```

dpobel:
it seems I was unclear, sorry. You should remove only the following mock configuration:
```js
            Mock.expect(this.location, {
                method: 'deleteContent',
                args: [Mock.Value.Object, Mock.Value.Function],
                run: function (options, callback){
                    callback(true);
                }
            });
```

the other comment is about putting the common mock configuration in a dedicated method to not repeat that all over the test.

yannickroger:
+1
Cool to see you on the JS side of the force :smile:

crevillo:
ok, i'll try. thanks.

crevillo:
thanks @yannickroger, i think is not bad for my first js code in almost 7 years :dancer:

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -186,6 +226,50 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+         * @param {Boolean} error
+         */
+        _afterDeleteCallback: function (parentLocation, content, error) {
+            var app = this.get('app'),
+                contentName = content.get('name');
+
+            if (error) {
+                this._notify(
+                    'An error occurred when deleting "' + contentName + '"',
+                    'content-' + content.get('id'),
```
StephaneDiot: Shouldn't it be `'delete-' + content.get('id')` here. Or your 'started' notification will not be updated and disapear.

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -186,6 +226,50 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+                    'error',
+                    0
+                );
+                return;
+            }
+
+            this._notify(
+                '"' + contentName + '" was deleted',
+                'content-' + content.get('id'),
```
StephaneDiot: same here

Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
```diff
@@ -1368,10 +1368,520 @@ YUI.add('ez-locationviewviewservice-tests', function (Y) {
...
+                args: [Mock.Value.String],
+                run: function (attr) {
+                    if (attr === "name") {
+                        return contentName;
+                    } else if (attr === "id") {
+                        return contentId;
+                    } else {
+                        Y.fail("Unexpected parameter for content mock");
+                    }
+
```
StephaneDiot: niptick: unnecessary empty line

Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
```diff
@@ -1368,10 +1368,520 @@ YUI.add('ez-locationviewviewservice-tests', function (Y) {
...
```
StephaneDiot: same niptick

StephaneDiot:
besides last comments +1, and good job ! For the icon thing, I don't know if you had your answer but you should ask to @dpobel .

Resources/public/js/views/services/ez-locationviewviewservice.js
crevillo: you're right @StephaneDiot. My fault. i'll add these and blank lines fixes to last changes proposed by @dpobel. 
For the icon part, yep, @dpobel suggested me to use the same icon as the send to trash for now. 
Thanks for your input!.

crevillo:
@dpobel @StephaneDiot @yannickroger , added last comments
@dpobel: i finally added a `_initializeMocks` method that always do almost the same except the `true` or `false` for the content delete part. that's why i added another argument `callbackArgument`
maybe naming of those needs to be improved... suggestions?
thanks!.

yannickroger:
Maybe `error` ? As if `true` it will error.

crevillo:
you mean the callbackArgument param?

dpobel:
looks good to me. +1

(TravisCI failure does not seem to be related to your change, I guess it's one of our random failure... I restarted the job)

PR #539, opened 2016-04-04
[travis] disabled xdebug

bdunogier:
Travis is failing on [master](https://travis-ci.org/ezsystems/PlatformUIBundle/builds/119515363) because of memory limit on composer install.

Disabling xdebug might be sufficient. If not, we will have to disable memory_limit on the composer command line.

PR #540, opened 2016-04-05
Travis memory errors

bdunogier:
> 1.2 version of https://github.com/ezsystems/PlatformUIBundle/pull/539 (passed on master)
> Approved "if travis passes"

Travis is failing on [master](https://travis-ci.org/ezsystems/PlatformUIBundle/builds/119515363) because of memory limit on composer install.

This change disables xdebug, and runs composer without a memory limit (many alternatives were tried until that solution was reached. it works, unlike the many alternatives).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,9 @@ matrix:
         - php: 7.0
           env: BEFORE="./bin/travis/prepare_behat.sh" TEST_CMD="./bin/travis/runbehat.sh" AFTER_SUCCESS='echo "After success"'
         - php: 5.6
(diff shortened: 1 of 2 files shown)


andrerom:
I should have linked to it earlier, but assumed you might remember how we fixed in in solr bundle:
https://github.com/ezsystems/ezplatform-solr-search-engine/blob/master/.travis.yml#L27

bdunogier:
Damn... i've checked at least 3 repos, but not this one.

Even weirder is that I did test with phpenv (with an extra config file, but it should still have worked).

bdunogier:
What is this `travis_retry` command btw ?

andrerom:
Google knows ;) But basically a command _(afaik only available in .travis.yml btw, never got it to work in scripts)_ that retries another one 3 times, for use on commands that sometimes might get net timeouts.

PR #542, opened 2016-04-11
Fix EZP-25653: Use eZSupportTools in PlatformUI

glye:
> Implement https://jira.ez.no/browse/EZP-25653
> Status: Ready for review

- [x] Implemented SystemInfo view pjax config & templates, adapted controller
- [x] Deprecated SystemInfoHelper interface and class, not needed after EZP-25653

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,6 +11,7 @@
         "ezsystems/platform-ui-assets-bundle": "^2.0.0",
         "ezsystems/repository-forms": "^1.0",
         "ezsystems/ezpublish-kernel": "^6.2",
+        "ezsystems/ez-support-tools": "^0.1-dev",
         "zetacomponents/system-information": "^1.1"
     },
     "require-dev": {
--- a/Controller/SystemInfoController.php
+++ b/Controller/SystemInfoController.php
@@ -10,24 +10,17 @@ namespace EzSystems\PlatformUIBundle\Controller;
 
 use eZ\Publish\Core\MVC\Symfony\Security\Authorization\Attribute;
 use Symfony\Component\HttpFoundation\Response;
-use EzSystems\PlatformUIBundle\Helper\SystemInfoHelperInterface;
 
 class SystemInfoController extends Controller
 {
     /**
-     * @var \EzSystems\PlatformUIBundle\Helper\SystemInfoHelperInterface
+     * @var \EzSystems\EzSupportToolsBundle\SystemInfo\SystemInfoCollectorRegistry
      */
-    protected $systemInfoHelper;
+    private $collectorRegistry;
 
-    /**
-     * @var string
-     */
-    private $installDir;
-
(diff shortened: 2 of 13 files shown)


glye:
@andrerom Would the travis failure "Could not open input file: bin/behat" be resolved by adding behat to require-dev? (I don't think the PR itself can be the cause of this.)

composer.json
```diff
@@ -11,6 +11,7 @@
         "ezsystems/platform-ui-assets-bundle": "^2.0.0",
         "ezsystems/repository-forms": "^1.0",
         "ezsystems/ezpublish-kernel": "^6.2",
+        "ezsystems/ez-support-tools": "^0.1-dev",
```
bdunogier: `^0.1@dev`, not dash dev.

bdunogier:
Looks okay.

About behat failing:
1. The version requirement in composer.json is wrong (see comment)
2. ezsystems/ezplatform needs to be configured for this to work: load bundle in `app/AppKernel.php`, load support tools routes from `app/config/routing.yml`

The easiest would be to open branch/PR against ezplatform that configures those. They will have to be configured for this to work in any case. Then in the platform-ui PR, you can change the branch at https://github.com/ezsystems/PlatformUIBundle/blob/master/bin/travis/prepare_behat.sh#L10 to yours, so that your version is used for the install.

bdunogier:
P.S. A hint: in your ezplatform branch, set the requirements for support tools and platformui using `dev-branchname as vX.Y.Z-dev`, where x.y.z is the current dev version for each package. Then run composer update, and see if it installs / works as you expect it to.

(something like `"ezsystems/ez-support-tools": "v0.1.0-dev"` and `"ezsystems/platform-ui-bundle": "v1.3.0-dev"`)

composer.json
andrerom: If ez-support-tool is not meant to make it to 1.0.0-beta until next monday, then I suggest we hold this PR back until after 16.04 freeze.

glye:
Continued in https://github.com/ezsystems/PlatformUIBundle/pull/545

bdunogier:
Why did you close this one ?

If the goal was to rename `ezp-25653_use_ezsupporttools_in_platformui` to `dev-ezp-25653_use_ezsupporttools_in_platformui`, then I guess I've misled you: `dev-ezp-25653_use_ezsupporttools_in_platformui` is the way to reference a branch with composer. With the new name, the branch now needs to be referenced as `dev-dev-ezp-25653_use_ezsupporttools_in_platformui`.

glye:
Yes, I mistook it as a branch naming convention :)

composer.json
bdunogier: I'd really prefer having it as 0.1 than 1.0.0-beta.

What is the problem with 0.x ?

PR #546, opened 2016-04-14
EZP-25461: Display all versions of a content item

yannickroger:
Link: https://jira.ez.no/browse/EZP-25461

## Description
Adds a new tab to display version information.

## Screenshot
versions

## Tests
Manual and update unit test


Resources/public/js/models/ez-versioninfomodel.js
```diff
@@ -0,0 +1,186 @@
...
+        },
+
+        /**
+         * Checks whether the version is a draft
+         *
+         * @method isDraft
+         * @return {Boolean}
+         */
+        isDraft: function () {
+            return (this.get('status') === 'DRAFT');
```
crevillo: isn't there a constant for this?

Resources/public/js/views/tabs/ez-locationviewversionstabview.js
```diff
@@ -0,0 +1,142 @@
...
+        },
+
+        render: function () {
+            var container = this.get('container'),
+                draft = [],
+                published = [],
+                archived = [];
+
+            Y.Array.each(this.get('versions'), function (version) {
+                var status = version.get('status');
```
crevillo: so, `version.get('status')` returns strings? i though it would return integers at first :)

Resources/public/js/models/ez-contentmodel.js
```diff
@@ -278,6 +278,29 @@ YUI.add('ez-contentmodel', function (Y) {
...
+         * @param {Object} options
+         * @param {Object} options.api (required) the JS REST client instance
+         * @param {Function} callback
+         */
+        loadVersions: function (options, callback) {
+            var versions = [],
+                contentService = options.api.getContentService();
+
+            contentService.loadVersions(this.get('id'), function (error, response) {
+                Y.Array.each(response.document.VersionList.VersionItem, function (versionItemHash) {
```
sunpietro: You don't have to use `Y.Array.each` you can use native `forEach` function.

andrerom: +1 :)

yannickroger:
Ready for review @dpobel @StephaneDiot

Resources/public/js/views/tabs/ez-locationviewversionstabview.js
```diff
@@ -0,0 +1,140 @@
...
+                published = [],
+                archived = [],
+                hasDraft = false,
+                hasPublished = false,
+                hasArchived = false;
+
+            Y.Array.each(this.get('versions'), function (version) {
+                var status = version.get('status');
+
+                if(status === 'ARCHIVED') {
```
StephaneDiot: missing space ! :smiling_imp:

Resources/public/js/views/tabs/ez-locationviewversionstabview.js
```diff
@@ -0,0 +1,140 @@
...
+        render: function () {
+            var container = this.get('container'),
+                draft = [],
+                published = [],
+                archived = [],
+                hasDraft = false,
+                hasPublished = false,
+                hasArchived = false;
+
+            Y.Array.each(this.get('versions'), function (version) {
```
dpobel: splitting the versions by status is a tad too much logic for a render method. I think you should split the versions by status in `eZ.Content#loadVersions` so that its callback directly receives an object where the versions are already sorted out.

Here you would just have to iterate over the 3 arrays to call `toJSON` and maybe to compute the `has*` flags.

Resources/public/js/views/ez-locationviewview.js
```diff
@@ -336,6 +336,14 @@ YUI.add('ez-locationviewview', function (Y) {
                             selected: false,
                             bubbleTargets: this,
                         }),
+                        new Y.eZ.LocationViewVersionsTabView({
+                            content: this.get('content'),
+                            location: this.get('location'),
```
dpobel: the Location seems useless so I guess you can get rid of it.

Resources/public/js/views/services/plugins/ez-versionsloadplugin.js
```diff
@@ -0,0 +1,64 @@
...
+                capi = service.get('capi'),
+                options = {
+                    api: capi,
+                };
+
+            e.content.loadVersions(options, function (error, versions) {
+                if (error) {
+                    e.target.set('loadingError', true);
+                } else {
+                    e.target.set('versions', versions);
```
dpobel: I think you also need to set `loadingError` to `false` for the case where an error happened previously and the user hits the retry button.

Resources/public/js/views/tabs/ez-locationviewversionstabview.js
dpobel: actually, I've just realized that this split should probably be done in the VersionsLoad plugin instead.

sunpietro: Nitpick: instead of using YUI3 wrapper for native array methods you can use native array methods (`forEach` in this case): `this.get('versions').forEach(, function (version) {`

Tests/js/views/assets/ez-locationviewview-tests.js
```diff
@@ -314,6 +314,15 @@ YUI.add('ez-locationviewview-tests', function (Y) {
...
 
+            Y.eZ.LocationViewVersionsTabView = Y.Base.create('locationViewVersionsTabView', Y.View, [], {}, {
+                ATTRS: {
+                    content: {},
+                    contentType: {},
+                    config: {},
+                    priority: {},
+                }
+            });
+
```
dpobel: in addition to that, I think you also have to add some tests to check that the versions tab view receives the expected parameters.

yannickroger:
Feedback taken into account @dpobel (and others)

yannickroger:
I am still missing a tiny bit of test coverage. Updating in a few minutes.

dpobel:
+1

StephaneDiot:
+1

yannickroger:
@miguelcleverti is there any valid reason why behat would fail ?

miguelcleverti:
@yannickroger not really sure what is causing it most likely a 'waiting' problem. However there is already an issue([EZP-25836](https://jira.ez.no/browse/EZP-25836)) related to this and I will look at it ASAP

PR #548, opened 2016-04-15
EZP-25641: Image missing when translating content

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-25641

## Description

Image, media file and binary file fields were disapearing when translating a content.
This behaviour came from the fact that currently if we are not updating the file attached to one of these fields, this means "no changes" and we were not returning a new fieldValue (so we are using the already filled one). This was done not to send the same file again and again.

But when translating a content the field value is not filled, but the file attached to the field is loaded (without activating the 'updated' flag). So when we are publishing the content with the new translation, those fields, were not saved.

To make it work with translation, we now check that the file is not updated **and that the language code of the current exist in the translations list**

## Tests

- manually and unit tested

### DIFF
--- a/Resources/public/js/views/fields/ez-binarybase-editview.js
+++ b/Resources/public/js/views/fields/ez-binarybase-editview.js
@@ -114,9 +114,14 @@ YUI.add('ez-binarybase-editview', function (Y) {
          */
         _getFieldValue: function () {
             var file = this.get('file'),
+                actualLanguageCode = this.get('languageCode'),
+                currentVersion = this.get('content').get('currentVersion'),
+                translationExist =  Y.Array.some(currentVersion.getTranslationsList(), function (languageCode) {
+                    return languageCode === actualLanguageCode;
+                }),
                 fieldValue;
 
-            if ( !this.get('updated') ) {
+            if ( !this.get('updated') && translationExist) {
                 return undefined;
             }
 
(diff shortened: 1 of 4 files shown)


Resources/public/js/views/fields/ez-binarybase-editview.js
```diff
@@ -116,7 +116,7 @@ YUI.add('ez-binarybase-editview', function (Y) {
             var file = this.get('file'),
                 fieldValue;
 
-            if ( !this.get('updated') ) {
+            if ( !this.get('updated') && this.get('field').fieldValue && (this.get('field').fieldValue.fileName == file.name) ) {
```
dpobel: comparing file names to check if the file has been updated is a bad idea, what if I update the field with a different file but with the same name ?

StephaneDiot:
ping @dpobel, @yannickroger

Resources/public/js/views/fields/ez-binarybase-editview.js
```diff
@@ -114,9 +114,14 @@ YUI.add('ez-binarybase-editview', function (Y) {
          */
         _getFieldValue: function () {
             var file = this.get('file'),
+                actualLanguageCode = this.get('languageCode'),
```
yannickroger: Maybe put all this block in a dedicated method

Resources/public/js/views/fields/ez-binarybase-editview.js
```diff
@@ -114,9 +114,14 @@ YUI.add('ez-binarybase-editview', function (Y) {
...
+                currentVersion = this.get('content').get('currentVersion'),
+                translationExist =  Y.Array.some(currentVersion.getTranslationsList(), function (languageCode) {
```
dpobel: I would put this logic in a dedicated method. Also, what you are checking here is not exactly that the translation exist but rather that the user is creating a new translation, so this method would be called `_isCreatingTranslation`.
Nitpick: you can also avoid using Y.Array.some by just using indexOf on the array.

StephaneDiot:
re-opened ( based on 1.2 ) --->  https://github.com/ezsystems/PlatformUIBundle/pull/554

PR #549, opened 2016-04-15
EZP-25666: Can't edit some fields after switching language

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25666

# Description

When switching language after creating a content, the field edit views are *rerendered* to take into account the new language but some field edit views do not support this *rerendering*. This patch fixes that by adding a lifecycle method to the base field edit view which is called after the rerendering of a field edit view when it is already in the active state. This gives an opportunity to correctly reinitialize any UI widget used for input. This is what is done for RichText, MapLocation, Selection (Country) and Date (for browsers not supporting date input). For Image, this is just about updating the template to take that case into account.

## Tasks

* [x] Add a way for the field edit views to deal with that case
* [x] Fix RichText edit view
* [x] Fix MapLocation edit view
* [x] Fix Selection/Country edit view
* [x] Fix Date edit view (in browsers not supporting date input like Firefox)
* [x] Fix Image edit view

# Tests

manual tests + unit test

### DIFF
--- a/Resources/public/js/views/ez-fieldeditview.js
+++ b/Resources/public/js/views/ez-fieldeditview.js
@@ -61,7 +61,10 @@ YUI.add('ez-fieldeditview', function (Y) {
         /**
          * Default implementation of the field edit view render. By default, it
          * injects the field definition, the field, the content and the content
-         * type
+         * type. If the view is already active, it calls `_afterActiveReRender`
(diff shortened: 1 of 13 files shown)


dpobel:

yannickroger:
+1

StephaneDiot:
+1

PR #550, opened 2016-04-15
Fix EZP-25672: Content Type controller should use Languages more wisely

glye:
> Fixes https://jira.ez.no/browse/EZP-25672
> Status: Ready for review

Ensure that the language we use exists in the content type.

- [x] Content type view: Uses the first of specified/prioritised languages that exist in the content type
- [x] Content type edit: Uses the first of specified/prioritised languages that exist in the content type
  - [x] ~~Possible improvement: Add language if it doesn't exist in the content type~~ Postponing until we have language selection https://jira.ez.no/browse/EZP-24451
- [x] ~~Content type group view: Uses the first of prioritised languages that exist in each content type~~ No change needed here.
- [x] Content type create: Use specified language, or first prioritised language if none is specified. (Does not check if language exists in repo.)
  - [x] ~~Possible improvement: Give user a choice of existing repo languages~~ See https://jira.ez.no/browse/EZP-24451

### DIFF
--- a/Controller/ContentTypeController.php
+++ b/Controller/ContentTypeController.php
@@ -13,6 +13,7 @@ namespace EzSystems\PlatformUIBundle\Controller;
 use eZ\Publish\API\Repository\ContentTypeService;
 use eZ\Publish\API\Repository\Exceptions\InvalidArgumentException;
 use eZ\Publish\API\Repository\Exceptions\NotFoundException;
+use eZ\Publish\API\Repository\LanguageService;
 use eZ\Publish\API\Repository\SearchService;
 use eZ\Publish\API\Repository\UserService;
 use eZ\Publish\API\Repository\Values\Content\Query;
@@ -61,7 +62,15 @@ class ContentTypeController extends Controller
      */
     private $fieldTypeMapperRegistry;
 
+    /**
+     * @var string[]
+     */
     private $prioritizedLanguages = [];
+
+    /**
+     * @var \eZ\Publish\API\Repository\LanguageService
+     */
+    private $languageService;
 
     public function __construct(
         ContentTypeService $contentTypeService,
@@ -69,7 +78,8 @@ class ContentTypeController extends Controller
         UserService $userService,
         ActionDispatcherInterface $contentTypeGroupActionDispatcher,
         ActionDispatcherInterface $contentTypeActionDispatcher,
-        FieldTypeFormMapperRegistryInterface $fieldTypeMapperRegistry
+        FieldTypeFormMapperRegistryInterface $fieldTypeMapperRegistry,
+        LanguageService $languageService
     ) {
         $this->contentTypeService = $contentTypeService;
         $this->searchService = $searchService;
@@ -77,6 +87,7 @@ class ContentTypeController extends Controller
         $this->contentTypeGroupActionDispatcher = $contentTypeGroupActionDispatcher;
         $this->contentTypeActionDispatcher = $contentTypeActionDispatcher;
         $this->fieldTypeMapperRegistry = $fieldTypeMapperRegistry;
+        $this->languageService = $languageService;
     }
 
     /**
@@ -118,6 +129,7 @@ class ContentTypeController extends Controller
         $canDelete = $this->isGranted(new Attribute('class', 'delete'));
         $deleteFormsById = [];
         $canDeleteById = [];
+        $languageById = [];
 
         foreach ($contentTypes as $contentType) {
             $contentTypeId = $contentType->id;
@@ -127,6 +139,11 @@ class ContentTypeController extends Controller
             )->createView();
 
             $canDeleteById[$contentTypeId] = $canDelete && !$this->contentTypeService->isContentTypeUsed($contentType);
+
+            $languageById[$contentTypeId] = $this->getPrioritizedLanguage(
+                array_keys($contentType->names),
+                $contentType->mainLanguageCode
+            );
         }
 
         return $this->render('eZPlatformUIBundle:ContentType:view_content_type_group.html.twig', [
@@ -137,6 +154,7 @@ class ContentTypeController extends Controller
             'can_delete_by_id' => $canDeleteById,
             'create_form' => $createForm->createView(),
             'delete_forms_by_id' => $deleteFormsById,
+            'language_by_id' => $languageById,
         ]);
     }
 
@@ -207,7 +225,6 @@ class ContentTypeController extends Controller
 
     public function viewContentTypeAction($contentTypeId, $languageCode = null)
     {
-        $languageCode = $languageCode ?: $this->prioritizedLanguages[0];
         $contentType = $this->contentTypeService->loadContentType($contentTypeId);
         $countQuery = new Query([
             'filter' => new Query\Criterion\ContentTypeId($contentTypeId),
@@ -215,6 +232,34 @@ class ContentTypeController extends Controller
         ]);
         $contentCount = $this->searchService->findContent($countQuery, [], false)->totalCount;
         $deleteForm = $this->createForm(new ContentTypeDeleteType(), ['contentTypeId' => $contentTypeId]);
+
+        if (isset($languageCode) && !isset($contentType->names[$languageCode])) {
+            $changeToLanguageCode = $this->getPrioritizedLanguage(
+                array_keys($contentType->names),
+                $contentType->mainLanguageCode
+            );
+
+            $this->notify(
+                'content_type.notification.translation_does_not_exist',
+                [
+                    '%requestedLanguage%' => $languageCode,
+                    '%contentTypeLanguage%' => $changeToLanguageCode,
+                ],
+                'content_type'
+            );
+
+            return $this->redirectToRouteAfterFormPost(
+                'admin_contenttypeView',
+                ['contentTypeId' => $contentTypeId, 'languageCode' => $changeToLanguageCode]
+            );
+        }
+
+        if (!isset($languageCode)) {
+            $languageCode = $this->getPrioritizedLanguage(
+                array_keys($contentType->names),
+                $contentType->mainLanguageCode
+            );
+        }
 
         return $this->render('eZPlatformUIBundle:ContentType:view_content_type.html.twig', [
             'language_code' => $languageCode,
@@ -232,7 +277,23 @@ class ContentTypeController extends Controller
         $createForm = $this->createForm(new ContentTypeCreateType($this->contentTypeService), ['contentTypeGroupId' => $contentTypeGroupId]);
         $createForm->handleRequest($request);
         if ($createForm->isValid()) {
-            $languageCode = $languageCode ?: $this->prioritizedLanguages[0];
+            $repositoryLanguages = $this->getEnabledRepositoryLanguageCodes();
+            if (isset($languageCode) && !in_array($languageCode, $repositoryLanguages)) {
+                $this->notify(
+                    'content_type.notification.language_does_not_exist',
+                    [
+                        '%requestedLanguage%' => $languageCode,
+                        '%fallbackLanguage%' => $repositoryLanguages[0],
+                    ],
+                    'content_type'
+                );
+                $languageCode = null;
+            }
+
+            if (!isset($languageCode)) {
+                $languageCode = $this->getPrioritizedLanguage($repositoryLanguages, $repositoryLanguages[0]);
+            }
+
             $contentTypeGroup = $this->contentTypeService->loadContentTypeGroup($createForm->getData()['contentTypeGroupId']);
 
             $contentTypeCreateStruct = new ContentTypeCreateStruct([
@@ -266,15 +327,31 @@ class ContentTypeController extends Controller
 
     public function updateContentTypeAction(Request $request, $contentTypeId, $languageCode = null)
     {
-        $languageCode = $languageCode ?: $this->prioritizedLanguages[0];
+        $contentType = $this->contentTypeService->loadContentType($contentTypeId);
+
+        $repositoryLanguages = $this->getEnabledRepositoryLanguageCodes();
+        if (isset($languageCode) && !in_array($languageCode, $repositoryLanguages)) {
+            $this->notify(
+                'content_type.notification.language_does_not_exist',
+                [
+                    '%requestedLanguage%' => $languageCode,
+                    '%fallbackLanguage%' => $repositoryLanguages[0],
+                ],
+                'content_type'
+            );
+            $languageCode = null;
+        }
+
+        if (!isset($languageCode)) {
+            $languageCode = $this->getPrioritizedLanguage($repositoryLanguages, $repositoryLanguages[0]);
+        }
+
         // First try to load the draft.
         // If it doesn't exist, create it.
         try {
             $contentTypeDraft = $this->contentTypeService->loadContentTypeDraft($contentTypeId);
         } catch (NotFoundException $e) {
-            $contentTypeDraft = $this->contentTypeService->createContentTypeDraft(
-                $this->contentTypeService->loadContentType($contentTypeId)
-            );
+            $contentTypeDraft = $this->contentTypeService->createContentTypeDraft($contentType);
         }
 
         $contentTypeData = (new ContentTypeDraftMapper())->mapToFormData($contentTypeDraft);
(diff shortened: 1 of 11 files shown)


Controller/ContentTypeController.php
```diff
@@ -61,22 +63,32 @@ class ContentTypeController extends Controller
...
+
     public function __construct(
         ContentTypeService $contentTypeService,
         SearchService $searchService,
         UserService $userService,
         ActionDispatcherInterface $contentTypeGroupActionDispatcher,
         ActionDispatcherInterface $contentTypeActionDispatcher,
-        FieldTypeFormMapperRegistryInterface $fieldTypeMapperRegistry
+        FieldTypeFormMapperRegistryInterface $fieldTypeMapperRegistry,
+        LanguagesConfigProvider $languagesConfigProvider
```
andrerom: In retro you can just inject LanguageService like LanguagesConfigProvider does and load `$this->languageService->loadLanguages()` when needed, the abstraction does not provide any value here and just complicates reading the code.

Controller/ContentTypeController.php
```diff
@@ -127,6 +140,8 @@ public function viewContentTypeGroupAction($contentTypeGroupId)
             )->createView();
 
             $canDeleteById[$contentTypeId] = $canDelete && !$this->contentTypeService->isContentTypeUsed($contentType);
+
+            $languageById[$contentTypeId] = $this->getVerifiedContentTypeLanguageCode($contentType);
```
andrerom: Think we need to simplify this. Because of the extra helper functions here it is basically loading all languages over and over again for each content type.

So what do we need to extract here? Afaik you are only interested in which languages content type is already translated to right? So this can then simply extract this from `array_keys($contentType->names)` if we can live with referring to language by language code.

If not, why do you need id? :P

Controller/ContentTypeController.php
```diff
@@ -217,7 +232,7 @@ public function viewContentTypeAction($contentTypeId, $languageCode = null)
         $deleteForm = $this->createForm(new ContentTypeDeleteType(), ['contentTypeId' => $contentTypeId]);
 
         return $this->render('eZPlatformUIBundle:ContentType:view_content_type.html.twig', [
-            'language_code' => $languageCode,
+            'language_code' => $this->getVerifiedContentTypeLanguageCode($contentType, $languageCode),
```
andrerom: How is this used?

Can't we just simply verify  $languageCode against translations contentType is in and give error/warning to user and fallback to first language? aka

```php
if (!isset($content->names[$languageCode])) {
    // do something else then prev logic as we can't use $languageCode
}
```

Controller/ContentTypeController.php
```diff
@@ -232,7 +247,24 @@ public function createContentTypeAction(Request $request, $contentTypeGroupId, $
         $createForm = $this->createForm(new ContentTypeCreateType($this->contentTypeService), ['contentTypeGroupId' => $contentTypeGroupId]);
         $createForm->handleRequest($request);
         if ($createForm->isValid()) {
-            $languageCode = $languageCode ?: $this->prioritizedLanguages[0];
+            $verifiedLanguageCodes = $this->getVerifiedLanguageCodes(
```
andrerom: I think we can simplify, assuming:
- ui should provide drop down of languages to create in based on data in LanguageService, so that is already verified
- if prioritizedLanguages[] is wrong, it will fail later, however it a mistake of developer that he can fix, so we can live with that.

then the logic can be simplified to be what is was potentially, however lest take a skype call on this, I need to get the bigger picture :)

Controller/ContentTypeController.php
```diff
@@ -266,15 +298,15 @@ public function createContentTypeAction(Request $request, $contentTypeGroupId, $
 
     public function updateContentTypeAction(Request $request, $contentTypeId, $languageCode = null)
     {
-        $languageCode = $languageCode ?: $this->prioritizedLanguages[0];
+        $contentType = $this->contentTypeService->loadContentType($contentTypeId);
+        $languageCode = $this->getVerifiedContentTypeLanguageCode($contentType, $languageCode);
```
andrerom: hmm, API wise the language might be part of the draft but not the previous published version, so might not be the way to do it.

Controller/ContentTypeController.php
glye: @andrerom Right, we should only load the languages once. But we can't just choose a language at random, so we need the list of prioritised languages, and we also need the main CT language in case none of the prioritised ones exist in the CT. I can rework the helpers to avoid multiple loading.

Controller/ContentTypeController.php
glye: OK, i.e. ignore prioritisedLanguages, use $languageCode if valid, otherwise use main CT language and notify about it. Should be fine.

Controller/ContentTypeController.php
andrerom: Ok, I see I misunderstood the code a bit, id is referring to content type, so lets take prioritized languages into account, then we end up with something like

```php
function getPrioritizedLanguage($languageCodes, $fallbackLanguageCode)
{
    // the first possible natch in $this->prioritizedLanguages, fallback to $fallbackLanguageCode
}
```


use:
```php
 $languageById[$contentTypeId] = $this->getPrioritizedLanguage($languages = array_keys($contentType->names), $languages[0]);
```

glye: Thanks, sounds good. (Only I'd use `$contentType->mainLanguageCode`) instead of `$languageCodes[0]` as I'm not sure that the first language is the main one.)

glye:
Update: Injecting LanguageService instead of Config/Language. Added `getPrioritizedLanguage()` & test for `viewContentTypeGroupAction()`.

Controller/ContentTypeController.php
glye: Done.

Controller/ContentTypeController.php
glye: Done.

Controller/ContentTypeController.php
glye: Done, using `getPrioritizedLanguage()` now.

glye:
Update:
Create/update CT: Show a notification when language does not exist in repo, and fall back to first prioritised language that exists in repo. When we have the dropdown to select language, we can remove the notification & fallback, and throw an exception instead.
Removed helpers that are not needed anymore.
Fixed a bug when updating a content type that is not yet published (separate commit)

Controller/ContentTypeController.php
glye: Simplified, see https://github.com/ezsystems/PlatformUIBundle/pull/550#issuecomment-211966115

Controller/ContentTypeController.php
glye: Simplified, see https://github.com/ezsystems/PlatformUIBundle/pull/550#issuecomment-211966115

andrerom:
+1

Can be merged to 1.2 and up.

glye:
Damn, the notification on CT view still shows up on the next pageload only, need to fix that before merge.

Controller/ContentTypeController.php
```diff
@@ -216,6 +233,25 @@ public function viewContentTypeAction($contentTypeId, $languageCode = null)
         $contentCount = $this->searchService->findContent($countQuery, [], false)->totalCount;
         $deleteForm = $this->createForm(new ContentTypeDeleteType(), ['contentTypeId' => $contentTypeId]);
 
+        if (isset($languageCode) && !isset($contentType->names[$languageCode])) {
+            $this->notify(
```
glye: This notification shows up not on the rendered page, but on the next pageload. Do you know what's wrong @dpobel?

glye: As an alternative I tried doing a `redirectToRouteAfterFormPost()` after the notification. This generates a HTTP 205 with the correct new pjax-location (with changed language), looks good. But the page never loads, it just spins the loader icon. It's not an infinite loop, I've checked that.

glye:
Fixup: Made sure the notification displays the prioritised language (minor correction). Rebased on master.

Still suffering from delayed-notification-syndrome, see above.

andrerom:
>Still suffering from delayed-notification-syndrome, see above.

ping @dpobel

dpobel:
I see 3 call to `notify`, which one is delayed ?

glye:
@dpobel Sorry, the clarification was hidden by a commit. It's the notification in `viewContentTypeAction()`, which is followed by a `render()`.

dpobel:
for the delay notification, I think this is the first time we try to issue a notification and render a Response in the same request (usually there's a redirection in the middle). And this is happening because [we put the notifications in the session on the event KernelEvents::RESPONSE](https://github.com/ezsystems/PlatformUIBundle/blob/master/Notification/NotificationPool.php#L45) so after the rendering. I don't know if there's a reason for that, @lolautruche any hint ?

glye:
Yep, I didn't find any other cases of this, so I also tried doing a redirect. Since the language (and the URL) changes, that's makes sense anyway. But the redirect never happened, it got stuck on the spinning loader graphic. Will update the PR to that if you prefer.

glye:
Just pushed https://github.com/ezsystems/PlatformUIBundle/pull/550/commits/580b0da643ea648e43db990951e6d7b9cfcc9cd2 which does a redirect instead. This fails to load as I mentioned, loader spins into infinity. The result array from the redirect looks ok to me: https://gist.github.com/glye/a7707e2cf57b99d66b67b39c7daac500
Any idea @dpobel @andrerom?

andrerom:
For redirect ping, maybe rather ping @lolautruche, @bdunogier and @yannickroger, I'm not the one to ping here :)

glye:
I have verified that the redirect array is well formed. The same redirect, when performed from deleteContentTypeAction, is working. Perhaps the problem is that viewContentTypeAction is not a POST action, so redirectToRouteAfterFormPost fails.

dpobel:
I never said you should do a redirection :)
that said, this makes me realize that I don't really like the approach of adding so much logic in each action. Given that at some point, we'll have to add the ability for the user to choose the language in which the content type is edited, why don't we always pass a language ? Of course, we have to check that this language exists/is valid but if it's not, you just throw an exception and don't try to guess a language there.
@bdunogier @yannickroger @glye (or anybody else) any opinion on that ?

glye:
@dpobel I know, I'm just illustrating different approaches here. Simply throwing an exception would be an easy way out. (Though it would be nice to find out why the redirect fails.)
@andrerom was involded in the current approach, CC.

bdunogier:
I have posted a long'ish comment about the topic on JIRA: https://jira.ez.no/browse/EZP-25672?focusedCommentId=189061&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-189061.

Without further information, i'm -1: I think this does a bit too much, and not always in the right place.

andrerom:
You are right, it can be simplify, expanded on that here: https://jira.ez.no/browse/EZP-25672?focusedCommentId=189065&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-189065

With that we can get rid of the notifications, the redirects and the repository loads, and just solve the customer issue.

glye:
Status: Kinda stuck. JV suggested to verify language config in `EzPublishCoreExtension`, fine. In order to verify, we need to load the repository for every site access, and check if configured languages for each site access exist in the repo. However, the repo factories are supposed to be called by the `ezpublish.api.repository` service, giving you only the repo of the current site access:
https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Bundle/EzPublishCoreBundle/ApiLoader/RepositoryFactory.php#L80

We could make a repo factory that accepts site access as parameter, and use it manually, I guess. Or some other, manual-ish way. Though it sounds like the kind of kludge reviewers are not thrilled about. Opinions welcome.

andrerom:
@glye as implied by both myself and @bdunogier you don't need to verify at all, that is not what the bug is about. See discussion on issue: https://jira.ez.no/browse/EZP-25672?focusedCommentId=189065&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-189065

glye:
Simplified as discussed. Usage of repository and languageservice is removed from PR. Uses the first of specified/prioritised languages that exist in the content type. When creating CTs, just use the first of specified/prioritised languages.

Possible improvement is when updating a CT and the speficied language doesn't exist - we should add it to the CT then. I elected to postpone that, as it looked like a larger change and we wanted to simplify this PR. Must be done when we add language selection.

joaoinacio:
Looks good, IMHO what this does is to actually fix a bug (I'd get an exception otherwise, even if eng-GB was in the SA languages but not prioritized ).
+1 FWIW

glye:
Good morning! Review ping @andrerom @bdunogier 
Thanks @joaoinacio

Controller/ContentTypeController.php
```diff
@@ -350,4 +372,23 @@ public function deleteContentTypeAction(Request $request, $contentTypeId)
...
+
+    /**
+     * Return the highest prioritized language that exists in $languageCodes, or if none, return $fallbackLanguageCode.
+     *
+     * @param string[] $languageCodes Language codes
+     * @param string $fallbackLanguageCode Fallback language code
+     *
+     * @return string Language code
+     */
+    private function getPrioritizedLanguage($languageCodes, $fallbackLanguageCode)
```
bdunogier: The calls to `getPrioritzedLanguage` all read the arguments from the same `ContentTypeDraft`, unless I have missed something.

You could maybe simplify by making this method accept a ContentTypeDraft instead of language codes. It could then take care of the `array_keys` logic.

Tests/Controller/ContentTypeControllerTest.php
```diff
@@ -0,0 +1,114 @@
...
+                'fre-FR',
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider getPrioritizedLanguageData
+     * @covers \EzSystems\PlatformUIBundle\Controller\ContentTypeController::getPrioritizedLanguage()
+     */
+    public function testGetPrioritizedLanguage($prioritizedLanguages, $languageCodes, $fallbackLanguageCode, $expected)
```
bdunogier: Argh, I'm dying. If I'm not, please finish me :-)

Need to think about it, but as some used to say, _I don't like it_.

Controller/ContentTypeController.php
glye: Yeah, the usage has changed, so this is possible now (though must be ContentType, not ContentTypeDraft, since it must support both).

Tests/Controller/ContentTypeControllerTest.php
glye: Wat...? :)

Controller/ContentTypeController.php
andrerom: you can indeed change to that now

andrerom:
+1 Looks good now with the change to take type as argument which is possible now to reduce repetition.

Controller/ContentTypeController.php
glye: Done.

glye:
Update: Simplified `getPrioritizedLanguage` to take just the type as argument.

Resources/views/ContentType/view_content_type_group.html.twig
```diff
@@ -36,7 +36,7 @@
                     {% for content_type in content_types %}
                         <tr>
                             <td>{{ content_type.id }}</td>
-                            <td><a href="{{ path('admin_contenttypeView', {'contentTypeId': content_type.id}) }}">{{ ez_trans_prop(content_type, "name") }}</a></td>
+                            <td><a href="{{ path('admin_contenttypeView', {'contentTypeId': content_type.id}) }}">{{ ez_trans_prop(content_type, "name", language_by_id[content_type.id]) }}</a></td>
```
andrerom: I might be wrong, but I don't think this is needed, [ez_trans_prop](https://doc.ez.no/display/TECHDOC/ez_trans_prop) will pick name by prioritized languages and fallback to main language on it's own afaik.

glye: Yep, seems to work actually. Done.

glye:
Update: Removed changes to the view CT group action, as they are not needed.

Tests/Controller/ContentTypeControllerTest.php
bdunogier: Using reflection to unit test a protected / private method. We do not do that.

Being "forced" to do that is a clear case of an insufficient factoring. On the other hand, there was no unit testing whatsoever before, and you can't make it up. That's why I'm not asking that this is changed :-)

andrerom:
+1 still from me :)

@bdunogier WDYT?

Controller/ContentTypeController.php
```diff
@@ -350,4 +360,23 @@ public function deleteContentTypeAction(Request $request, $contentTypeId)
...
+     * If there is no translation for a prioritized language, return $contentType's main language.
+     *
+     * @param ContentType $contentType Content type (or content type draft)
+     *
+     * @return string Language code
+     */
+    private function getPrioritizedLanguage(ContentType $contentType)
+    {
+        foreach ($this->prioritizedLanguages as $prioritizedLanguage) {
+            if (in_array($prioritizedLanguage, array_keys($contentType->names))) {
```
andrerom: Come to think about, it could have been `if (isset($contentType->names[$prioritizedLanguage]))` more readable imo

bdunogier:
Thinking about it, this needs to be reopened against older branches, as we are fixing a bug. That would be at least platform-ui-bundle 1.3.

andrerom:
@glye Can you take care about reopening and if you somewhat agree with review comment adjust FTW?

bdunogier:
And I'll be +1 against 1.3.

PR #551, opened 2016-04-15
Fix EZP-25530: Allow saving DateAndTime on Firefox

yannickroger:
Link: https://jira.ez.no/browse/EZP-25530

## Description
Firefox doesn't support (yet ?)  `get('valueAsNumber')`. This patch refactors a bit to use the input's value and parses it until we improve the general behavior of DateAnd time field on Firefox in [EZP-23925](https://jira.ez.no/browse/EZP-23925)

## Tests
Manual test on FF and chrome.

### DIFF
--- a/Resources/public/js/views/fields/ez-dateandtime-editview.js
+++ b/Resources/public/js/views/fields/ez-dateandtime-editview.js
@@ -268,7 +268,7 @@ YUI.add('ez-dateandtime-editview', function (Y) {
          *
          * @protected
          * @method _getUtcTimeStamp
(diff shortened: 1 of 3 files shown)


andrerom:
*Side: Would this have been detected if we had unt tests running on more js/browser engines?*

yannickroger:
@andrerom most likely yes.

glye:
+0.5 (JS handicap). Weird that I could not reproduce this, am using FF.

dpobel:
+1

StephaneDiot:
+1

PR #553, opened 2016-04-18
[Behat] Add Behat test for issue EZP-25642 users 'Delete' button

miguelcleverti:
Behat BDD tests for issue https://jira.ez.no/browse/EZP-25642

This test adds the verification that the in the PlatformUI Action bar menu there is a 'Delete' button and not a 'Send to trash'. Since the users should be deleted and not sent to trash.

### DIFF
--- a/Features/Context/SubContext/CommonActions.php
+++ b/Features/Context/SubContext/CommonActions.php
@@ -313,6 +313,20 @@ trait CommonActions
         $this->clickOnTreePath($path);
     }
 
+
+    /**
+     * @Then I should see a :button button in the action bar
+     *
+     * @param  string   $button     Text of the element in the action bar
+     */
+    public function iShouldSeeActionBarButton($button)
+    {
+        $element = $this->getElementByText($button, '.ez-actionbar-container .ez-action', '.action-label');
+        if (!$element) {
+            throw new \Exception("Action bar button '$button' not found");
+        }
+    }
+
     /**
      * @Then I am on the :name location view
      */
--- a/Features/Context/Users.php
+++ b/Features/Context/Users.php
@@ -36,6 +36,15 @@ class Users extends PlatformUI
     }
 
     /**
+     * @When I go to (the) User :username page
+     */
+    public function goToUserPage($username)
+    {
+        $this->clickOnTreePath("$username $username");
+        $this->sleep(); //safegaurd for application delays
+    }
+
+    /**
      * @When I edit user :username
      */
     public function editUserUser($username)
--- a/Features/Users/users.feature
+++ b/Features/Users/users.feature
@@ -67,3 +67,10 @@ Feature: Use the eZ Users field
         And I fill in "Password" with "12345"
         And I fill in "Confirm password" with "123456"
         Then I should see error messages
+
+    @javascript @edge
+    Scenario: Validate that users have the "Delete" button available
+        Given I am on the Users page
+        And there is a User with name "One"
+        When I go to User "One" page
+        Then I should see a "Delete" button in the action bar


miguelcleverti:
@andrerom @bdunogier @dpobel @joaoinacio @pcardiga

Features/Context/SubContext/CommonActions.php
```diff
@@ -313,6 +313,20 @@ private function goToContentWithPath($path)
         $this->clickOnTreePath($path);
     }
 
+
+    /**
+     * @Then I should see a :button button in the action bar
+     *
+     * @param  string   $button     Text of the element in the action bar
```
bdunogier: Extra spaces ?

Features/Context/SubContext/CommonActions.php
```diff
@@ -313,6 +313,20 @@ private function goToContentWithPath($path)
...
+     */
+    public function iShouldSeeActionBarButton($button)
```
bdunogier: Argument naming: `$buttonText` ?

Features/Users/users.feature
```diff
@@ -67,3 +67,10 @@ Feature: Use the eZ Users field
         And I fill in "Password" with "12345"
         And I fill in "Confirm password" with "123456"
         Then I should see error messages
+
+    @javascript @edge
+    Scenario: Validate that users have the "Delete" button available
+        Given I am on the Users page
```
bdunogier: We could wrap the first three sentences into something like `When I view a (valid) User". How we get there does not really matter in this context.

pcardiga: The objective of the scenario is not to view a users' data, but to be on a user's page, so 'view' might mislead the reader.
But yes, we might shorten the sentences. Suggestion:
```Given I am on a user's page```
```Then I should see a "Delete" button in the action bar```

Comment (miguelcleverti) [inline Features/Users/users.feature]: `When I am on (valid) User page` it is not as generic and it really is what happened. I navigated to the User page so I am on the said user page. However this sentence might cause conflicts with the `Given I am on the Users page` behat does not like similar sentences.

Comment (bdunogier) [inline Features/Users/users.feature]: > Given I am on a user's page

Well, "page" is  not a term from the platform / platform ui domain. But it is at least a generic web term. Opinion, @dpobel ?

> But yes, we might shorten the sentences.

Note that the goal ain't really to shorten, but to limit the context to what matters in, well, the context.

File: Features/Users/users.feature [hunk 4]
```diff
@@ -67,3 +67,10 @@ Feature: Use the eZ Users field
...
+        And there is a User with name "One"
+        When I go to User "One" page
+        Then I should see a "Delete" button in the action bar
```
bdunogier: We may want to complete that: what we are testing here is indeed that we are getting a delete button on users, but _also_ that we are _not_ getting a "send to trash" button.

What about

```
Then I should not see a "Send to trash" button in the action bar
 But I should see a "Delete" button in the action bar
```

(what about changing the above with `I should see a "Delete" action bar button` ? Less clutter :-)

Features/Context/Users.php
```diff
@@ -36,6 +36,15 @@ public function iCreateUser(TableNode $users = null)
     }
 
     /**
+     * @When I go to (the) User :username page
+     */
+    public function goToUserPage($username)
+    {
+        $this->clickOnTreePath("$username $username");
+        $this->sleep(); //safegaurd for application delays
```
crevillo: nitpick: `safeguard`?

Features/Users/users.feature
crevillo: i tend to agree with @bdunogier. to testing this delete button existence there is no need to check where are you coming front. in fact, you could access this kind of `pages` from other places and button should be there too :)

Features/Context/SubContext/CommonActions.php
```diff
@@ -314,6 +314,32 @@ private function goToContentWithPath($path)
...
     /**
+     * @Then I should see a :button action bar button
+     *
+     * @param  string   $buttonText     Text of the element in the action bar
+     */
+    public function iShouldSeeActionBarButton($buttonText)
+    {
+        $element = $this->getElementByText($buttonText, '.ez-actionbar-container .ez-action', '.action-label');
+        if (!$element) {
+            throw new \Exception("Action bar button '$buttonText' not found");
```
joaoinacio: File seems to be using `EzAssertion` already, should it keep doing so, for consistency?
```
EzAssertion::assertElementFound($buttonText, $element, null, 'Action bar button');
```

Features/Context/SubContext/CommonActions.php
```diff
@@ -314,6 +314,32 @@ private function goToContentWithPath($path)
...
+    /**
+     * @Then I should not see a :button action bar button
+     *
+     * @param  string   $buttonText     Text of the element in the action bar
+     */
+    public function iShouldNotSeeActionBarButton($buttonText)
+    {
+        $element = $this->getElementByText($buttonText, '.ez-actionbar-container .ez-action', '.action-label');
+        if ($element) {
+            throw new \Exception("Action bar button '$buttonText' was found");
```
joaoinacio: same as above

miguelcleverti:
as suggested by @joaoinacio, depends now on [EzAssertion](https://github.com/ezsystems/BehatBundle/pull/55)

joaoinacio:
> as suggested by @joaoinacio, depends now on EzAssertion
It already did ;)

+1

Features/Context/Users.php
```diff
@@ -36,6 +36,17 @@ public function iCreateUser(TableNode $users = null)
     }
 
     /**
+     * @When I go to (the) User :username page
+     */
+    public function goToUserPage($username)
+    {
+        $this->iAmOnPage('Users');
+        $this->waitWhileLoading();
+        $this->clickOnTreePath("$username $username");
```
joaoinacio: actually, won't this work solely for users created in behat (same first/last name)?
what about admin, etc?

Correct/long approach would be to load the User's first/last name then use those values.

andrerom:
+1 *(however should pass on travis, trying to restart)*

miguelcleverti:
@joaoinacio now the User first and last name is loaded and used instead of assuming they are always the same.
However now this PR depends on these two:
* https://github.com/ezsystems/ezplatform/pull/109
* https://github.com/ezsystems/ezpublish-kernel/pull/1584

bdunogier:
There is a point that I had raised, and that got hidden by a commit, but that still stands: https://github.com/ezsystems/PlatformUIBundle/pull/553#discussion-diff-60112174.

I don't think we should have a username at all here, as it doesn't matter to the situation (happens on any user, without any particular characteristic).

Features/Users/users.feature
```diff
@@ -67,3 +67,10 @@ Feature: Use the eZ Users field
         And I fill in "Password" with "12345"
         And I fill in "Confirm password" with "123456"
         Then I should see error messages
+
+    @javascript @edge
+    Scenario: Validate that users have the "Delete" button available
+        Given there is a User with name "One"
+        When I go to User "One" page
+        Then I should not see a "Send to thrash" action bar button
```
crevillo: Typo: thrash should be trash.

Features/Users/users.feature
```diff
@@ -67,3 +67,10 @@ Feature: Use the eZ Users field
         And I fill in "Password" with "12345"
         And I fill in "Confirm password" with "123456"
         Then I should see error messages
+
+    @javascript @edge
+    Scenario: Validate that users have the "Delete" button available
```
bdunogier: The scenario describes something a bit different. What about `Validates that the "Send to trash" button is replaced with "Delete" when browsing users` ?

Features/Context/Users.php
bdunogier: This is a context problem, connected to what I have said above.

We should/could store the created user's reference, so that we can easily navigate to it. It would avoid having to reference the same fake string, and we could use `uniqid()` in generated names to ensure uniqueness.

miguelcleverti:
@bdunogier regarding the not having the username on the sentence,... What would you suggest, having a new sentence that uses the admin user or creates a new user just for the test?

bdunogier:
> What would you suggest, having a new sentence that uses the admin user or creates a new user just for the test?

Well, in that scenario, we could use the admin. The problem is that if somebody uses the same sentence to test removal somehow, we will end up in trouble :-)

I'd rather make the sentence create a user.

Features/Context/Users.php
bdunogier: But besides that, @joaoinacio is right. 

1. Either the function accepts a username as an argument, in which case it must use the userService to load the user, and get the content's name.
2. or the function accepts a user _content name_, and doesn't have to load the user, but it needs to be specified in the sentence.

Features/Users/users.feature
```diff
@@ -67,3 +67,12 @@ Feature: Use the eZ Users field
         And I fill in "Password" with "12345"
         And I fill in "Confirm password" with "123456"
         Then I should see error messages
+
+    @javascript @edge
+    Scenario: Validate "Send to trash" button is replaced with "Delete" when browsing users
+    Scenario: Validate that users have the "Delete" button available
```
bdunogier: I was really suggesting that the sentence was _replaced_. As I see it, the scenario covers the fact that "Send to trash" is replaced with "Delete". We can easily formulate that as one sentence. Or did you have something else in mind ?

Features/Users/users.feature
```diff
@@ -67,3 +67,9 @@ Feature: Use the eZ Users field
         And I fill in "Password" with "12345"
         And I fill in "Confirm password" with "123456"
         Then I should see error messages
+
+    @javascript @edge
+    Scenario: Validate "Send to trash" button is replaced with "Delete" when browsing users
+        When I go to a valid User page
+        Then I should not see a "Send to trash" action bar button
+        And I should see a "Delete" action bar button
```
bdunogier: We could replace "and" with "but", I think it works too. And it sounds a bit better :-)

Features/Context/SubContext/CommonActions.php
```diff
@@ -314,6 +314,28 @@ private function goToContentWithPath($path)
     }
 
     /**
+     * @Then I should see a :button action bar button
+     *
+     * @param  string   $buttonText     Text of the element in the action bar
+     */
+    public function iShouldSeeActionBarButton($buttonText)
```
bdunogier: We should move this to an `ActionBarContext`.

PR #554, opened 2016-04-19
EZP-25641: Image, binary and media file missing when translating a conten

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-25641

## Description

Image, media file and binary file fields were disapearing when translating a content.
This behaviour came from the fact that currently if we are not updating the file attached to one of these fields, this means "no changes" and we were not returning a new fieldValue (so we are using the already filled one). This was done not to send the same file again and again.

But when translating a content the field value is not filled, but the file attached to the field is loaded (without activating the 'updated' flag). So when we are publishing the content with the new translation, those fields, were not saved.

To make it work with translation, we now check that the file is not updated **and that the language code of the current exist in the translations list**

## Tests

- manually and unit tested

### DIFF
--- a/Resources/public/js/views/fields/ez-binarybase-editview.js
+++ b/Resources/public/js/views/fields/ez-binarybase-editview.js
@@ -116,7 +116,7 @@ YUI.add('ez-binarybase-editview', function (Y) {
             var file = this.get('file'),
                 fieldValue;
 
-            if ( !this.get('updated') ) {
+            if ( !this.get('updated') && !this._isCreatingTranslation()) {
                 return undefined;
             }
 
@@ -131,6 +131,19 @@ YUI.add('ez-binarybase-editview', function (Y) {
                 fieldValue.data = file.data;
             }
             return this._completeFieldValue(fieldValue);
+        },
+
+        /**
+         * Check if a new translation is being created.
+         *
+         * @method _isCreatingTranslation
+         * @protected
+         * @return {Boolean}
+         */
+        _isCreatingTranslation: function () {
+            var currentVersion = this.get('content').get('currentVersion');
+
+            return currentVersion.getTranslationsList().indexOf(this.get('languageCode')) === -1;
         },
 
         /**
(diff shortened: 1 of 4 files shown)


Resources/public/js/views/fields/ez-binarybase-editview.js
```diff
@@ -134,6 +134,20 @@ YUI.add('ez-binarybase-editview', function (Y) {
...
+         *
+         * @method _isCreatingTranslation
+         * @protected
+         * @return {Boolean}
+         */
+        _isCreatingTranslation: function () {
+            var actualLanguageCode = this.get('languageCode'),
+                currentVersion = this.get('content').get('currentVersion');
+
+                return currentVersion.getTranslationsList().indexOf(actualLanguageCode) >= 0;
```
dpobel: this method does exactly the opposite of what its name suggests :) If the language code is in the list, then you are not creating a translation :-)

nitpick: indentation level

Resources/public/js/views/fields/ez-binarybase-editview.js
```diff
@@ -116,7 +116,7 @@ YUI.add('ez-binarybase-editview', function (Y) {
             var file = this.get('file'),
                 fieldValue;
 
-            if ( !this.get('updated') ) {
+            if ( !this.get('updated') && this._isCreatingTranslation()) {
```
dpobel: the condition should be *not updated and not creating a translation*

Resources/public/js/views/fields/ez-binarybase-editview.js
```diff
@@ -134,6 +134,20 @@ YUI.add('ez-binarybase-editview', function (Y) {
         },
 
         /**
+         * Check if the user is creating a new translation
```
yannickroger: Nitpick: if a new translation is being created

StephaneDiot:
updated

dpobel:
+1

yannickroger:
+1

PR #555, opened 2016-04-19
EZP-25349: Editing with country loads value incorrectly

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-25349
## Description

When editing a content with a country field the selection filter was displaying the alpha code instead of the the country name. This bad behaviour came from the fact that the country list (with the names of the countries) was loaded a bit after the field was filled.

To avoid that, we, now, use directly the countryList of the config instead of filling (too lately !) an attribute for the country list. This also means that the attribute countryList ins now **deprecated** and will be removed in PlatformUI 2.0


StephaneDiot:
ping @dpobel @yannickroger

dpobel:
+1

yannickroger:
+1

PR #556, opened 2016-04-19
EZP-25425: Make SubitemListView more robust against inconsistencies

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25425

# Description

After sending a Location to the trash, the parent Location HTTP cache is not correctly cleared, as a result the REST API reports a wrong number of child and this puts the subitem list view in an inconsistent state. The cache issue is being handled in [EZP-25003](https://jira.ez.no/browse/EZP-25003) but this patch makes the subitem list a bit more robust against such inconsistencies.

# Tests

manual + unit tests

### DIFF
--- a/Resources/public/css/theme/views/subitem/list.css
+++ b/Resources/public/css/theme/views/subitem/list.css
@@ -10,6 +10,7 @@
 
 .ez-view-subitemlistview.is-page-loading {
     opacity: 0.3;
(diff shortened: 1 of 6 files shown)


dpobel:

StephaneDiot:
+1

yannickroger:
+1

PR #558, opened 2016-04-19
EZP-25332: When closing language selection box it change URI

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-25332
## Description

The cancel handler of the language selection box was missing a 'prevent default' behaviour. the result was that it was changing the URI

### DIFF
--- a/Resources/public/js/views/ez-languageselectionboxview.js
+++ b/Resources/public/js/views/ez-languageselectionboxview.js
@@ -274,13 +274,15 @@ YUI.add('ez-languageselectionboxview', function (Y) {
          *
          * @protected
          * @method _cancelLanguageSelection
+         * @param {EventFacade} e
          */
-        _cancelLanguageSelection: function () {
+        _cancelLanguageSelection: function (e) {
             /**
              * Fired when language selection is being cancelled
              *
              * @event cancelLanguageSelection
              */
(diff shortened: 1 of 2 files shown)


Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -275,12 +275,14 @@ YUI.add('ez-languageselectionboxview', function (Y) {
          * @protected
```
dpobel: missing doc update

Resources/public/js/views/ez-languageselectionboxview.js
```diff
@@ -275,12 +275,14 @@ YUI.add('ez-languageselectionboxview', function (Y) {
...
          * @method _cancelLanguageSelection
          */
-        _cancelLanguageSelection: function () {
+        _cancelLanguageSelection: function (e) {
             /**
              * Fired when language selection is being cancelled
              *
              * @event cancelLanguageSelection
+             * @param {EventFacade} e
```
dpobel: wrong like, the new param is for the method not the event...

dpobel:
besides the doc issue, +1

StephaneDiot:
ping @yannickroger

yannickroger:
+1

PR #562, opened 2016-04-20
Fix EZP-25426: Mashing Publish creates more than one content

yannickroger:
Link: https://jira.ez.no/browse/EZP-25426

# Description
When clicking multiple times on the publish button, multiple content could be created.
This patch disabled the button after the first click and reenables it after the loading.

# Tests
Manual and unit test update

### DIFF
--- a/Resources/public/js/views/services/plugins/ez-publishdraftplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-publishdraftplugin.js
@@ -38,7 +38,8 @@ YUI.add('ez-publishdraftplugin', function (Y) {
             var service = this.get('host'),
                 content = service.get('content'),
                 notificationIdentifier,
-                app = service.get('app');
+                app = service.get('app'),
+                buttonActionView = e.target;
 
             if ( !e.formIsValid ) {
                 return;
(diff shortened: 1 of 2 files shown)


Resources/public/js/views/services/plugins/ez-publishdraftplugin.js
```diff
@@ -38,7 +38,8 @@ YUI.add('ez-publishdraftplugin', function (Y) {
             var service = this.get('host'),
                 content = service.get('content'),
                 notificationIdentifier,
-                app = service.get('app');
+                app = service.get('app'),
+                target = e.target;
```
dpobel: not the best name ever :) `view` or even `buttonActionView` would greatly improve the semantic

Resources/public/js/views/services/plugins/ez-publishdraftplugin.js
```diff
@@ -59,6 +60,12 @@ YUI.add('ez-publishdraftplugin', function (Y) {
             });
 
             app.set('loading', true);
+            target.set("disabled", true);
+
+            app.after('loadingChange', function () {
```
dpobel: you should use `onceAfter` as you only want to do that once.

yannickroger:
@dpobel 's feedback taken into account.

dpobel:
+1

StephaneDiot:
+1

PR #563, opened 2016-04-20
Fix EZP-25404: Broken content tree when displaying the Trash

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25404

# Description

This patch fixes the content tree which was broken when displaying the Trash. This was happening because the Content Tree was build upon the assumption that the main view would always provide a Location and its path which could be used to build the Content Tree. With the Trash, this assumption is not satisfied so basically this patch adds the handling of the case where we have no Location provided by the main view and in such case, we build the tree from the Location id provided in the root REST response.

# Tests

manual test + unit tests (the coverage of the DiscoveryBar Content Tree plugin is now decent, around 85%).

### DIFF
--- a/Resources/public/js/views/services/plugins/ez-discoverybarcontenttreeplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-discoverybarcontenttreeplugin.js
@@ -26,6 +26,22 @@ YUI.add('ez-discoverybarcontenttreeplugin', function (Y) {
         },
 
         /**
+         * Retrieves the root location id so that we can fallback on it as a
+         * starting point to build the tree.
+         *
+         * @method parallelLoad
+         * @param {Function} callback
+         */
+        parallelLoad: function (callback) {
+            var discoveryService = this._getCAPI().getDiscoveryService();
+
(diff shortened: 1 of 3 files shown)


dpobel:

dpobel:

Resources/public/js/views/services/plugins/ez-discoverybarcontenttreeplugin.js
```diff
@@ -26,6 +26,22 @@ YUI.add('ez-discoverybarcontenttreeplugin', function (Y) {
...
+         * Retrieves the root location id so that we can fallback on it as a
+         * starting point to build the tree.
+         *
+         * @method parallelLoad
+         * @param {Function} callback
+         */
+        parallelLoad: function (callback) {
+            var discoveryService = this._getCAPI().getDiscoveryService();
+
+            discoveryService.getInfoObject("rootLocation", Y.bind(function (error, rootLocation) {
```
yannickroger: Should we do anything if an error happens ?

yannickroger:
+1

StephaneDiot:
+1

PR #564, opened 2016-04-20
EZP-25597: Disable send to Trash action if root location

crevillo:
Link: https://jira.ez.no/browse/EZP-25597

It disables the send to trash action button when viewing a location with depth === 1. 

Added two more tests. 

p.s. should i try to add behat tests?

ping @andrerom @bdunogier @dpobel @SylvainGuittard @yannickroger

### DIFF
--- a/Resources/public/js/models/ez-locationmodel.js
+++ b/Resources/public/js/models/ez-locationmodel.js
@@ -303,6 +303,16 @@ YUI.add('ez-locationmodel', function (Y) {
         },
 
         /**
+         * Returns if the location is a root one
+         * 
+         * @method isRootLocation
+         * @return {boolean}
+         */
+        isRootLocation: function() {
+            return this.get('depth') === 1;
+        },
+
+        /**
          * Overrides the RestModel implementation to also deal with the `path` attribute
          *
          * @method toJSON
--- a/Resources/public/js/views/ez-actionbarview.js
+++ b/Resources/public/js/views/ez-actionbarview.js
@@ -77,7 +77,7 @@ YUI.add('ez-actionbarview', function (Y) {
                         actionList.push(
                             new Y.eZ.ButtonActionView({
                                 actionId: 'sendToTrash',
-                                disabled: false,
+                                disabled: this.get('location').isRootLocation(),
                                 label: 'Send to Trash',
                                 priority: 10
                             })
--- a/Tests/js/models/assets/ez-locationmodel-tests.js
+++ b/Tests/js/models/assets/ez-locationmodel-tests.js
@@ -3,7 +3,7 @@
  * For full copyright and license information view LICENSE file distributed with this source code.
  */
 YUI.add('ez-locationmodel-tests', function (Y) {
-    var modelTest, trashTest, moveTest, hideTest, removeTest, loadPathTest, toJSONTest, updateSortingTest, updatePriorityTest,
+    var modelTest, trashTest, moveTest, hideTest, removeTest, loadPathTest, toJSONTest, updateSortingTest, updatePriorityTest, isRootTest,
         Assert = Y.Assert, Mock = Y.Mock;
 
     modelTest = new Y.Test.Case(Y.merge(Y.eZ.Test.ModelTests, {
@@ -905,6 +905,35 @@ YUI.add('ez-locationmodel-tests', function (Y) {
         },
     });
 
+    isRootTest = new Y.Test.Case({
+        name: "eZ Location Model is Root test",
+
+        setUp: function () {
+            this.model = new Y.eZ.Location();
+        },
+
+        tearDown: function () {
+            this.model.destroy();
+            delete this.model;
+        },
+
+        "Should return true if root location": function () {
+            this.model._set('depth', 1);
+            Assert.isTrue(
+                this.model.isRootLocation(),
+                "The isRootLocation method should return true"
+            );
+        },
+
+        "Should return false if not root location": function () {
+            this.model._set('depth', 2);
+            Assert.isFalse(
+                this.model.isRootLocation(),
+                "The isRootLocation method should return false"
+            );
+        },
+    });
+
     Y.Test.Runner.setName("eZ Location Model tests");
     Y.Test.Runner.add(modelTest);
     Y.Test.Runner.add(trashTest);
@@ -915,4 +944,5 @@ YUI.add('ez-locationmodel-tests', function (Y) {
     Y.Test.Runner.add(updateSortingTest);
     Y.Test.Runner.add(updatePriorityTest);
     Y.Test.Runner.add(toJSONTest);
+    Y.Test.Runner.add(isRootTest);
 }, '', {requires: ['test', 'model-tests', 'ez-locationmodel', 'ez-restmodel']});
--- a/Tests/js/views/assets/ez-actionbarview-tests.js
+++ b/Tests/js/views/assets/ez-actionbarview-tests.js
@@ -5,13 +5,13 @@
 YUI.add('ez-actionbarview-tests', function (Y) {
     var Assert = Y.Assert,
         Mock = Y.Mock,
-        viewTest, userViewTest;
+        viewTest, userViewTest, rootLocationTest;
 
     viewTest = new Y.Test.Case({
         name: "eZ Action Bar View test",
 
         setUp: function () {
-            this.location = {};
+            this.locationMock = new Mock();
             this.content = {};
             this.contentTypeMock = new Mock();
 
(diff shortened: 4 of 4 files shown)


crevillo:
looks build fails with behat test... can't it be related to this pr?

Resources/public/js/views/ez-actionbarview.js
```diff
@@ -77,7 +77,7 @@ YUI.add('ez-actionbarview', function (Y) {
                         actionList.push(
                             new Y.eZ.ButtonActionView({
                                 actionId: 'sendToTrash',
-                                disabled: false,
+                                disabled: this.get('location').get('depth') === 1,
```
yannickroger: I would put this in a private method.

crevillo: called `_isDisabled` maybe, or...?

yannickroger: `_isRootNode` ?

Tests/js/views/assets/ez-actionbarview-tests.js
```diff
@@ -84,6 +89,19 @@ YUI.add('ez-actionbarview-tests', function (Y) {
             );
         },
 
+        _isActionDisabled: function (actionId) {
```
yannickroger: is this code used anywhere ?

Tests/js/views/assets/ez-actionbarview-tests.js
```diff
@@ -229,7 +247,139 @@ YUI.add('ez-actionbarview-tests', function (Y) {
...
+                'The ' + actionId + ' should be disabled'
+            );
+        },
+
+        "Should disable sendToTrashActionView": function () {
+            this._isActionDisabled('sendToTrash', {});
+        },
+    });
+
+    nonRootLocationViewTest = new Y.Test.Case({
```
yannickroger: You should merge this test into the previous one. They are two use cases of the previous test.
That would also help to mutualize some code.

Tests/js/views/assets/ez-actionbarview-tests.js
crevillo: Sorry, it's not, you're right. i'll delete it.

Tests/js/views/assets/ez-actionbarview-tests.js
crevillo: this is interesting. but how can a expect that this https://github.com/ezsystems/PlatformUIBundle/pull/564/files/cd8165ef05e9e9704d96e4ecbff6164ba1b5138a#diff-321204d5f5f2385b4110ce3cf0fc08a2R345 returns 1 or 2 in the same test? Actually this mocks are initialized in the setup function... should, would i need to reinit them in each of the test?

Resources/public/js/views/ez-actionbarview.js
crevillo: could be... but should it be here or maybe in the `location`model itself? 
nitpick: then `_isRootLocation` as we don't have nodes anymore? ;)

Tests/js/views/assets/ez-actionbarview-tests.js
yannickroger: Yes you mock this one differently in each test function.

Tests/js/views/assets/ez-actionbarview-tests.js
crevillo: @yannickroger  now it's used, because i applied your suggestion of adding the test to the existing suite. 
added the `_isRootLocation` method to the location model... is that correct? or were you meaning adding in the actionbarview itself?

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -303,6 +303,15 @@ YUI.add('ez-locationmodel', function (Y) {
         },
 
         /**
+         * @method _ïsRootLocation
```
yannickroger: there is a typo in the method name. The comments explaining what it does is missing.
Also since you moved it in the model (which is not what I meant, but it seems like a good idea) it needs to be public and unit tested in the location model tests.

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -905,6 +905,35 @@ YUI.add('ez-locationmodel-tests', function (Y) {
...
+        setUp: function () {
+            this.model = new Y.eZ.Location();
+        },
+
+        tearDown: function () {
+            this.model.destroy();
+            delete this.model;
+        },
+
+        "_isLocationRoot Should return true if depth is 1": function () {
```
yannickroger: "Should return true if location is root node"

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -905,6 +905,35 @@ YUI.add('ez-locationmodel-tests', function (Y) {
...
+            this.model._set('depth', 1);
+            Assert.isTrue(
+                this.model._isRootLocation(),
+                "The path attribute should be jsonified"
```
yannickroger: Update message.

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -905,6 +905,35 @@ YUI.add('ez-locationmodel-tests', function (Y) {
...
+            );
+        },
+
+        "_isLocationRoot Should return false if depth is greater than 1": function () {
```
yannickroger: Same as above

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -905,6 +905,35 @@ YUI.add('ez-locationmodel-tests', function (Y) {
...
+            this.model._set('depth', 2);
+            Assert.isFalse(
+                this.model._isRootLocation(),
+                "The path attribute should be jsonified"
```
yannickroger: Same as above

Tests/js/views/assets/ez-actionbarview-tests.js
yannickroger: I meant in the actionbar, but it sounds good in the model. We need just need @dpobel to confirm it is not a problem layerwise.

Tests/js/models/assets/ez-locationmodel-tests.js
crevillo: ok... but in that sentence i still miss *what* should return true (the method in this case)

crevillo:
applied your last comments @yannickroger

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -303,6 +303,17 @@ YUI.add('ez-locationmodel', function (Y) {
         },
 
         /**
+         * Returns if the location is a root one
+         * 
+         * @method _isRootLocation
+         * @return {boolean}
+         * @public
+         */
+        _isRootLocation: function() {
```
yannickroger: public method should not start by `_` sorry… I should have mentioned it.
You don't need the `@public`

Tests/js/views/assets/ez-actionbarview-tests.js
```diff
@@ -116,6 +146,36 @@ YUI.add('ez-actionbarview-tests', function (Y) {
         "Should not instantiate deleteActionView": function () {
             this._isNotActionCreated('deleteContent', {});
         },
+
+        "Should disable sendToTrashActionView if root location": function() {
```
yannickroger: Regarding merging the two test I told you yesterday, I meant merging them into one `rootLocationTest`, not into the existing viewTest.
You can call or extend the existing `viewTest` to avoid code duplication, let me try to find an example.

Tests/js/views/assets/ez-actionbarview-tests.js
```diff
@@ -84,6 +88,32 @@ YUI.add('ez-actionbarview-tests', function (Y) {
...
+                return actionView.get('actionId') === actionId;
+            });
+
+            Assert.isTrue(
+                !!actionFound.get('disabled'),
+                'The ' + actionId + ' action should be disabled for this view'
+            );
+        },
+
+        _isActionEnabled: function (actionId) {
```
yannickroger: Nitpick: You could even pass an `exepected` param and use Assert.areSame to avoid having 2 method almost identical.

Tests/js/views/assets/ez-actionbarview-tests.js
crevillo: ah, ok. sorry, 
interested in the example about extend...

yannickroger: Found it, so it's more about reusing: https://github.com/ezsystems/PlatformUIBundle/blob/306ff4b8c42ab0126bf6111fbb14f71ac393d83e/Tests/js/views/fields/assets/ez-richtext-editview-tests.js#L602

yannickroger:
+1 with feedback taken into account.
Great work @crevillo

crevillo:
Feedback taken in account, @yannickroger 
thanks a lot for your help.

dpobel:
+1

PR #567, opened 2016-04-21
EZP-25704: removed deprecated twig initRuntime() usage

bdunogier:
> JIRA: http://jira.ez.no/browse/EZP-25704

See https://github.com/ezsystems/ezpublish-kernel/pull/1637 for a description.


yannickroger:
+1

andrerom:
+1 when travis passes, seems tests or impl change needs to be adopted.

bdunogier:
Unit test should be fixed.

PR #572, opened 2016-04-25
Fix EZP-25459: Date edit view considers 0 as empty

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25459

# Description

The Date edit view does not correctly handle a timestamp to zero. This patch makes sure, we correctly handle that case even if at the moment, it's impossible to get that because in such case case [the REST API gives a null fieldValue](https://jira.ez.no/browse/EZP-25719).

# Tests

unit tests and manual tests

### DIFF
--- a/Resources/public/js/views/fields/ez-date-editview.js
+++ b/Resources/public/js/views/fields/ez-date-editview.js
@@ -55,6 +55,19 @@ YUI.add('ez-date-editview', function (Y) {
         },
 
         /**
+         * Checks whether the field contains a valid timestamp.
+         *
+         * @method _containsValidTimestamp
+         * @protected
+         * @return {Boolean}
+         */
+        _containsValidTimestamp: function () {
+            var field = this.get('field');
+
+            return (field && field.fieldValue && field.fieldValue.timestamp !== null);
+        },
+
+        /**
          * Synchronize Date attribute
          *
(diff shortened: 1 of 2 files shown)


dpobel:

Resources/public/js/views/fields/ez-date-editview.js
```diff
@@ -55,6 +55,19 @@ YUI.add('ez-date-editview', function (Y) {
         },
 
         /**
+         * Checks whether the field contains a valid timestamp.
+         *
+         * @method _containsValidTimestamp
+         * @protected
+         * @return {Boolean}
+         */
+        _containsValidTimestamp: function () {
```
yannickroger: Nitpick: You should check how it is done in Time and DateAndTime with the `_isFieldEmpty` method. Maybe it could be nice to have have some coherency between them.

yannickroger:
+1 (it's cool if the nipick is taken into account)

Resources/public/js/views/fields/ez-date-editview.js
yannickroger: After discussing with @dpobel it's not the exact same behavior. Forget about the previous comment.

StephaneDiot:
+1

PR #573, opened 2016-04-26
EZP-25671: Duplicates item when using grid view

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-25671
## Description

Items were duplicated up to 10 when using the "loading more" button for less than 10 items.

## Tests

- Unit tests & manual tests

### DIFF
--- a/Resources/public/js/views/subitem/ez-subitemgridview.js
+++ b/Resources/public/js/views/subitem/ez-subitemgridview.js
@@ -103,9 +103,9 @@ YUI.add('ez-subitemgridview', function (Y) {
                     this._disableLoadMore();
                 }
             });
-            this.after('subitemsChange', function () {
+            this.after('subitemsChange', function (e) {
                 this._uiUpdatePagination();
-                this._appendGridItem();
+                this._appendGridItem(this._getNewlyAddedSubitems(e.newVal, e.prevVal));
             });
             this.after(['subitemsChange', 'loadingErrorChange'], this._uiEndLoading);
         },
@@ -277,11 +277,12 @@ YUI.add('ez-subitemgridview', function (Y) {
          *
          * @method _appendGridItem
          * @protected
+         * @param {Array} newSubitems
          */
-        _appendGridItem: function () {
+        _appendGridItem: function (newSubitems) {
             var gridContent = this.get('container').one('.ez-subitemgrid-content');
 
-            this.get('subitems').slice(-1 * this.get('limit')).forEach(function (struct) {
+            newSubitems.forEach(function (struct) {
                 var itemView = new SubitemGridItemView(struct);
 
                 this._gridItemViews.push(itemView);
@@ -289,6 +290,20 @@ YUI.add('ez-subitemgridview', function (Y) {
                     itemView.render().get('container')
                 );
             }, this);
+        },
+
+        /**
+         * Return an array containing the subitems that we want to append
+         * in the grid view.
+         *
+         * @method _getNewlyAddedSubitems
+         * @private
+         * @param {EventFacade} e
+         */
+        _getNewlyAddedSubitems: function (subitemNewCount, subitemPrevCount) {
+            var subitemPreviousCount = subitemPrevCount ? subitemPrevCount.length : 0;
+console.log(subitemPreviousCount, subitemNewCount.length);
+            return this.get('subitems').slice( (subitemPreviousCount - subitemNewCount.length));
         },
 
         /**
--- a/Tests/js/views/subitem/assets/ez-subitemgridview-tests.js
+++ b/Tests/js/views/subitem/assets/ez-subitemgridview-tests.js
@@ -5,6 +5,7 @@
 YUI.add('ez-subitemgridview-tests', function (Y) {
     var renderTest, subitemSetterTest, loadSubitemsTest, gridItemTest,
         loadingStateTest, paginationUpdateTest, loadMoreTest, errorHandlingTest,
+        gridItemsDoNotDuplicateTest,
         Assert = Y.Assert,
         getSubItemStructs = function (count) {
             return (new Array(count)).map(function () {
@@ -21,7 +22,7 @@ YUI.add('ez-subitemgridview-tests', function (Y) {
 
         setUp: function () {
             this.location = new Y.Model({
-                childCount: 5,                          
+                childCount: 5,
             });
             this.view = new Y.eZ.SubitemGridView({
                 location: this.location,
@@ -108,7 +109,7 @@ YUI.add('ez-subitemgridview-tests', function (Y) {
 
         setUp: function () {
             this.location = new Y.Model({
-                childCount: 5,                          
+                childCount: 5,
             });
             this.view = new Y.eZ.SubitemGridView({
                 location: this.location,
@@ -276,7 +277,7 @@ YUI.add('ez-subitemgridview-tests', function (Y) {
 
         setUp: function () {
             this.location = new Y.Model({
-                childCount: 5,                          
+                childCount: 5,
             });
             this.view = new Y.eZ.SubitemGridView({
                 location: this.location,
@@ -330,7 +331,7 @@ YUI.add('ez-subitemgridview-tests', function (Y) {
                     gridContainer.one('.contentType').getContent(),
                     "The grid item view template should receive the contentType"
                 );
-                
+
                 i++;
             }, this);
         },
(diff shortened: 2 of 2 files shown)


Resources/public/js/views/subitem/ez-subitemgridview.js
```diff
@@ -103,9 +103,9 @@ YUI.add('ez-subitemgridview', function (Y) {
                     this._disableLoadMore();
                 }
             });
-            this.after('subitemsChange', function () {
+            this.after('subitemsChange', function (e) {
                 this._uiUpdatePagination();
-                this._appendGridItem();
+                this._appendGridItem(e);
```
dpobel: instead of passing the EventFacade, here you should compute an array containing the newly added subitems and then pass this to `_appendGridItem` which should stay focused on appending grid item as its name suggests.

Tests/js/views/subitem/assets/ez-subitemgridview-tests.js
```diff
@@ -276,11 +276,13 @@ YUI.add('ez-subitemgridview-tests', function (Y) {
 
         setUp: function () {
             this.location = new Y.Model({
-                childCount: 5,                          
+                childCount: 5,
             });
             this.view = new Y.eZ.SubitemGridView({
                 location: this.location,
+                container: '.container',
```
dpobel: why ?

Tests/js/views/subitem/assets/ez-subitemgridview-tests.js
```diff
@@ -330,10 +332,63 @@ YUI.add('ez-subitemgridview-tests', function (Y) {
...
+            this.view.set('subitems', subitems);
+            gridItems = container.all('.ez-subitemgrid-item');
+            Assert.areEqual(
+                this.view.get('limit'),
+                this.view.get('container').all('.ez-subitemgrid-content').size(),
+                "OoO"
+            );
+
+            button.set('disabled', false);
+            button.simulateGesture('tap', this.next(function () {
```
dpobel: using `tap` for this test complicates a bit your test. Why don't you change directly ̀the `subitems` attribute since the code you want to test is called in a `subitemsChange` event handler ?

dpobel: in addition, you change the `subitems` attribute 9 lines above, I'm pretty sure this is why you have trouble testing that.

StephaneDiot:
ping @dpobel @yannickroger

Tests/js/views/subitem/assets/ez-subitemgridview-tests.js
```diff
@@ -276,10 +277,11 @@ YUI.add('ez-subitemgridview-tests', function (Y) {
 
         setUp: function () {
             this.location = new Y.Model({
-                childCount: 5,                          
+                childCount: 5,
             });
             this.view = new Y.eZ.SubitemGridView({
                 location: this.location,
+                container: '.container',
```
dpobel: don't add if it's not needed (and it's not here)

Tests/js/views/subitem/assets/ez-subitemgridview-tests.js
```diff
@@ -597,6 +598,70 @@ YUI.add('ez-subitemgridview-tests', function (Y) {
...
+
+            this.view.set('subitems', subitems);
+            gridItems = container.all('.ez-subitemgrid-item');
+            Assert.areEqual(
+                subitems.length,
+                gridItems.size(),
+                "There should be one grid item per subitem"
+            );
+
+            this.view.set('subitems', this._getSubItemStruct(11));
```
dpobel: this is working because you are lucky (or because `subitems` should have a validator), but the `subitems` is supposed to accept only arrays. So this should be:
```js
this.view.set('subitems', [this._getSubItemStruct(11)]);
```

Tests/js/views/subitem/assets/ez-subitemgridview-tests.js
```diff
@@ -597,6 +598,70 @@ YUI.add('ez-subitemgridview-tests', function (Y) {
...
+    gridItemsDoNotDuplicateTest = new Y.Test.Case({
+        name: "eZ Subitem Grid View grid items do not duplicates test",
+
+        setUp: function () {
+            this.location = new Y.Model({
+                childCount: 5,
+            });
+            this.view = new Y.eZ.SubitemGridView({
+                location: this.location,
+                container: '.container',
```
dpobel: nitpick: not needed

Tests/js/views/subitem/assets/ez-subitemgridview-tests.js
```diff
@@ -597,6 +598,70 @@ YUI.add('ez-subitemgridview-tests', function (Y) {
...
+            );
+
+            this.view.set('subitems', this._getSubItemStruct(11));
+
+            gridItems = container.all('.ez-subitemgrid-item');
+
+            Assert.areEqual(
+                subitems.length + 1,
+                gridItems.size(),
+                "The subitems should not be duplicate and ther should be only one more grid item"
```
dpobel: `There should be one grid item per subitem` (ie no need to repeat the name of the test)

Tests/js/views/subitem/assets/ez-subitemgridview-tests.js
```diff
@@ -597,6 +598,70 @@ YUI.add('ez-subitemgridview-tests', function (Y) {
...
+                gridItems.size(),
+                "There should be one grid item per subitem"
+            );
+
+            this.view.set('subitems', this._getSubItemStruct(11));
+
+            gridItems = container.all('.ez-subitemgrid-item');
+
+            Assert.areEqual(
+                subitems.length + 1,
```
dpobel: that's odd, `this.view.set('subitems').length` would make more sense

Resources/public/js/views/subitem/ez-subitemgridview.js
```diff
@@ -292,6 +293,21 @@ YUI.add('ez-subitemgridview', function (Y) {
...
+         *
+         * @method _getSubItemsToAppend
+         * @protected
+         * @param {EventFacade} e
+         */
+        _getSubItemsToAppend: function (e) {
+            var subitemNewCount = e.newVal ? e.newVal.length : 0,
+                subitemPreviousCount = e.prevVal ? e.prevVal.length : 0;
+
+            return this.get('subitems').slice(-1 * (subitemNewCount - subitemPreviousCount));
```
dpobel: `this.get('subitems').slice(subitemPreviousCount - subitemNewCount);` :)

Resources/public/js/views/subitem/ez-subitemgridview.js
```diff
@@ -292,6 +293,21 @@ YUI.add('ez-subitemgridview', function (Y) {
...
 
         /**
+         * Return an array containing the subitems that we want to append
+         * in the grid view.
+         *
+         * @method _getSubItemsToAppend
+         * @protected
+         * @param {EventFacade} e
+         */
+        _getSubItemsToAppend: function (e) {
```
dpobel: what you actually want is an array containing the newly added subitems, so this method would be better named `_getNewlyAddedSubitems` and should directly receive the `previousValue` and the `newValue`. You should also mark it as private, since it's really some internal plumbing.

Resources/public/js/views/subitem/ez-subitemgridview.js
```diff
@@ -292,6 +293,21 @@ YUI.add('ez-subitemgridview', function (Y) {
...
+            var subitemNewCount = e.newVal ? e.newVal.length : 0,
```
dpobel: this condition, is not needed (and thus the 0 case is not covered), the `subitems` array can only grow, there's no way to have a new value to something else than a filled array.

Resources/public/js/views/subitem/ez-subitemgridview.js
```diff
@@ -277,11 +277,12 @@ YUI.add('ez-subitemgridview', function (Y) {
          *
          * @method _appendGridItem
          * @protected
+         * @param {Array} subitemsToAppend
          */
-        _appendGridItem: function () {
+        _appendGridItem: function (subitemsToAppend) {
```
dpobel: s/subitemsToAppend/newSubitems/

dpobel:
Replaced by https://github.com/ezsystems/PlatformUIBundle/pull/576

PR #574, opened 2016-04-27
EZP-25577: When clicking on the word preview I do not get preview

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-25577

## Description

Clicking on preview in the edit action bar isn't doing anything. Before cursor was a "hand" (meaning you should be able to click) when hovering "Preview", now it's the standard pointer to avoid confusion

## Tests

- manually tested


StephaneDiot:
re opened on 1.3

PR #577, opened 2016-04-29
Fix EZP-25718: Unable to remove inline style

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25718

# Description

This patch add a `beforeCommandExec` handler to remove the auto-generated ids added by YUI3. This is needed to fix EZP-25718, otherwise the CKEditor engine won't be able to remove some inline styles if you select the whole element representing the style because the engine considers the element with the id as representing another style.

See: http://dev.ckeditor.com/ticket/8369#comment:14

# Tests

manual + unit test

### DIFF
--- a/Resources/public/js/alloyeditor/plugins/yui3.js
+++ b/Resources/public/js/alloyeditor/plugins/yui3.js
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+/* global CKEDITOR */
+YUI.add('ez-alloyeditor-plugin-yui3', function (Y) {
+    "use strict";
+
+    if (CKEDITOR.plugins.get('yui3')) {
+        return;
+    }
+
+    function cleanUpIds(editor) {
+        Array.prototype.forEach.call(
+            editor.element.$.querySelectorAll('[id]'),
+            function (element) {
+                element.removeAttribute('id');
+            }
+        );
+    }
+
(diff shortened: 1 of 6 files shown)


dpobel:

Resources/public/js/alloyeditor/plugins/yui3.js
```diff
@@ -0,0 +1,37 @@
...
+
+    if (CKEDITOR.plugins.get('yui3')) {
+        return;
+    }
+
+    function cleanUpIds(editor) {
+        Array.prototype.forEach.call(
+            editor.element.$.querySelectorAll('[id]'),
+            function (element) {
+                element.removeAttribute('id');
```
andrerom: id isn't used for anything else? Like the id on links or embed tags?

dpobel: that's a good remark. For now, there's no way to set an id on a link or an embed (I'm not even sure the PM wants that back). That said, I could have checked for the `yui_3_18` prefix but the same problem can also occur if the user paste some code where a strong (or any inline style element) has an id. So for now, IMHO, I think this is enough and if we add the id features on links and/or embed then, we'll have to improve the code.

andrerom: ok

andrerom:
+1

Re-Ping @StephaneDiot @yannickroger

StephaneDiot:
+1

dpobel:
Closing this should be fixed in 1.3 first

PR #580, opened 2016-05-04
Fix EZP-25718: Unable to remove inline style

dpobel:
Replaces https://github.com/ezsystems/PlatformUIBundle/pull/577 to first fix the issue in 1.3


dpobel:
(just waiting for TravisCI)

PR #581, opened 2016-05-05
Fix EZP-25678: respect default content language on content create

joaoinacio:
https://jira.ez.no/browse/EZP-25678

When creating new content, the primary (top priority) language defined in the siteaccess should be respected, even more so if it's the only one.

TODO:
- [x] Also respect default name for new contents based on the language, if possible
- [x] Update tests


joaoinacio:
ping @dpobel

Resources/public/js/views/services/ez-contentcreateviewservice.js
```diff
@@ -56,9 +56,14 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
...
                 version = new Y.eZ.Version(),
                 type = this.get('contentType'),
-                defaultFields = {};
+                defaultFields = {},
+                contentTypeNames,
+                contentTypeName;
 
-            content.set('name', 'New "' + this.get('contentType').get('names')['eng-GB'] + '"');
+            contentTypeNames = type.get('names');
+            contentTypeName = contentTypeNames[this.get('app').get('defaultLanguageCode')] || contentTypeNames[0];
```
andrerom: This seems incomplete, we need some js service that behaves like we do on PHP side:
Loop over prioritized languages and pick first match, and if none fallback to main language of the given object.

andrerom: So in other words: there are no default language, just the list and fallback.

ApplicationConfig/Providers/SiteAccesses.php
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\PlatformUIBundle\ApplicationConfig\Providers;
+
+use eZ\Publish\Core\MVC\ConfigResolverInterface;
+use eZ\Publish\Core\MVC\Symfony\SiteAccess;
+use EzSystems\PlatformUIBundle\ApplicationConfig\Provider;
+
+class SiteAccesses implements Provider
```
andrerom: I think we already injection SA and prioritized languages somehow but might be wrong, @dpobel?

Resources/public/js/views/services/ez-contentcreateviewservice.js
joaoinacio: Indeed, I see how this would be too simplistic. 
So IIUC, replace app attribute with `prioritizedLanguages`, and use first one as default when necessary 
(such as `ez-contentcreateplugin`)

What about the method for retrieving the prioritizedLanguages?
Getting the info from siteaccesses list seems overkill but OTOH it gives a lot of flexibility, not certain if there's any explicit reason not to do this?

ApplicationConfig/Providers/SiteAccesses.php
joaoinacio: The Languages provider lists them by their natural (ID) order, AFAICT, so not really useful here.

andrerom: I was not takling about that one, that is full list Of installer languages from repo.

Resources/public/js/views/services/ez-contentcreateviewservice.js
andrerom: I'm not following. Besides, still relevant? Given this is now marked as outdated.

joaoinacio: > I'm not following. Besides, still relevant? Given this is now marked as outdated.

My question was about how the priorizedLanguages info retrieved, I didn't find any suitable method for this and could have implemented a provider differently (not listing all siteaccesses).

Still relevant, I just commited a fix that resolves an issue in the content create page title being set to f.e. "New fre-FR" see [150fff4179878e1b](https://github.com/ezsystems/PlatformUIBundle/pull/581/commits/150fff4179878e1b)

glye:
Related: Discussion in https://github.com/ezsystems/PlatformUIBundle/pull/550 about how/where to fail when configured siteaccess language(s) don't exist in repository.

@bdunogier may also be interested btw.

ApplicationConfig/Providers/SiteAccesses.php
dpobel: @andrerom for now, PlatformUI does not receive the list of siteaccess. Also, so far I don't know what are the plans regarding PlatformUI vs. siteaccesses (cf. the EPIC https://jira.ez.no/browse/EZP-25411)

Instead of a creating a new provider, I would enhance the Language provider to add a `default` key in https://github.com/ezsystems/PlatformUIBundle/blob/master/ApplicationConfig/Providers/Languages.php#L33 to flag the default language. From that, setting the `defaultLanguageCode` app attribute will be trivial.

Resources/public/js/views/services/ez-contentcreateviewservice.js
```diff
@@ -56,9 +56,15 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
...
                 version = new Y.eZ.Version(),
                 type = this.get('contentType'),
-                defaultFields = {};
+                defaultFields = {},
+                contentTypeNames,
+                contentTypeName;
 
-            content.set('name', 'New "' + this.get('contentType').get('names')['eng-GB'] + '"');
+            contentTypeNames = this.get('contentType').get('names');
+            contentTypeName = contentTypeNames[this.get('app').get('defaultLanguageCode')]
```
dpobel: I would put that a in a dedicated method

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -1031,6 +1041,19 @@ YUI.add('ez-platformuiapp', function (Y) {
                 readOnly: true,
                 value: {}
             },
+
+            /**
+             * Default language code, as defined by the current siteaccess language config.
+             *
+             * @attribute defaultLanguageCode
+             * @default "eng-GB"
+             * @type {Object}
```
dpobel: it's a string

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -1031,6 +1041,19 @@ YUI.add('ez-platformuiapp', function (Y) {
...
+             * @readOnly
+             */
+            defaultLanguageCode: {
```
dpobel: this name can be confusing, since it's only for content creation, this should appear somehow in the name (`contentCreationDefaultLanguageCode` ?)

Resources/public/js/views/services/plugins/ez-contentcreateplugin.js
```diff
@@ -46,10 +46,13 @@ YUI.add('ez-contentcreateplugin', function (Y) {
...
         setNextViewServiceParameters: function (service) {
-            if ( this.get('contentType') && this.get('languageCode') && this.get('parentLocation') ) {
+            var app = this.get('host').get('app'),
+                defaultLanguageCode = app.get('defaultLanguageCode');
+
+            if ( this.get('contentType') && this.get('parentLocation') ) {
                 service.setAttrs({
                     contentType: this.get('contentType'),
-                    languageCode: this.get('languageCode'),
+                    languageCode: defaultLanguageCode,
```
dpobel: with this change, the `languageCode` attribute becomes useless even if it is filled with the language code coming from the `createContent` event or other said, the language code in `createContent` event is completely ignored.

That's not necessarily a bad thing (the whole `createContent` event is badly designed...) but I think you should add/do the following:

1. document the `createContent` event in https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/ez-contenttypeselectorview.js#L195 and mark the `languageCode` parameter as deprecated.
1. Remove that line https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/services/plugins/ez-contentcreateplugin.js#L121 to not use the deprecated event parameter
1. instead of changing the code here, you should just add a `valueFn` to the `languageCode` attribute to pick its default value from the app.

dpobel:
@joaoinacio very nice first step, I'm glad you jumped in PlatformUI JavaScript code :)
(Note: You should open the PR against the 1.3 branch instead)

ApplicationConfig/Providers/SiteAccesses.php
andrerom: but there is no default language, there is a list of prioritised languages to try to match against.

dpobel: I know but we can take the first one as the default language when creating a content (basically same logic as in https://github.com/ezsystems/PlatformUIBundle/pull/581/commits/7a8c177f5a8ffa8bfe566adfb4910dd3fba3d14b#diff-4e897328f624be930d091d1ce688a3ffR233 but here instead)

Resources/public/js/views/services/ez-contentcreateviewservice.js
andrerom: Here logic is wrong.

It should iterate over prioritized languages until match is found, otherwise pick the one set as mainLanguage.
Just like in #550 

See also: https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/Helper/TranslationHelper.php#L164

ApplicationConfig/Providers/SiteAccesses.php
andrerom: ok, makes sense to use first one if no languageCode is set on creation, however for picking content type name language the logic should use whole prioritized language list and fallback to main language.

Resources/public/js/views/services/ez-contentcreateviewservice.js
dpobel: @andrerom  one problem at a time please :) You are right but keep in mind that:

* this `contentTypeName` is only used as a title in the form so if we pick the wrong content type name ok that's a bug but a minor one in this case and a different one than https://jira.ez.no/browse/EZP-25678
* having a prioritized language list basically means we are on a siteaccess but until the PlatformUI siteaccess handling, this is a very unclear until the corresponding EPIC correctly specified.
* we have the same problem in a lots of places, so the solution is not to fix it there and that will be a quite large fix

ApplicationConfig/Providers/SiteAccesses.php
dpobel: > makes sense to use first one if no languageCode is set on creation

that's exactly the point of the issue (reminder: unlike legacy, when creating a Content item in PlatformUI, you don't choose a language so the content creation is always initiated in a "default language").

For the Content Type, that's right but that's a different issue (see https://github.com/ezsystems/PlatformUIBundle/pull/581#discussion_r62660891).

Resources/public/js/views/services/ez-contentcreateviewservice.js
glye: This relates somewhat to https://jira.ez.no/browse/EZP-25765 - content type names not shown when not eng-GB. May want to close/change that if we're going to fix all such cases in one go.

dpobel: @glye this is exactly the same bug just in a different context, thanks for the links, I'm adding a comment.

andrerom: > @andrerom one problem at a time please :) You are right but keep in mind that:

ok

joaoinacio:
Updated according to feedback, still need to identify why there are now multiple `eZ Content Create Plugin tests` failing.

@dpobel any hints on this?

joaoinacio:
closing in favor of #593

PR #582, opened 2016-05-05
[Behat] Update CommonActions trait to use Behat BasicContent Context

miguelcleverti:
The changes in PR https://github.com/ezsystems/BehatBundle/pull/59 cause platformUI behat test the use the `BasicContent` Context to break.
This PR takes the changes into account and registers the `BasicContent` in the needed suite

### DIFF
--- a/Features/Context/SubContext/CommonActions.php
+++ b/Features/Context/SubContext/CommonActions.php
@@ -10,9 +10,28 @@
 namespace EzSystems\PlatformUIBundle\Features\Context\SubContext;
 
 use EzSystems\BehatBundle\Helper\EzAssertion;
+use Behat\Behat\Hook\Scope\BeforeScenarioScope;
 
 trait CommonActions
 {
+    /**
+     * @var EzSystems\BehatBundle\Context\Object\BasicContent
+     */
+    private $basicContentContext;
+
+    /** @BeforeScenario */
+    public function gatherContexts(BeforeScenarioScope $scope)
+    {
+        // not all context need the BasicContent Context
+        $this->basicContentContext = null;
+
+        $environment = $scope->getEnvironment();
+        $basicContextNamespace = 'EzSystems\BehatBundle\Context\Object\BasicContent';
+        if ($environment->hasContextClass($basicContextNamespace)) {
+            $this->basicContentContext = $environment->getContext($basicContextNamespace);
+        }
+    }
+
     /**
      * @Given I clicked on/at (the) :link link
      * @When  I click on/at (the) :link link
@@ -299,8 +318,12 @@ trait CommonActions
      */
     public function onFullView($name)
     {
-        $path = $this->getBasicContentManager()->getContentPath($name);
-        $this->goToContentWithPath($path);
+        if ($this->basicContentContext) {
+            $path = $this->basicContentContext->getContentPath($name);
+            $this->goToContentWithPath($path);
+        } else {
+            throw new \Exception('BasicContent Context must be registered in the suite environment');
+        }
     }
 
     /**
--- a/Features/Context/SubContext/CommonActions.php
+++ b/Features/Context/SubContext/CommonActions.php
@@ -10,28 +10,9 @@
 namespace EzSystems\PlatformUIBundle\Features\Context\SubContext;
 
 use EzSystems\BehatBundle\Helper\EzAssertion;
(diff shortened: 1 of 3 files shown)


miguelcleverti:
ping @bdunogier @joaoinacio @andrerom

Features/Context/SubContext/CommonActions.php
```diff
@@ -10,10 +10,29 @@
...
 {
     /**
+     * @var EzSystems\BehatBundle\Context\Object\BasicContent
+     */
+    private $basicContentContext;
+
+    /** @BeforeScenario */
+    public function gatherContexts(BeforeScenarioScope $scope)
+    {
+        // not all context need the BasicContent Context
```
joaoinacio: nitpick: typo: `contexts`

Features/Context/SubContext/CommonActions.php
```diff
@@ -10,10 +10,29 @@
...
+        $this->basicContentContext = null;
+
+        $environment = $scope->getEnvironment();
+        $basicContextNamespace = 'EzSystems\BehatBundle\Context\Object\BasicContent';
+        if ($environment->hasContextClass($basicContextNamespace)) {
+            $this->basicContentContext = $environment->getContext($basicContextNamespace);
+        }
```
joaoinacio: If I understand correctly, this looks OK but what happens if/when more contexts are needed (even if on instances outside the trait)?

I would perhaps refactor it to hold an indexed array of contexts (`private $contextInstances`).

OTOH, the contexts could be fetched directly from the environment, which already has the logic for checking and exception handling (see https://github.com/Behat/Behat/blob/master/src/Behat/Behat/Context/Environment/InitializedContextEnvironment.php#L110 )

I just find odd that InitializedContextEnvironment does not return and indexed array of contexts (only keys OR values), but in any case the `getContext($class)` function should suffice.

To sum, I would suggest:
1. in `gatherContexts()`, storing either the indexed contexts or the environment itself.
2. Implementing a `getContext()` wrapper that returns the context instance (again if fetched from $environment validation/exception is already done., only thing necessary is adding the full NameSpace to the class name)
3. on methods that use alt. contexts, call $this->getContext( name )->callSomeMethod();

joaoinacio:
If I understand correctly, this looks OK but what happens if/when more contexts are needed (even if on instances outside the trait)?

I would perhaps refactor it to hold an indexed array of contexts (private $contextInstances).

OTOH, the contexts could be fetched directly from the environment, which already has the logic for checking and exception handling (see https://github.com/Behat/Behat/blob/master/src/Behat/Behat/Context/Environment/InitializedContextEnvironment.php#L110 )

I just find odd that InitializedContextEnvironment does not return and indexed array of contexts (only keys OR values), but in any case the getContext($class) function should suffice.

To sum, I would suggest:
1. in gatherContexts(), storing either the indexed contexts or the environment itself.
2. Implementing a getContext() wrapper that returns the context instance (again if fetched from $environment validation/exception is already done., only thing necessary is adding the full NameSpace to the class name)
3. on methods that use alt. contexts, call $this->getContext( name )->callSomeMethod();

miguelcleverti:
as suggested by @joaoinacio the code was refactored and the environment is stored instead of the context.
This way the code should be easier to manage in the future.

bdunogier:
I wouldn't factor this that way. I have used context import a few times myself, and I didn't even think of organizing things that way.

It is interesting on some levels, but we are in my opinion falling for added complexity again.

joaoinacio:
> I wouldn't factor this that way. I have used context import a few times myself, and I didn't even think of organizing things that way.
> 
> It is interesting on some levels, but we are in my opinion falling for added complexity again.

TBH I don't see how this adds complexity, rather it removes it. The reasoning for using the getContext() "proxy" as suggested was to take advantage of existing features/exception handling in environment.
Plus the code looks much simpler ;)

As it is/was: 
- gatherContexts will need to be expanded in the future for added dependencies.
- any time a context needs to be used exception handling will also need to be there (unless we assume configuration is always correct for all cases all the time), otherwise there is the potential for fatal errors. see [onFullView](https://github.com/ezsystems/PlatformUIBundle/pull/582/commits/71ebaaf23338358773d291813ac6bf4cda0c2f06#diff-148a46c387798264a02f4a131e96298fL300)

Obviously both ways would work so i'm OK with either...

miguelcleverti:
This is outdated and no longer needed. Closing.

PR #583, opened 2016-05-09
[WIP] EZP-25760: As a user, I want to manage and view my content from my dashboard

sunpietro:
**WORK IN PROGRESS**

https://jira.ez.no/browse/EZP-25760

The code can be splitted into separate commits/PR if needed. I kept all of the code here for discussion later.

## The new extensible Dashboard concept
It allows to add new blocks to dashboard by creating plugins applied to the `eZ.DashboardView` instance.
Each block will require its own view followed by a `eZ.DashboardViewService` plugin that handles actions of getting data to blocks. The service plugin is optional.

To add a new block you have to define:
- [x] block view,
- [x] `eZ.DashboardView` plugin that adds a block to a dashboard view instance,
- [x] `eZ.DashboardViewService` plugin required for fetching data used by a block (**optional**).

In a `eZ.DashboardView` plugin a new block can be added by using `addBlock` method: `this.get('host').addBlock(blockIdentifier, blockViewInstance);`

The `this.get('host')` refers to `eZ.Dashboard` view instance.

Any block can be removed from the dashboard by calling: `this.get('host').removeBlock(blockIdentifier);`

Any block can be hidden by calling:
`this.get('host').hideBlock(blockIdentifier);`

Any block can be displayed again by calling:
`this.get('host').showBlock(blockIdentifier);`

**Important:** This will require Behat tests rebuilding as it changes how the dashboard behaves.

**TO DO:**
- [x] create extensibility points for adding/removing/showing/hiding blocks in the dashboard,
- [x] create the dashboard view service,
- [x] create basic dashboard block view class to be extended in specific blocks,
- [x] create basic dashboard block view plugin to be attached to `eZ.DashboardView`. The plugin is a base to be extended in specific blocks.
- [x] create **All content** block as a sample block,
- [x] implement block ordering in the dashboard,
- [x] implement redirection to content editor,
- [x] implement redirection to content preview,
- [ ] JS unit tests


crevillo:
hasn't github a clap icon to add reaction? :)

dpobel:
> The code can be splitted into separate commits/PR if needed. I kept all of the code here for discussion later.

Please try to keep the PR size as small as possible, it's already close to max acceptable size (~1500 changes) in only one commit (!) while you have only covered half of your TODO list and no review has been done.

As I suggested the very first thing to focus on is a dashboard implementation which will accept some blocks. This is can be build in a very similar way to the tabs in the LocationView. So in a first PR, you could do:

1. deprecate `eZ.DashboardView`: add some `console.warn` in the dashboard view module and a note in the description of eZ.DashboardView (example https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/services/plugins/ez-locationsearchplugin.js)
1. create a `eZ.DashboardBlockView` which will serve as a basis for the blocks, it should only a define a 2 attributes:
  * an `identifier`
  * a `priority`
1. create a `eZ.DashboardBlockBasedView` (that way we should avoid most BC breaks), it's a quite simple view with:
  * a `blockViews` attribute
  * a method to add a block view (that makes sure to keep the block sorted by priority)
  * a method to remove a block view using a block identifier
  * a `render` method that renders the view itself with a template and then renders each block and then append the block view container to the DOM
1. make sure `app.views.dashboardView.type` points to the new dashboard implementation

sunpietro:
@dpobel I know this PR is a huge one. I just keep it that size to have all in one place. When we decide the way we should go, then I'll split it into smaller ones.

dpobel:
@sunpietro no, that a bad approach that makes efficient reviews close to impossible.

Resources/public/js/views/ez-dashboardview.js
```diff
@@ -33,35 +39,161 @@ YUI.add('ez-dashboardview', function (Y) {
...
+        },
+
+        /**
+         * Hides a block on the dashboard
+         *
+         * @method hideBlock
+         * @param identifier {String} block identifier
+         * @return {eZ.DashboardView} the view itself
+         */
+        hideBlock: function (identifier) {
```
dpobel: where does this come from ? this feature is not specified anywhere.

Resources/public/js/views/ez-dashboardview.js
```diff
@@ -33,35 +39,161 @@ YUI.add('ez-dashboardview', function (Y) {
...
-             * @default 'http://ez.no/in-product/eZ-Platform'
-             * @readOnly
+             * @attribute blocks
+             * @type Object
+             * @example {
+             *     key1<String>: viewInstance1<Y.eZ.DashboardBlockView>,
+             *     key2<String>: viewInstance2<Y.eZ.DashboardBlockView>,
+             * }
              */
-            iframeSource: {
```
dpobel: for the record, you can not remove an attribute like that, that's a BC break. That's why to easily avoid that, you should create another view implementation for the dashboard.

sunpietro:
@dpobel about the 2. point: each block has content and title so the list above I would add title and content attributes. What do you think?

Resources/public/js/views/ez-dashboardview.js
sunpietro: This comes from the talks with Jince and FT guys.

dpobel:
@sunpietro for the title maybe even if you can imagine some blocks without a title, so I would rather not put it in the base class for now. For the content, it's completely up to each block to handle its own content in its own way so I see no reason to add a `content` attribute in the block base view.

Resources/config/yui.yml
```diff
@@ -292,11 +296,45 @@ system:
...
+                    type: 'template'
+                    path: %ez_platformui.public_dir%/templates/dashboard/allcontent-dashboard-block-row.hbt
+                ez-allcontent-dashboardblockserviceplugin:
+                    requires:
+                        - 'ez-viewservicebaseplugin'
+                        - 'ez-pluginregistry'
+                        - 'ez-contentmodel'
+                        - 'model-list'
+                        - 'promise'
+                    dependencyOf: ['ez-dashboardviewservice']
```
dpobel: we don't do that in PlatformUI since you can directly change the dependencies of `ez-dashboardviewservice`

sunpietro:
@dpobel I don't agree with you about handling content in blocks. Block without content handling is just a template-based view so I see no reason to create any base _class_ for dashboard blocks. Every block have content, no matter whether it is a list of articles, image, dummy text or a link. All of that is a content.
The same goes for title. Blocks should have a title. In case of any developer wants to hide then it might be done by developer itself.
All other sugestions will be applied soon.

lserwatka:
ping @rolandbenedetti 

What is your POV here? Ref. to block title and content.

dpobel:
> Block without content handling is just a template-based view so I see no reason to create any base class for dashboard blocks.

there is a simple reason. This is what you would call an abstract class or an interface in others languages. This is basically to make sure each block has the necessary minimum attributes/properties/methods to be correctly handled. It's more or less to apply https://en.wikipedia.org/wiki/SOLID_(object-oriented_design)

> Every block have content, no matter whether it is a list of articles, image, dummy text or a link. All of that is a content.

Yes every block have content, but that content will be created when rendering. There's no need to force a `content` attribute. If a block needs it, of course it can have one but that won't be the default case as most block will just be a simple view with a template.

> The same goes for title. Blocks should have a title. In case of any developer wants to hide then it might be done by developer itself.

I already explain that in https://github.com/ezsystems/PlatformUIBundle/pull/583#issuecomment-217857131

>  All other sugestions will be applied soon.

>  ping @rolandbenedetti
> What is your POV here? Ref. to block title and content.

I though I was the lead dev of PlatformUI so in charge of technical decisions.

lserwatka:
@dpobel question to Roland has nothing to do with you being a leader. It is feature related question that has impact on the technical implementation, whether both fields are mandatory or not.

dpobel:
@lserwatka not at all, it's a pure technical decision on how to **implement** the blocks.

lserwatka:
@dpobel my question about feature was if some blocks might not have title at all. This will have impact on UI. I was not looking form technical point of view how to implement blocks. I hope this is more clear now.

bdunogier:
If I understand correctly the conversation above, about blocks having or not a title, I want to remind one of the basic rules of agile methodologies: when facing two choices, pick the most open one. In that case, not enforcing a title (or the "content" property) on a block doesn't prevent anyone from having another block that inherits this one, with a title and/or a content property.

dpobel:
@bdunogier thanks, you got the point. Also to complete Bertrand's comment, the fact that our blocks all have a title does not mean our customer will want to have a title by default! In addition, if we want to not repeat ourselves for this title thing, we have others ways than forcing that in the base/interface/abstract class.

sunpietro:
Let's have no convention then.

bdunogier:
> Let's have no convention then.

Not enforcing a convention at that level does not mean we can't have conventions lower down the chain.

sunpietro:
Closing this PR in favour of: ezsystems/PlatformUIBundle#584 (this is the first of 3)

rolandbenedetti:
Sorry for late reply.
To your question @lserwatka , I don't foresee now coming with blocks without titles any time soon, but it could come later. If it doesn't add complexity, I think it's good to keep the door open for blocks without titles, but only if it doesn't add complexity.

I can confirm that I am the wrong person to speak about any technical details on the discussion above :-)

PR #584, opened 2016-05-10
EZP-25760: Implement a block based extensible dashboard

sunpietro:
https://jira.ez.no/browse/EZP-25760

## The new extensible Dashboard
It allows to add new blocks to dashboard by creating plugins applied to the `eZ.DashboardView` instance.
Each block will require its own view followed by a `eZ.DashboardViewService` plugin that handles actions of getting data to blocks. The service plugin is optional.

To add a new block you have to define:
- [x] block view,
- [x] `eZ.DashboardBlockBasedView` plugin that adds a block to a dashboard view instance,
- [x] `eZ.DashboardBlockBasedViewService` plugin required for fetching data used by a block (**optional**).

In a `eZ.DashboardBlockBasedView` plugin a new block can be added by using `addBlock` method: `this.get('host').addBlock(blockViewInstance);`

The `this.get('host')` refers to `eZ.DashboardBlockBasedView` view instance.

Any block can be removed from the dashboard by calling: `this.get('host').removeBlock(blockIdentifier);`

The `blockIdentifier` comes from `blockViewInstance.get('identifier');`

**Important:** This will require Behat tests rebuilding as it changes how the dashboard behaves.

**TO DO:**
- [x] create extensibility points for adding/removing/showing/hiding blocks in the dashboard,
- [x] create the dashboard view service,
- [x] JS unit tests


Resources/public/css/views/dashboard/dashboard.css
```diff
@@ -0,0 +1,26 @@
...
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+.ez-view-dashboardblockbasedview,
+.ez-view-dashboardblockbasedview * {
+    box-sizing: border-box;
+}
+
+.ez-view-dashboardblockbasedview,
+.ez-view-dashboardblockbasedview > .pure-u {
```
dpobel: we try to get rid of PureCSS and for now this is clearly not needed so you should remove the selector with `pure-u` and the corresponding class in the template.

Resources/public/css/views/dashboard/dashboard.css
```diff
@@ -0,0 +1,26 @@
+/**
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+.ez-view-dashboardblockbasedview,
+.ez-view-dashboardblockbasedview * {
```
dpobel: that's a bit wide and not needed for now

Resources/public/css/views/dashboard/dashboard.css
sunpietro: This is required to remove ugly margin from the top of the view container. When we remove PureCSS then we'll remove this line.

Resources/public/css/views/dashboard/dashboard.css
sunpietro: I don't see any problems with that and applying the same CSS property for each CSS selector is far less effective than this.

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -87,6 +87,10 @@ YUI.add('ez-platformuiapp', function (Y) {
                 type: Y.eZ.DashboardView,
                 parent: 'loginFormView',
             },
+            dashboardBlockBasedView: {
```
dpobel: I think you could reuse app.views.dashboardView first because we still have a dashboard view and also because otherwise app.views.dashboardView becomes completely useless since it references Y.eZ.DashboardView which is undefined (ez-dashboardview is not loaded anymore).

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,217 @@
...
+     *
+     * @namespace eZ
+     * @class DashboardBlockBasedView
+     * @constructor
+     * @extends eZ.TemplateBasedView
+     */
+    Y.eZ.DashboardBlockBasedView = Y.Base.create('dashboardBlockBasedView', Y.eZ.TemplateBasedView, [], {
+        initializer: function () {
+            this.after('activeChange', function () {
+                this.after('blocksChange', this._renderBlocks, this);
```
dpobel: that's a bad idea, that basically means the blocks will be rendered twice. Also that's not needed for now, rendering the blocks while rendering the dashboard is enough.

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,217 @@
...
+         * Activates all the blocks to allow communication between blocks and service
+         *
+         * @method _makeBlocksActive
+         * @protected
+         */
+        _makeBlocksActive: function () {
+            var blocks = this.get('blocks');
+
+            Object.keys(blocks).forEach(function (key) {
+                blocks[key].set('active', true);
```
dpobel: actually you should forward the `active` value of the dashboard, not hardcoding the true.

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,217 @@
...
+        /**
+         * Renders blocks on the dashboard
+         *
+         * @method _renderBlocks
+         * @protected
+         */
+        _renderBlocks: function () {
+            var fragment = Y.one(document.createDocumentFragment());
+
+            this._sortBlocks(this.get('blocks')).forEach(Y.bind(function (block) {
```
dpobel: this method does too many things. Also from a test perspective, it would be easier to move the sorting and the `addTarget` call to `addView`. So basically, this method only renders the blocks (as its name suggests).

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,217 @@
...
+                var targets = block.getTargets();
+
+                if (targets.indexOf(this) === -1) {
```
dpobel: you don't need to do that, `addTarget` is already smart enough (and that removes a code branch to test)

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,217 @@
...
+             */
+            dashboardTitle: {
+                value: 'Dashboard'
+            },
+
+            /**
+             * Hash containing information about all blocks in the dashboard
+             *
+             * @attribute blocks
+             * @type Object
```
dpobel: You should store the blocks as an array that would ease the addition of new blocks while not making the removal complicated (just filter the array by identifier).
Also, this attribute should be read only as we want the external developers to use `addView`. That also means we don't need the setter.

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,217 @@
...
+            });
+
+            list.sort(function (a, b) {
+                var priorityA = a.get('priority'),
+                    priorityB = b.get('priority');
+
+                if (priorityA === priorityB) {
+                    // sort alphabetically (from A to B)
+                    return a.get('identifier') < b.get('identifier');
+                } else {
```
dpobel: that's very picky and that also means more unit tests to cover that case. To be honest, I would not cover this case for now (we'll make sure to let enough room between the priority of our blocks)

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,217 @@
...
+         *
+         * @method addBlock
+         * @param view {Y.eZ.BaseDashboardBlockView} dashboard block view instance
+         * @return {Object} the hash of blocks
+         */
+        addBlock: function (view) {
+            var blocks = this.get('blocks'),
+                identifier = view.get('identifier');
+
+            if (!identifier) {
```
dpobel: the base dashboard block view is missing but if you mark the `identifier` as required you can skip this check (again less code to test and to maintain)

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,217 @@
...
+
+            return blocks;
+        },
+
+        /**
+         * Adds a block to the dashboard
+         *
+         * @method addBlock
+         * @param view {Y.eZ.BaseDashboardBlockView} dashboard block view instance
+         * @return {Object} the hash of blocks
```
dpobel: not really needed so better not do that

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,217 @@
...
+            });
+
+            return blocks;
+        },
+
+        /**
+         * Adds a block to the dashboard
+         *
+         * @method addBlock
+         * @param view {Y.eZ.BaseDashboardBlockView} dashboard block view instance
```
dpobel: `eZ.DashboardBlockBaseView` so that everything dashboard related starts with eZ.Dashboard (easier in API doc)

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,217 @@
...
+         */
+        addBlock: function (view) {
+            var blocks = this.get('blocks'),
+                identifier = view.get('identifier');
+
+            if (!identifier) {
+                return blocks;
+            }
+
+            if (blocks[identifier]) {
```
dpobel: I would not check that as well because the identifier represents an implementation but you can imagine usecases where you want 2 instances of the same block implementations and again this check is not needed now, if it happens to be, we'll add later.

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,217 @@
...
+         * @param identifier {String} block identifier
+         * @param preventUpdating {Boolean} should the `blocks` attribute value be updated?
+         * @return {Object} the hash of blocks
+         */
+        removeBlock: function (identifier, preventUpdating) {
+            var blocks = this.get('blocks'),
+                block = blocks[identifier];
+
+            block.removeTarget(this);
+            block.destroy({remove: true});
```
dpobel: IMHO, this should not be done in this method. for consistency this method should do exactly the opposite of `addBlock` so remove the block from the list and call `removeTarget`. The proper destruction of the block view is up to the user of this method.

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,217 @@
...
+
+            delete blocks[identifier];
+
+            if (!preventUpdating) {
```
dpobel: I don't really get the use case for that flag. API wise you end up with a `removeBlock` method that does not what it is supposed to do according to its name. With https://github.com/ezsystems/PlatformUIBundle/pull/584/files#r62699537 taken into account, this parameter becomes useless anyway.

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,217 @@
...
+        },
+
+        /**
+         * Renders the dashboard view
+         *
+         * @method render
+         * @return {eZ.DashboardBlockBasedView} the view itself
+         */
+        render: function () {
+            this.get('container').setHTML(this.template({dashboardTitle: this.get('dashboardTitle')}));
```
dpobel: Why not directly write "Dashboard" in the template and get rid of the `dashboardTitle` attribute ? Again, it's to simplify the code so the tests and the maintenance.

Resources/public/js/views/services/ez-dashboardblockbasedviewservice.js
```diff
@@ -0,0 +1,190 @@
...
+     * The dashboard block based view service
+     *
+     * @namespace eZ
+     * @class DashboardBlockBasedViewService
+     * @constructor
+     * @extends eZ.ViewService
+     */
+    Y.eZ.DashboardBlockBasedViewService = Y.Base.create('dashboardBlockBasedViewService', Y.eZ.ViewService, [], {
+        _load: function (next) {
+            this._loadRootLocation()
```
dpobel: not needed for now so better to skip that for now. (Yes the view service will be empty)

dpobel:
Overall this can be simplified a lot and that will make the unit tests super simple. Also as mentioned above, the base dashboard block view is missing.

Resources/public/css/views/dashboard/dashboard.css
dpobel: no you remove the `pure-u` class which is not supposed to be used outside of a `pure-g` (in that sense it's a hack). I guess the margin issue actually comes from the top margin of the `h1`, so normally you can fix that way.

Resources/public/css/views/dashboard/dashboard.css
dpobel: is it needed to set the box-sizing on everything at moment ?

sunpietro:
@dpobel the base dashboard block view will be in separate PR. The same will happen with All Content Dashboard block. There will be 2 additional PR.

sunpietro:
@dpobel you don't have to write any stories. I created subtasks here: https://jira.ez.no/browse/EZP-25760 so it's already done.

Resources/public/js/views/services/ez-dashboardblockbasedviewservice.js
sunpietro: I don't agree. If you want to fetch all content then you'll have to write it over and over again. This is basic information and it has to be stored here. For multiple blocks you'll have to run it for each block separately. It's better to load it once and keep it here.

Resources/public/js/views/ez-dashboardblockbasedview.js
sunpietro: Why? If you want to translate the title, then you have to store it in some kind of variable. If you hardcode the `Dashboard` text inside the template, then it will be impossible.

Resources/public/js/views/ez-dashboardblockbasedview.js
sunpietro: The identifier should be unique. If someone wants to use 2 instances of the same block view, then the second instance should get a different identifier.

Resources/public/css/views/dashboard/dashboard.css
sunpietro: good point

Resources/public/css/views/dashboard/dashboard.css
sunpietro: Yes, it is. I want to have all elements inside the dashboard to behave the same way.

Resources/public/js/views/ez-dashboardblockbasedview.js
dpobel: not at all, we have a lots of cases where we can not really get rid of the "template only" strings. And that's why, in most projects, to translate strings in template, there's a template operator/function. This is most likely what we'll do as well.

So there's absolutely no problem in writing "Dashboard" in the template and even better, this removes some code.

Resources/public/js/views/services/ez-dashboardblockbasedviewservice.js
dpobel: At this step, it's too early to decide that. At the moment, those Locations are used any where => useless code => remove code.

dpobel:
> @dpobel the base dashboard block view will be in separate PR. The same will happen with All Content Dashboard block. There will be 2 additional PRs.

No. The dashboard block base view is definitively part of this PR since it's mentioned in the documentation of the `addView` method and the `blocks` attribute. This is the core of the feature.

> @dpobel you don't have to write any stories. I created subtasks here: https://jira.ez.no/browse/EZP-25760 so it's already done.

not this split is not correct, this is not how we work. The EPIC has now a set a stories, please follow them. They are split so that each story has a very limited scope and so that 1 story => 1 (reviewable) pull request.

Resources/public/js/views/ez-dashboardblockbasedview.js
sunpietro: How can you set an attribute as required? I can only set a default value.

Resources/public/css/views/dashboard/dashboard.css
dpobel: why ? you don't set any size nor padding => useless => removed.

Resources/public/js/views/ez-dashboardblockbasedview.js
dpobel: You don't set the attribute as required, you document it as required with `@required`, then if the external dev does not fill it, it's his problem not ours.

Resources/public/js/views/ez-dashboardblockbasedview.js
dpobel: good point, without unique identifier, `removeBlock` will have problems :) that said, I would still not check that in `addView` but if you want to keep it just make sure it's properly unit tested.

Resources/public/js/views/ez-dashboardblockbasedview.js
sunpietro: In this case this might lead to errors. See the block-based dashboard view is named `eZ.DashboardBlockBasedView` and the base dashboard block view you want to name: `eZ.DashboardBlockBaseView`

Resources/public/css/views/dashboard/dashboard.css
sunpietro: @dpobel the CSS is updated and now you should get better sense of why I need this.

sunpietro:
ping @dpobel

bdunogier:
Not sure this has been said before, but please don't amend commits when applying review requests. It makes it very complicated to track down what was changed as the code was rebase.

We now tend to use `git commit --fixup`, in order to fixup the relevant commit, and rebase with autosquash when the review is done.

Resources/config/css.yml
```diff
@@ -90,6 +90,8 @@ system:
                 - 'bundles/ezplatformui/css/views/fields/edit/richtext.css'
                 - 'bundles/ezplatformui/css/views/fields/edit/isbn.css'
                 - 'bundles/ezplatformui/css/views/fields/edit/user.css'
+                - 'bundles/ezplatformui/css/views/dashboard/dashboard.css'
+                - 'bundles/ezplatformui/css/views/dashboard/base-dashboard-block.css'
```
dpobel: we'll never instantiate the base block view so this file is not needed.

Resources/config/yui.yml
```diff
@@ -204,6 +205,9 @@ system:
                 ez-loginformviewservice:
                     requires: ['ez-viewservice']
                     path: %ez_platformui.public_dir%/js/views/services/ez-loginformviewservice.js
+                ez-dashboardblockbasedviewservice:
+                    requires: ['ez-viewservice', 'ez-locationmodel']
```
dpobel: `'ez-locationmodel'` not needed

Resources/config/yui.yml
```diff
@@ -297,6 +301,23 @@ system:
...
                     type: 'template'
                     path: %ez_platformui.public_dir%/templates/dashboard.hbt
+                ez-dashboardblockbasedview:
+                    requires:
+                        - 'ez-templatebasedview'
+                        - 'dashboardblockbasedview-ez-template'
+                    path: %ez_platformui.public_dir%/js/views/ez-dashboardblockbasedview.js
+                dashboardblockbasedview-ez-template:
+                    type: 'template'
+                    path: %ez_platformui.public_dir%/templates/dashboard/dashboard.hbt
```
dpobel: to be consistent with the path of the view, this template should not be in the dashboard directory (but indeed this directory will be useful the block templates).

Resources/config/yui.yml
```diff
@@ -297,6 +301,23 @@ system:
...
+                ez-base-dashboardblockview:
```
dpobel: as stated in a previous comment, since the view is called `eZ.DashboardBlockBaseView` the corresponding module should be called `ez-dashboardblockbaseview`

Resources/config/yui.yml
```diff
@@ -297,6 +301,23 @@ system:
...
+                    requires: ['ez-templatebasedview', 'basedashboardblockview-ez-template']
```
dpobel: no template for this view, it will never be instantiated directly

Resources/config/css.yml
```diff
@@ -90,6 +90,8 @@ system:
                 - 'bundles/ezplatformui/css/views/fields/edit/richtext.css'
                 - 'bundles/ezplatformui/css/views/fields/edit/isbn.css'
                 - 'bundles/ezplatformui/css/views/fields/edit/user.css'
+                - 'bundles/ezplatformui/css/views/dashboard/dashboard.css'
```
dpobel: same remark as in https://github.com/ezsystems/PlatformUIBundle/pull/584/files#r62842220, the path of the view and its stylesheets (and the tests) should be consistent. So no `dashboard/` level here.

Resources/public/css/views/dashboard/dashboard.css
dpobel: it's not that I don't understand. I'm saying that this is not needed **right now** and it is still not (for now). Again, the point of this story/PR is to setup the extensibility point for the dashboard. Don't worry, you'll have time to add some CSS but only when there's something to style.

Resources/public/js/views/dashboard/ez-base-dashboardblockview.js
```diff
@@ -0,0 +1,59 @@
...
+     */
+    Y.eZ.DashboardBlockBaseView = Y.Base.create('dashboardBlockBaseView', Y.eZ.TemplateBasedView, [], {
+        /**
+         * Renders the dashboard block view
+         *
+         * @method render
+         * @return {eZ.DashboardBlockView} the view itself
+         */
+        render: function () {
+            this.get('container').setHTML(this.template(this.getAttrs()));
```
dpobel: this view will not be rendered so no render here. Also, the rendering logic is completely up to each implementation.

Resources/public/js/views/dashboard/plugins/ez-base-dashboardblockviewplugin.js
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * The dashboard base block view plugin
+     *
+     * @namespace eZ.Plugin
+     * @class DashboardBlockBaseView
+     * @constructor
+     * @extends Base
+     */
+    Y.eZ.Plugin.DashboardBlockBaseView = Y.Base.create('baseDashboardBlockViewPlugin', Y.Base, [], {
```
dpobel: what's the point this plugin ? no needed and a bit off the point with its `block` attribute instantiating the base class.

sunpietro: The reason I created this plugin is to show how the block can be added to dashboard by third-party developers.
This way, I provide the way to add new blocks to dashboard easily. Third-party developers will just extend this plugin and set different `block` attribute value to contain the block view instance they want to add.
So this is the extension point.

sunpietro:
@dpobel Should I write empty tests for empty classes like `eZ.DashboardBlockBasedView`, `eZ.DashboardBlockBaseView`, etc?

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,167 @@
...
+     * The dashboard view
+     *
+     * @namespace eZ
+     * @class DashboardBlockBasedView
+     * @constructor
+     * @extends eZ.TemplateBasedView
+     */
+    Y.eZ.DashboardBlockBasedView = Y.Base.create('dashboardBlockBasedView', Y.eZ.TemplateBasedView, [], {
+        initializer: function () {
+            this.onceAfter('activeChange', function (event) {
```
dpobel: this is still not good, the blocks will still be rendered twice. Just remove that. It's up to each block to decide if it needs to do something when active is changing.

sunpietro: Twice? I don't see any case like that. Can you explain this case a little bit more?

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,167 @@
...
+         * Sorts the provided list of blocks by priority.
+         *
+         * @method _sortBlocks
+         * @protected
+         * @param blocks {Array} list of blocks
+         * @return {Array} sorted list of blocks
+         */
+        _sortBlocks: function (blocks) {
+            return blocks.sort(function (a, b) {
+                return a.get('priority') > b.get('priority');
```
dpobel: according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort the compare function should return a number not a boolean so you should return `b.get('priority') - a.get('priority')` instead (higher priority first).

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,167 @@
...
+         *
+         * @method addBlock
+         * @param view {Y.eZ.DashboardBlockBaseView} dashboard block view instance
+         * @return {Array} the list of blocks
+         */
+        addBlock: function (view) {
+            var blocks = this.get('blocks'),
+                identifier = view.get('identifier');
+
+            if (!identifier || !(view instanceof Y.eZ.DashboardBlockBaseView)) {
```
dpobel: this condition is still not needed. let's consider the identifier is required and the passed view is a block view.

Resources/public/js/views/dashboard/ez-base-dashboardblockview.js
```diff
@@ -0,0 +1,59 @@
...
+
+            /**
+             * The block identifier
+             *
+             * @attribute identifier
+             * @type String
+             * @default 'dashboard-base-block'
+             */
+            identifier: {
+                value: 'dashboard-base-block'
```
dpobel: don't put any default value but add the tag `@required` on the attribute

Resources/public/js/views/dashboard/plugins/ez-base-dashboardblockviewplugin.js
dpobel: that's not a bad idea but that's dead code and this is not needed right now again don't add such code. Also if that was really needed, this should have been tested.

Resources/public/js/views/ez-dashboardblockbasedview.js
dpobel: first the blocks are rendered in render, then the dashboard becomes active, the blocks are rendered again.

sunpietro: No, you're wrong. I just add the listener then.

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,167 @@
...
+                return a.get('priority') > b.get('priority');
+            });
+        },
+
+        /**
+         * Adds a block to the dashboard
+         *
+         * @method addBlock
+         * @param view {Y.eZ.DashboardBlockBaseView} dashboard block view instance
+         * @return {Array} the list of blocks
```
dpobel: still the same as https://github.com/ezsystems/PlatformUIBundle/pull/584#discussion_r62698596 don't return anything, if you want the block you can use the blocks attribute.

Resources/public/js/views/ez-dashboardblockbasedview.js
dpobel: @sunpietro I missed that comment and that's a very valid point. Maybe we can call the new dashboard implementation `eZ.DashboardBlocksView` ?

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,167 @@
...
+
+            return blocks;
+        },
+
+        /**
+         * Finds block with a given identifier
+         *
+         * @method _findBlocks
+         * @param identifier {String} block identifier
+         * @return {Y.eZ.DashboardBlockBaseView} a block view
```
dpobel: `@return {eZ.DashboardBlockBaseView|Null}`

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,167 @@
...
+         * @method removeBlock
+         * @param identifier {String} block identifier
+         * @return {Array} the list of blocks
+         */
+        removeBlock: function (identifier) {
+            var blocks = this.get('blocks'),
+                existingBlock = this._findBlock(identifier);
+
+            existingBlock.removeTarget(this);
+            existingBlock.destroy({remove: true});
```
dpobel: still to be removed https://github.com/ezsystems/PlatformUIBundle/pull/584#discussion_r62700715

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,167 @@
...
+                return block.get('identifier') === identifier;
+            })[0];
+        },
+
+        /**
+         * Removes a block from the dashboard
+         *
+         * @method removeBlock
+         * @param identifier {String} block identifier
+         * @return {Array} the list of blocks
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/584#discussion_r62698596 and https://github.com/ezsystems/PlatformUIBundle/pull/584/files#r62849728 don't return anything.

Resources/public/js/views/ez-dashboardblockbasedview.js
sunpietro: I don't agree. If you remove that block then you don't need it anywhere else. Anyway, I'll remove this line.

Resources/public/js/views/ez-dashboardblockbasedview.js
```diff
@@ -0,0 +1,167 @@
...
+
+            return blocks;
+        }
+    }, {
+        ATTRS: {
+            /**
+             * An array dashboard blocks view instances
+             *
+             * @attribute blocks
+             * @type Array
```
dpobel: missing the `@readOnly` tag and maybe a warning in the description saying that to manipulate, you have to use `addBlock` and `removeBlock`.

Resources/public/templates/dashboard/dashboard.hbt
```diff
@@ -0,0 +1,2 @@
+<h1 class="dashboard-title">Dashboard</h1>
```
dpobel: please prefix the classes with `ez-`.

Tests/js/views/assets/ez-dashboardblockbasedview-tests.js
```diff
@@ -0,0 +1,222 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-dashboardblockbasedview-tests', function (Y) {
+    var renderTest,
+        eventsTest,
+        publicMethodsTest,
+        VIEW_CONFIG = {container: '.container'},
```
dpobel: nitpick: that's not even needed

Tests/js/views/assets/ez-dashboardblockbasedview-tests.js
```diff
@@ -0,0 +1,222 @@
...
+            var templateCalled = false,
+                origTpl = this.view.template;
+
+            this.view.template = function () {
+                templateCalled = true;
+
+                return origTpl.apply(this, arguments);
+            };
+
+            this.view._set('blocks', [
```
dpobel: use `addBlock`, it's meant for that.

Tests/js/views/assets/ez-dashboardblockbasedview-tests.js
```diff
@@ -0,0 +1,222 @@
...
+            ]);
+
+            this.view.render();
+
+            Y.Assert.isTrue(templateCalled, 'The template should have been used to render view');
+            Y.Assert.areSame(1, this.view.get('container').one(SELECTOR_CONTENT).get('children').size(), 'Should render one block inside');
+        }
+    });
+
+    publicMethodsTest = new Y.Test.Case({
```
dpobel: please write one test case per public method so one for `render`, one for `addBlock`, one for `removeBlock`.

Tests/js/views/assets/ez-dashboardblockbasedview-tests.js
sunpietro: No, not in this case.

Resources/public/js/views/ez-dashboardblockbasedview.js
dpobel: yes I just realized, I misread the following line :)
But I think we should not try to do that this way. We should only cover the case where the block are added before the rendering of the dashboard (ie what happens when you add the block in the initializer of a plugin). I don't see any usecase where you would need to add a block after that point.

Tests/js/views/assets/ez-dashboardblockbasedview-tests.js
dpobel: what ? isn't addBlock supposed to add a block to the list ?

sunpietro: It covers the case when the dashboard view is initialized with predefined blocks.

Tests/js/views/assets/ez-dashboardblockbasedview-tests.js
```diff
@@ -0,0 +1,222 @@
...
+            blocks = view.addBlock(block1);
+
+            Y.Assert.areNotSame(view, block1.getTargets()[0], 'The dashboard view should be a target of a first block');
+            Y.Assert.areSame(0, view.get('container').one(SELECTOR_CONTENT).get('children').size(), 'Should render one block inside');
+            Y.Assert.isArray(blocks, 'The `addBlock` method should return an array of blocks');
+            Y.Assert.areSame(0, blocks.length, 'The `addBlock` method should return an array of blocks containing 2 blocks');
+        },
+    });
+
+    eventsTest = new Y.Test.Case({
```
dpobel: events ? which events ? here it seems you are testing that the `active` flag is correctly forwarded so it could be named `forwardActiveTest`.

Tests/js/views/assets/ez-dashboardblockbasedview-tests.js
```diff
@@ -0,0 +1,222 @@
...
+            view.addBlock(block1);
+            view.set('active', false);
+
+            Y.Assert.isFalse(block1.get('active'), 'A block view instance should get inactive');
+        }
+    });
+
+    Y.Test.Runner.setName('eZ Dashboard Block Based View tests');
+    Y.Test.Runner.add(renderTest);
+    Y.Test.Runner.add(publicMethodsTest);
```
dpobel: `eventsTest` is missing

Tests/js/views/ez-dashboardblockbasedview.html
```diff
@@ -0,0 +1,50 @@
...
+<html>
+<head>
+<meta charset="utf-8">
+<title>eZ Dashboard Block Based view tests</title>
+</head>
+<body>
+<div class="container"></div>
+
+<script type="text/x-handlebars-template" id="dashboardblockbasedview-ez-template" >
+    <h1 class="dashboard-title">Dashboard</h1>
```
dpobel: please always put the minimal needed template for the test to run and not a copy paste of the "normal" template, so here you can remove the `h1`

Tests/js/views/assets/ez-dashboardblockbasedview-tests.js
dpobel: and ? addBlock should be usable before rendering. Actually, this is connected to this thread above https://github.com/ezsystems/PlatformUIBundle/pull/584#discussion_r62852782 and as written there using it before rendering is the main use case for now.

Remove the `blockChange` event handler, and you'll see, this will become easier. (that's exactly the point of all my comments on removing not necessary code...).

dpobel:
> @dpobel Should I write empty tests for empty classes like eZ.DashboardBlockBasedView, eZ.DashboardBlockBaseView, etc?

for `eZ.DashboardBlockBasedView` it's not empty, you wrote it :) For `eZ.DashboardBlockBaseView` as long as we only have statements, it's not needed to write a test for it. if we introduce code in it, then a test will be needed.

sunpietro:
ping @dpobel 
I hope I didn't miss anything

sunpietro:
ping @dpobel

Resources/public/js/views/dashboard/plugins/ez-dashboardblockbaseviewplugin.js
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * The dashboard base block view plugin
+     *
+     * @namespace eZ.Plugin
+     * @class DashboardBlockBaseView
+     * @constructor
+     * @extends Base
+     */
+    Y.eZ.Plugin.DashboardBlockBaseView = Y.Base.create('baseDashboardBlockViewPlugin', Y.Base, [], {
```
dpobel: you got me wrong it seems, this plugin is outside of the scope of EZP-25760 so please remov it from  the PR

Resources/config/css.yml
```diff
@@ -90,6 +91,7 @@ system:
                 - 'bundles/ezplatformui/css/views/fields/edit/richtext.css'
                 - 'bundles/ezplatformui/css/views/fields/edit/isbn.css'
                 - 'bundles/ezplatformui/css/views/fields/edit/user.css'
+                - 'bundles/ezplatformui/css/views/dashboard/base-dashboard-block.css'
```
dpobel: this stylesheet is useless at the moment, there's no block to style

Resources/config/yui.yml
```diff
@@ -297,6 +301,23 @@ system:
...
+                ez-dashboardblocksview:
+                    requires:
+                        - 'ez-templatebasedview'
+                        - 'dashboardblocksview-ez-template'
+                    path: %ez_platformui.public_dir%/js/views/ez-dashboardblocksview.js
+                dashboardblocksview-ez-template:
+                    type: 'template'
+                    path: %ez_platformui.public_dir%/templates/dashboard-blocks.hbt
+                ez-dashboardblockbaseview:
+                    requires: ['ez-templatebasedview', 'dashboardblockbaseview-ez-template']
```
dpobel: again the base block is not supposed to be rendered, so it does not need a template. Please remove the template from the PR.

Resources/public/css/views/dashboard-blocks.css
```diff
@@ -0,0 +1,35 @@
...
+}
+
+.ez-view-dashboardblocksview .ez-dashboard-content {
+    display: flex;
+    justify-content: space-between;
+    flex-direction: row;
+    flex-wrap: wrap;
+}
+
+.ez-view-dashboardblocksview .ez-view-dashboardblockbaseview {
```
dpobel: the base block is not supposed to be rendered so `ez-view-dashboardblockbaseview` is not supposed to be used anywhere.

Resources/public/css/views/dashboard-blocks.css
```diff
@@ -0,0 +1,35 @@
+/**
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+.ez-view-dashboardblocksview,
+.ez-view-dashboardblocksview * {
```
dpobel: again, the wildcard selector is still not needed

Resources/public/js/views/ez-dashboardblocksview.js
```diff
@@ -0,0 +1,159 @@
...
+     *
+     * @namespace eZ
+     * @class DashboardBlocksView
+     * @constructor
+     * @extends eZ.TemplateBasedView
+     */
+    Y.eZ.DashboardBlocksView = Y.Base.create('dashboardBlocksView', Y.eZ.TemplateBasedView, [], {
+        initializer: function () {
+            this.onceAfter('activeChange', function (event) {
+                this.after('blocksChange', this._renderBlocks, this);
```
dpobel: as I was writing in https://github.com/ezsystems/PlatformUIBundle/pull/584#discussion_r62852782, please remove that, for now we'll only cover the case where the block are added before rendering. We'll see later if we need more than that.

Resources/public/js/views/ez-dashboardblocksview.js
```diff
@@ -0,0 +1,159 @@
...
+            var blocks = this.get('blocks'),
+                existingBlock = this._findBlock(identifier);
+
+            existingBlock.removeTarget(this);
+
+            blocks.splice(blocks.indexOf(existingBlock), 1);
+
+            this._set('blocks', blocks);
+
+            return blocks;
```
dpobel: you forgot to remove the return statement with the doc update.

Tests/js/views/assets/ez-dashboardblocksview-tests.js
```diff
@@ -0,0 +1,217 @@
...
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-dashboardblocksview-tests', function (Y) {
+    var renderTest,
+        forwardActiveTest,
+        addBlockTest,
+        removeBlockTest,
+        VIEW_CONFIG = {container: '.container'},
+        SELECTOR_CONTENT = '.dashboard-content',
```
dpobel: should be `.ez-dashboard-content`, because of that the tests are failing it seems https://travis-ci.org/ezsystems/PlatformUIBundle/jobs/129451719#L383

Tests/js/views/assets/ez-dashboardblocksview-tests.js
```diff
@@ -0,0 +1,217 @@
...
+            this.view.template = function () {
+                templateCalled = true;
+
+                return origTpl.apply(this, arguments);
+            };
+
+            this.view.render();
+
+            Y.Assert.isTrue(templateCalled, 'The template should have been used to render view');
+        }
```
dpobel: You should have a test where you check that the blocks are rendered as well (this is what `render` does).

Tests/js/views/assets/ez-dashboardblocksview-tests.js
```diff
@@ -0,0 +1,217 @@
...
+    });
+
+    addBlockTest = new Y.Test.Case({
+        name: 'eZ Dashboard Blocks View add block tests',
```
dpobel: the tests in this test case are wrongly written. You are testing `addBlock` not `render`, so you should basically get 3 tests here:
* `Should add the block to the list` where you add a block and assert it is in the `blocks` attribute, its bubble target was set to the dashboard view and the active flag value was on the block.
* `Should remove an existing block with the same identifier` to check the case where you add a block with an existing identifier (I would to not have some code for that, but you insist on having, so you test to it explicitly)
* `Should keep the block list sorted` where you add several block with different priorities to make sure the sorting is done

Tests/js/views/assets/ez-dashboardblocksview-tests.js
```diff
@@ -0,0 +1,217 @@
...
+                    priority: 1,
+                    identifier: 'A'
+                },
+                block1 = new Y.eZ.DashboardBlockBaseView(block1Config),
+                blocks;
+
+            view.set('active', true);
+            blocks = view.addBlock(block1);
+            blocks = view.removeBlock(block1Config.identifier);
+
```
dpobel: an assertion about the `removeTarget` is missing.

dpobel:
@sunpietro still too many things in this PR but that's better. Also please, take @bdunogier's comment into account https://github.com/ezsystems/PlatformUIBundle/pull/584#issuecomment-218449691 (ie do not amend yet your commit).

Tests/js/views/assets/ez-dashboardblocksview-tests.js
sunpietro: The rendering test when blocks are added is done within test case for adding blocks.

Tests/js/views/assets/ez-dashboardblocksview-tests.js
sunpietro: There should be also the test for testing whether the it doesn't render blocks with the same identifiers. So 4 tests. All of them are done.

sunpietro:
ping @dpobel

Tests/js/views/assets/ez-dashboardblocksview-tests.js
dpobel: Stop! Please read the code you wrote:
```js
        render: function () {
            this._renderBlocks();

            return this;
},
```
 the `render` methods does 2 things:
1. render the view it self
2. renders the blocks in a specific container. **This is not tested**

sunpietro: It is tested in the test case below. So, maybe it would be better idea to remove the render test case?

Tests/js/views/assets/ez-dashboardblocksview-tests.js
dpobel: `addBlock` does not render anything. **It adds a block in the list**, that why the tests should NOT be render centric but should focus on what happens in the `blocks` attribute. There should be no call to render at all in those tests.

sunpietro:
ping @dpobel

Tests/js/views/dashboard/plugins/ez-dashboardblockbaseviewplugin.html
```diff
@@ -0,0 +1,33 @@
+<!doctype html>
```
dpobel: Travis CI fails because you removed the plugin but not the corresponding test.

Tests/js/views/assets/ez-dashboardblocksview-tests.js
```diff
@@ -0,0 +1,204 @@
...
+            Y.Assert.areSame(view, block1.getTargets()[0], 'The dashboard view should be a target of a first block');
+            Y.Assert.areSame(view, block2.getTargets()[0], 'The dashboard view should be a target of a second block');
+            Y.Assert.areSame(viewActiveState, block1.get('active'), 'A first block should have the same active state as dashbaord view');
+            Y.Assert.areSame(viewActiveState, block2.get('active'), 'A second block should have the same active state as dashbaord view');
+            Y.Assert.areSame(2, blocks.length, 'There should be 2 block view instances available');
+
+            // PhantomJS 1.9.8 swap params in `sort` function of Array.
+            // This is fixed in PhantomJS 2.0
+            // See: https://github.com/ariya/phantomjs/issues/11090
+            // if (!Y.UA.phantomjs) {
```
dpobel: don't comment code, if that's not needed, remove the code

sunpietro:
ping @dpobel

Gruntfile.js
```diff
@@ -9,6 +9,8 @@ module.exports = function(grunt) {
             "./Resources/public/js/apps/extensions/*.js",
             "./Resources/public/js/apps/plugins/*.js",
             "./Resources/public/js/views/*.js",
+            "./Resources/public/js/views/dashboard/*.js",
+            "./Resources/public/js/views/dashboard/plugins/*.js",
```
dpobel: this folder is not there (yet)

Resources/public/templates/dashboard/base-dashboard-block.hbt
```diff
@@ -0,0 +1 @@
+{{ identifier }}
```
dpobel: this file should be removed as well.

dpobel:
besides the last 2 comments, this looks good. +1 when those comments are taken into account.
@StephaneDiot can you please have a look ?

dpobel:
/!\ self note: do not merge when having the necessary +1, but create the feature branch for the following stories.

sunpietro:
@dpobel should I ammend commits now?

sunpietro:
Also, please create a feature branch. I'm going to make a PR with _Everyone's Content_ block.

StephaneDiot:
+1

dpobel:
@sunpietro I can not create the feature before the review is finalized. Now that, it's done, I can do it.

So I created `feature/ezp-25766_dashboard` ([Diff](https://github.com/ezsystems/PlatformUIBundle/compare/feature/ezp-25766_dashboard)). So the following pull requests for this EPIC should be done against this branch.

dpobel:
Closing, we'll merge `feature/ezp-25766_dashboard` when it contains a useful feature set.

PR #585, opened 2016-05-11
Fix EZP-25672: Content Type controller should use Languages more wisely

glye:
> Fixes https://jira.ez.no/browse/EZP-25672
> Status: Ready for review

Ensure that the language we use exists in the content type.

- [x] Content type view: Uses the first of specified/prioritised languages that exist in the content type
- [x] Content type edit: Uses the first of specified/prioritised languages that exist in the content type
  - [x] ~~Possible improvement: Add language if it doesn't exist in the content type~~ Postponing until we have language selection https://jira.ez.no/browse/EZP-24451
- [x] ~~Content type group view: Uses the first of prioritised languages that exist in each content type~~ No change needed here.
- [x] Content type create: Use specified language, or first prioritised language if none is specified. (Does not check if language exists in repo.)
  - [x] ~~Possible improvement: Give user a choice of existing repo languages~~ See https://jira.ez.no/browse/EZP-24451

Migrated from https://github.com/ezsystems/PlatformUIBundle/pull/550 on master.

### DIFF
--- a/Controller/ContentTypeController.php
+++ b/Controller/ContentTypeController.php
@@ -278,7 +278,9 @@ class ContentTypeController extends Controller
         try {
             $contentTypeDraft = $this->contentTypeService->loadContentTypeDraft($contentTypeId);
         } catch (NotFoundException $e) {
-            $contentTypeDraft = $this->contentTypeService->createContentTypeDraft($contentType);
+            $contentTypeDraft = $this->contentTypeService->createContentTypeDraft(
+                $this->contentTypeService->loadContentType($contentTypeId)
+            );
         }
 
         if (!isset($languageCode) || !isset($contentTypeDraft->names[$languageCode])) {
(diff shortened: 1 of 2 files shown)


glye:
@andrerom @bdunogier Moved to 1.3, and micro-optimisation accomplished :)

bdunogier:
You could have `git rebase -i --autosquash` _before_ reopening ;-)

And github could also have that as a merge option. Can you do it please ? Then we'll merge.

glye:
I could, but since we can do that on merge now I didn't bother. Whatevz, coming up...

andrerom:
+1, and yes we can do that on merge now so not a must that you do it as long as end result is supposed to be 1 commit, in cases where there should be several commits, then you'll have to take care about it ;)

glye:
Ok, then I have to since this should be 2 commits.

bdunogier:
> And github could also have that as a merge option.

I did mean by that that github can't do autosquash (it can squash everything into one commit, as André said, but it is not what we want here).

bdunogier:
Merged, I'll take care of merging to master.

PR #586, opened 2016-05-11
Add README instructions on how to test new asset builds

andrerom:


README.md
```diff
@@ -131,6 +131,19 @@ interesting are:
...
+  
+* To test updating custom AlloyEditor skin against custom platform-ui-assets-bundle builds:
+
+  ```bash
+  $ composer install --prefer-sopurce
+  $ cd vendor/ezsystems/platform-ui-assets-bundle/
+  $ git checkout <your_branch>
+  $ ./bin/prepare_release.sh -v X.Y.X
+  $ git checkout vX.Y.X
+  $ echo -e `cd Resources/public/vendors/alloy-editor/ && npm install`
```
dpobel: echo -e ? why not just `cd Resources/public/vendors/alloy-editor/ && npm install` ?

dpobel:
besides the nitpick, +1

README.md
andrerom: I wanted to make sure next line works, so placed it in subshell and added echo to make sure output is displayed :)

suggestions on other ways welcome, after all you are the shell master here :)

dpobel: lol thanks for the new title ;)
oh yeah the `cd -` that's a good reason, then that's fine

andrerom: BTW, while I remember it: be aware these instructions will fail if `<your_branch>` contains changes to for instance `package.json` because `./bin/prepare_release.sh` always swaps to master and builds from it. It didn't cause problems for me as the script was already loaded when this was done, but for changes in `package.json` it would.

PR #588, opened 2016-05-12
[test] disable stash in memory cache for behat

bdunogier:
I feel like there is a possibility that some of our errors come from in-memory persistence cache: the behat process is a long running one. In-memory cache will remain for the duration of the execution, and won't be cleared by external events such as everything that is done on the UI.

The goal is to estimate the stability when disabling it.

### DIFF
--- a/bin/travis/prepare_behat.sh
+++ b/bin/travis/prepare_behat.sh
@@ -7,7 +7,7 @@ git fetch --unshallow && git checkout -b tmp_travis_branch
 cd "$HOME/build"
 
 # Checkout meta repo, change the branch and/or remote to use a different ezpublish branch/distro
-git clone --depth 1 --single-branch --branch master https://github.com/ezsystems/ezplatform.git
+git clone --depth 1 --single-branch --branch behat_disable_in_memory_cache https://github.com/ezsystems/ezplatform.git
 cd ezplatform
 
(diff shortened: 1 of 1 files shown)


joaoinacio:
On a side note, is there a stash issue for the unrelated rmdir warning?

```
 Warning: rmdir(app/cache/behat/stash/0fea6a13c52b4d47/25368f24b045ca84/c21f969b5f03d33d
/43e04f8f136e7682/4a1c6be177996f9e/03934ae1c1c02ffc): Directory not empty in vendor/tedivm
/stash/src/Stash/Utilities.php line 170
```

miguelcleverti:
I've only seem this issue in travis, never locally on my machines.

bdunogier:
> IIRC mostly in travis
> I've only seem this issue in travis

So, mostly, or only ? :)

joaoinacio:
> So, mostly, or only ? :)
Not certain about local tbh, can only say I've definitely seen this around. stash 0.13 .1seems to try to address this with https://github.com/tedious/Stash/commit/0c8742af6d9a583866c82fb015c24399dd0fb9f7 , though they do refer to hhvm the error suppressor `@` should still apply

andrerom:
Didn't pass from what I could see, reopen if still relevant.

PR #589, opened 2016-05-13
EZP-25767: Implement the Everyone's Content dashboard block

sunpietro:
https://jira.ez.no/browse/EZP-25767

**TO DO:**
- [x] create a new block view - `Y.eZ.DashboardBlockAllContentView`,
- [x] create a view for each row in a block  - `Y.eZ.DashboardBlockAllContentRowView`,
- [x] add a block view to dashboard,
- [x] display row options after clicking on a row,
- [x] redirect to content location view,
- [x] redirect to content edit view,
- [x] JS unit tests


sunpietro:
ping @dpobel 
I have some issues with displaying the loading state inside the block. Can you give me some suggestions on it?

sunpietro:
One comment about the block template: the reason I used divs instead of table is that row options were misaligned when trying to position them absolutely within a div inside the table cell. This approach solves that issue.

sunpietro:
Another thing is that I would make more generic view for block view and block row view with tabular data that will be extended by other block views and block row views, but I'm afraid you don't not like this approach.

sunpietro:
ping @dpobel

dpobel:
> One comment about the block template: the reason I used divs instead of table is that row options were misaligned when trying to position them absolutely within a div inside the table cell. This approach solves that issue.

I did not look at the code yet, but that seems a bad reason. If you want to represent tabular data, just use a table. The behavior you add around that table should not imply changing anything on the table.

> Another thing is that I would make more generic view for block view and block row view with tabular data that will be extended by other block views and block row views, but I'm afraid you don't not like this approach.

That's a bit early to do that. I don't even see the added value of having a view per row given at the moment the columns are defined and fixed. So this block could be very simple in terms of rendering ie it receives a list Location and Content Type and it displays that with the template.

Resources/public/js/views/ez-dashboardblocksview.js
```diff
@@ -145,7 +145,14 @@ YUI.add('ez-dashboardblocksview', function (Y) {
              * @readOnly
              */
             blocks: {
-                value: [],
+                valueFn: function () {
+                    return [
+                        new Y.eZ.DashboardBlockAllContentView({
+                            priority: 1,
+                            identifier: 'everyones-content'
+                        }).addTarget(this).set('active', this.get('active'))
```
dpobel: you should write:

```js
new Y.eZ.DashboardBlockAllContentView({
    priority: 1,
    bubbleTargets: this,
});
```

the `identifier` is read only (and should be set in the code of `eZ.DashboardBlockAllContentView` itself. It's useless to forward the `active` flag, we know it's not active and there's an event handler that takes care of keeping active up to date.
it's not needed to set the active flag, we know the view is not active yet

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,228 @@
...
+
+            /**
+             * The block identifier
+             *
+             * @attribute identifier
+             * @type String
+             * @default 'dashboard-all-content-block'
+             */
+            identifier: {
+                value: 'dashboard-all-content-block'
```
dpobel: I would not redefine the attribute here but rather set its value in `initializer` (with `_set` since it is supposed to be `readOnly`). Also `all-content` (or `everyone-s-content` to fit with the specs) would be enough.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,228 @@
...
+         * @param event {Object} event facade
+         */
+        _getContentData: function () {
+            /**
+             * Makes request for all content dashboard block data.
+             * Listened in {eZ.Plugin.DashboardBlockAllContentService}
+             *
+             * @event getAllContent
+             */
+            this.fire('getAllContent');
```
dpobel: hum that's complicated, you end up first loading the root location then firing the search event from the plugin but as if it was done from the view.

Instead, you should:

* in the dashboard view service, load the root location id (only the id, we don't need the full Location model) and make sure the Dashboard view receive the `rootLocationId`
* inject this `rootLocationId` when instantiating the everyone's block
* and here use it to fire the search event

That way, we don't need the `eZ.Plugin.DashboardBlockAllContentService`

dpobel: ah actually, the Location model is needed :) so just replace `rootLocatioId` by `rootLocation` in my previous comment.

dpobel: actually it depends on what we call *everyone's content* (I'm asking for clarification in the JIRA issue).

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,228 @@
...
+        },
+
+        /**
+         * Renders the dashboard block view
+         *
+         * @method render
+         * @return {eZ.DashboardBlockView} the view itself
+         */
+        render: function () {
+            this.get('container').setHTML(this.template(this.getAttrs()));
```
dpobel: please specify the template variables you really need.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,228 @@
...
+     * @extends eZ.DashboardBlockBaseView
+     */
+    Y.eZ.DashboardBlockAllContentView = Y.Base.create('dashboardBlockAllContentView', Y.eZ.DashboardBlockBaseView, [Y.eZ.AsynchronousView], {
+        initializer: function () {
+            var parentView = this.constructor.superclass;
+
+            this._fireMethod = this._getContentData;
+            this._watchAttribute = 'contentList';
+
+            this.get('container').addClass(parentView._generateViewClassName(parentView._getName()));
```
dpobel: I'm surprised this is working :) `parentView` is not a view instance, I would more write it like:

```js
this.get('container').addClass(this._generateViewClassName(Y.eZ.DashboardBlockBaseView.NAME));
```

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,228 @@
...
+
+            /**
+             * The table header text - title
+             *
+             * @attribute headTextTitle
+             * @type String
+             * @default 'Title'
+             */
+            headTextTitle: {
+                value: 'Title'
```
dpobel: same reasoning as in https://github.com/ezsystems/PlatformUIBundle/pull/584#discussion_r62701449 just put those directly in the template

Resources/public/js/views/services/plugins/ez-dashboardblockallcontentserviceplugin.js
```diff
@@ -0,0 +1,310 @@
...
+         * Redirects user to content edit
+         *
+         * @method _goToContentEdit
+         * @protected
+         * @param event {Object} event facade
+         * @param event.contentId {String} REST content id
+         * @param event.languageCode {String} content language
+         */
+        _goToContentEdit: function (event) {
+            this.get('host').get('app').navigateTo('editContent', {
```
dpobel: I guess it's time to have something a bit more generic. We have https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/services/ez-navigationhubviewservice.js#L25 in navigation hub view service, I think you should move this system to the app so that any component can fire `navigateTo` if there's a need to navigate somewhere. (Also keep the `_navigateTo` method in the navigation hub view service but add in it a deprecation warning and add `@deprecated` tag in the doc).

Resources/public/js/views/services/plugins/ez-dashboardblockallcontentserviceplugin.js
```diff
@@ -0,0 +1,310 @@
...
+        NS: 'DashboardBlockAllContentServicePlugin',
+        ATTRS: {
+            /**
+             * Root location model
+             *
+             * @attribute rootLocation
+             * @type Y.eZ.Location
+             */
+            rootLocation: {
+                cloneDefaultValue: false,
```
dpobel: nitpick: with a `valueFn`, that's useless.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentrowview.js
```diff
@@ -0,0 +1,189 @@
...
+            options: {
+                valueFn: function () {
+                    return [
+                        new Y.eZ.ButtonActionView({
+                            actionId: 'edit',
+                            label: 'Edit'
+                        }),
+                        new Y.eZ.ButtonActionView({
+                            actionId: 'preview',
+                            label: 'Preview'
```
dpobel: it's not a *preview*, it's *view* (according the specification).
Also, initially the button action view were introduced to be used inside an action bar so that it's easy add new action. The context here is quite different, so I would rather not use that here (I'll detail an alternative in a following comment)

dpobel:
Alright now I read the whole patch.

First, for the list:

* to answer, your first question, the easiest here is to add a *loading* class on the container when rendering the view without a list of contents and to remove that class when the views gets a list of contents.
* there's no need for any view besides `eZ.DashboardBlockAllContentView` (the row view actually makes thing harder without bringing much value). It should generate a true table
* as written in https://github.com/ezsystems/PlatformUIBundle/pull/589#discussion_r63186170, the loading work should be spread between the view service and the search plugin. Actually what does the view service depends on the answer [to this comment](https://jira.ez.no/browse/EZP-25767?focusedCommentId=189714&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-189714). If we are interested only by the content items under "Content structure", then there you'll have to load the Location model for the *rootLocation* but if we really want to request all content items, then the starting is the *virtual root* and I guess you'll have to hardcode it in a similar way than [this code](https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/services/plugins/ez-universaldiscoverycontenttreeplugin.js#L47) (we'll see later how to unify that)

For the edit and views buttons appearing when clicking a row:

* as written in https://github.com/ezsystems/PlatformUIBundle/pull/589#discussion_r63193939, you can not really use the button action view in that context, so for now you should just go for some buttons generated by the template.
* to simplify that, just put those buttons in the last cell of the row and make it visible when the user taps on the row. (I know this won't play well with the potential scrollbar represented in the wireframe but for now there's not scrollbar...)
* the last point also means you don't even have to do what is written in https://github.com/ezsystems/PlatformUIBundle/pull/589#discussion_r63190680, you can directly use a regular `a` element and generates the URL to edit or view with `path` in the template and the routing will be done by the browser.

Resources/public/js/views/services/plugins/ez-dashboardblockallcontentserviceplugin.js
```diff
@@ -0,0 +1,310 @@
...
+         */
+        _getContentList: function (target) {
+            this.get('host').fire('contentSearch', {
+                viewName: 'all-content',
+                loadContentType: true,
+                search: {
+                    criteria: {
+                        SubtreeCriterion: this.get('rootLocation').get('pathString')
+                    },
+                    sortClauses: {
```
dpobel: just one thing on that: this is not yet supported server side. Meaning the block will not display the modified content but a *random* list of content item. See https://jira.ez.no/browse/EZP-24315 (added it as a blocker).

Resources/public/js/views/services/plugins/ez-dashboardblockallcontentserviceplugin.js
```diff
@@ -0,0 +1,310 @@
...
+                        DateModifiedClause: 'DESC'
+                    },
+                    limit: 15
```
dpobel: it's 10 according to the specs

Resources/public/js/views/services/plugins/ez-dashboardblockallcontentserviceplugin.js
```diff
@@ -0,0 +1,310 @@
...
+         *
+         * @method _setContentList
+         * @protected
+         * @param target {eZ.DashboardBlockAllContentView} a view requesting a content
+         * @param error {false|Error} error
+         * @param result {Array} array of struct objects containing a `content` property
+         */
+        _setContentList: function (target, error, result) {
+            if (error) {
+                this._handleError({
```
dpobel: better set the `loadingError` attribute provided by AsynchronousView. That way, the retry button (you added it in the template in https://github.com/ezsystems/PlatformUIBundle/pull/589/files#diff-4f07963fcef784ab36aeb705d5588b1eR10) can work.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
sunpietro: @dpobel yes, the location model is needed. The _Everyone's content_ means content created by all users.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
sunpietro: Why? I don't see any difference when specifying template variables here. I would say, that my approach is simpler and less complicated.

Resources/public/js/views/services/plugins/ez-dashboardblockallcontentserviceplugin.js
sunpietro: Do you think it should be done in this PR? I think, it would be better to make it in separate PR.

Resources/public/js/views/services/plugins/ez-dashboardblockallcontentserviceplugin.js
sunpietro: Thanks, good point.

Resources/public/js/views/ez-dashboardblocksview.js
sunpietro: Good idea. I didn't know about the `bubbleTargets` as it wasn't very well documented in YUI docs.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
sunpietro: @dpobel could you tell me more about the advantages of that approach (setting identifier in `initializer`? From my perspective these two approaches are the same.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
sunpietro: Good idea, I didn't like my way of solving it, but I had no idea how to make it better.

sunpietro:
First, for the list:

> - to answer, your first question, the easiest here is to add a loading class on the container when rendering the view without a list of contents and to remove that class when the views gets a list of contents.

thanks

> - there's no need for any view besides eZ.DashboardBlockAllContentView (the row view actually makes thing harder without bringing much value). It should generate a true table
    as written in #589 (comment), 

I believe that keeping row views is a good idea. This way, the block makes a request for data and spreads the data into child views (row views) and renders them. 
It would be too much if a block has to render the block container and then render the rows and populate them with data when data comes.
Having row views, also brings the potential for extending it in the future by adding more buttons and custom actions.
Do you still disagree with me?
I reverted the view template into clasic table.

> - the loading work should be spread between the view service and the search plugin. Actually what does the view service depends on the answer to this comment. If we are interested only by the content items under "Content structure", then there you'll have to load the Location model for the rootLocation but if we really want to request all content items, then the starting is the virtual root and I guess you'll have to hardcode it in a similar way than this code (we'll see later how to unify that)

This is why I wanted to have the root detection in the dashboard service, in my previous PR. 

> For the edit and views buttons appearing when clicking a row:

> - as written in #589 (comment), you can not really use the button action view in that context, so for now you should just go for some buttons generated by the template.
    to simplify that, just put those buttons in the last cell of the row and make it visible when the user taps on the row. (I know this won't play well with the potential scrollbar represented in the wireframe but for now there's not scrollbar...)
> - the last point also means you don't even have to do what is written in #589 (comment), you can directly use a regular a element and generates the URL to edit or view with path in the template and the routing will be done by the browser.

This makes me wondering, what is `eZ.ButtonActionView` for, if it's not for any kind of buttons?

Resources/public/js/views/dashboard/ez-dashboardblockallcontentrowview.js
sunpietro: If I remember correctly, when I was talking about this feature with Jince. The user should be redirected to content preview in eZ Studio. In PlatformUI there's nothing like that, so user is redirected to location view.

dpobel: Reminder, this is PlatformUI.
*Preview* has a meaning in PlatformUI but this is not what we'll be showed to the user when he'll use the corresponding link so as mentioned in the specification, use *View*.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
dpobel: @sunpietro thanks for repeating https://github.com/ezsystems/PlatformUIBundle/pull/589#discussion_r63186756 and the conclusion of the discussion on JIRA. Now, you can apply https://github.com/ezsystems/PlatformUIBundle/pull/589#discussion_r63186170 and https://github.com/ezsystems/PlatformUIBundle/pull/589#discussion_r63186756

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
dpobel: putting all attributes is indeed more simple in the first shot but this has 2 issues:
* it makes more complicated to keep the BC (remember the template can be customized so we have to be very careful on what is available from there and we have to keep that)
* by passing all attributes like that, you might pass data to the template that can not be used in a template (like a model or an array of models instead of the result of `toJSON`

Resources/public/js/views/services/plugins/ez-dashboardblockallcontentserviceplugin.js
dpobel: in any case, given the end of https://github.com/ezsystems/PlatformUIBundle/pull/589#issuecomment-219079121, that's not needed anymore.

dpobel:
> I believe that keeping row views is a good idea. This way, the block makes a request for data and spreads the data into child views (row views) and renders them.

I'm not saying it's a bad idea, I'm saying it's not needed at this point.

> It would be too much if a block has to render the block container and then render the rows and populate them with data when data comes.

this only adds 2 loops:
* one to generate an array containing the result of `toJSON` on the Location and the ContentType
* one in the template to loop over that array to display a table row per item.

Nothing really complicated.

> Having row views, also brings the potential for extending it in the future by adding more buttons and custom actions.

AFAIK, this has not been asked anywhere. And doing it without the row views does not really prevent the extensibility. That's why at this point, the row views are not needed. If we need the row views, we could easily re-introduce them after without problems.

> This is why I wanted to have the root detection in the dashboard service, in my previous PR. 

and I did not say that was a bad code, I said, that was not needed at that point. Now it is needed, so it's time to re-introduce that.

> This makes me wondering, what is `eZ.ButtonActionView` for, if it's not for any kind of buttons?

as written in the above comment, those were made to represent the button and the corresponding action in the bar views. Here, it's a different context, where we don't really want to act on the Location but more just navigate somewhere else, that's why regular links would work well for now and makes the whole thing a lot easier since we have everything in place to generate those links and the rest is done by the browser and the app without changing a single line.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
dpobel: Well, that's not really a technical thing (besides not repeating the attribute definition) but it's more a matter of semantic or of splitting the responsibility.
Here, the base view defines an attribute with a set of properties (readOnly and required) which is basically a contract. It's completely up to the implementation to comply with that contract but the implementation should not change anything to that contract.

To me, the best way to clearly reflect that is to write:

* in the base view the contract, so only the declaration of the attribute with the necessary config and doc tag
* in the child view, to comply to that contract by setting only the value since we are not supposed there to change that contract.

sunpietro:
ping @dpobel

sunpietro:
ping @dpobel

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,160 @@
...
+    /**
+     * The all content dashboard block view
+     *
+     * @namespace eZ
+     * @class DashboardBlockAllContentView
+     * @constructor
+     * @extends eZ.DashboardBlockBaseView
+     */
+    Y.eZ.DashboardBlockAllContentView = Y.Base.create('dashboardBlockAllContentView', Y.eZ.DashboardBlockBaseView, [Y.eZ.AsynchronousView], {
+        events: EVENTS,
```
dpobel: as far as I can remember, when using the asynchronous view extension, you can not define the dom event handlers this way otherwise the ones defined in asynchronous view (for the retry button) are completely overriden. You have to write it like https://github.com/ezsystems/PlatformUIBundle/blob/206d055a33cc735cfd050a90bf708113d5f8ed77/Resources/public/js/views/tabs/ez-locationviewlocationstabview.js#L39

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,160 @@
...
+
+        initializer: function () {
+            this._fireMethod = this._getContentData;
+            this._watchAttribute = 'contentList';
+            this._clickOutsideHandler = null;
```
dpobel: this property should be documented

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,160 @@
...
+         * @return {eZ.DashboardBlockAllContentView} the view itself
+         */
+        render: function () {
+            var rows = [],
+                methodName = 'addClass',
+                contentList = this.get('contentList');
+
+            if (contentList) {
+                methodName = 'removeClass';
+                rows = this._prepareRowsData(contentList);
```
dpobel: `rows` is a quite low level term. We are  actually displaying a list of Content items. Also I see in `_prepareRowsData` that for each Content item, you build a custom object. Instead, it would be better to build what we call a *content struct* and to benefit from the `toJSON` method in the models. So basically, each Content item should be represented by a *content struct* object in the template build like:
```js
{
    content: content.toJSON(),
    contentType: contentType.toJSON(),
}
```

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,160 @@
...
+         * Shows the row options
+         *
+         * @method _uiShowOptions
+         * @protected
+         * @param event {Object} event facade
+         */
+        _uiShowRowOptions: function (event) {
+            this._uiHideRowOptions();
+
+            event.target.ancestor(SELECTOR_ROW).addClass(CLASS_ROW_SELECTED);
```
dpobel: given you subscribe to the event on the row, I think `event.target` is directly the row, so there's no need for the `ancestor` call.

sunpietro: Not in this case. You're clicking on `td` inside the `tr`. You can't click on a `tr` element itself.

Resources/public/js/views/ez-dashboardblocksview.js
```diff
@@ -145,7 +145,14 @@ YUI.add('ez-dashboardblocksview', function (Y) {
              * @readOnly
              */
             blocks: {
-                value: [],
+                valueFn: function () {
+                    return [
+                        new Y.eZ.DashboardBlockAllContentView({
+                            priority: 1,
+                            bubbleTargets: this
```
dpobel: You only took half of the https://github.com/ezsystems/PlatformUIBundle/pull/589/files#r63186170 into account so now that the root Location is loaded in the view service:

1. the dashboard blocks view should receive it (ie you have to implement a `_getViewParameters` in the view service and add the corresponding attribute in the dashboard blocks view)
1. then here, the `eZ.DashboardBlockAllContentView` instance should receive that Location, that way the view has everything to directly fire the `contentSearch`  event.
1. with that in place, the plugin becomes useless and can be removed.

(nitpick: always add `,` at the end of those kind of plain object, that makes future diff cleaner and sometimes avoid stupid conflict)

Resources/public/js/views/services/ez-dashboardblocksviewservice.js
```diff
@@ -20,5 +23,163 @@ YUI.add('ez-dashboardblocksviewservice', function (Y) {
      * @constructor
      * @extends eZ.ViewService
      */
-    Y.eZ.DashboardBlocksViewService = Y.Base.create('dashboardBlocksViewService', Y.eZ.ViewService, [], {});
+    Y.eZ.DashboardBlocksViewService = Y.Base.create('dashboardBlocksViewService', Y.eZ.ViewService, [], {
+        _load: function (next) {
+            this._loadRootLocation()
+                .then(Y.bind(this._loadRootLocationModelData, this))
+                .then(next)
+                .catch(Y.bind(this._handleError, this));
```
dpobel: the base view service already provide a `_error` method, please use it

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
dpobel: ah your right, I mixed two things. The `tr` node is directly available in `event.currentTarget`

Resources/public/js/views/services/ez-dashboardblocksviewservice.js
```diff
@@ -20,5 +23,163 @@ YUI.add('ez-dashboardblocksviewservice', function (Y) {
...
+
+            return new Y.Promise(Y.bind(function (resolve, reject) {
+                rootLocation.load({api: this.get('capi')}, function (error, response) {
+                    if (error) {
+                        reject(errorData, error);
+
+                        return;
+                    }
+
+                    resolve(response.document.Location);
```
dpobel: you should not put `response.document.Location` in parameter of resolve because this kind of duplicate the response parsing done in the Location model.

Resources/public/css/views/dashboard/allcontent-dashboard-block.css
```diff
@@ -0,0 +1,65 @@
...
+    border: 1px solid;
+    position: relative;
+}
+
+.ez-view-dashboardblockallcontentview.is-loading .ez-allcontent-block-content {
+    height: 3em;
+    display: block;
+}
+
+.ez-view-dashboardblockallcontentview .ez-allcontent-block-header th:nth-child(1),
```
dpobel: that's not super readable. You should instead add a class on each `th` specific to the column (`ez-allcontent-block-header-title`, `ez-allcontent-block-header-contenttype`, ...) and then style the width of the column based on that.

Resources/public/css/views/dashboard/allcontent-dashboard-block.css
```diff
@@ -0,0 +1,65 @@
...
+
+
+.ez-view-dashboardblockallcontentview .ez-allcontent-block-cell {
+    position: relative;
+    z-index: 1;
+    box-sizing: border-box;
+}
+
+.ez-view-dashboardblockallcontentview .ez-allcontent-block-row-options {
+    transform: translate3d(100%, -50%, 0) scaleX(0);
```
dpobel: transforms should be defined in the theme stylesheet.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,160 @@
...
+         * Hides the row options
+         *
+         * @method _uiHideOptions
+         * @protected
+         */
+        _uiHideRowOptions: function () {
+            this.get('container').all(SELECTOR_ROW).removeClass(CLASS_ROW_SELECTED);
+
+            if (this._clickOutsideHandler) {
+                this._clickOutsideHandler.detach();
```
dpobel: the same cleanup should be made in the `destructor` otherwise the `clickoutside` event handler might be left

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,160 @@
...
+            contentList: {},
+
+            /**
+             * The block identifier
+             *
+             * @attribute identifier
+             * @type String
+             * @readOnly
+             */
+            identifier: {}
```
dpobel: no need to redefine the attribute

Resources/public/js/views/services/plugins/ez-dashboardblockallcontentserviceplugin.js
```diff
@@ -0,0 +1,103 @@
...
+         */
+        _getAllContent: function (event) {
+            var host = this.get('host');
+
+            host.fire('contentSearch', {
+                viewName: 'all-content',
+                loadContentType: true,
+                search: {
+                    criteria: {SubtreeCriterion: host.get('rootLocation').get('pathString')},
+                    sortClauses: {DateModifiedClause: 'DESC'},
```
dpobel: for please leave that commented out and add the same todo as in https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/services/plugins/ez-contenttreeplugin.js#L62 for instance so that when it's supported we can easily find relevant places where this has to be put in place.

Resources/public/templates/dashboard/allcontent-dashboard-block.hbt
```diff
@@ -0,0 +1,36 @@
+<h3 class="ez-allcontent-block-title">All Content</h3>
```
dpobel: since the dashboard title is h1, this should be a h2

Tests/js/views/assets/ez-dashboardblocksview-tests.js
```diff
@@ -78,7 +76,7 @@ YUI.add('ez-dashboardblocksview-tests', function (Y) {
             blocks = view.get('blocks');
 
             Y.Assert.areSame(view, block1.getTargets()[0], 'The dashboard view should be a target of a block');
-            Y.Assert.areSame(1, blocks.length, 'There should be 1 block view instance available');
+            Y.Assert.areSame(2, blocks.length, 'There should be 2 block view instances available');
```
dpobel: to avoid the need to do the very same changes when adding a new default block, I would make sure the `blocks` attribute is empty (a `this.view._set('blocks', []);` call in `setUp` is perfectly acceptable in that case).

dpobel: also, you have to add a test case for the default value of `blocks`, for now you have to make sure a `eZ.DashboardBlockAllContentView` is instantiated, that the bubble target is correctly defined and once https://github.com/ezsystems/PlatformUIBundle/pull/589/files#r64005450 is taken into account that it receives the root location.

Tests/js/views/dashboard/ez-dashboardblockallcontentview.html
```diff
@@ -0,0 +1,84 @@
...
+<head>
+<meta charset="utf-8">
+<title>eZ Dashboard All Content Block View tests</title>
+</head>
+<body>
+<div class="container"></div>
+<div class="outside"></div>
+
+<script type="text/x-handlebars-template" id="dashboardblockallcontentview-ez-template" >
+    <div class="ez-allcontent-block-wrapper ez-asynchronousview">
```
dpobel: as said in a previous PR, please use the simplest template possible in tests. In this template, most classes are useless from a test point of view.

Tests/js/views/dashboard/assets/ez-dashboardblockallcontentview-tests.js
```diff
@@ -0,0 +1,245 @@
...
+
+        setUp: function () {
+            this.view = new Y.eZ.DashboardBlockAllContentView(VIEW_CONFIG);
+        },
+
+        tearDown: function () {
+            this.view.destroy();
+        },
+
+        'Should render view correctly with loading state': function () {
```
dpobel: *correctly* ? That's very inaccurate. This test should be called something like 'Should initially render in loading state'. Also the template variable are not tested.

Tests/js/views/dashboard/assets/ez-dashboardblockallcontentview-tests.js
```diff
@@ -0,0 +1,245 @@
...
+            };
+
+            view.render();
+
+            Y.Assert.isTrue(templateCalled, 'The template should have been used to render view');
+            Y.Assert.areSame(0, container.one(SELECTOR_CONTENT).all('tr').size(), 'Should not render any row');
+            Y.Assert.isTrue(container.hasClass(CLASS_LOADING), 'Should add the loading state CSS class to the view container');
+        },
+
+        'Should render view correctly with content inside': function () {
```
dpobel: same as in the previous test, should be something like 'Should pass the content item structs to the template'

Tests/js/views/dashboard/assets/ez-dashboardblockallcontentview-tests.js
```diff
@@ -0,0 +1,245 @@
...
+                return origTpl.apply(this, arguments);
+            };
+
+            view.set('contentList', [{
+                content: contentMock,
+                contentType: contentTypeMock
+            }]);
+
+            Y.Assert.isTrue(templateCalled, 'The template should have been used to render view');
+            Y.Assert.areSame(1, container.one(SELECTOR_CONTENT).all('tr').size(), 'Should render 1 row');
```
dpobel: by asserting that you're not testing the view but its template. Instead, you should test the variable available in the template to make it's consistent with the `contentList` you set. The actual output of the view will be tested in Behat tests.

Tests/js/views/dashboard/assets/ez-dashboardblockallcontentview-tests.js
```diff
@@ -0,0 +1,245 @@
...
+                row;
+
+            view.set('contentList', [{
+                content: this.contentMock,
+                contentType: this.contentTypeMock
+            }]);
+
+            row = view.get('container').one(SELECTOR_ROW);
+
+            row.one('td').simulateGesture('tap', Y.bind(function () {
```
dpobel: just a small tip, you can write the same with a bit lighter syntax:
```js
row.one('td').simulateGesture('tap', this.next(function () {
    // content of your function
}, this);
this.wait();
```

Tests/js/views/dashboard/assets/ez-dashboardblockallcontentview-tests.js
```diff
@@ -0,0 +1,245 @@
...
+                this.resume(Y.bind(function () {
+                    Y.Assert.isTrue(row.hasClass(CLASS_ROW_SELECTED), 'The row should be selected');
+
+                    Y.Mock.verify(this.contentMock);
```
dpobel: those asserts have nothing to do with the row becoming selected so they should be removed.

Tests/js/views/dashboard/assets/ez-dashboardblockallcontentview-tests.js
```diff
@@ -0,0 +1,245 @@
...
+                isEventFired = true;
+            });
+
+            view.set('active', true);
+
+            Y.Assert.isTrue(isEventFired, 'Should fire the `getAllContent` event');
+        }
+    });
+
+    uiEventsTest = new Y.Test.Case({
```
dpobel: actually this test is completely about the selecting a row so the variable name and its name should reflect that.

dpobel: also I would add a test with 2 contents item and to check that a previously selected row is unselected when you select a new one.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
sunpietro: I'm not sure about this idea. Please, correct me if I'm wrong but creating content struct for this particular case seems like a building a cannon to kill the fly. I believe this approach is much simpler. Anyway, I'm going to use `toJSON` method to avoid calling `get` method multiple times.

Resources/public/js/views/services/ez-dashboardblocksviewservice.js
sunpietro: We should also be able to provide more details about the error. The error message might be not enough to debug an issue.

Resources/public/js/views/services/ez-dashboardblocksviewservice.js
sunpietro: I'm not sure what you mean.

Resources/public/css/views/dashboard/allcontent-dashboard-block.css
sunpietro: Could you explain why? Transform is used to correct position of an element.

Resources/public/templates/dashboard/allcontent-dashboard-block.hbt
sunpietro: Should the module titles be treaten as page subtitles?
The reason I put module title in `h3` is that module title is less imporant semantically  than page title or content title.

sunpietro:
@dpobel please, rebase feature branch with master

Resources/public/js/views/services/ez-dashboardblocksviewservice.js
sunpietro: @dpobel can you explain me what did you mean?

sunpietro:
ping @dpobel @yannickroger @StephaneDiot @dew326

dpobel:
> @dpobel please, rebase feature branch with master

@sunpietro actually you should rebase your branch against `feature/ezp-25766_dashboard` because it also contains the changes already in that branch and you are doing a PR against this one.

sunpietro:
Ok @dpobel, rebased against feature branch.
The downside is that all the previous coomits have been squashed into one.

dpobel:
@sunpietro rebase does not mean you have to squash your commits, but anyway I'll manage.

Resources/public/css/views/dashboard/allcontent-dashboard-block.css
dpobel: not exactly, that's right for the translate3d but not for the scaleX part which actually means hide the element. But you hide it in a way that makes possible to animate the display of the options. This is typically something that should go in the theme stylesheet.
Instead, here you should have a `display: none` (or `visibility: hidden`) and in the theme stylesheet, you should them write a `display: flex` (or `visibility: visible`) + the transform + the transition.

sunpietro: @dpobel Is it the only thing to correct?

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,157 @@
...
+
+        /**
+         * Makes request for data
+         *
+         * @method _getContentData
+         * @protected
+         * @param event {Object} event facade
+         */
+        _getContentData: function () {
+            this.fire('contentSearch', {
```
dpobel: given your criteria is on the subtree, you should rather do a Location search. That way, you get a list of Location in return (which embeds the Content Info so you have everything in hand), this will avoid to use the main Location while displaying the result since this main Location is maybe not the correct one regarding that criteria.

Also, using the `locationSearch` allows to configure an attribute (instead of the `callback` a few lines below) that will receive the search results to fully benefit from the asynchronous view extension. With that, you also have the error handling for *free* (which is missing at the moment).

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,157 @@
...
+         * @param error {Boolean} is error?
+         * @param list {Array} rows content data
+         */
+        _renderContentRows: function (error, list) {
+            var rows = list.map(function (data) {
+                var content = data.content.toJSON(),
+                    contentTypeName = data.contentType.get('names')[content.mainLanguageCode];
+
+                return {
+                    title: content.name,
```
dpobel: you got me wrong in https://github.com/ezsystems/PlatformUIBundle/pull/589#discussion_r64001614, the point of calling `toJSON` is not to avoid calling `get` multiple times, it's to provide a full representation of the object in the template. That way, integrators of eZ Platform can customize the block just by using another template (for instance to add new column for another property we don't display by default). So as I was writing in the initial comment, here you should have something like (without the comments of course):
```js
return {
    location: location.toJSON(), // because of https://github.com/ezsystems/PlatformUIBundle/pull/589/files#r65134592
    contentInfo: location.get('contentInfo').toJSON(), // to ease the template a bit
    contentType: contentType.toJSON(),
};
```

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,157 @@
...
+        },
+
+        /**
+         * Makes request for data
+         *
+         * @method _getContentData
+         * @protected
+         * @param event {Object} event facade
+         */
+        _getContentData: function () {
```
dpobel: the name of the method is wrong, it suggests it returns something while it does not return anything. It would be better named `_fireSearchEvent` or something like that.

Resources/public/js/views/services/ez-dashboardblocksviewservice.js
dpobel: the response parsing is the responsibility of the Location model. Writing `response.document.Location` is the very beginning of parsing the response so that should not appear here, I think you should write:
```js
resolve(rootLocation);
```

since the result of the Promise here is supposed to be a fully loaded root Location.

Resources/public/templates/dashboard/allcontent-dashboard-block.hbt
dpobel: well in that case, the module title are clearly an important thing from a structure point of view. One of the aspect of the *hN* structure is to be able to build a table of contents that can be used to guess the page structure. So here, the block title is definitively the second level in the heading hierarchy.

Resources/public/templates/dashboard/allcontent-dashboard-block.hbt
```diff
@@ -0,0 +1,36 @@
...
+    {{#if loadingError}}
+    <p class="ez-asynchronousview-error ez-font-icon">
+        An error occurred while content list.
+        <button class="ez-asynchronousview-retry ez-button ez-font-icon pure-button">Retry</button>
+    </p>
+    {{else}}
+    <table class="pure-table">
+        <thead class="ez-allcontent-block-header">
+            <tr>
+                <th class="ez-allcontent-block-head--title">Title</th>
```
dpobel: we don't use such convention (the `--`) in PlatformUI, for consistency with the rest of the code, please to stick to `-`.

Resources/public/css/views/dashboard/allcontent-dashboard-block.css
```diff
@@ -0,0 +1,65 @@
...
+    position: relative;
+}
+
+.ez-view-dashboardblockallcontentview.is-loading .ez-allcontent-block-content {
+    height: 3em;
+    display: block;
+}
+
+.ez-view-dashboardblockallcontentview .ez-allcontent-block-head--title,
+.ez-view-dashboardblockallcontentview .ez-allcontent-block-cell--title { width: 30%; }
```
dpobel: isn't it enough to set the width of the head in order to set the width of the column ?

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
sunpietro: In the context of dashboard the location of content is not important. You just have to list content no matter the location. That's why I decided to use main location.

Using `locationSearch` just to update the attribute value is just a different way to proceed further with data. IMO, these 2 approaches do the same and they both do it good. In this case, changing it to something else seems like unnecessary thing to do.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
sunpietro: If I understood correctly, in situation when integrators wants something different displayed in the block view, then they should simply create a new block view.
Anyway, I'll follow your suggestions.

sunpietro:
@dpobel changes applied

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
dpobel: > In the context of dashboard the location of content is not important. You just have to list content no matter the location. That's why I decided to use main location.

the block is supposed to list the Content items located under the root Location, so the Location does matter. I repeat, by doing a Content search, you get some Content and you then use the main Location to build the link to it in PlatformUI but for a Content with multiple Location, you do not know whether the main Location is under the root Location or somewhere else.

So by doing a Content search, you avoid this bug. In addition, by doing a location search, you take benefit from the asynchronous view system, in this case, you have 2 benefits:

* less code with the rendering logic located in one place (while at the moment you have both `render` and `_renderContentRows`)
* you get the error handling which is only half implemented in the current code

so no, it's no *unnecessary thing to do*, one option gives a better result than the other. So please, implement that.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
dpobel: > If I understood correctly, in situation when integrators wants something different displayed in the block view, then they should simply create a new block view.

no, in this case the point is exactly the opposite. By providing a full representation of the objects (ContentInfo, Location, ContentType), we open the ability to customize a block by just replacing the *template* without the need to implement a new block.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,149 @@
...
+         * @method _renderContentRows
+         * @protected
+         * @param error {Boolean} is error?
+         * @param list {Array} rows content data
+         */
+        _renderContentRows: function (error, list) {
+            var rows = list.map(function (data) {
+                return {
+                    content: data.content.toJSON(),
+                    contentType: data.contentType.get('names')[data.content.get('mainLanguageCode')]
```
dpobel: `contentType: data.contentType.toJSON()`

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,149 @@
...
+         * Renders content rows in a block
+         *
+         * @method _renderContentRows
+         * @protected
+         * @param error {Boolean} is error?
+         * @param list {Array} rows content data
+         */
+        _renderContentRows: function (error, list) {
+            var rows = list.map(function (data) {
+                return {
```
dpobel: missing one level of indentation

andrerom: because of the map + anonymous function?
if so there are a few in `Resources/public/js/views/services/ez-dashboardblocksviewservice.js`

andrerom: And a lot of other places in the code base, like:
```
YUI.add('ez-texthelper', function (Y) {
6	    var win = Y.config.win;
7	
8	    Y.Handlebars.registerHelper('urlEncode', function (txt) {
9	        return win.encodeURIComponent(txt);
10	    });
11	
12	});
```

sunpietro: @andrerom I'm not sure what you mean

dpobel: @andrerom no because of the `var` statement + function, so one level to align with `var` + one level because of a block (the function code). example:
```js
var rows = {
        whatever: 1,
        somethingElse: 2,
    },
    things = [1, 2, 3, 4],
    func = function (param1, param2) {
        console.log('Hey!', param1, param2);
    },
    double = things.map(function (item) {
        return item*2;
    });

func(rows, double);
```

in the ez-texthelper example, there's only one `var` (notice the `;` after `Y.config.win`), the rest is just some code, so the current code is OK.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
sunpietro: @dpobel in this case, how can I get a correct value inside the template? `{{ contentType.names[content.mainLanguageCode] }}` doesn't work. Is there any (unknown to me) Handlebar helper that would help me in this situation?

sunpietro: ok, got it: `{{ lookup contentType.names content.mainLanguageCode }}`

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
sunpietro: I'm still not convinced. The root location is always the top node of content tree. If there's anything else above the top node, then that _over-top-node_ node is not reachable anyway.
Of course, I can create a new attribute in the views but the question is: is really needed? you just load content to the block once. You don't refresh it. So implementing new attribute and handling attribute value change event is not reasonable in this case.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
sunpietro: @dpobel in case when there is only one variable introduced the indentation in the code above is correct. If there were more variables, then it would be incorrect.

sunpietro:
ping @dpobel

sunpietro:
Sorry, closed by accident.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
dpobel: I didn't know this one :) in most case we hardcode 'eng-GB' (which is as wrong `content.mainLanguageCode`), this is a bug https://jira.ez.no/browse/EZP-25765 we'll probably fix in the next certification period (most likely with a handlebars function).

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
dpobel: @sunpietro no, the point of this convention to avoid reindenting when doing an unrelated change, so please add an indentation level.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
bdunogier: > I'm still not convinced. The root location is always the top node of content tree.

There are also content items under media and users. We have also mentioned, together with André, that it would technically be possible in some implementations to have _more_, custom top level trees. 

How come this hasn't been solved ? I think that Damien has provided plenty of arguments, and explicitly _asked_ that it is done differently. Unless there is a good _technical_ reason not to, it should be done.

sunpietro: @bdunogier as far as I know, in that block no other content than ones from the *Content* node should be visible in this block.

bdunogier: To be honest, I don't really care about these nitpicks anymore. Damien has clearly stated his preference, and he's the lead developer and maintainer of the package this is being submitted to.

(I'll add that the tree is _also_ used to show users right now, meaning that the argument from april does sound irrelevant)

sunpietro: I've followed your suggestions on using `locationSearch`:
```javascript
this.fire('locationSearch', {
    viewName: 'all-content-' + this.get('rootLocation').get('locationId'),
    resultAttribute: 'items',
    loadContentType: true,
    loadContent: true,
    search: {
        criteria: {
            SubtreeCriterion: this.get('rootLocation').get('pathString'),
        },
        limit: 10,
    },
});
```
Results seem to be incorrect. Location models are pointing to different place in the content tree than they should. For instance, when I get information about the *Home* content from Content model and get its location from Location model, then `{{ path "viewLocation" id=location.id languageCode=content.mainLanguageCode }}` points to an image I've created not long time ago. What am I doing wrong?

sunpietro: Forgot to ping @dpobel about the question above /\

sunpietro:
@dpobel I've applied the `locationSearch`.
I found that `content.resources.MainLocation` and `location.id` point to different unrelated places. 
The `content.resources.MainLocation` points to `/api/ezp/v2/content/locations/1/2/95/70/64` which is an article, but the `location.id` points to `/api/ezp/v2/content/locations/1/2` which is a homepage. I'm not sure what is the cause of this issue. This behaviour applies for all content found using `locationSearch`

dpobel:
@sunpietro I've just tested your branch and I don't see the issue about a wrong Location id so I guess you maybe have an index issue somewhere. Also it's normal to not have the same result as before since the sort clause is not implemented, the order is a bit random.
(I'm reviewing the changes now)

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,167 @@
...
+             */
+            this._clickOutsideHandler = null;
+            this.events = Y.merge(this.events, EVENTS);
+
+            this._set('identifier', BLOCK_IDENTIFIER);
+            this.get('container')
+                .addClass(this._generateViewClassName(Y.eZ.DashboardBlockBaseView.NAME))
+                .addClass(CLASS_LOADING);
+
+            this.after('itemsChange', this.render);
```
dpobel: better write:
```js
this._watchAttribute = 'items';
```

that way the AsynchronousView extension does that for you (see https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/extensions/ez-asynchronousview.js#L82)

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,167 @@
...
+         * @param event {Object} event facade
+         */
+        _fireSearchEvent: function () {
+            var rootLocation = this.get('rootLocation');
+
+            this.fire('locationSearch', {
+                viewName: 'all-content-' + rootLocation.get('locationId'),
+                resultAttribute: 'items',
+                loadContentType: true,
+                loadContent: true,
```
dpobel: too bad the `currentVersionNo` is not available in the ContentInfo (see https://jira.ez.no/browse/EZP-25842), that would avoid having to load the Content. Can you add a TODO to remove the `loadContent` flag with a link to that JIRA issue, that way if manage to that in a very near feature, we can remove that.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,167 @@
...
+         *
+         * @method render
+         * @return {eZ.DashboardBlockAllContentView} the view itself
+         */
+        render: function () {
+            var items = this.get('items').map(function (item) {
+                    return {
+                        content: item.content.toJSON(),
+                        contentType: item.contentType.toJSON(),
+                        location: item.location.toJSON()
```
dpobel: to allow to quickly fix the TODO mentioned in https://github.com/ezsystems/PlatformUIBundle/pull/589/files#r65525193, you should add the contentInfo as well:
```js
contentInfo: item.location.get('contentInfo').toJSOn(),
```

and use it in the template instead of the Content when possible (ie everywhere except for the currentVersionNo), that way, the TODO can be very quickly solved.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -0,0 +1,167 @@
...
+                    };
+                });
+
+            this.get('container').setHTML(this.template({items: items}));
```
dpobel: you should also pass the `loadingError` attribute value otherwise, the corresponding condition in the template is useless and a potential loading error is not handled.

Tests/js/views/assets/ez-dashboardblocksview-tests.js
```diff
@@ -15,41 +15,42 @@ YUI.add('ez-dashboardblocksview-tests', function (Y) {
...
 
-            view.addBlock(new Y.eZ.DashboardBlockBaseView({
-                priority: 1,
-                identifier: 'A'
-            }));
             this.view.render();
 
             Y.Assert.isTrue(templateCalled, 'The template should have been used to render view');
             Y.Assert.areSame(1, view.get('container').one(SELECTOR_CONTENT).get('children').size(), 'Should render one block inside');
+            Y.Assert.areSame(view, block.getTargets()[0], 'The view should be a block events target');
```
dpobel: this is completely unrelated to the `render` method, this should be tested in a test case dedicated to the `blocks` attribute default value.

Tests/js/views/assets/ez-dashboardblocksview-tests.js
```diff
@@ -78,7 +80,7 @@ YUI.add('ez-dashboardblocksview-tests', function (Y) {
             blocks = view.get('blocks');
 
             Y.Assert.areSame(view, block1.getTargets()[0], 'The dashboard view should be a target of a block');
-            Y.Assert.areSame(1, blocks.length, 'There should be 1 block view instance available');
+            Y.Assert.areSame(1, blocks.length, 'There should be 2 block view instances available');
```
dpobel: The error message is not consistent with the actual assertion

Tests/js/views/assets/ez-dashboardblocksview-tests.js
```diff
@@ -152,7 +155,7 @@ YUI.add('ez-dashboardblocksview-tests', function (Y) {
             view.addBlock(block1);
             view.removeBlock(block1Config.identifier);
 
-            Y.Assert.areSame(0, view.get('blocks').length, 'There should be no blocks defined');
+            Y.Assert.areSame(0, view.get('blocks').length, 'There should only one block rendered');
```
dpobel: same here, the error message should not be changed

Tests/js/views/dashboard/assets/ez-dashboardblockallcontentview-tests.js
```diff
@@ -0,0 +1,194 @@
...
+            view.template = function () {
+                templateCalled = true;
+
+                return origTpl.apply(this, arguments);
+            };
+
+            view.render();
+
+            Y.Assert.isTrue(templateCalled, 'The template should have been used to render view');
+            Y.Assert.areSame(0, container.one(SELECTOR_CONTENT).all('tr').size(), 'Should not render any row');
```
dpobel: by doing that, you are testing the template not the rendering logic. Instead you should check that the template receives the expected variables so here an empty array `items` and the `loadingError` set to `false`

Tests/js/views/dashboard/assets/ez-dashboardblockallcontentview-tests.js
```diff
@@ -0,0 +1,194 @@
...
+            Y.Assert.isTrue(container.hasClass(CLASS_LOADING), 'Should add the loading state CSS class to the view container');
+        }
+    });
+
+    getDataTest = new Y.Test.Case({
```
dpobel: there's no *getData*, I would call it `searchEventTest` and fix the same way any *get data* usage in this test case.

Tests/js/views/dashboard/assets/ez-dashboardblockallcontentview-tests.js
```diff
@@ -0,0 +1,194 @@
...
+                pathString: PATH_STRING
+            });
+            this.view = new Y.eZ.DashboardBlockAllContentView(Y.merge(VIEW_CONFIG, {rootLocation: this.rootLocation}));
+        },
+
+        tearDown: function () {
+            this.view.destroy();
+        },
+
+        _getRowsMock: function () {
```
dpobel: it does not return anything, it can not be called *get* something

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
sunpietro: @dpobel is it what you want?
```javascript
var items = this.get('items').map(function (item) {
        return {
            content: item.content.toJSON(),
            /*
             * @TODO load contentInfo model
             * see  https://jira.ez.no/browse/EZP-25842
             *
             * contentInfo: item.location.get('contentInfo').toJSON(),
             */
            contentType: item.contentType.toJSON(),
            location: item.location.toJSON()
        };
    });
```
If I pass the contentInfo into template, then nothing will be displayed in the block. Should I use `contentInfo` data in the template anyway?

Tests/js/views/dashboard/ez-dashboardblockallcontentview.html
```diff
@@ -0,0 +1,75 @@
...
+<body>
+<div class="container"></div>
+<div class="outside"></div>
+
+<script type="text/x-handlebars-template" id="dashboardblockallcontentview-ez-template" >
+    <table class="pure-table">
+        <tbody class="ez-allcontent-block-content">
+        {{#each items}}
+            <tr class="ez-allcontent-block-row">
+                <td class="ez-allcontent-block-cell--title">{{ content.name }}</td>
```
dpobel: no need for so many classes, please use minimal template so that the view can be rendered

Tests/js/views/services/assets/ez-dashboardblocksviewservice-tests.js
```diff
@@ -0,0 +1,186 @@
...
+
+        setUp: function () {
+            this.app = new Y.Mock();
+            this.capi = new Y.Mock();
+            this.discoveryService = new Y.Mock();
+            this.rootLocationModel = new Y.Mock();
+
+            Y.Mock.expect(this.capi, {
+                method: 'getDiscoveryService',
+                run: Y.bind(function () {
```
dpobel: nitpick: in this case, you can directly write:
```
Mock.expect(this.capi, {
    method: 'getDiscoveryService',
    returns: this.discoveryService,
});
```

Tests/js/views/services/assets/ez-dashboardblocksviewservice-tests.js
```diff
@@ -0,0 +1,186 @@
...
+            });
+        },
+
+        tearDown: function () {
+            this.service.destroy();
+        },
+
+        'Should load root location data when being loaded': function () {
+            var rootLocationId = 'root-location',
+                loadLocationResponse = {
```
dpobel: not needed any more, you can remove this variable and its use a few line below

Tests/js/views/services/assets/ez-dashboardblocksviewservice-tests.js
```diff
@@ -0,0 +1,186 @@
...
+
+        setUp: function () {
+            this.app = new Y.Mock();
+            this.capi = new Y.Mock();
+            this.discoveryService = new Y.Mock();
+            this.rootLocationModel = new Y.Mock();
+
+            Y.Mock.expect(this.capi, {
+                method: 'getDiscoveryService',
+                run: Y.bind(function () {
```
dpobel: same nitpick as in https://github.com/ezsystems/PlatformUIBundle/pull/589/files#r65530839

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
dpobel: no the point here is to always inject the `contentInfo` (as my previous comment suggests, it is part of the Location so loaded with it) and to use it as much as possible, that way when we remove the `loadContent` flag, you'll only have a to change a very few things.

Resources/public/css/views/dashboard/allcontent-dashboard-block.css
```diff
@@ -0,0 +1,69 @@
...
+    z-index: 2;
+    top: 50%;
+    right: 0;
+    display: flex;
+    height: 100%;
+    box-sizing: border-box;
+}
+
+.ez-view-dashboardblockallcontentview .ez-allcontent-block-row.is-row-selected .ez-allcontent-block-row-options {
+    transform: translate3d(100%, -50%, 0) scaleX(1);
```
dpobel: you forgot this one when moving the transform from this stylesheet to the theme one

Tests/js/views/services/assets/ez-dashboardblocksviewservice-tests.js
dpobel: same nitpick as in https://github.com/ezsystems/PlatformUIBundle/pull/589/files#r65530839

dpobel:
In addition to the inline comments, some tests are missing:

* a test checking that the SearchPlugin is plugged to the dashboard view service
* a test in `renderTest` of the block view for the case where the view is rendered because it received some items
* a test for `eZ.DashboardBlocksViewService#getViewParameters` 
* a test for the `rootLocation` default value of `eZ.DashboardBlocksViewService`

sunpietro:
@dpobel how can I test anything like that: `a test checking that the SearchPlugin is plugged to the dashboard view service`? I can't find any samples of such testing.

dpobel:
@sunpietro like that: https://github.com/ezsystems/PlatformUIBundle/blob/master/Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js#L779 which illustrate quite well that you're not the first one to forgot that kind of test :)

sunpietro:
@dpobel the test: `a test in renderTest of the block view for the case where the view is rendered because it received some items` is done in `uiEventsTest` test case. I would be not able to test UI events without getting rows rendered. Is it ok for you?

sunpietro:
ping @dpobel

Tests/js/views/services/assets/ez-dashboardblocksviewservice-tests.js
```diff
@@ -0,0 +1,226 @@
...
+        'Should return a correct set of data when running `getViewParameters` method': function () {
+            Y.Assert.areSame(this.rootLocationModel, this.service.getViewParameters().rootLocation, 'Should return `rootLocation` model');
+        }
+    });
+
+    rootLocationDefaultValueTest = new Y.Test.Case({
+        name: 'eZ Dashboard Blocks View Service root location default value test',
+
+        setUp: function () {
+            this.rootLocationModel = new Y.Mock();
```
dpobel: not needed can be removed

Tests/js/views/services/assets/ez-dashboardblocksviewservice-tests.js
```diff
@@ -0,0 +1,226 @@
...
+
+            this.service = new Y.eZ.DashboardBlocksViewService();
+        },
+
+        tearDown: function () {
+            this.service.destroy();
+        },
+
+        'Should return a model': function () {
+            Y.Assert.isInstanceOf(Y.Model, this.service.get('rootLocation'), 'Should return a model');
```
dpobel: Actually, we are expecting a `Y.eZ.Location`, what you can do is to define a *fake* `Y.eZ.Location` in `setUp` (like `Y.eZ.Location = Y.Model` for instance), change the assertion here and of course delete `Y.eZ.Location` in `tearDown`.

Tests/js/views/services/assets/ez-dashboardblocksviewservice-tests.js
```diff
@@ -0,0 +1,226 @@
...
+
+    var successTest,
+        errorTests,
+        getViewParametersTest,
+        searchPluginRegisterTest,
+        rootLocationDefaultValueTest,
+        loadRootLocationErrorMessage = 'Cannot load root location',
+        loadLocationModelErrorMessage = 'Cannot load root location data into model';
+
+    Y.eZ.Location = Y.Model;
```
dpobel: that's a bit dirty, this should be done in each `setUp` where that's needed and the corresponding clean up should be done in the corresponding `tearDown`

Tests/js/views/services/assets/ez-dashboardblocksviewservice-tests.js
```diff
@@ -0,0 +1,226 @@
...
+        tearDown: function () {
+            this.service.destroy();
+        },
+
+        'Should return a model': function () {
+            Y.Assert.isInstanceOf(Y.Model, this.service.get('rootLocation'), 'Should return a model');
+        }
+    });
+
+    searchPluginRegisterTest = new Y.Test.Case(Y.eZ.Test.PluginRegisterTest);
```
dpobel: this is a wrong place to do that, since you are testing some code added in the search plugin module, the corresponding test should go the search plugin module test so here: https://github.com/ezsystems/PlatformUIBundle/blob/master/Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js#L779 (where while at it you can add the others view services on which the search plugin is plugged)

Tests/js/views/dashboard/assets/ez-dashboardblockallcontentview-tests.js
```diff
@@ -0,0 +1,265 @@
...
+                content: contentMock,
+                contentType: contentTypeMock,
+                location: locationMock
+            }];
+
+            view.template = function (params) {
+                templateCalled = true;
+
+                Y.Assert.isArray(params.items, 'The `items` variable should be an array');
+                Y.Assert.areSame(list.length, params.items.length, 'Should provide data of 1 item');
```
dpobel: that's not enough, you also have to check that the element in `params.items` is an object and it contains the `content`, `contentInfo`, `location`, `contentType` property with the correct value ie the return values of the `toJSON` methods.

Tests/js/views/dashboard/assets/ez-dashboardblockallcontentview-tests.js
```diff
@@ -0,0 +1,265 @@
...
+                Y.Assert.isFalse(params.loadingError, 'The `loadingError` should not be enabled');
+
+                return origTpl.apply(this, arguments);
+            };
+
+            view.on('locationSearch', function (event) {
```
dpobel: to only focus on the rendering logic, I would even just write:
```js
view.set('items', list);
```
to trigger the rendering. The fact this is done on the `locationSearch` event after becoming `active` is an implementation detail from the `render` method point of view.

Tests/js/views/dashboard/ez-dashboardblockallcontentview.html
```diff
@@ -0,0 +1,75 @@
...
+        <tbody class="ez-allcontent-block-content">
+        {{#each items}}
+            <tr class="ez-allcontent-block-row">
+                <td>{{ contentInfo.name }}</td>
+                <td>{{ lookup contentType.names contentInfo.mainLanguageCode }}</td>
+                <td>{{ content.currentVersion.versionNo }}</td>
+                <td>
+                    {{ contentInfo.lastModificationDate }}
+                    <div class="ez-allcontent-block-row-options">
+                        <a class="ez-allcontent-block-option-edit ez-font-icon" href="#"></a>
```
dpobel: you can even remove those 2 `a`

Tests/js/views/dashboard/assets/ez-dashboardblockallcontentview-tests.js
sunpietro: Missing comparisons added

sunpietro:
ping @dpobel

Tests/js/views/dashboard/assets/ez-dashboardblockallcontentview-tests.js
dpobel: well yes, but the comparisons are not correct, you are not really checking that the objects returned by the `toJSON` methods are the same in the template. (Also you forgot the `contentInfo`)
When defining the mocks above, you should have something like:
```js
contentInfoJSON = {};
Mock.expect(contentInfoMock, {
    method: 'toJSON',
    returns: contentInfoJSON,
});
```

and then here, you should write:

```js
Assert.areSame(
    contentInfoJSON, params.items.contentInfo,
    "The contentInfo toJSON result should have available in the template"
);
```

And of course, the same needs to be done for `content`, `location` and `contentType`.

Tests/js/views/dashboard/assets/ez-dashboardblockallcontentview-tests.js
```diff
@@ -0,0 +1,269 @@
...
+
+            view.template = function (params) {
+                templateCalled = true;
+
+                Y.Assert.isArray(params.items, 'The `items` variable should be an array');
+                Y.Assert.areSame(list.length, params.items.length, 'Should provide data of 1 item');
+                Y.Assert.areSame(contentJSON, params.items[0].content, 'Should provide content data of 1 item');
+                Y.Assert.areSame(contentTypeJSON, params.items[0].contentType, 'Should provide content type data of 1 item');
+                Y.Assert.areSame(locationJSON, params.items[0].location, 'Should provide location data of 1 item');
+                Y.Assert.areSame(contentInfoJSON, params.items[0].location.contentInfo, 'Should provide content info data of 1 item');
```
dpobel: this one is incorrect, in this context, the point is to check that the `contentInfo` is directly available without using the Location (basically the result of https://github.com/ezsystems/PlatformUIBundle/pull/589/files#diff-b26270100a3f8b2b43dbc1f0c0ed3604R70)

So this should be written:
```js
Y.Assert.areSame(contentInfoJSON, params.items[0].contentInfo, 'Should provide content info data of 1 item');
```

sunpietro:
ping @dpobel

dpobel:
+1
@StephaneDiot can you have a look please ?

StephaneDiot:
+1

sunpietro:
@dpobel this PR has just been rebased with feature branch

PR #591, opened 2016-05-13
[WIP] EZP-25770: Implement the My Drafts dashboard block

sunpietro:
https://jira.ez.no/browse/EZP-25770

**TO DO:**
- [x] created `Y.eZ.DashboardBlockAsynchronousView`,
- [x] created `Y.eZ.DashboardBlockMyDraftsView`,
- [x] created `Y.eZ.Plugin.UserDrafts`,
- [x] updated `Y.eZ.DashboardBlockAllContentView` by extending `Y.eZ.DashboardBlockAsynchronousView` instead of `Y.eZ.DashboardBlockBaseView`,
- [x] corrected dashboard layout styles,
- [x] added **My Drafts** block to the dashboard,
- [ ] JS unit tests.


sunpietro:
ping @dpobel 
Please review ezsystems/PlatformUIBundle#589 first, then I'll reimplement your suggestions here, if there will be any.

sunpietro:
**Important:** This PR requires ezsystems/ez-js-rest-client#77 merged first, in order to make it working.

sunpietro:
@dpobel Reimplemented here your suggestions from ezsystems/PlatformUIBundle#589

dpobel:
Now that https://github.com/ezsystems/PlatformUIBundle/pull/589 is closed to be merged, this one is the next target but it's a bit more complicated unlike in #589, several piece are missing. So you have to:

1. in the user model, add a `loadDrafts` methods that will use the method you added in https://github.com/ezsystems/ez-js-rest-client/pull/77. This method will provide to its a callback a list of `eZ.VersionInfo` (that's why I rebased `feature/ezp-25766_dashboard` and that's why there are some conflicts now)
1. in a `eZ.Plugin.UserDraft` plugin for the Dashboard View Service, you'll use the `loadDrafts` method. (The plugin can retrieve the currently logged in user from the `user` app attribute) This plugin will subscribe to a `loadUserDrafts` event. And for each draft, it will have to load the corresponding `eZ.ContentInfo` and the `eZ.ContentType`.
1. the block itself is very close to the *all content* one. To reorganize a bit the code, I think you should create an extension to externalize the code that handles the toolbar when a row is clicked.

sunpietro:
Thanks for suggestions, I'm waiting for #589 to merged into feature branch so I can start changing My drafts block to meet the newest standards.

andrerom:
@sunpietro I guess you have rebased, is it ready for review already or are you WIP with the changes to spec?

sunpietro:
@andre I'm going to push latest changes within 30 minutes

sunpietro:
@dpobel before I push latest changes. I'm thinking about creating asynchronous dashboard block view which could be extended by both: the `Y.eZ.DashboardBlockAllContentView` and `Y.eZ.DashboardBlockMyDraftsView`. Also, I found that the styling of these two blocks is almost the same. Should I create some more generic CSS styles for that kind of blocks?

sunpietro:
ping @dpobel

sunpietro:
Requires #607 to be merged first. Without it the block will throw errors.

dpobel:
@sunpietro please rebase against the last `feature/ezp-25766_dashboard` so that I can review that without the previous changes (see https://github.com/ezsystems/PlatformUIBundle/pull/607#issuecomment-225112171)

Other than that a base asynchronous dashboard block view is a good idea. In https://github.com/ezsystems/PlatformUIBundle/pull/591#issuecomment-223915274 I was also mentioning a view extension to handle the toolbar on each row but actually at this point, this behavior can perfectly be in such class.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -29,26 +22,10 @@ YUI.add('ez-dashboardblockallcontentview', function (Y) {
...
-            this._clickOutsideHandler = null;
-            this.events = Y.merge(this.events, EVENTS);
-
             this._set('identifier', BLOCK_IDENTIFIER);
-            this.get('container')
-                .addClass(this._generateViewClassName(Y.eZ.DashboardBlockBaseView.NAME))
-                .addClass(CLASS_LOADING);
-
-            this.after('itemsChange', this._uiEndLoading);
+            this.get('container').addClass(this._generateViewClassName(Y.eZ.DashboardBlockAsynchronousView.NAME));
```
dpobel: this should go the block asynchronous view

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -29,26 +22,10 @@ YUI.add('ez-dashboardblockallcontentview', function (Y) {
      * @constructor
      * @extends eZ.DashboardBlockBaseView
```
dpobel: you need to update that as well

Resources/public/js/views/dashboard/ez-dashboardblockasynchronousview.js
```diff
@@ -0,0 +1,134 @@
...
+        },
+
+        /**
+         * Makes request for data
+         *
+         * @method _getContent
+         * @protected
+         * @param event {Object} event facade
+         */
+        _getContent: function (event) {},
```
dpobel: this is a bad name, this method does not return anything so it can not be named `get` something. It would call it `_fireLoadEvent` or something similar.

Resources/public/js/views/dashboard/ez-dashboardblockasynchronousview.js
```diff
@@ -0,0 +1,134 @@
...
+         *
+         * @method _uiEndLoading
+         * @protected
+         */
+        _uiEndLoading: function () {
+            this.get('container').removeClass(CLASS_LOADING);
+        },
+
+        /**
+         * Makes request for data
```
dpobel: related to https://github.com/ezsystems/PlatformUIBundle/pull/591/files#r66582700, should be something like:
```
Fires an event to request the data to display. This method must reimplemented                                         
when extending eZ.DashboardBlockAsynchronousView.
```

Resources/public/js/views/dashboard/ez-dashboardblockasynchronousview.js
```diff
@@ -0,0 +1,134 @@
...
+            this.after('contentChange', this._uiEndLoading);
+        },
+
+        /**
+         * Renders the dashboard block view
+         *
+         * @method render
+         * @return {eZ.DashboardBlockAllContentView} the view itself
+         */
+        render: function () {
```
dpobel: you can completely remove that method, `Y.View` already provides exactly the same implementation.

Resources/public/js/views/dashboard/ez-dashboardblockasynchronousview.js
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }, {
+        ATTRS: {
+            /**
+             * The block content
+             *
+             * @attribute content
+             * @type Any
+             */
+            content: {}
```
dpobel: `items` used in the all content block is much better name as this attribute is supposed to contain an array of Object representing the items to display.

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -158,20 +86,16 @@ YUI.add('ez-dashboardblockallcontentview', function (Y) {
...
         ATTRS: {
             /**
-             * The items list.
+             * The block content - items list.
              *
-             * @attribute items
+             * @attribute content
              * @type Array of objects containing location structs
              */
-            items: {
```
dpobel: actually the dashboard asynchronous view now provide the attribute, so this can be completely removed

Resources/public/js/views/services/plugins/ez-userdraftsplugin.js
```diff
@@ -0,0 +1,245 @@
...
+         */
+        _resetTempVariables: function () {
+            /**
+             * Holds a temporary reference to `loadUserDrafts` event target
+             *
+             * @property _target
+             * @type {Y.View}
+             * @protected
+             */
+            this._target = null;
```
dpobel: you should not store a state like that in the object because you are forced to reset the state in several places and that makes the whole code flow is bit cumbersome. Instead, you should pass those values as parameters when needed. For instance the attribute name is only needed at the very end of the process, the limit is only needed at the very beginning.

Resources/public/js/views/services/plugins/ez-userdraftsplugin.js
```diff
@@ -0,0 +1,245 @@
...
+
+            versions.sort(this._sortByModificationDate);
+            versions.length = this._targetLimit;
+
+            Y.Promise
+                .resolve(versions)
+                .then(Y.bind(this._loadDraftContentInfo, this))
+                .then(Y.bind(this._loadDraftContentType, this))
+                .then(Y.bind(this._fillDraftContentTypeModel, this))
+                .catch(Y.bind(host._error, host));
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/591/files#r66585610

Resources/public/js/views/services/plugins/ez-userdraftsplugin.js
```diff
@@ -0,0 +1,245 @@
...
+        ATTRS: {
+            /**
+             * The drafts list
+             *
+             * @attribute drafts
+             * @type Y.ModelList
+             */
+            drafts: {
+                valueFn: function () {
+                    return new Y.ModelList({
```
dpobel: we can not use a ModelList because what you'll return to the view is not just a model list and a list of Object containing a VersionInfo model, a ContentInfo model and a ContentType model

Resources/public/js/views/services/plugins/ez-userdraftsplugin.js
```diff
@@ -0,0 +1,245 @@
...
+         * @param data {Array} list of content
+         * @return {Y.Promise}
+         */
+        _loadDraftContentType: function (data) {
+            var contentTypeService = this.get('host').get('capi').getContentTypeService(),
+                drafts = this.get('drafts'),
+                promises = [];
+
+            data.forEach(Y.bind(function (contentInfo, index) {
+                drafts.item(index).set('contentInfo', contentInfo);
```
dpobel: the `VersionInfo` has no `contentInfo` attribute (and should not have one)

Resources/public/js/views/services/plugins/ez-userdraftsplugin.js
```diff
@@ -0,0 +1,245 @@
...
+         */
+        _fillDraftContentTypeModel: function (data) {
+            var drafts = this.get('drafts');
+
+            drafts.each(function (version, index) {
+                var contentType = new Y.eZ.ContentType();
+
+                contentType.loadFromHash(data[index]);
+
+                version.set('contentType', contentType);
```
dpobel: the `VersionInfo` has no `contentType` attribute (and should not have one). This is related to https://github.com/ezsystems/PlatformUIBundle/pull/591/files#r66586529 and https://github.com/ezsystems/PlatformUIBundle/pull/591/files#r66586850. The plugin should build an array of Object which should contain the VersionInfo and the corresponding `ContentInfo` and `ContentType`

Resources/public/js/views/services/plugins/ez-userdraftsplugin.js
```diff
@@ -0,0 +1,245 @@
...
+                contentService = this.get('host').get('capi').getContentService(),
+                promises = [];
+
+            versions.forEach(function (version) {
+                var contentId = version.get('resources.Content');
+
+                drafts.add(version);
+
+                promises.push(new Y.Promise(Y.bind(function (resolve, reject) {
+                    contentService.loadContentInfo(contentId, Y.bind(function (error, response) {
```
dpobel: please use `ContentInfo#load` instead of duplicating it

Resources/public/js/views/services/plugins/ez-userdraftsplugin.js
```diff
@@ -0,0 +1,245 @@
...
+        _loadDraftContentType: function (data) {
+            var contentTypeService = this.get('host').get('capi').getContentTypeService(),
+                drafts = this.get('drafts'),
+                promises = [];
+
+            data.forEach(Y.bind(function (contentInfo, index) {
+                drafts.item(index).set('contentInfo', contentInfo);
+
+                promises.push(new Y.Promise(Y.bind(function (resolve, reject) {
+                    contentTypeService.loadContentType(contentInfo.ContentType._href, Y.bind(function (error, response) {
```
dpobel: please use `ContentType#load` instead of duplicating it

Resources/public/js/views/services/plugins/ez-userdraftsplugin.js
```diff
@@ -0,0 +1,245 @@
...
+                        if (error) {
+                            reject('Cannot load content type data: ' + contentInfo.ContentType._href);
+                            this._target.set('loadingError', true);
+                            this._resetTempVariables();
+
+                            return;
+                        }
+
+                        resolve(response.document.ContentType);
```
dpobel: such response parsing is the responsibility of the ContentType model, the Promise should be resolved to a ContentType model

Resources/public/js/views/services/plugins/ez-userdraftsplugin.js
```diff
@@ -0,0 +1,245 @@
...
+                    contentService.loadContentInfo(contentId, Y.bind(function (error, response) {
+                        if (error) {
+                            reject('Cannot load content info: ' + contentId);
+                            this._target.set('loadingError', true);
+                            this._resetTempVariables();
+
+                            return;
+                        }
+
+                        resolve(response.document.Content);
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/591/files#r66587338, don't parse the response here and resolve the Pomise with the ContentInfo model

Resources/public/js/views/ez-dashboardblocksview.js
```diff
@@ -152,6 +152,10 @@ YUI.add('ez-dashboardblocksview', function (Y) {
                             bubbleTargets: this,
                             rootLocation: this.get('rootLocation'),
                         }),
+                        new Y.eZ.DashboardBlockMyDraftsView({
+                            priority: 1,
```
dpobel: just to make sure the block are correctly sorted and to let the opportunity for the integrators to put block in between, it's better to let a gap between priorities. For instance, All Content could have a priority of 100 and My Draft a priority of 200.

Resources/public/js/views/services/plugins/ez-userdraftsplugin.js
dpobel: also you can not write `contentInfo.ContentType._href`, you instead here have the ContentInfo model and get the ContentType from it.

Resources/public/js/views/dashboard/ez-dashboardblockmydraftsview.js
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * @namespace eZ
+     * @class DashboardBlockMyDraftsView
+     * @constructor
+     * @extends eZ.DashboardBlockBaseView
+     */
+    Y.eZ.DashboardBlockMyDraftsView = Y.Base.create('dashboardBlockMyDraftsView', Y.eZ.DashboardBlockAsynchronousView, [], {
+        initializer: function () {
+            this._set('identifier', BLOCK_IDENTIFIER);
+            this.get('container').addClass(this._generateViewClassName(Y.eZ.DashboardBlockAsynchronousView.NAME));
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/591/files#r66582432

Resources/public/js/views/dashboard/ez-dashboardblockmydraftsview.js
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }, {
+        ATTRS: {
+            /**
+             * The dashboard block content. The versions list.
+             *
+             * @attribute content
+             * @type Y.ModelList
+             */
+            content: {},
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/591/files#r66584120, no need to redeclare an attribute defined in the parent class

Resources/public/js/views/dashboard/ez-dashboardblockmydraftsview.js
```diff
@@ -0,0 +1,81 @@
...
+         * @return {eZ.DashboardBlockView} the view itself
+         */
+        render: function () {
+            var drafts = this.get('content') || [];
+
+            if ('toJSON' in drafts) {
+                drafts = drafts.toJSON();
+            }
+
+            this.get('container').setHTML(this.template({drafts: drafts}));
```
dpobel: the template should also receive `loadingError`

Resources/public/js/views/dashboard/ez-dashboardblockasynchronousview.js
dpobel: actually, after thinking a bit about that, this base class should provide a `render` method because the rendering logic will be the same, the only difference will be how the `items` content is transformed to something the template can understand. So I would write the `render` method like:
```js
        render: function () {
            var items = this.get('items').map(function (item) {
                    return this._getTemplateItem(item);
                }, this);

            this.get('container').setHTML(this.template({
                items: items,
                loadingError: this.get('loadingError'),
            }));

            return this;
        },
```

and declare `_getTemplateItem` the same way as the method which fires the event to retrieve the data.
That way, the blocks using this class will just have to implement those 2 methods.

Resources/public/js/views/dashboard/ez-dashboardblockasynchronousview.js
sunpietro: If someone is going to build a new view by extending the `Y.eZ.DashboardBlockAsynchronousView` then it might be something different than items. That's why I decided to use more generic name for the attribute.

Resources/public/js/views/services/plugins/ez-userdraftsplugin.js
sunpietro: I'm not sure what to use instead in this case.

Resources/public/js/views/services/plugins/ez-userdraftsplugin.js
sunpietro: Of course it doesn't have anything like this. Where should I store required data?

sunpietro:
ping @dpobel

Resources/public/js/views/services/plugins/ez-userdraftsplugin.js
dpobel: well a plain Array would do the trick

sunpietro:
ping @dpobel

Resources/public/js/views/dashboard/ez-dashboardblockallcontentview.js
```diff
@@ -12,130 +12,49 @@ YUI.add('ez-dashboardblockallcontentview', function (Y) {
...
+                 */
+                content: item.content.toJSON(),
+                contentType: item.contentType.toJSON(),
+                location: item.location.toJSON(),
+                contentInfo: item.location.get('contentInfo').toJSON(),
+            };
         },
 
         /**
          * Makes request for data
```
dpobel: this should be adjusted to explain that in this implementation, the location search event is fired.

Resources/public/js/views/dashboard/ez-dashboardblockasynchronousview.js
```diff
@@ -0,0 +1,152 @@
...
+         * @method _uiEndLoading
+         * @protected
+         */
+        _uiEndLoading: function () {
+            this.get('container').removeClass(CLASS_LOADING);
+        },
+
+        /**
+         * Fires an event to request the data to display.
+         * This method must reimplemented when extending eZ.DashboardBlockAsynchronousView.
```
dpobel: nitpick: "must be reimplemented"

Resources/public/js/views/dashboard/ez-dashboardblockasynchronousview.js
```diff
@@ -0,0 +1,152 @@
...
+         *
+         * @method _fireLoadDataEvent
+         * @protected
+         * @param event {Object} event facade
+         */
+        _fireLoadDataEvent: function (event) {},
+
+        /**
+         * Gets item template data.
+         * This method must reimplemented when extending eZ.DashboardBlockAsynchronousView.
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/591/files#r66628977

Resources/public/js/views/dashboard/ez-dashboardblockmydraftsview.js
```diff
@@ -0,0 +1,67 @@
...
+        initializer: function () {
+            this._set('identifier', BLOCK_IDENTIFIER);
+        },
+
+        /**
+         * Gets item template data.
+         *
+         * @method _getTemplateItem
+         * @protected
+         * @param item {Y.Model} draft item model
```
dpobel: it's called `draft` in the implementation, not `item`. I would call it `draftStruct` and it's not a Y.Model any longer.

Resources/public/js/views/dashboard/ez-dashboardblockmydraftsview.js
```diff
@@ -0,0 +1,67 @@
...
+                contentInfo: draft.contentInfo.toJSON()
+            };
+        },
+
+        /**
+         * Makes request for data
+         *
+         * @method _getDraftsData
+         * @protected
+         * @param event {Object} event facade
```
dpobel: there's no event in the function signature

Resources/public/js/views/dashboard/ez-dashboardblockmydraftsview.js
```diff
@@ -0,0 +1,67 @@
...
+                version: draft.version.toJSON(),
+                contentType: draft.contentType.toJSON(),
+                contentInfo: draft.contentInfo.toJSON()
+            };
+        },
+
+        /**
+         * Makes request for data
+         *
+         * @method _getDraftsData
```
dpobel: wrong name, also the description should be a bit more precise since it's a reimplementation of a parent class method.

Resources/public/js/views/services/plugins/ez-userdraftsplugin.js
```diff
@@ -0,0 +1,204 @@
...
+         * @param data {Array} list of content types
+         */
+        _fillDraftContentTypeModel: function (target, attributeName, data) {
+            var drafts = this.get('drafts');
+
+            data.forEach(function (contentType, index) {
+                drafts[index].contentType = contentType;
+            });
+
+            drafts.sort(function (a, b) {
```
dpobel: since the VersionInfo array is sorted in _collectDraftsData, I'm pretty sure it's not needed to sort again here.

Resources/public/js/views/services/plugins/ez-userdraftsplugin.js
```diff
@@ -0,0 +1,204 @@
...
+         * @return {Y.Promise}
+         */
+        _loadDraftContentType: function (data) {
+            var drafts = this.get('drafts'),
+                promises = [];
+
+            data.forEach(Y.bind(function (contentInfo, index) {
+                var contentType = new Y.eZ.ContentType({id: contentInfo.get('resources.ContentType')});
+
+                drafts[index].contentInfo = contentInfo;
```
dpobel: it's weird to do that in a method called `_loadDraftContentType`

dpobel:
overall that looks OKish but I think we'll have to change a bit the plugin because the some methods tend to do much more than what their name suggests. Thanks @sunpietro, we'll take that over with @StephaneDiot next week.

dpobel:
Closing, replaced by https://github.com/ezsystems/PlatformUIBundle/pull/609

PR #592, opened 2016-05-15
EZP-25501: As an editor i want to be able to Swap locations

crevillo:
Implements https://jira.ez.no/browse/EZP-25501

### Ready for review

Actually the feature seems to be working.

Here's an screencast
[Screencast](https://www.youtube.com/watch?v=wblOWpBpEt0)

ping @dpobel @StephaneDiot @yannickroger @rolandbenedetti

### DIFF
--- a/Resources/public/js/models/ez-locationmodel.js
+++ b/Resources/public/js/models/ez-locationmodel.js
@@ -180,6 +180,21 @@ YUI.add('ez-locationmodel', function (Y) {
         },
 
         /**
+         * Swap the location with other location
+         * 
+         * @method swap
+         * @param {Object} options the required for the swap         *
+         * @param {Object} options.api (required) the JS REST client instance
+         * @param {eZ.Location} destinationLocation Location where the given one will be relocated
+         * @param {Function} callback a callback executed when the operation is finished
+         */
+        swap: function(options, destinationLocation, callback) {
+            var contentService = options.api.getContentService();
+
+            contentService.swapLocation(this.get('id'), destinationLocation.get('id'), callback);
+        },
+
+        /**
          * Deletes the location in the repository.
          *
          * @protected
--- a/Resources/public/js/views/services/plugins/ez-locationswapplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-locationswapplugin.js
@@ -1,0 +1,146 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-locationswapplugin', function (Y) {
+    "use strict";
+    /**
+     * Provides the plugin for swap location
+     *
+     * @module ez-locationswapplugin
+     */
+    Y.namespace('eZ.Plugin');
+
+    /**
+     * Swap location plugin. It sets an event handler to swap location
+     *
+     * In order to use it you need to fire `swapLocation` event with parameter
+     * `location` containing the eZ.Location object you want swap with other.
+     *
+     * @namespace eZ.Plugin
+     * @class LocationSwap
+     * @constructor
+     * @extends eZ.Plugin.ViewServiceBase
+     */
+    Y.eZ.Plugin.LocationSwap = Y.Base.create('locationswapplugin', Y.eZ.Plugin.ViewServiceBase, [], {
+
+        initializer: function () {
+            this.onHostEvent('*:swapLocation', this._selectDestinationLocation);
+        },
+
+        /**
+         * swapLocation event handler, launch the universal discovery widget
+         * to choose the location you want to swap with the given one
+         *
+         * @method _selectDestinationLocation
+         * @private
+         * @param {EventFacade} e swapLocation event facade
+         */
+        _selectDestinationLocation: function (e) {
+            var service = this.get('host'),
+                originLocationId = e.location.get('id'),
+                originLocationCanHaveChildren = service.get('contentType').get('isContainer'),
+                originLocationHasChildren = originLocationCanHaveChildren &&
+                    (service.get('location').get('childCount') > 0);
+
+            service.fire('contentDiscover', {
+                config: {
+                    title: "Select the location you want to swap with the given one",
+                    contentDiscoveredHandler: Y.bind(this._swapLocations, this, e.location),
+                    multiple: false,
+                    isSelectable: function(contentStruct) {
+                        if (originLocationId === contentStruct.location.get('id')) {
+                            return false;
+                        }
+
+                        if (originLocationHasChildren && !contentStruct.contentType.get('isContainer')) {
+                            return false;
+                        }
+
+                        if (!originLocationCanHaveChildren && (contentStruct.location.get('childCount') > 0)) {
+                            return false;
+                        }
+
+                        return true;
+                    }
+                },
+            });
+        },
+
+        /**
+         * Swaps the given location with the selected one
+         *
+         * @method _swapLocations
+         * @protected
+         * @param {EventFacade} e
+         */
+        _swapLocations: function (location, e) {
+            var service = this.get('host'),
+                app = service.get('app'),
+                capi = service.get('capi'),
+                notificationIdentifier = 'swap-location-' + location.get('id'),
+                destinationLocation = e.selection.location,
+                that = this;
+
+            this._notify(
+                "Swaping location for '" + location.get('contentInfo').get('name') + "'",
+                notificationIdentifier,
+                'started',
+                5
+            );
+
+            location.swap({api: capi}, destinationLocation, function (error) {
+                if (error) {
+                    that._notify(
+                        "Swaping location for '" + location.get('contentInfo').get('name') + "' failed",
+                        notificationIdentifier,
+                        'error',
+                        0
+                    );
+                }
+                else {
+                    var contentInfo = location.get('contentInfo');
+                    that._notify(
+                        "Location for '" + contentInfo.get('name') + "' " +
+                        "has been swapped with '" + destinationLocation.get('contentInfo').get('name') + "'",
+                        notificationIdentifier,
+                        'done',
+                        5
+                    );
+
+                    app.navigateTo('viewLocation', {
+                        id: location.get('id'),
+                        languageCode: contentInfo.get('mainLanguageCode')
+                    });
+                }
+            });
+        },
+
+        /**
+         * Fire 'notify' event
+         *
+         * @method _notify
+         * @protected
+         * @param {String} text the text shown during the notification
+         * @param {String} identifier the identifier of the notification
+         * @param {String} state the state of the notification
+         * @param {Integer} timeout the number of second, the notification will be shown
+         */
+        _notify: function (text, identifier, state, timeout) {
+            this.get('host').fire('notify', {
+                notification: {
+                    text: text,
+                    identifier: identifier,
+                    state: state,
+                    timeout: timeout,
+                }
+            });
+        },
+    }, {
+        NS: 'swapLocation'
+    });
+
+    Y.eZ.PluginRegistry.registerPlugin(
+        Y.eZ.Plugin.LocationSwap, ['locationViewViewService']
+    );
+});
--- a/Resources/public/js/views/tabs/ez-locationviewlocationstabview.js
+++ b/Resources/public/js/views/tabs/ez-locationviewlocationstabview.js
@@ -23,6 +23,9 @@ YUI.add('ez-locationviewlocationstabview', function (Y) {
             },
             '.ez-remove-locations-button': {
                 'tap': '_removeSelectedLocations'
+            },
+            '.ez-locations-swap-button': {
+                'tap': '_swapLocation'
             }
         };
 
@@ -102,6 +105,26 @@ YUI.add('ez-locationviewlocationstabview', function (Y) {
             this.fire('createLocation', {
                 content: this.get('content'),
                 afterCreateCallback: Y.bind(this._refresh, this)
+            });
+        },
+
+        /**
+         * Tap event handler on the `Swap location` button. It fires the
+         * `swapLocation` event
+         *
+         * @method _swapLocation
+         * @protected
+         * @param {EventFacade} e
+         */
+        _swapLocation: function (e) {
+            /**
+             * Fired when the user clicks on `Swap location` button
+             *
+             * @event swapLocation
+             * @param {eZ.Location} location the content to be swapped
+             */
+            this.fire('swapLocation', {
+                location: this.get('location'),
             });
         },
 
--- a/Resources/public/templates/tabs/locations.hbt
+++ b/Resources/public/templates/tabs/locations.hbt
@@ -73,3 +73,14 @@
         </div>
     </div>
 </div>
+
+<div class="ez-locations-box">
+    <h2 class="ez-locations-box-title">Location Content Swap</h2>
+    <p class="ez-subitem-swap-locations">
+        Swap the content item at this location with another
+        <button
+            class="ez-locations-swap-button pure-button ez-button">
+            Select Content Item
+        </button>
+    </p>
+</div>
--- a/Tests/js/models/assets/ez-locationmodel-tests.js
+++ b/Tests/js/models/assets/ez-locationmodel-tests.js
@@ -3,7 +3,7 @@
  * For full copyright and license information view LICENSE file distributed with this source code.
  */
 YUI.add('ez-locationmodel-tests', function (Y) {
-    var modelTest, trashTest, moveTest, hideTest, removeTest, loadPathTest, toJSONTest, updateSortingTest, updatePriorityTest, isRootTest,
+    var modelTest, trashTest, moveTest, hideTest, removeTest, loadPathTest, toJSONTest, updateSortingTest, updatePriorityTest, isRootTest, swapTest,
         Assert = Y.Assert, Mock = Y.Mock;
 
     modelTest = new Y.Test.Case(Y.merge(Y.eZ.Test.ModelTests, {
@@ -934,6 +934,45 @@ YUI.add('ez-locationmodel-tests', function (Y) {
         },
     });
 
+    swapTest = new Y.Test.Case({
+        "name": "eZ Location Swap tests",
+
+        setUp: function () {
+            this.capiMock = new Y.Mock();
+            this.capiGetService = 'getContentService';
+            this.locationId = 'location/man-city/sergio-aguero';
+            this.destinationLocationId = 'location/fcb/arda-turam';
+            this.model = new Y.eZ.Location({id: this.locationId});
+            this.destinationLocation = new Y.eZ.Location({id: this.destinationLocationId});
+            this.contentServiceMock = new Y.Mock();
+        },
+
+        tearDown: function () {
+            this.model.destroy();
+            delete this.model;
+        },
+
+        "Should swap the locations": function () {
+            var callback = function () {},
+                options = {api: this.capiMock};
+
+            Y.Mock.expect(this.capiMock, {
+                method: 'getContentService',
+                returns: this.contentServiceMock
+            });
+            Y.Mock.expect(this.contentServiceMock, {
+                method: 'swapLocation',
+                args: [this.locationId, this.destinationLocationId, callback],
+            });
+
+            this.model.swap(options,this.destinationLocation, callback);
+
+            Y.Mock.verify(this.contentServiceMock);
+            Y.Mock.verify(this.capiMock);
+
+        },
+    });
+
     Y.Test.Runner.setName("eZ Location Model tests");
     Y.Test.Runner.add(modelTest);
     Y.Test.Runner.add(trashTest);
@@ -945,4 +984,5 @@ YUI.add('ez-locationmodel-tests', function (Y) {
     Y.Test.Runner.add(updatePriorityTest);
     Y.Test.Runner.add(toJSONTest);
     Y.Test.Runner.add(isRootTest);
+    Y.Test.Runner.add(swapTest);
 }, '', {requires: ['test', 'model-tests', 'ez-locationmodel', 'ez-restmodel']});
--- a/Tests/js/views/services/plugins/assets/ez-locationswapplugin-tests.js
+++ b/Tests/js/views/services/plugins/assets/ez-locationswapplugin-tests.js
@@ -1,0 +1,467 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-locationswapplugin-tests', function (Y) {
+    var tests, registerTest, swapLocationTest,
+        Assert = Y.Assert, Mock = Y.Mock,
+        getModelMock = function(attributes) {
+            var mock = new Mock();
+
+            Mock.expect(mock, {
+                method: 'get',
+                args: [Mock.Value.String],
+                run: function (attr) {
+                    if (attributes.hasOwnProperty(attr)) {
+                        return attributes[attr];
+                    } else {
+                        Assert.fail("Unexpected call to get('" + attr + "')");
+                    }
+                },
+            });
+
+             return mock;
+        };
+
+    tests = new Y.Test.Case({
+        name: "eZ Location Swap Plugin event tests",
+
+        setUp: function () {
+            this.service = new Y.Base();
+            this.view = new Y.View();
+            this.locationJsonWithChildren = {
+                'id': '/location/Sergio/Aguero',
+                'childCount': 5
+            };
+            this.locationJsonWithoutChildren = {
+                'id': '/location/Sergio/Aguero',
+                'childCount': 0
+            };
+            this.containerContentTypeJson = {
+                'id': 'contentType',
+                'isContainer': true
+            };
+            this.nonContainerContentTypeJson = {
+                'id': 'contentType',
+                'isContainer': false
+            };
+
+            this.locationWithChildren = getModelMock(this.locationJsonWithChildren);
+            this.locationWithoutChildren = getModelMock(this.locationJsonWithoutChildren);
+            this.containerContentType = getModelMock(this.containerContentTypeJson);
+            this.nonContainerContentType = getModelMock(this.nonContainerContentTypeJson);
+
+            this.plugin = new Y.eZ.Plugin.LocationSwap({
+                host: this.service
+            });
+        },
+
+        tearDown: function () {
+            this.plugin.destroy();
+            this.view.destroy();
+            this.service.destroy();
+            delete this.plugin;
+            delete this.view;
+            delete this.service;
+        },
+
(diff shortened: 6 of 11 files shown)


Resources/public/js/models/ez-locationmodel.js
```diff
@@ -180,6 +180,27 @@ YUI.add('ez-locationmodel', function (Y) {
         },
 
         /**
+         * Swap the location with other location
+         * 
+         * @method swap
+         * @param {Object} options the required for the swap
+         * @param {eZ.Location} Location where the given one will be relocated
```
yannickroger: params are wrong

crevillo: you mean order or...?

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
```diff
@@ -0,0 +1,134 @@
...
+                that = this;
+
+            this._notify(
+                "Swaping location for '" + location.get('contentInfo').get('name') + "'",
+                notificationIdentifier,
+                'started',
+                5
+            );
+
+            var destinationLocation = e.selection.location,
```
yannickroger: all var should be instantiated at the beginning

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
```diff
@@ -0,0 +1,134 @@
...
+                            errNotificationIdentifier,
+                            'error',
+                            0
+                        );
+                        return;
+                    }});
+
+            location.swap({api: capi}, destinationLocation, end);
+
+            var msg = "Location for '" + location.get('contentInfo').get('name') + "' " +
```
yannickroger: same here

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
```diff
@@ -0,0 +1,134 @@
...
+                        that._notify(
+                            "Swaping location for '" + location.contentInfo.get('name') + "' failed",
+                            errNotificationIdentifier,
+                            'error',
+                            0
+                        );
+                        return;
+                    }});
+
+            location.swap({api: capi}, destinationLocation, end);
```
yannickroger: You might not need to create a variable to store the callback

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
yannickroger: and maybe you don't need to create a variable for that.

Resources/public/js/views/tabs/ez-locationviewlocationstabview.js
```diff
@@ -106,6 +109,29 @@ YUI.add('ez-locationviewlocationstabview', function (Y) {
...
+         * @protected
+         * @param {EventFacade} e
+         */
+        _swapLocation: function (e) {
+            /**
+             * Fired when the user clicks on `Swap location` button
+             *
+             * @event swapLocation
+             * @param {eZ.Location} location the content to be swaped
+             * @param {Function} afterCreateCallback callback function that will be called after
```
yannickroger: error in param.
Should we just call it `callback`

Resources/public/js/models/ez-locationmodel.js
yannickroger: * Order should be fixed: `options.api` should be right after `options`
* Missing name of varible: `@param {eZ.Location} destinationLocation`

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
crevillo: much of the code is adapted from create-location plugin. :)

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
crevillo: i don't understand this...

Resources/public/js/views/tabs/ez-locationviewlocationstabview.js
crevillo: for this, as said above, this is more or less taken from another already existing methods... https://github.com/crevillo/PlatformUIBundle/blob/0337d024d0cc838a1401ee38ba3a43c0768cdb89/Resources/public/js/views/tabs/ez-locationviewlocationstabview.js#L96-L109

you mean param should be called `callback` instead of `afterSwapLoctionCallback`?

yannickroger: The main problem is that the name of the variable is wrong in the JSdoc.
Regarding the naming of the callback, I am nitpicking, both names are ok

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
yannickroger: It doesn't really matter :)

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
yannickroger:
```
location.swap({api: capi}, destinationLocation, stack.add(function (error) {
    if (error) {
        that._notify(
            "Swaping location for '" + location.contentInfo.get('name') + "' failed",
            errNotificationIdentifier,
            'error',
            0
        );
        return;
    }})
 );
```

And remove the `end` variable

crevillo: ah, ok, understood now. thanks.

crevillo:
comments taken in account, @yannickroger . now i think i only miss swap test for location model... any idea on how to do it? better asked... how can i test that? :)

crevillo:
Rebased and squashed. Added swap location model test based on the moved one. it's the more similar i've found :)

Ready for review i think.

rolandbenedetti:
Hey Carlos, great to see that work. So cool to get a contribution on that that fast.
We were chatting with @inakijv , I think we'd love to bring a few improvements to the UI.
In current state, it's not optimal from a UX standpoint.
Can we hold on that before merging?

crevillo:
Sure @rolandbenedetti. No problem with that on my side.

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -180,6 +180,27 @@ YUI.add('ez-locationmodel', function (Y) {
...
+         * Swap the location with other location
+         * 
+         * @method swap
+         * @param {Object} options the required for the swap         *
+         * @param {Object} options.api (required) the JS REST client instance
+         * @param {eZ.Location} destination Location where the given one will be relocated
+         * @param {Function} callback a callback executed when the operation is finished
+         */
+        swap: function(options, destinationLocation, callback) {
+            var locationService = options.api.getContentService();
```
dpobel: should be called `contentService` (there's no location service in the JS REST Client)

crevillo: ok, i'll change it.

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -180,6 +180,27 @@ YUI.add('ez-locationmodel', function (Y) {
...
+
+            locationService.swapLocation(this.get('id'), destinationLocation.get('id'), function(error, response) {
+                if ( error ) {
+                    callback(error);
```
dpobel: in both cases you should provide the `response`, so this can even be written:
```js
locationService.swapLocation(this.get('id'), destinationLocation.get('id'), callback);
```

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
```diff
@@ -0,0 +1,137 @@
...
+
+        /**
+         * swapLocation event handler, launch the universal discovery widget
+         * to choose the location you want to swap with the given one
+         *
+         * @method _swapLocationSelect
+         * @private
+         * @param {EventFacade} e swapLocation event facade
+         */
+        _swapLocationSelect: function (e) {
```
dpobel: it would be better called `_selectLocation` or something like that.

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
```diff
@@ -0,0 +1,137 @@
...
+        _swapLocation: function (e) {
+            var service = this.get('host'),
+                capi = service.get('capi'),
+                location = service.get('location'),
+                notificationIdentifier = 'swap-location-' + location.get('id'),
+                data = e.target.get('data'),
+                destinationLocation = e.selection.location,
+                errNotificationIdentifier = 'swap-location-' + location.get('id'),
+                locationSwaped = false,
+                stack = new Y.Parallel(),
```
dpobel: since you only have one asynchronous operation, the use of Y.Parallel is not needed. You should write something like:

```js
location.swap({api: capi}, destinationLocation, function (error) {
    if ( error ) {
        // error notification
    } else {
        // success notification
    }
});
```

crevillo: aha... ok, I will change this.

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
```diff
@@ -0,0 +1,137 @@
...
+         * @param {EventFacade} e
+         */
+        _swapLocation: function (e) {
+            var service = this.get('host'),
+                capi = service.get('capi'),
+                location = service.get('location'),
+                notificationIdentifier = 'swap-location-' + location.get('id'),
+                data = e.target.get('data'),
+                destinationLocation = e.selection.location,
+                errNotificationIdentifier = 'swap-location-' + location.get('id'),
```
dpobel: this is a simple case where one notification has to be generated so no need for a custom error notification identifier.

Resources/public/js/views/tabs/ez-locationviewlocationstabview.js
```diff
@@ -106,6 +109,29 @@ YUI.add('ez-locationviewlocationstabview', function (Y) {
...
+             * Fired when the user clicks on `Swap location` button
+             *
+             * @event swapLocation
+             * @param {eZ.Location} location the content to be swaped
+             * @param {Function} afterSwapLocationCallback callback function that will be called after
+             *                   swaping location
+             */
+            this.fire('swapLocation', {
+                location: this.get('location'),
+                afterSwapLocationCallback: Y.bind(this._afterSwapCallback, this)
```
dpobel: not sure we need a callback from the view in the case of swap Location. I guess we should just redirect to the new Location after swapping.

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -934,6 +934,45 @@ YUI.add('ez-locationmodel-tests', function (Y) {
...
+            var callback = function () {},
+                options = {api: this.capiMock};
+
+            Y.Mock.expect(this.capiMock, {
+                method: 'getContentService',
+                returns: this.contentServiceMock
+            });
+            Y.Mock.expect(this.contentServiceMock, {
+                method: 'swapLocation',
+                args: [this.locationId, this.destinationLocationId, Y.Mock.Value.Function],
```
dpobel: with https://github.com/ezsystems/PlatformUIBundle/pull/592/files#r63542286 taken into account, you should replace `Y.Mock.Value.Function` by `callback` to make sure the callback is passed to the JS REST Client method.

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
```diff
@@ -0,0 +1,137 @@
...
+         */
+        _swapLocationSelect: function (e) {
+            var service = this.get('host');
+
+            service.fire('contentDiscover', {
+                config: {
+                    title: "Select the location you want to swap with the given one",
+                    contentDiscoveredHandler: Y.bind(this._swapLocation, this),
+                    multiple: false,
+                    isSelectable: true,
```
dpobel: shouldn't we check whether the `isContainer`/`childCount` are compatible ? for instance what should happen if we swap two Location, one holding a container Content with children with another one which is not container (and so do not have children) ?

crevillo: i really can't say, but at first, i didn't see any check on the api. so, don't know what it should do, but may i guess what it does is just swaping locations?
so, if you had
```
A
------B
------C
D
```
and you swap A with B, you will end with
```
D
------  B
------ C 
A
`````
This indeed looks weird in you think in a common article / comments structure, as looks like comments for A will now be children of D?

But on the other side... isn't it the way it should go when you try to change root location by swaping it with any other given one?

Comment (rolandbenedetti) [inline Resources/public/js/views/services/plugins/ez-locationswapplugin.js]: @crevillo I tihnk you meant "and you swap A with D".
My understanding of the feature is that we swap the content item attached to the location, sub-items remain under the location.

eg:
```
Folder A ( [folder] objid=10 locationid=1)
|---Image1 ([image] objid=11 locationid=2)
|---Image2 ([image] objid=13 locationid=3)
Gallery A ( [folder] objid=10 locationid=1)
```

if you swap Gallery A location content with Folder A, you get to:

Comment (crevillo) [inline Resources/public/js/views/services/plugins/ez-locationswapplugin.js]: i think this is what it does... but take in account that from here, platform ui, we're doing nothing else than calling the api... https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/REST/Server/Controller/Location.php#L285

i mean, if this is not behaving as expected right now, should be changed at that level?

Comment (andrerom) [inline Resources/public/js/views/services/plugins/ez-locationswapplugin.js]: From UX perspective it would be better if UI can make sure you can't swap when one is container with children, and one which is non container. But yes the API should also check this, so if it does not, thats a bug.

Comment (crevillo) [inline Resources/public/js/views/services/plugins/ez-locationswapplugin.js]: maybe we could then ellaborate use cases to know how to proceed. let's see cases...

* Container with children with non container => don't allow
* Container without children with non container => allow
* container with container => allow
* non container with container without children => allow
* non container with container with children => i'd say allow... 
* non container with non container => allow
Am I right? Do i miss something? 

About the api, it ends doing some update queries but as said, i can't see any other check related to containers / non container. Should i open an issue with that and deal with it?

Comment (SylvainGuittard) [inline Resources/public/js/views/services/plugins/ez-locationswapplugin.js]: > non container with container with children => i'd say allow...
I'd say don't allow

Comment (crevillo) [inline Resources/public/js/views/services/plugins/ez-locationswapplugin.js]: probably :). yep, seems you're right.

**Comment by crevillo** [comment]
@dpobel, I worked on your comments. let me know how this is look now:
@rolandbenedetti, @inakijv , worked on the ux part too. here's the screencast (added it to the description of the pr too). 
https://youtu.be/wblOWpBpEt0

**Comment by SylvainGuittard** [comment]
Thanks @crevillo for this great contribution

**Comment by crevillo** [comment]
added conditions to not allow select locations in the discovery depending on the origin one
@dpobel @yannickroger probably you see any better way to do this?

tests are missing yet.

**Comment by crevillo** [comment]
tests for selections added... maybe code can be reduced due duplicates? 
re-ping @yannickroger @dpobel @inakijv @rolandbenedetti @andrerom

File: Tests/js/views/services/plugins/assets/ez-locationswapplugin-tests.js [hunk 14]
```diff
@@ -0,0 +1,623 @@
...
+
+        "Should swap locations and fire notifications": function () {
+            var startNotificationFired = false,
+                successNotificationFired = false,
+                errorNotificationFired = false,
+                appNavigateCalled = false,
+                that = this;
+
+            this.service.set('app', {
+                navigateTo: function (routeName, params) {
```
dpobel: it would be better to use real mock for the app instead

Tests/js/views/services/plugins/assets/ez-locationswapplugin-tests.js
```diff
@@ -0,0 +1,623 @@
...
+                        get: function () {}
+                    }
+                };
+
+                e.config.contentDiscoveredHandler(config);
+            });
+
+            Mock.expect(this.location, {
+                'method': 'swap',
+                'args': [Mock.Value.Object, Mock.Value.Object, Mock.Value.Function],
```
dpobel: for the second argument, you should either give a real value (the Location that was selected) or add an `Assert.areSame` call in the run function to make sure the correct Location is used there.

Tests/js/views/tabs/assets/ez-locationviewlocationstabview-tests.js
```diff
@@ -845,6 +846,65 @@ YUI.add('ez-locationviewlocationstabview-tests', function (Y) {
...
+        "Should fire `swapLocation` event": function () {
+            var eventFired = false,
+                that = this,
+                locationIdForSwap,
+                swapButton;
+
+            this.view.render();
+
+            swapButton = this.view.get('container').one('.ez-locations-swap-button');
+            locationIdForSwap = swapButton.getAttribute('data-location-id');
```
dpobel: not used anywhere, I guess it's a left over of a previous of the patch. You should remove that variable and the corresponding data attribute from the template.

Tests/js/views/services/plugins/assets/ez-locationswapplugin-tests.js
```diff
@@ -0,0 +1,623 @@
...
+                }
+            });
+
+            return locationMock;
+        },
+
+        _getContentInfoMock: function (attrs) {
+            var contentInfoMock = new Mock();
+
+            Mock.expect(contentInfoMock, {
```
dpobel: this very same code structure is repeated all over the test. You should define a function like the following to define a mock with a `get` method:
```js
function getModelMock(attributes) {
    var mock = new Mock();

    Mock.expect(mock, {
        method: 'get',
        args: [Mock.Value.String],
        run: function (attr) {
             if ( attributes[attr] ) {
                 return attributes[attr];
             } else {
                 Assert.fail("Unexpected call to get('" + attr + "')");
             }
        },
    });
    return mock;
}
```

Then for instance here, the method would become:
```js
_getContentInfoMock: function () {
    return getModelMock(attrs);
}
```

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -180,6 +180,21 @@ YUI.add('ez-locationmodel', function (Y) {
         },
 
         /**
+         * Swap the location with other location
+         * 
+         * @method swap
+         * @param {Object} options the required for the swap         *
+         * @param {Object} options.api (required) the JS REST client instance
+         * @param {eZ.Location} destination Location where the given one will be relocated
```
dpobel: `@param {eZ.Location} destinationLocation Location where the given one will be relocated`

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
```diff
@@ -0,0 +1,148 @@
...
+                            return false;
+                        }
+
+                        if (!originLocationCanHaveChildren && (contentStruct.location.get('childCount') > 0)) {
+                            return false;
+                        }
+
+                        return true;
+                    },
+                    data: {}
```
dpobel: not needed, no ?

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
```diff
@@ -0,0 +1,148 @@
...
+         *
+         * @method _swapLocation
+         * @protected
+         * @param {EventFacade} e
+         */
+        _swapLocations: function (e) {
+            var service = this.get('host'),
+                app = service.get('app'),
+                capi = service.get('capi'),
+                location = service.get('location'),
```
dpobel: since the `swapLocation` event provides a Location model, it would be better to use it instead of retrieving the Location from the service.

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
```diff
@@ -0,0 +1,148 @@
...
+         * swapLocation event handler, launch the universal discovery widget
+         * to choose the location you want to swap with the given one
+         *
+         * @method _selectDestinationLocation
+         * @private
+         * @param {EventFacade} e swapLocation event facade
+         */
+        _selectDestinationLocation: function (e) {
+            var service = this.get('host'),
+                originLocationId = service.get('location').get('id'),
```
dpobel: since the swapLocation event provides a Location model, it would be better to use it instead of retrieving the Location from the service.

crevillo: @dpobel i don't get this. `e.location` returns undefined for me...

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
```diff
@@ -0,0 +1,148 @@
...
+        /**
+         * swapLocation event handler, launch the universal discovery widget
+         * to choose the location you want to swap with the given one
+         *
+         * @method _selectDestinationLocation
+         * @private
+         * @param {EventFacade} e swapLocation event facade
+         */
+        _selectDestinationLocation: function (e) {
+            console.log(e);
```
StephaneDiot: don't forget this one ;)

crevillo: yep... don't worry. it was just a try to see what i had in e... as said, i can't find the location from there...

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
```diff
@@ -0,0 +1,148 @@
...
+                        return true;
+                    }
+                },
+            });
+        },
+
+        /**
+         * Swaps the given location with the selected one
+         *
+         * @method _swapLocation
```
StephaneDiot: _swapLocations

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
StephaneDiot: e.location should be available from what I see... Let's wait for @dpobel to give you an indication

crevillo: Ok. i have other things pending. fire in the office these days :).

crevillo:
think i have it now. @dpobel, @StephaneDiot.

crevillo:
Applied a new reorganization of code for reducing it.

Resources/public/js/views/tabs/ez-locationviewlocationstabview.js
```diff
@@ -106,6 +109,27 @@ YUI.add('ez-locationviewlocationstabview', function (Y) {
...
+         * @param {EventFacade} e
+         */
+        _swapLocation: function (e) {
+            /**
+             * Fired when the user clicks on `Swap location` button
+             *
+             * @event swapLocation
+             * @param {eZ.Location} location the content to be swapped
+             */
+            console.log(this.get('location'));
```
yannickroger: :)
You should use breakpoints

Resources/public/templates/tabs/locations.hbt
```diff
@@ -73,3 +73,14 @@
         </div>
     </div>
 </div>
+
+<div class="ez-locations-box">
```
yannickroger: Not a big fan of the class name. Maybe something like `ez-locations-swap-box` would be a bit better (and adapt below).

crevillo: No problem for me, but note i just followed what i saw in https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/templates/tabs/details.hbt#L54-L68 :)

Thought that was the convention :)

yannickroger: You are right, leave it the way it is :smile:

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -934,6 +934,45 @@ YUI.add('ez-locationmodel-tests', function (Y) {
...
+            Y.Mock.expect(this.contentServiceMock, {
+                method: 'swapLocation',
+                args: [this.locationId, this.destinationLocationId, callback],
+            });
+
+            this.model.swap(options,this.destinationLocation, callback);
+
+            Y.Mock.verify(this.contentServiceMock);
+            Y.Mock.verify(this.capiMock);
+
```
yannickroger: nitpick empty line

Tests/js/views/services/plugins/assets/ez-locationswapplugin-tests.js
```diff
@@ -0,0 +1,499 @@
...
+            var startNotificationFired = false,
+                successNotificationFired = false,
+                errorNotificationFired = false,
+                selection = this._getSelection(),
+                that = this;
+
+            Mock.expect(this.app, {
+                method: 'navigateTo',
+                args: [Mock.Value.String, Mock.Value.Object],
+                run: Y.bind(function (routeName, params) {
```
yannickroger: You are binding to have `this` and you use `that`

Tests/js/views/services/plugins/assets/ez-locationswapplugin-tests.js
```diff
@@ -0,0 +1,499 @@
...
+                    Assert.areSame(
+                        options.api,
+                        that.capi,
+                        'CAPI should be passed as param'
+                    );
+                    callback(true);
+                }
+            });
+
+            this.service.on('notify', function (e) {
```
yannickroger: Is that a copy paste from the previous method, if yes could we put this in a protected method and call it in both cases.

yannickroger:
+1 with feedback taken into account.

Resources/public/js/views/services/plugins/ez-locationswapplugin.js
```diff
@@ -0,0 +1,147 @@
...
+            var service = this.get('host'),
+                originLocationId = e.location.get('id'),
+                originLocationCanHaveChildren = service.get('contentType').get('isContainer'),
+                originLocationHasChildren = originLocationCanHaveChildren &&
+                    (service.get('location').get('childCount') > 0);
+
+            service.fire('contentDiscover', {
+                config: {
+                    title: "Select the location you want to swap with the given one",
+                    contentDiscoveredHandler: Y.bind(this._swapLocations, this),
```
dpobel: you have to pass the *originLocation* (ie `e.location`) in the parameter of this function and use it in `_swapLocations` instead of `service.get('location')` to make sure the Location provided in the event is always used as the *source*, so here you can write:
```js
Y.bind(this._swapLocations, this, e.location),
```

and the `_swapLocations` signature becomes:
```js
_swapLocations: function (location, e) {
```

joaoinacio:
@crevillo: Looks like there are now conflicts that should be resolved, in any case looks very nice indeed, thanks!

crevillo:
conflict resolved... did i do it ok?

crevillo:
i did it wrong, I've rebased instead of mergin as suggested by @yannickroger

crevillo:
Sorry, Re-rebased and looking better now.

dpobel:
+1
Very good work Carlos! Thanks a lot for this contribution.

andrerom:
Congratulations Carlos!! 🎉  🎯

StephaneDiot:
:clap:

PR #593, opened 2016-05-16
Fix EZP-25678: respect default content language on content create

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-25678

#581 against 1.3

### DIFF
--- a/ApplicationConfig/Providers/Languages.php
+++ b/ApplicationConfig/Providers/Languages.php
@@ -25,7 +25,7 @@ class Languages implements Provider
     public function __construct(LanguageService $languageService, $languages)
     {
         $this->languageService = $languageService;
-        $this->defaultLanguageCode = $languages[0]['languageCode'];
+        $this->defaultLanguageCode = $languages[0];
     }
 
     public function getConfig()
--- a/ApplicationConfig/Providers/Languages.php
+++ b/ApplicationConfig/Providers/Languages.php
@@ -37,7 +37,7 @@ class Languages implements Provider
                 'name' => $language->name,
                 'languageCode' => $language->languageCode,
                 'enabled' => $language->enabled,
-                'default' => $language->languageCode === $this->defaultLanguageCode
+                'default' => $language->languageCode === $this->defaultLanguageCode,
             );
         }
 
--- a/ApplicationConfig/Providers/Languages.php
+++ b/ApplicationConfig/Providers/Languages.php
@@ -17,9 +17,15 @@ class Languages implements Provider
      */
     private $languageService;
 
-    public function __construct(LanguageService $languageService)
+    /**
+     * @var string
+     */
+    private $defaultLanguageCode;
+
+    public function __construct(LanguageService $languageService, $languages)
     {
         $this->languageService = $languageService;
+        $this->defaultLanguageCode = $languages[0]['languageCode'];
     }
 
     public function getConfig()
@@ -31,6 +37,7 @@ class Languages implements Provider
                 'name' => $language->name,
                 'languageCode' => $language->languageCode,
                 'enabled' => $language->enabled,
+                'default' => $language->languageCode === $this->defaultLanguageCode
             );
         }
 
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -227,10 +227,9 @@ YUI.add('ez-platformuiapp', function (Y) {
                     }
                 });
                 this._set('systemLanguageList', systemLanguageList);
-                this._set('contentCreationDefaultLanguageCode', defaultLanguageCode );
+                this._set('contentCreationDefaultLanguageCode', defaultLanguageCode);
                 delete config.languages;
             }
-
         },
 
         /**
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -194,7 +194,8 @@ YUI.add('ez-platformuiapp', function (Y) {
          */
         _dispatchConfig: function () {
             var config = this.get('config'),
-                systemLanguageList = {};
+                systemLanguageList = {},
+                defaultLanguageCode;
 
             if ( !config ) {
                 return;
@@ -221,10 +222,15 @@ YUI.add('ez-platformuiapp', function (Y) {
             if ( config.languages ) {
                 Y.Array.each(config.languages, function (language) {
                     systemLanguageList[language.languageCode] = language;
+                    if ( language.default ) {
+                        defaultLanguageCode = language.languageCode;
+                    }
                 });
                 this._set('systemLanguageList', systemLanguageList);
+                this._set('contentCreationDefaultLanguageCode', defaultLanguageCode );
                 delete config.languages;
             }
+
         },
 
         /**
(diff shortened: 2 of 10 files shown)


joaoinacio:
ping @dpobel @andrerom

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -221,10 +222,15 @@ YUI.add('ez-platformuiapp', function (Y) {
             if ( config.languages ) {
                 Y.Array.each(config.languages, function (language) {
                     systemLanguageList[language.languageCode] = language;
+                    if ( language.default ) {
+                        defaultLanguageCode = language.languageCode;
+                    }
                 });
                 this._set('systemLanguageList', systemLanguageList);
+                this._set('contentCreationDefaultLanguageCode', defaultLanguageCode );
```
dpobel: nitpick cs: no space before `)`

Resources/public/js/views/ez-contenttypeselectorview.js
```diff
@@ -194,7 +194,6 @@ YUI.add('ez-contenttypeselectorview', function (Y) {
 
             this.fire('createContent', {
                 contentType: type,
-                languageCode: 'eng-GB',
```
dpobel: that's a BC break, you can not do that, we have to keep `languageCode`. As I was writing in https://github.com/ezsystems/PlatformUIBundle/pull/581#discussion_r62484406, you should properly document the event and explain that the `languageCode` is deprecated (example of doc block for the event can be found in http://ezsystems.github.io/platformui-javascript-api/files/Resources_public_js_views_ez-serversideview.js.html#l72 for example).

Resources/public/js/views/services/ez-contentcreateviewservice.js
```diff
@@ -56,9 +56,15 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
...
                 version = new Y.eZ.Version(),
                 type = this.get('contentType'),
-                defaultFields = {};
+                defaultFields = {},
+                contentTypeNames,
+                contentTypeName;
 
-            content.set('name', 'New "' + this.get('contentType').get('names')['eng-GB'] + '"');
+            contentTypeNames = this.get('contentType').get('names');
+            contentTypeName = contentTypeNames[this.get('app').get('contentCreationDefaultLanguageCode')]
```
dpobel: please create a protected method that returns the content type name.

Resources/public/js/views/services/plugins/ez-contentcreateplugin.js
```diff
@@ -144,8 +142,13 @@ YUI.add('ez-contentcreateplugin', function (Y) {
              * @default undefined
              * @type String
              */
-            languageCode: {},
-
+            languageCode: {
+                readOnly: true,
```
dpobel: that's not really needed (and a *small* BC break)

dpobel: also you should update the default value in the doc block

Tests/js/views/services/plugins/assets/ez-contentcreateplugin-tests.js
```diff
@@ -272,7 +290,6 @@ YUI.add('ez-contentcreateplugin-tests', function (Y) {
             this.service.set('location', location);
             this.service.fire('whatever:createContent', {
                 contentType: type,
-                languageCode: languageCode,
```
dpobel: given https://github.com/ezsystems/PlatformUIBundle/pull/593/files#r63476850, this should be kept but you'll have to change the assertion to check that this language code is ignored.

dpobel:
In addition to the inline comments, you also need to test the new code in the app.

joaoinacio:
updated as per feedback. Not really certain about how to doc the deprecated argument besides describing it (there isn't a standard way of doing that in jsdoc?)

Resources/public/js/views/ez-contenttypeselectorview.js
```diff
@@ -192,6 +192,13 @@ YUI.add('ez-contenttypeselectorview', function (Y) {
                 return t.get('id') === typeId;
             });
 
+            /**
+             * Fired when creating a new content of a given type
+             *
+             * @event createContent
+             * @param {Object} contentType
```
dpobel: `@param {eZ.ContentType} contentType`

Resources/public/js/views/ez-contenttypeselectorview.js
```diff
@@ -192,6 +192,13 @@ YUI.add('ez-contenttypeselectorview', function (Y) {
...
+             * @param {String} languageCode - deprecated
```
dpobel: there's no standard way to document that but I guess can be a bit more precise here with something like:
`@param {String} languageCode (always set to "eng-GB"). This parameter is deprecated and will be removed in PlatformUI 2.0`

Resources/public/js/views/services/plugins/ez-contentcreateplugin.js
```diff
@@ -144,8 +142,12 @@ YUI.add('ez-contentcreateplugin', function (Y) {
              * @default undefined
```
dpobel: this should be updated to mention this attribute default value the value of the `contentCreationDefaultLanguageCode` app attribute.

dpobel:
+1 when the last 3 comments are fixed.

andrerom:
+1  _(and all tests pass but behat that fails for some random reason)_ ping @StephaneDiot @yannickroger

yannickroger:
+1

StephaneDiot:
+1

bdunogier:
For some reason, behat fails to output anything: https://travis-ci.org/ezsystems/PlatformUIBundle/jobs/130794307 (it had failed because of "no output for 10 minutes". Restarted it, and it seems to be doing do the same thing).

andrerom:
@joaoinacio did you get behat tests to run locally with these changes?

joaoinacio:
Not behat tests, but will do them.
The issue does appears to be related to travis somehow as the timeout happens while opening the browser (AFAIK at that stage no test code has even begun to run, not even BeforeScenario).

In any case deserves further attention, probably also another test PR to verify if the behavior is the same.

bdunogier:
I'm afraid it is running on your test PR, @joaoinacio... coincidence ?

bdunogier:
Re-restarted the job, and it seems that it will fail again the same way... maybe not a coincidence...

ApplicationConfig/Providers/Languages.php
```diff
@@ -17,9 +17,15 @@ class Languages implements Provider
...
-    public function __construct(LanguageService $languageService)
+    /**
+     * @var string
+     */
+    private $defaultLanguageCode;
+
+    public function __construct(LanguageService $languageService, $languages)
     {
         $this->languageService = $languageService;
+        $this->defaultLanguageCode = $languages[0]['languageCode'];
```
dpobel: this line is responsible for:
```
Warning: Illegal string offset 'languageCode' in /home/dp/dev/PlatformUIBundle/ApplicationConfig/Providers/Languages.php on line 28
```

bdunogier:
The issue has been identified, @joaoinacio will push a fix soon.

joaoinacio:
Indeed behat refused to let this one pass, though it could have been more informative...
ping @miguelcleverti this case would really benefit from some actual information, possible issue

andrerom:
Seems to pass now :)

bdunogier:
Great. Will squash & merge from github as it only has one real commit. Thank you @joaoinacio and all the reviewers.

miguelcleverti:
I was running the tests on `prod` environment where it passes because it does not show the warning. however in the `behat` it platformUI hangs in the 'Loading the application...'. However locally behat triggers a timeout after 15 seconds while it seems in travis it does not...

bdunogier:
@miguelcleverti I had the same issue with the timeout, and had to increase our own internal timeout to get it to pass at all...

PR #596, opened 2016-05-19
Fix EZP-25705: Can't add anything between 2 embeds

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25705

# Description

The JIRA issue is actually about two issues:

1. it's not possible to add anything between 2 embeds
2. it's not possible to complete an unordered list if you put an embed in the last list item.

This patch only fixes the first issue for now and this can be seen as 3 steps:

1. make sure the *add* toolbar (the plus button on the left) is available and correctly positioned when an embed receives the focus. This was not the case previously because an embed is represented in the editor by a CKEditor widget and this widget is not editable (`contenteditable` set to false) so the default AlloyEditor `add` toolbar assumes we can't add anything. To avoid that, we now use an `ezadd` toolbar which basically extend the default one to not check the content editable property.
1. When adding a simple element after an embed (paragraph, heading, ...) we have to make sure the element added after the embed
1. When adding another embed after an existing one, we also have to make sure the new embed is added in the right place. By default, the CKEditor engine was running the *editing process* of the widget.

This can be seen in the following screencast:

[](http://www.youtube.com/watch?v=mRAJcITt6yY "Click to play on Youtube.com")

# Tests

manual tests + unit tests

### DIFF
--- a/Resources/config/yui.yml
+++ b/Resources/config/yui.yml
@@ -14,6 +14,9 @@ system:
                 ez-alloyeditor:
                     requires: ['alloyeditor']
                     path: %ez_platformui.public_dir%/js/external/ez-alloyeditor.js
+                ez-alloyeditor-toolbar-ezadd:
+                    requires: ['ez-alloyeditor']
+                    path: %ez_platformui.public_dir%/js/alloyeditor/toolbars/ezadd.js
                 ez-alloyeditor-toolbar-config-link:
                     requires: ['ez-alloyeditor']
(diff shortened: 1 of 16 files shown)


dpobel:

yannickroger:
+1

dpobel:
echo ping @StephaneDiot ;-)

StephaneDiot:
+1

dpobel:
EZP-25705 has been split to separate the case where we want to add something between 2 embeds (EZP-25850) from the unsupported case where the embed is added in a list item.

PR #598, opened 2016-05-26
Fix EZP-25804: Updating main location doesn't refresh the view

yannickroger:
Link: https://jira.ez.no/browse/EZP-25804

## Description
When updating the main location id of a content, the view was not refreshed.
This was happening because we would not update (or reload) the content (which stores the mainLocationId information), but only the locations. This patch updates the content with the content in order to refresh the main location id selector.

## Test
Manual and unit tests.


dpobel:
+1

yannickroger:
ping @StephaneDiot

StephaneDiot:
+1

PR #600, opened 2016-06-01
EZP-25833: Unable to set custom redirection URLs after creating a content

sunpietro:
https://jira.ez.no/browse/EZP-25833

**TO DO:**
- [x] check whether any custom URLs were provided to `Y.eZ.ContentCreateViewService`,
- [x] if not, then set default redirection URLs,
- [x] reset redirection attrs in `Y.eZ.ContentEditViewService` when redirecting user somewhere else,
- [x] updated JS unit tests.

### DIFF
--- a/Resources/public/js/views/services/ez-contenteditviewservice.js
+++ b/Resources/public/js/views/services/ez-contenteditviewservice.js
@@ -400,6 +400,10 @@ YUI.add('ez-contenteditviewservice', function (Y) {
         _redirectToAttribute: function (attr) {
             var attrRedirectionUrl = this.get(attr);
 
+            this.set('publishRedirectionUrl');
+            this.set('closeRedirectionUrl');
+            this.set('discardRedirectionUrl');
+
             if ( attrRedirectionUrl ) {
                 return this.get('app').navigate(attrRedirectionUrl);
             }
--- a/Resources/public/js/views/services/ez-contentcreateviewservice.js
+++ b/Resources/public/js/views/services/ez-contentcreateviewservice.js
@@ -95,21 +95,29 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
          */
         _setRedirectionUrls: function () {
             var app = this.get('app'),
-                viewParent;
-
-            viewParent = app.routeUri('viewLocation', {
-                id: this.get('parentLocation').get('id'),
-                languageCode: this.get('parentContent').get('mainLanguageCode')
-            });
-            this.set('discardRedirectionUrl', viewParent);
-            this.set('closeRedirectionUrl', viewParent);
-            this.set('publishRedirectionUrl', function () {
-                var content = this.get('content');
-                return app.routeUri('viewLocation', {
-                    id: content.get('resources').MainLocation,
-                    languageCode: content.get('mainLanguageCode'),
+                viewParent = app.routeUri('viewLocation', {
+                    id: this.get('parentLocation').get('id'),
+                    languageCode: this.get('parentContent').get('mainLanguageCode')
                 });
-            });
+
+            if (!this.get('discardRedirectionUrl')) {
+                this.set('discardRedirectionUrl', viewParent);
+            }
+
+            if (!this.get('closeRedirectionUrl')) {
+                this.set('closeRedirectionUrl', viewParent);
+            }
+
+            if (!this.get('publishRedirectionUrl')) {
+                this.set('publishRedirectionUrl', function () {
+                    var content = this.get('content');
+
+                    return app.routeUri('viewLocation', {
+                        id: content.get('resources').MainLocation,
+                        languageCode: content.get('mainLanguageCode'),
+                    });
+                });
+            }
         },
 
         /**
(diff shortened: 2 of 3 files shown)


sunpietro:
ping @yannickroger @dpobel @StephaneDiot

andrerom:
Cool, and seems behat failures are something else.

Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -400,6 +400,10 @@ YUI.add('ez-contenteditviewservice', function (Y) {
         _redirectToAttribute: function (attr) {
             var attrRedirectionUrl = this.get(attr);
 
+            this.set('publishRedirectionUrl');
```
dpobel: I'm a bit confused by those 3 lines.
First because, it seems you set the value of those attributes to `undefined` while they expect either a function or string (ok that's badly documented), second because you *set* those 3 attributes while `_redirectToAttribute` expects one of those attribute identifier and last because you do after getting the actual value of the attribute. So clearly, something is wrong here.

sunpietro: The idea behind this is to reset (somehow) the attribute value. Using `this.reset('publishRedirectionUrl');` didn't work correctly. I found it many times, thus that's why I'm setting undefined here.
Without reseting the values of these attributes a user might be unexpectedly redirected somewhere else after creating content. I found this as an issue when trying to redirect a user to the dashboard after editing content from dashboard block.

sunpietro:
ping @dpobel

dpobel:
well not much to add, the problems raised in https://github.com/ezsystems/PlatformUIBundle/pull/600#discussion_r65858502 are still there but I'm not enough *into* the issue to propose an alternate solution yet.

sunpietro:
ping @dpobel

andrerom:
> ping @dpobel

@sunpietro did you resolve the issue he raised? No trace of that here.

Resources/public/js/views/services/ez-contenteditviewservice.js
sunpietro: @dpobel I have to clean these attributes up somehow. They store previously set, obsolete data that affects other use cases. By resetting attributes values to `undefined` I'm able to update them later like in: https://github.com/ezsystems/PlatformUIBundle/pull/600/files#diff-f806c7b7ec53181d7b83e4396abd100aR103
I found no other way to fix that behaviour. Any suggestions are welcome.

sunpietro:
ping @dpobel

Resources/public/js/views/services/ez-contenteditviewservice.js
dpobel: @sunpietro there are still 2 problems here:

1. as said before the way the attributes are reset (by setting `undefined`) is wrong
1. the place were you do it is wrong as well. As the name `_redirectToAttribute` suggests (and as written in its doc description), this method is supposed to redirect the user to the URI contained in the `attr` attribute passed in parameter. Resetting the 3 attributes while doing that is clearly not in the scope of that method.

sunpietro: @dpobel 
1. the default value is `undefined` so I'm setting the default value. We set the redirection URLs when needed. There's no need to keep them after redirecting a user. Keeping it after user redirection makes the app behaviour unpredictable when someone wants to update it and redirect user somewhere else. If it still doesn't convinvce you, then what you propose to do in this case?
2. moving the reset of values into a separate method would work for you?

andrerom:
@sunpietro can you expand on high level use case you are trying to archive in description? neither PR or issue contains this.

Resources/public/js/views/services/ez-contenteditviewservice.js
dpobel: > the default value is undefined so I'm setting the default value.

which is indeed true but this info is completely absent when you read the code you are adding (or the existing code, while doing that the attribute doc block should be fixed as well). So it's clear for you now, not for me nor anybody else (even you) in a near future.

> There's no need to keep them after redirecting a user.

yeah but you do it before redirecting not after.

> Keeping it after user redirection makes the app behaviour unpredictable when someone wants to update it and redirect user somewhere else.

that's right but you are assuming here that using one of this attribute is the only way to navigate away from the current route associated with the view service. If you navigate differently (for instance for whatever reason, the editor is using a simple link available in the page), the attribute won't be reseted because you tie the reset logic with the navigation logic with one of those attributes.

> If it still doesn't convinvce you, then what you propose to do in this case?

the problem is not convincing me, the problem is fixing the issue correctly both in terms of code organization/quality (yes create a separate method to reset the redirection attributes with a good name and a few lines of description is a good first step in that direction) and in terms of application behavior.
For the later point, it's a bit more complicated because I realize that in a view service, we have no easy way detect when it becomes the active view service and when it's not anymore, `setNextViewServiceParameters` is called in that case, but this is not the right place to do such cleaning. This is a need we had before but we were able to workaround but given this issue, I think it's time to improve that. So I would suggest the following:

1. the base view service should define a boolean `active` attribute
1. when the app switch from one view service to the other (see the `activeViewService` attribute in the app), the `active` flag is updated
1. then the redirection attributes reset logic should be called when the create/edit view service becomes inactive.

andrerom:
@sunpietro still valid?

sunpietro:
@andrerom I'm going to close this one as I did the same with another approach.

PR #601, opened 2016-06-01
EZP-25794: Display the subitem grid view by default when in media

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25794

# Description

This patch changes the subitem so that the subitem grid view is displayed by default when the Content Type of the Content item being viewed belongs to the *Media* Content Type Group (phew :)).

Note: the *Media* Content Type Group id is hardcoded in the Subitem Box view (the view responsible for allowing the user to change the type of subitem view to display and to decide which one to display by default), I assume it's safe to do that this way since it's something that can not really change. Also, it's perfectly possible to write a plugin for the Subitem Box view to implement additional rules if needed.

* [x] Behat scenario
* [x] Add an option to load the Content Type Group ids when loading a Content Type
* [x] Load the Content Type Group Ids in the LocationView view service
* [x] Use the `grid` view when the Content Type belongs to the *Media* Content Type Group

# Test

unit tests + manual tests

### DIFF
--- a/Features/Standard/SubitemGrid.feature
+++ b/Features/Standard/SubitemGrid.feature
@@ -1,0 +1,9 @@
+Feature: Subitem Grid
+
+    @javascript
+    Scenario: The Subitem Grid view is used when viewing a Content item of a Type that belongs to the Media Group
+        Given a Content Type of the Media Group flagged as container
+          And a Content item of this type
+          And this Content item has a subitem
+         When I view this Content
+         Then The subitems view mode is set to grid
--- a/Resources/public/js/models/ez-contenttypemodel.js
+++ b/Resources/public/js/models/ez-contenttypemodel.js
@@ -32,18 +32,75 @@ YUI.add('ez-contenttypemodel', function (Y) {
          * @param {String} action the action, currently only 'read' is supported
          * @param {Object} options the options for the sync.
          * @param {Object} options.api (required) the JS REST client instance
+         * @param {Boolean} [options.loadGroups=false] also load the Content
+         *        Type Groups id the Content Type belongs to (only applicable for
+         *        the 'read' operation).
          * @param {Function} callback a callback executed when the operation is finished
          */
         sync: function (action, options, callback) {
-            var api = options.api;
-
             if ( action === 'read' ) {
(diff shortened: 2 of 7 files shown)


dpobel:
@bdunogier there's also a Behat scenario if you want to have look (you can look at the code as well ;-))

yannickroger:
+1

StephaneDiot:
+1

Features/Standard/SubitemGrid.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Subitem Grid
+
+    @javascript
+    Scenario: The Subitem Grid view is used when viewing a Content item which Content Type belongs to the Media Content Type Group
+        Given a Content Type of the Media Content Type Group marked as container
```
bdunogier: I think that our domain would still work phrased as "of the Media Group".

I'd maybe prefer "flagged" over "marked". Both are correct, but I like "flagged" for boolean options.

Features/Standard/SubitemGrid.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Subitem Grid
+
+    @javascript
+    Scenario: The Subitem Grid view is used when viewing a Content item which Content Type belongs to the Media Content Type Group
```
bdunogier: Shortened: "a Content item of a Type that belongs to the Media Group ?" ?

Features/Standard/SubitemGrid.feature
```diff
@@ -0,0 +1,9 @@
...
+        Given a Content Type of the Media Group flagged as container
+          And a Content item of this type
+          And this Content item has a subitem
+         When I view this Content
+         Then The Subitem Grid is visible
```
bdunogier: I don't think that "visible" is a good choice there. To me, it implies that it is otherwise "hidden". It's not really hidden/visible, more of a choice out of a list.

"The the Subitems are shown using the Grid view"

bdunogier:
We should remember to ping @miguelcleverti so that he looks into implementing the behat scenario.

Maybe there are implementations hints you could provide Miguel with @dpobel ? Iirc, there were a couple ones you had a clear plan for.

PR #602, opened 2016-06-01
[Behat] WIP Improve structure and usability of PlatformUI behat Contexts

miguelcleverti:
**NOT YET FOR MERGE**

This PR refactors the platfomui behat contexts in order to make them more decoupled and usable.
Main objectives:
- [x] Remove the BehatBubdle Browser Context dependency from PlatformUI Context
- [x] Remove all traits and make them into Contexts classes
- [x] Separate CommonsActions trait into specific Contexts 
- [ ] Improve namespaces/organization?
- [ ] Improve/shorten behat_suite.yml?

Also added a easier way to gather contexts by using annotations in the properties specifying the context needed.
There are a lot of changes here but most of it is code that has already been reviewed.

depends on: https://github.com/ezsystems/ezpublish-kernel/pull/1654


miguelcleverti:
@bdunogier @joaoinacio @andrerom

Features/Context/ContentActions.php
```diff
@@ -9,9 +9,32 @@
...
+     */
+    protected $dashboardContext;
+
+    /**
+     */
+    private function iSeeNotification($message)
+    {
+        $this->sleep();
+        $result = $this->getElementByText($message, '.ez-notification-text');
+        Assertion::AssertNotNull($result);
```
joaoinacio: instance var not really necessary, if not for consistency maybe?

Features/Context/ContentActions.php
```diff
@@ -25,10 +48,11 @@ public function iSeeCopiedNotification($name, $destiny)
      */
     public function moveInto($name, $destiny)
     {
-        $this->onFullView($name);
-        $this->clickActionBar('Move');
-        $this->selectFromUniversalDiscovery("eZ Platform/$destiny");
-        $this->confirmSelection();
+        $this->contentEditContext->onFullView($name);
+        $this->waitWhileLoading();
```
joaoinacio: seems that the `onFullView()` needs to wait afterwards.
If so, adding it there instead would allow removing all waits after using it.

Features/Context/Fields.php
```diff
@@ -161,9 +170,8 @@ public function publishContent()
     public function seeRequiredFieldtOfType($label)
     {
         if ($this->platformStatus == self::WAITING_FOR_PUBLISHING) {
-            $fieldManager = $this->getFieldTypeManager();
             // $type = ...
-            //$name = $fieldManager->getThisFieldTypeName();
+            //$name = $fieldtypeContext->getThisFieldTypeName();
```
joaoinacio: cleanup comments instead of updating them

Features/Context/PlatformUI.php
```diff
@@ -106,6 +102,22 @@ public function __construct($uri = self::PLATFORM_URI, $user = null, $password =
     /**
      * @BeforeScenario
```
joaoinacio: This would need some doc.
something in the lines of:
```
 *
 * Parses Context properties typehints and automatically assigns them the context instances, if any.
 */
```

Features/Context/PlatformUI.php
```diff
@@ -106,6 +102,22 @@ public function __construct($uri = self::PLATFORM_URI, $user = null, $password =
...
      */
+    public function gatherContexts(BeforeScenarioScope $scope)
+    {
+        $refClass = new ReflectionClass($this);
+        $refProperties = $refClass->getProperties();
+        foreach ($refProperties as $refProperty) {
+            preg_match_all('#@(.*?)\n#s', $refProperty->getDocComment(), $matches);
+            $contexts = $this->parseAnnotations($matches[1]);
+            foreach ($contexts as $property => $context) {
+                $this->$property = $scope->getEnvironment()->getContext($context);
```
joaoinacio: If I read correctly the $context holds the full class Namespace to the context class.

It should be verified that this class implements a `Context`, otherwise this will try to assign *some* class instance to properties

Features/Context/PlatformUI.php
```diff
@@ -408,33 +466,30 @@ protected function closeEditView()
...
-     * Attaches a file to a input field on the HTML.
+     * Returns an array with the properties contexts,
+     * if the properties use the Context Annotation.
      *
-     * @param   string $file        file name relative to mink definitions
-     * @param   string $selector    CSS file upload element selector
+     * @return array array of methods and their service dependencies
      */
-    protected function attachFile($fileName, $selector)
+    private function parseAnnotations($annotations)
```
joaoinacio: name should be more descriptive of actual use.
Since it parses only properties...

Features/Context/PlatformUI.php
```diff
@@ -408,33 +466,30 @@ protected function closeEditView()
...
-            if (is_file($fullPath)) {
-                $fileInput = 'input[type="file"]' . $selector;
-                $field = $this->getSession()->getPage()->find('css', $fileInput);
-
-                if (null === $field) {
-                    throw new Exception("File input $selector is not found");
-                }
-                $field->attachFile($fullPath);
+        // parse array from (numeric key => 'annotation <value>') to (annotation => value)
+        $propertiesContexts = [];
```
joaoinacio: why contexts? maybe `$propertyAnnotations` ?

Features/Context/PlatformUI.php
```diff
@@ -408,33 +466,30 @@ protected function closeEditView()
...
+        foreach ($annotations as $annotation) {
+            if (!preg_match('/^(\w+)\s+\$(\w+)\s+([\w\.\\\\]+)/', $annotation, $matches)) {
+                continue;
+            }
+
+            array_shift($matches);
+            $tag = array_shift($matches);
+            if ($tag == 'Context') {
```
joaoinacio: The `Context` tag shouldn't really be necessary, 
An `@var` should suffice, as (see above) it should be checked if the class NS implements a `Context`

PR #604, opened 2016-06-02
EZP-25821: Add search icon

StephaneDiot:
## Description

The goal here is only to add an icon in platformUI. Currently this icon will be useful for the platformUISearchPrototypeBundle, but in the future we will probably use it in PlatformUI. That's why I added here directly

related to https://jira.ez.no/browse/EZP-25821


StephaneDiot:
ping @dpobel @yannickroger

dpobel:
+1 of course :) I'm merging it right away.

PR #606, opened 2016-06-08
EZP-25794: Avoid searching for subitems when there's none

dpobel:
Additional patch needed for https://jira.ez.no/browse/EZP-25794 (and https://jira.ez.no/browse/EZP-25662)

# Description

This patch makes sure we don't try to search for the subitems on a Location that does not have any. This is a small performance improvement but this also fixes a JavaScript error issued because we were trying to disable the *Load More* button which is not generated in that case.

# Tests

unit test + manual test


andrerom:
+1 I guess

dpobel:
ping @StephaneDiot

StephaneDiot:
+1

PR #611, opened 2016-06-14
EZP-25464: Create new draft from archived version

yannickroger:
Link: https://jira.ez.no/browse/EZP-25464

## Description
Allows creating new draft based on an archived version.

Screencast: https://drive.google.com/file/d/0B5amepf6EkXbZ3Z4YjlzNGdIU1E/view

## Test
Manual and unit test update


Resources/public/js/models/ez-contentmodel.js
```diff
@@ -339,6 +339,24 @@ YUI.add('ez-contentmodel', function (Y) {
...
+         * @param {String} versionNo (optional) Version Number for the draft to be based on. If not
+         *                                      provided, the current version of the content will
+         *                                      be used. Ex: 42
+         * @param {Function} callback
+         */
+        createDraft: function (options, versionNo, callback) {
+            var capi = options.api,
+                contentService = capi.getContentService();
+
+            contentService.createContentDraft(this.get('id'), versionNo, callback);
```
dpobel: you should probably parse the response and provide a VersionInfoModel to the callback instead of the response Object

Resources/public/js/views/services/plugins/ez-versionsplugin.js
```diff
@@ -0,0 +1,155 @@
...
+                    api: this.get('host').get('capi'),
+                },
+                app = this.get('host').get('app'),
+                content = e.content;
+
+            content.createDraft(options, e.versionNo, Y.bind(function (error, response) {
+                var routeName = 'editContentVersion',
+                    routeParams = {
+                        id: content.get('id'),
+                        languageCode: response.document.Version.VersionInfo.initialLanguageCode,
```
dpobel: related to https://github.com/ezsystems/PlatformUIBundle/pull/611/files#r67123398, then here you could just write `version.get('initialLanguageCode');`

Resources/public/js/views/tabs/ez-locationviewversionstabview.js
```diff
@@ -91,6 +101,70 @@ YUI.add('ez-locationviewversionstabview', function (Y) {
...
+
+                /**
+                 * Fired when the user clicks on the Create draft from archived version button
+                 *
+                 * @event createDraftFromArchivedVersion
+                 * @param {eZ.Content} content
+                 *        {String} versionNo of the archived version. Ex: 42
+                 *
+                 */
+                this.fire('createDraftFromArchivedVersion', {
```
dpobel: the event name is maybe a tad long, you could just call it `createDraft`

Resources/public/js/views/services/plugins/ez-versionsplugin.js
```diff
@@ -0,0 +1,159 @@
...
+                },
+                app = this.get('host').get('app'),
+                content = e.content;
+
+            content.createDraft(options, e.versionNo, Y.bind(function (error, version) {
+                var routeName = 'editContentVersion',
+                    routeParams = {
+                        id: content.get('id'),
+                        languageCode: version.get('initialLanguageCode'),
+                        versionId: version.get('id'),
```
dpobel: you should check the error before using the version

Tests/js/views/services/plugins/assets/ez-versionsplugin-tests.js
```diff
@@ -0,0 +1,311 @@
...
+                    }
+                    Y.fail('Unexpected call to get("' + attr + '")');
+                }, this),
+            });
+
+            Mock.expect(this.content, {
+                method: 'createDraft',
+                args: [Mock.Value.Object, this.versionNo, Mock.Value.Function],
+                run: function (options,versionId, callback) {
+                    callback(error, version);
```
dpobel: in case of error, the second parameter of the callback, you should provide something that looks like a Response object (`{}` would be ok here)

dpobel:
+1 with the last comment fixed

PR #615, opened 2016-06-15
[Behat] Improve element finding with XPath instead of CSS selectors

miguelcleverti:
Using XPath to fetch the elements from the page improves stability of the tests.
Before CSS selectors were being used to fetch the all elements with a given class and then they were iterated until we found the element with the right Text value, since most elements are identified by their Text contents.
XPath give us the possibility to fetch an element directly by its Text so the time frame in which problems can occur is smaller, making the tests more stable, reducing the Selenium [Stale Element Reference Exception](http://docs.seleniumhq.org/exceptions/stale_element_reference.jsp)

**DO NOT MERGE**
Still some work is needed
Studio behat test will break if this is merged

### DIFF
--- a/Features/Context/ContentActions.php
+++ b/Features/Context/ContentActions.php
@@ -69,7 +69,7 @@ class ContentActions extends PlatformUI
      */
     public function confirmRemoval()
     {
-        $this->clickElementByText('Confirm', '.ez-confirmbox-confirm');
+        $this->clickElementByText('Confirm', ['ez-confirmbox-confirm']);
     }
 
     /**
@@ -78,7 +78,7 @@ class ContentActions extends PlatformUI
      */
     public function cancelRemoval()
     {
-        $this->clickElementByText('Cancel', '.ez-confirmbox-cancel');
+        $this->clickElementByText('Cancel', ['ez-confirmbox-cancel']);
     }
 
     /**
@@ -88,11 +88,8 @@ class ContentActions extends PlatformUI
     {
         $element = $this->getElementByText(
             'Are you sure you want to send this content to trash?',
-            '.ez-confirmbox-title'
+            ['ez-confirmbox-title']
         );
-        if (!$element) {
-            throw new \Exception('Confirmation box not found');
-        }
     }
 
     /**
@@ -100,10 +97,7 @@ class ContentActions extends PlatformUI
      */
     public function iSeeConfirmation()
     {
-        $element = $this->getElementByText('Confirm', '.ez-confirmbox-confirm');
-        if (!$element) {
-            throw new \Exception('Confirmation button not found');
-        }
+        $element = $this->getElementByText('Confirm', ['ez-confirmbox-confirm']);
     }
 
     /**
@@ -111,9 +105,6 @@ class ContentActions extends PlatformUI
      */
     public function iSeeCancel()
     {
-        $element = $this->getElementByText('Cancel', '.ez-confirmbox-cancel');
-        if (!$element) {
-            throw new \Exception('Cancel button not found');
-        }
+        $element = $this->getElementByText('Cancel', ['ez-confirmbox-cancel']);
     }
 }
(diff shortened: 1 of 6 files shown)


miguelcleverti:
ping @bdunogier what do you think of this approach to elemente fetching.
We can still use css selectors for some other more 'simple' cases, but this way the tests a much more stable and the fails we are still getting are not random which is good.

bdunogier:
> Before CSS selectors were being used to fetch the all elements with a given class and then they were iterated until we found the element with the right Text value, since most elements are identified by their Text contents.

This is I think the part that is important. We have talked about it with @dpobel, and I think we agree that it is fine to facilitate testing by tweaking the markup in the code PlatformUI generates. _Having_ to search for elements by text is I think a problem, and indeed, the time it takes probably makes the stale errors worse.

@dpobel ?

dpobel:
I fully agree with @bdunogier here I'm curious to know where/what you need to search by text and indeed it happens we don't have the choice yet, it's probably because of class and/or an attribute is missing somewhere in the markup generated by PlatformUI.

miguelcleverti:
@bdunogier @dpobel so how should we procede?
Having a direct way to fetch the elements from the page improves the tests greatly.
Should we wait until the markup is improved or do you want me to add what I think is needed in the markup for the tests?

joaoinacio:
my 2 cents:
Using XPath as implemented in this PR seems to greatly improve on the mostly random stale element ref errors, even if ideally most sentences shouldn't require searching elements by text.

In any case, if searching by text is wanted and/or needed, this seems to improve things a lot (and current failures at least look consistent AFAICT), as it's something that cannot be done as effectively in CSS.

The issues I see in the code are the changing of function signature in getElementByText whch replaces a css selector string with an array of classes.

In any case, this seems like a good approach for the time being, which doesn't really prevent any future improvements on adding additional markup and selecting elements on a different way.

bdunogier:
> Using XPath as implemented in this PR seems to greatly improve on the (mostly random) stale element ref errors, even if ideally most sentences shouldn't require searching elements by text.

Correct. However, we should avoid those as much as possible. And for that, we can and should ask developers to adapt the markup / dom so that it is easy to test, and doesn't require horrors like text search.

Could you, @miguelcleverti, tell to @dpobel where you have to use text search like this ?

miguelcleverti:
@bdunogier, @dpobel currently we are using search by text everywhere because it's the only way to identify the elements. Ex: Navigation items, Roles, Content types, Action bar...

There are some cases that we can avoid using text search by improving the markup, such as static interface elements, by using markup attributes to ease this.

However, to find elements that are dynamic in nature, (most objects that come from the database), we must use text search because we may want to select an element by it's name (or identifier, title, description, etc). Unless we load the markup with such information about the contents, searching by text is the only way.

dpobel:
> Unless we load the markup with such information about the contents, searching by text is the only way.

This is exactly what we should do. We even did that for instance in https://jira.ez.no/browse/EZP-25192.
Actually, you could perfectly do that while implementing the tests at least for simple cases for instance when there's just a PlatformUI template to change because you need an extra class/attribute. For more complicated cases, just create a JIRA issue (as usual).

bdunogier:
@miguelcleverti do you wanna give it a try for the elements you've mentioned ? @dpobel or @yannickroger can guide you, I think.

> Navigation items, Roles, Content types, Action bar...

While xpath performs better at this than CSS, I think that not having to do text value matching altogether will perform better than xpath. 

If you'd rather leave that for the PlatformUI team, it's okay as well. In that case, I suggest that you create an improvement JIRA ticket, with the list of elements you need easier matching for.

miguelcleverti:
Yes  I will add what I think is needed in the elements. If I have doubts I'll ping @dpobel or @yannickroger.

miguelcleverti:
Closing as obsolete

PR #617, opened 2016-06-15
Fix EZP-25884: My Draft block raises an error if the user has less than 10 drafts

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25884

# Description

This patch makes sure we correctly handle the case where the user has less than 10 drafts. Without this patch, the block just display an error.

# Tests

unit tests + manual tests


dpobel:

andrerom:
+1

StephaneDiot:
+1

PR #619, opened 2016-06-17
Fix EZP-25817: Invalid subtree path when assigning role with subtree limitation

yannickroger:
Link: https://jira.ez.no/browse/EZP-25817

## Description
The platform UI used to send invalid subtree limitation that were missing a `/` at the end. That is what this patch is fixing. It can easily be reproduced with https://github.com/ezsystems/ezpublish-kernel/pull/1671 applied (as the REST API will send an error).

## Test
Manual and unit test

### DIFF
--- a/Resources/public/js/views/services/ez-roleserversideviewservice.js
+++ b/Resources/public/js/views/services/ez-roleserversideviewservice.js
@@ -164,6 +164,9 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
             // Creating a limitation in the format described in the JS REST client's `UserService.prototype.newRoleAssignInputStruct()`
             var limitation,
                 limitationValues = Y.Array.map(data.subtreeIds, function(subtreeId) {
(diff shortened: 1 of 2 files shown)


Resources/public/js/views/services/ez-roleserversideviewservice.js
```diff
@@ -164,6 +164,9 @@ YUI.add('ez-roleserversideviewservice', function (Y) {
             // Creating a limitation in the format described in the JS REST client's `UserService.prototype.newRoleAssignInputStruct()`
             var limitation,
                 limitationValues = Y.Array.map(data.subtreeIds, function(subtreeId) {
+                    if (subtreeId.slice(-1) != "/") {
```
dpobel: I would write that with:
```js
subtreeId[subtreeId.length - 1] !== "/"
```

dpobel:
other than the nitpick, +1

StephaneDiot:
+1

glye:
+1

PR #620, opened 2016-06-17
EZP-25705: Prevent embed elements from being added in a list item

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25705

# Description

Make sure it's impossible to add an embed inside a list item.

## Tasks

* [x] Fix a JavaScript error introduced in aada099e550ff84e9d90074bc6d0455ebd3a5d09
* [x] Give the focus (visually) to the whole list when the user is editing a list item
* [x] Make sure to add an element after the list when using the *ezadd* toolbar from a list/list item
* [ ] ~~Try to better position the *ezadd* toolbar to indicate where the new element will be added~~ (created [EZP-25906](https://jira.ez.no/browse/EZP-25906))

# Tests

unit tests + manual tests


andrerom:
+1

PR #621, opened 2016-06-20
EZP-25885: My Drafts block displays an empty table if there's no drafts

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25885

# Description

This patch makes sure we don't display an empty table in the My Drafts dashboard block when the user has no draft. Actually, it introduces a `loading` flag in the Asynchronous Dashboard Block *abstract* view so that we can differentiate the case where we are waiting for the drafts to be loaded from the case where there's no draft.

Also while working on that issue, I've realized the My Content block has exactly the same problem, so this patch also fixes it the same way:

empty_drafts_mycontent

# Tests

manual tests + unit tests

### DIFF
--- a/Resources/public/js/views/dashboard/ez-dashboardblockasynchronousview.js
+++ b/Resources/public/js/views/dashboard/ez-dashboardblockasynchronousview.js
@@ -44,10 +44,19 @@ YUI.add('ez-dashboardblockasynchronousview', function (Y) {
 
             this.get('container')
                 .addClass(this._generateViewClassName(Y.eZ.DashboardBlockBaseView.NAME))
-                .addClass(this._generateViewClassName(Y.eZ.DashboardBlockAsynchronousView.NAME))
-                .addClass(CLASS_LOADING);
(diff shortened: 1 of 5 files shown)


Resources/public/js/views/dashboard/ez-dashboardblockasynchronousview.js
```diff
@@ -44,17 +44,27 @@ YUI.add('ez-dashboardblockasynchronousview', function (Y) {
...
-                .addClass(CLASS_LOADING);
-
-            this.after('itemsChange', this._uiEndLoading);
+                .addClass(this._generateViewClassName(Y.eZ.DashboardBlockAsynchronousView.NAME));
+
+            this.after('itemsChange', function () {
+                this._set('loading', false);
+            });
+            this.after('loadingChange', function () {
+                if ( this.get('loading') ) {
```
andrerom: cs?

andrerom: ignore, I see other places that cs implies spaces in ´if´nut not other places.

andrerom:
+1

PR #622, opened 2016-06-20
EZP-24998: Order of the elements when doing a search is wrong (and random)

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24998

# Description

This patch makes sure the Content Tree and the Subitem views receives the Location lists sorted as indicated by the parent Location `sortOrder` and `sortField` property.

Screencast:

[](http://www.youtube.com/watch?v=tDhFK7X5P5U "Click to play on Youtube.com")

5 sort clauses seem to not be fully implemented server side (see also https://github.com/ezsystems/PlatformUIBundle/pull/622/files#diff-3d4c2abab2b4c380f20d2e09ec804e3cR367):

* (Location)Priority https://jira.ez.no/browse/EZP-25907
* (Location)Path https://jira.ez.no/browse/EZP-25911
* (Location)Depth https://jira.ez.no/browse/EZP-25910 (ok this one is a bit stupid in that case, but can still be set the sort field for a Location)
* Content Type identifier https://jira.ez.no/browse/EZP-25909
* Content Type name https://jira.ez.no/browse/EZP-25908

So now the question is: how do we handle those before having the necessary server side code ?

## Tasks

* [x] Fix [EZP-25846](https://jira.ez.no/browse/EZP-25846)
* [x] Expose the Location search as a public method in addition to the event so that it can be used from others plugins
* [x] Allow to sort the search result based on a Location model
* [x] Fix the Content Tree
* [x] Fix the Subitem views
* [x] Make sure the Subitem views and tree are updated when a Location sort order and/or sort field are updated

# Tests

manual test + unit test

### DIFF
--- a/Resources/public/js/models/ez-locationmodel.js
+++ b/Resources/public/js/models/ez-locationmodel.js
@@ -345,6 +345,64 @@ YUI.add('ez-locationmodel', function (Y) {
             return attrs;
         },
 
+        /**
+         * Returns the REST API sort order based on the sortOrder attribute
+         *
+         * @method _getSortDirection
+         * @protected
+         * @return 'ascending' or 'descending'
+         */
+        _getSortDirection: function () {
+            return this.get('sortOrder') === 'ASC' ? 'ascending' : 'descending';
+        },
+
+        /**
+         * Returns the REST API sort clause identifier based on the sortField
+         * attribute.
+         *
+         * @method _getSortClauseIdentifier
+         * @protected
+         * @return {String}
+         */
+        _getSortClauseIdentifier: function () {
+            switch(this.get('sortField')) {
+                case 'MODIFIED':
+                    return 'DateModified';
+                case 'PUBLISHED':
+                    return 'DatePublished';
+                case 'PATH':
+                    return 'LocationPath'; // Not Implemented server side !?!
+                case 'SECTION':
+                    return 'SectionIdentifier';
+                case 'DEPTH':
+                    return 'LocationDepth';
+                case 'CLASS_IDENTIFIER':
+                    return 'ClassIdentifier'; // Not implemented server side !?!
+                case 'CLASS_NAME':
+                    return 'ClassName'; // Not Implemented server side !?!
+                case 'PRIORITY':
+                    return 'Priority'; // Not Implemented server side !?!
+                case 'NAME':
+                    return 'ContentName';
+                default:
+                    console.warn("Unknow sortField '" + this.get('sortField') + "'");
+                    return 'DateModified';
+            }
+        },
+
+        /**
+         * Returns the sort clause suitable for the Search Plugin based on the
+         * Location's sortOrder and sortField attribute.
+         *
+         * @method getSortClause
+         * @return {Object}
+         */
+        getSortClause: function () {
+            var clause = {};
+
+            clause[this._getSortClauseIdentifier()] = this._getSortDirection();
+            return clause;
+        },
     }, {
         REST_STRUCT_ROOT: "Location",
         ATTRS_REST_MAP: [
--- a/Resources/public/js/models/ez-locationmodel.js
+++ b/Resources/public/js/models/ez-locationmodel.js
@@ -362,26 +362,28 @@ YUI.add('ez-locationmodel', function (Y) {
          *
          * @method _getSortClauseIdentifier
          * @protected
-         * @return {String}
+         * @return {String|Null}
          */
         _getSortClauseIdentifier: function () {
-            switch(this.get('sortField')) {
+            var sortField = this.get('sortField');
+
+            switch(sortField) {
                 case 'MODIFIED':
                     return 'DateModified';
                 case 'PUBLISHED':
                     return 'DatePublished';
                 case 'PATH':
-                    return 'LocationPath'; // Not Implemented server side !?!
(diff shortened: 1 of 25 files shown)


dpobel:
(ping @andrerom @bdunogier for the question)

andrerom:
> So now the question is: how do we handle those before having the necessary server side code ?

got the impression you where handling this with @bdunogier this morning, but ok; how does it behave on unimplemented once? crashing or just sorting wrong?

dpobel:
well, in the subitem view it displays an error notification saying the subitems could not be loaded (the subitem view is an inconsistent state, I'm currently fixing that). In the tree, you can not unfold the element and there's a red cross next to the Content name.

andrerom:
> well, in the subitem view it displays an error notification saying the subitems could not be loaded (the subitem view is an inconsistent state, I'm currently fixing that). In the tree, you can not unfold the element and there's a red cross next to the Content name.

Ok, it's clear we need to add implementation on that before we merge this then.

dpobel:
for the unimplemented sort clauses, we can also force another sort clause to be used instead, then the sorting would be wrong but the interface would be usable.

dpobel:
I'm sorry but given the time frame regarding the freeze it's a product decision: either no sorting, either a buggy one.

bdunogier:
I don't see any reasons why those can't be implemented. I'll give it a shot.

Can I test the REST changes by pulling your PR in Damien ?

bdunogier:
Implemented the 3 SortClauses that exist in the Public API. Unfortunately, `ContentTypeName` and `ContentTypeIdentifier` don't exist in the Public API.

andrerom:
> Unfortunately, ContentTypeName and ContentTypeIdentifier don't exist in the Public API.

> for the unimplemented sort clauses, we can also force another sort clause to be used instead, then the sorting would be wrong but the interface would be usable.


@dpobel For the two mentioned by @bdunogier that we on purpose are missing, and any undefined once, map them to not use anything at all or sane default as you prefer.

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -345,6 +345,64 @@ YUI.add('ez-locationmodel', function (Y) {
...
+                case 'PATH':
+                    return 'LocationPath'; // Not Implemented server side !?!
+                case 'SECTION':
+                    return 'SectionIdentifier';
+                case 'DEPTH':
+                    return 'LocationDepth';
+                case 'CLASS_IDENTIFIER':
+                    return 'ClassIdentifier'; // Not implemented server side !?!
+                case 'CLASS_NAME':
+                    return 'ClassName'; // Not Implemented server side !?!
```
andrerom: These two will not be implemented.

Could for instance _(since warning is not needed in this case, swap out `console.log` if something else if better)_:
```diff
+                case 'CLASS_IDENTIFIER':
+                case 'CLASS_NAME':
+                    console.log("Deprecated sort field used '" + this.get('sortField') + "', falling back to 'DateModified'.");
+                    return 'DateModified';
```

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -345,6 +345,64 @@ YUI.add('ez-locationmodel', function (Y) {
...
+                case 'PRIORITY':
+                    return 'Priority'; // Not Implemented server side !?!
```
andrerom: I think @bdunogier renamed this to `LocationPriority` to clarify what it is (when using php api on this, namespace will reflect it).

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -345,6 +345,69 @@ YUI.add('ez-locationmodel', function (Y) {
...
+                case 'PATH':
+                    return 'LocationPath';
+                case 'SECTION':
+                    return 'SectionIdentifier';
+                case 'DEPTH':
+                    return 'LocationDepth';
+                case 'CLASS_IDENTIFIER':
+                case 'CLASS_NAME':
+                    console.warn(sortField + ' sort order is unsupported at the moment');
+                    return null;
```
andrerom: see: https://github.com/ezsystems/PlatformUIBundle/pull/622#discussion_r68007200

Resources/public/js/models/ez-locationmodel.js
```diff
@@ -345,6 +345,69 @@ YUI.add('ez-locationmodel', function (Y) {
...
+                    return 'DatePublished';
+                case 'PATH':
+                    return 'LocationPath';
+                case 'SECTION':
+                    return 'SectionIdentifier';
+                case 'DEPTH':
+                    return 'LocationDepth';
+                case 'CLASS_IDENTIFIER':
+                case 'CLASS_NAME':
+                    console.warn(sortField + ' sort order is unsupported at the moment');
```
andrerom: see: https://github.com/ezsystems/PlatformUIBundle/pull/622#discussion_r68007200

If you return null or not I'm not sure of, I assume null means no sorting, but could behave like default.

mathieupoulenard:
I test some diff yesterday and sorting was correct. Great job !

andrerom:
+1, but @StephaneDiot and/or @dew326 should have a look perhaps, as it's big.

dew326:
it's ok for me, +1

PR #623, opened 2016-06-21
Call api only if content id is not empty and reformat list

mathieupoulenard:
richtext embed : call api only if object contains ids
search-plugin : avoid list ids start with a comma (cause bad SQL request
and fail on postgresql databases)

### DIFF
--- a/Resources/public/js/views/fields/richtext/ez-richtext-resolveembed.js
+++ b/Resources/public/js/views/fields/richtext/ez-richtext-resolveembed.js
@@ -78,14 +78,16 @@ YUI.add('ez-richtext-resolveembed', function (Y) {
      * @param {eZ.FieldView|eZ.FieldEditView} view
      */
     ResolveEmbed.prototype._loadEmbeds = function (mapNode, view) {
-        view.fire('contentSearch', {
-            viewName: 'resolveembed-field-' + view.get('field').id,
-            search: {
-                criteria: {'ContentIdCriterion': Object.keys(mapNode).join(',')},
-                offset: 0,
-            },
-            callback: Y.bind(this._renderEmbed, this, mapNode),
-        });
+        if(0 !== Object.keys(mapNode).length) {
+            view.fire('contentSearch', {
+                viewName: 'resolveembed-field-' + view.get('field').id,
+                search: {
+                    criteria: {'ContentIdCriterion': Object.keys(mapNode).join(',')},
+                    offset: 0,
+                },
+                callback: Y.bind(this._renderEmbed, this, mapNode),
+            });
+        }
     };
 
     /**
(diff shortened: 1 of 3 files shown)


aymericDD:
:+1:

andrerom:
Thanks @mathieupoulenard, will you take care about the js style reports so we can get this in? Also if you'd like this to go into 1.3.x you'll need to rebase and open the PR against 1.3 branch.

dpobel:
Thanks @mathieupoulenard
Could you report an issue in https://jira.ez.no with a description of the bug you are fixing (also make sure to set the Components to PlatformUI) ?

Resources/public/js/views/fields/richtext/ez-richtext-resolveembed.js
```diff
@@ -78,14 +78,16 @@ YUI.add('ez-richtext-resolveembed', function (Y) {
...
     ResolveEmbed.prototype._loadEmbeds = function (mapNode, view) {
-        view.fire('contentSearch', {
-            viewName: 'resolveembed-field-' + view.get('field').id,
-            search: {
-                criteria: {'ContentIdCriterion': Object.keys(mapNode).join(',')},
-                offset: 0,
-            },
-            callback: Y.bind(this._renderEmbed, this, mapNode),
-        });
+        if(0 !== Object.keys(mapNode).length) {
```
dpobel: I think this is not the correct place for the condition. I would instead check `embeds.size()` in `ResolveEmbed.prototype.process` that way both resolver would be fixed with one change.

dpobel:
also I realized you are fixing two different issues, you should then report 2 issues in JIRA and open 2 pull requests for those. Also I'm missing the corresponding unit tests.

For the resolvers, you'll have to add a test case in https://github.com/ezsystems/PlatformUIBundle/blob/master/Tests/js/views/fields/richtext/assets/ez-richtext-resolveembed-tests.js and https://github.com/ezsystems/PlatformUIBundle/blob/master/Tests/js/views/fields/richtext/assets/ez-richtext-resolveimage-tests.js to make sure we don't fire the `contentSearch` event if there's no embed nor image.

For the other bug, I think you'll have to add an assertion in https://github.com/ezsystems/PlatformUIBundle/blob/master/Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js#L905 to make sure the criterion value does not start with a comma.

andrerom:
@mathieupoulenard Please note that we are aware this might feel like a lot of things to deal with as your first PR, so don't hesitate to ask questions or ask for help either her or on slack.  Also if you see that this is not something you can contribute on in the next weeks, give us a heads up and we'll see if we can find time to more activly handle this issue _(but that typically means at the expense of other issues)_.

mathieupoulenard:
What is the deadline for this PR ? I have not enough time this weeks to do all fixes :(

andrerom:
There is no deadline for PR's, but we have release code freezes, 1.4.0 being around lunch tomorrow, so realistically speaking to late for this one. And 1.4.1 before 12th of July.

dpobel:
I can maybe fix those (or one of those) before the freeze, but please create the JIRA issues :)

dpobel:
FYI, I created https://jira.ez.no/browse/EZP-25941 for the useless requests

mathieupoulenard:
Second issue created : https://jira.ez.no/browse/EZP-25961

dpobel:
Replaced by #627 and #632

PR #628, opened 2016-06-23
EZP-24941: UDW becomes visible when decreasing the height of the viewport

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24941

# Description

A part of the UDW can appear if the user decreases the height of the viewport. This is happening because of the way we are hiding the UDW (with the sliding transition). This patch just makes sure the UDW is really invisible for the user.

# Tests

manual tests


StephaneDiot:
+1

PR #629, opened 2016-06-23
EZP-25874 ESC and back while discovering exits content edit

StephaneDiot:
jira : https://jira.ez.no/browse/EZP-25874 
## Description

The fix here is that  when the UDW is launched, pressing esc in edit mode or pressing the back button of the browser don't change the view which is "behind" the UDW.

The hash is also keeped and the  browser history as well.

## Tests

Unit and manually tested

### DIFF
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -120,6 +120,26 @@ YUI.add('ez-platformuiapp', function (Y) {
         },
 
         /**
+         * The hash change event subscription.
+         *
+         * @property _hashSubscription
+         * @default null
+         * @type EventHandle
+         * @protected
+         */
+        _hashChangeSubscription: null,
+
+        /**
+         * The current location hash .
+         *
+         * @property _currentLocationHash
+         * @default null
+         * @type string
+         * @protected
+         */
+        _currentLocationHash: null,
+
+        /**
          * Initializes the application.
          *
          * @method initializer
@@ -142,6 +162,15 @@ YUI.add('ez-platformuiapp', function (Y) {
                 'loadingChange': this._loading,
                 'navigate': function () {
                     this.set('loading', true);
+                },
+            });
+            this.after('routingEnabledChange', function () {
+                if (!this.get('routingEnabled')) {
+                    this._preventHashChange();
+                    this._disableRouting();
+                } else {
+                    this._allowHashChange();
+                    this._enableRouting();
                 }
             });
 
@@ -152,6 +181,58 @@ YUI.add('ez-platformuiapp', function (Y) {
                 if (oldService && newService) {
                     oldService.setNextViewServiceParameters(newService);
                 }
+            });
+        },
+
+        /**
+         * Prevent hash changes by setting the old hash on hashChange event
+         *
+         * @method  _preventHashChange
+         * @protected
+         */
+        _preventHashChange: function() {
+            var app = this;
+
+            this._hashChangeSubscription = Y.on('hashchange', function (e) {
+                if ( e.newHash != app._currentLocationHash ) {
+                    Y.HistoryHash.setHash(e.oldHash);
+                    app._currentLocationHash = e.oldHash;
+                }
+            });
+        },
+
+        /**
+         * Allows hash changes by detaching hashChange subscription
+         *
+         * @method  _allowHashChange
+         * @protected
+         */
+        _allowHashChange: function() {
+            this._hashChangeSubscription.detach();
+        },
+
+        /**
+         * Disable routing by deleting the callbacks of the routes
+         *
+         * @method _disableRouting
+         * @protected
+         */
+        _disableRouting: function() {
+            this.get('routes').forEach(function(route) {
+                route._initialCallbacks = route.callbacks;
+                route.callbacks = [];
+            });
+        },
+
+        /**
+         * Enables routing by setting back the callbacks to the routes
+         *
+         * @method _enableRouting
+         * @protected
+         */
+        _enableRouting: function() {
+            this.get('routes').forEach(function(route) {
+                route.callbacks = route._initialCallbacks;
             });
         },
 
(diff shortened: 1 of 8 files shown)


Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -140,8 +142,20 @@ YUI.add('ez-platformuiapp', function (Y) {
...
+                '*:cancelDiscover': function() { this.set('isDiscovering', false)},
+                '*:contentDiscovered': function() { this.set('isDiscovering', false)},
+                '*:navigate': function () {
                     this.set('loading', true);
+                },
+
+            });
+
+            Y.on('hashchange', function (e) {
+                if ( app.get('isDiscovering') && e.newHash != app.get('hash') ) {
```
dpobel: If I understand correctly, this part is for closing the UDW when using the back button ? if so, this case should be handled separately in an improvement since it's an improvement not a bug. Let's focus only the navigation caused by the back button

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -140,8 +142,20 @@ YUI.add('ez-platformuiapp', function (Y) {
 
             this.on({
                 'loadingChange': this._loading,
-                'navigate': function () {
+                '*:contentDiscover': function() { this.set('isDiscovering', true)},
```
dpobel: hum, that's a bit too connected with the UDW, the app should not really know about those event, so the logic preventing the navigation should probably lie in the UDW view service and/or eZ.Plugin.UniversalDiscovery.

Resources/public/js/apps/ez-platformuiapp.js
StephaneDiot: this is for closing the UDW ( by firing cancelDiscover ) but mostly to keep the same hash (which is part of the bug and not for the improvement)

Resources/public/js/apps/ez-platformuiapp.js
dpobel: actually, after thinking about it, maybe you can try something:
in eZ.Plugin.UniversalDiscovery when showing the universal discovery you could manipulate the application's routes (either completely remove them or just changing the middlewares for a *no operation* middleware) so that the navigate becomes impossible. And of course, when hiding the UDW, you should restore the app that the ability to navigate is restored.

StephaneDiot:
updated @dpobel (without hashchange test)

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -120,6 +120,16 @@ YUI.add('ez-platformuiapp', function (Y) {
...
 
         /**
+         * The hash event subscription set to handle hash changes
+         *
+         * @property _hashSubscription
+         * @default null
+         * @type EventHandle
+         * @protected
+         */
+        _hashSubscription: null,
```
dpobel: should rather be called `_hashChangeSubscription`

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -120,6 +120,16 @@ YUI.add('ez-platformuiapp', function (Y) {
         },
 
         /**
+         * The hash event subscription set to handle hash changes
```
dpobel: hash change

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -140,9 +150,13 @@ YUI.add('ez-platformuiapp', function (Y) {
 
             this.on({
                 'loadingChange': this._loading,
-                'navigate': function () {
+                '*:navigate': function () {
```
dpobel: not needed I guess

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -140,9 +150,13 @@ YUI.add('ez-platformuiapp', function (Y) {
...
                     this.set('loading', true);
-                }
+                },
+            });
+            this.after('routingEnabledChange', function () {
+                this._handleHash();
```
dpobel: I would rather add the condition on the new value of `routingEnabled` here to have better method names, so `_handleHash` would be split into `_preventHashChange` and `_allowHashChange` and `_alterRoutesCallback` into `_disableRouting` and `_enableRouting` for instance.

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -156,6 +170,50 @@ YUI.add('ez-platformuiapp', function (Y) {
...
         /**
+         * Add or deletes handleSideViews and handleMainViews from the routes callbacks
+         *
+         * @method _alterRoutesCallback
+         * @protected
+         */
+        _alterRoutesCallback: function() {
+            var app = this;
+
+            Y.Object.each(app.get('routes'), function(route) {
```
dpobel: `routes` value is an array, no ? if so you can iterate with `this.get('routes').forEach()`

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -156,6 +170,50 @@ YUI.add('ez-platformuiapp', function (Y) {
...
+                var index = route.callbacks.indexOf('handleSideViews');
```
dpobel: that's a bad condition, actually each and every route should be disabled somehow. Most likely you need to store somewhere the initial `callbacks` values and restore them afterwards.

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -629,7 +687,7 @@ YUI.add('ez-platformuiapp', function (Y) {
                     });
                 };
 
-            if ( serviceInstance ) {
```
dpobel: the following changes in this file are completely useless, please remove

StephaneDiot:
ping @yannickroger @dpobel

StephaneDiot:
self +1 !

dpobel:
self +1 as well :) @yannickroger can you have a look please ?

yannickroger:
+1

andrerom:
+1, but seems it needs a rebase

PR #631, opened 2016-06-23
EZP-25323: ContentType name and description not showed if it is not translated in eng-GB

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25323

# Description

This patch removes various hardcoded `eng-GB` when dealing with Content Type. It introduces a new extension component to be apply on any component that need to pick a translation from a hash indexed by eZ Locale and the corresponding Handlebars helper. For now, we only use the language(s) configured in the browser to try to pick the best translation from the hash and in the last option, we take the first available string, that way, the Content Type name (or field definition name or description) will never be empty because of inexistant `eng-GB` version.

This also involves passing the Locales map defined in locales.yml and making sure the components that need to pick the right translation (the Content Type Selector at the moment) also receives this configuration.

## Tasks

* [x] Provide the Locales map to the JavaScript application
* [x] Implement a Handlebars function similar to `ez_trans_prop`
* [x] Change the templates to use this one
* [x] Find a solution for views and view services doing the same

# Tests

manual test + unit test


dpobel:

dpobel:
(self note: do NOT forward that blindly, 1.4/master needs additional changes like in the Subitem List More view or the dashboard)

Resources/public/js/extensions/ez-translateproperty.js
```diff
@@ -0,0 +1,154 @@
...
+
+    Y.eZ.TranslateProperty = Y.Base.create('translateProperty', Y.Base, [], {
+        /**
+         * Find the correct property translation according to browser's
+         * languages configuration. It basically iterates over the browser's
+         * language to find a matching translation, if none matches, it takes
+         * the first one.
+         *
+         * @method translateProperty
+         * @param {Object} fullLocalesMap TODO
```
StephaneDiot: missing param description

StephaneDiot:
+1

Resources/public/js/extensions/ez-translateproperty.js
```diff
@@ -0,0 +1,155 @@
...
+         * languages to find a matching translation, if none matches, it takes
+         * the first one.
+         *
+         * @method translateProperty
+         * @param {Object} fullLocalesMap the Locales map between POSIX Locales
+         * (e.g.) fr_FR and Locales used in eZ Platform (e.g. fre-FR)
+         * @param {Object} property the property to translate as a hash indexed
+         * by eZ Locale for example:
+         *
+         *   {'eng-GB': 'potatoes', 'fre-FR': 'pomme de terre'}
```
yannickroger: potato :)

yannickroger:
+1

PR #633, opened 2016-06-28
Fix EZP-25148: Stale draft after canceling a content creation

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25148

# Description

When a Content item only has a DRAFT version (right after creating a Content item and clicking on *Save* for instance), this version (and the corresponding Content item) stays in the repository even if the user clicks on *Discard draft*. This patch makes sure in such case we remove the Content item because removing the version alone is not allowed since it's the last version of the Content item.

# Tests

manual + unit tests


dpobel:

andrerom:
+1

yannickroger:
+1

PR #635, opened 2016-06-29
Fix EZP-25980: Trashing a container Content item without sub-items raises a JavaScript error

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25980

# Description

If you send to trash a container Content item without any sub-items, you'll get a JavaScript error while you are redirected to the parent Location. This is happening because the subitem views subscribe to the `sortFieldChange` and `sortOrderChange` events and the corresponding attribute are updated in the redirection process.
This patch makes sure the active state of the view is updated earlier so that before loading the parent Location, the views become inactive and then refresh is ignored.

# Tests

manual tests + unit tests

### DIFF
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -629,6 +629,9 @@ YUI.add('ez-platformuiapp', function (Y) {
                     });
                 };
 
+            if ( this.get('activeView') ) {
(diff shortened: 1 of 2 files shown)


yannickroger:
+1

dpobel:
ping @StephaneDiot

StephaneDiot:
+1

PR #636, opened 2016-06-29
Fix EZP-26006: Subitem order not correctly updated

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26006
Discovered while working on EZP-25980 / #635 

# Description

The subitem views are not correctly updated if they are not visible/active when changing the sort field or the sort order of a Location so you could have the subitem list view sorted by name and the subitem grid view sorted by priority while the Location sort order is set to modified date.

# Tests

manual + unit tests


yannickroger:
+1

dpobel:
ping @StephaneDiot

StephaneDiot:
+1

PR #640, opened 2016-07-04
EZP-25463: Delete drafts in version tab

yannickroger:
Link: https://jira.ez.no/browse/EZP-25463

## Description
Allows to delete drafts displayed in the version tab

delete_draft

## Test
Manual and unit test

## TODO
- [x] Feature
- [x] Unit test
- [x] Screenshot/screencast
- [x] Rebase and cleanup

### DIFF
--- a/Resources/public/js/views/services/plugins/ez-versionsplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-versionsplugin.js
@@ -19,6 +19,9 @@ YUI.add('ez-versionsplugin', function (Y) {
      * In order to use it you need to fire `loadVersions` event with parameter
      * `content` containing the eZ.Content object for which you want to load versions.
      *
+     * Deleting versions:
+     * In order to use it you need to fire `deleteVersionDraft`
+     *
      * Creating draft based on an archived version:
      * In order to use it you need to fire 'createDraft' with corresponding
      * parameters (see method).
@@ -33,6 +36,7 @@ YUI.add('ez-versionsplugin', function (Y) {
         initializer: function () {
             this.onHostEvent('*:loadVersions', this._loadVersions);
             this.onHostEvent('*:createDraft', this._createDraftFromArchivedVersion);
+            this.onHostEvent('*:deleteVersionDraft', this._deleteDraftConfirm);
         },
 
         /**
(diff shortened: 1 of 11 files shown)


Resources/public/js/views/services/plugins/ez-versionsplugin.js
```diff
@@ -132,6 +136,101 @@ YUI.add('ez-versionsplugin', function (Y) {
...
+                    });
+
+                version.destroy({remove: true, api: service.get('capi')}, end);
+            });
+
+            tasks.done(function () {
+                var errorNotificationIdentifier, successNotificationIdentifier,
+                    versionsRemoved = (countRemovedVersions > 0);
+
+                if (countRemovedVersions === versions.length) {
```
dpobel: that seems a bit complicated for nothing, the notification identifier should not depend on the result of the operation

Resources/public/js/views/tabs/ez-locationviewversionstabview.js
```diff
@@ -156,15 +180,90 @@ YUI.add('ez-locationviewversionstabview', function (Y) {
...
+
+        /**
+         * Callback function called after removing version(s).
+         *
+         * @method _afterDeleteVersionsCallback
+         * @protected
+         * @param {Boolean} versionsRemoved if TRUE the view is reloaded, if FALSE it just enables checkboxes
+         */
+        _afterDeleteVersionsCallback: function (versionsRemoved) {
+            if (versionsRemoved) {
```
dpobel: I would always refresh the version list even in case of error, so you can completely remove that method and use `this._refresh` as the `afterDeleteVersionsCallback`

Resources/public/js/views/tabs/ez-locationviewversionstabview.js
```diff
@@ -156,15 +180,90 @@ YUI.add('ez-locationviewversionstabview', function (Y) {
...
+         * Deletes one or several drafts
+         *
+         * @method _deleteDraft
+         * @protected
+         */
+        _deleteDraft: function () {
+            var c = this.get('container'),
+                versions = [];
+
+            versions = Y.Array.reject(this.get('versions').DRAFT, function (version) {
```
dpobel: this would deserve to be an external (private) property

Resources/public/js/views/services/plugins/ez-versionsplugin.js
```diff
@@ -132,6 +136,101 @@ YUI.add('ez-versionsplugin', function (Y) {
...
+                tasks = new Y.Parallel();
+
+            this._notify(
+                "Removing versions for '" + content.get('name') + "'",
+                notificationIdentifier,
+                'started',
+                5
+            );
+
+            Y.Array.each(versions, function (version) {
```
dpobel: `versions.forEach` :-p

Resources/public/js/views/services/plugins/ez-versionsplugin.js
```diff
@@ -132,6 +136,101 @@ YUI.add('ez-versionsplugin', function (Y) {
...
+        /**
+         * Removes given versions. After that, calls the callback function.
+         *
+         * @method _deleteVersion
+         * @protected
+         * @param {Array} versions array containing eZ.VersionInfo objects for removal
+         * @param {Function} callback
+         */
+        _deleteVersion: function (versions, callback) {
+            var that = this,
```
dpobel: please bind the function instead

Resources/public/js/views/services/plugins/ez-versionsplugin.js
yannickroger: @dpobel the behavior is copy pasted from remove location. The goal behind this is to know which notification replaces which. Especially to handle the case where only one (or a few) deletion have errors but not all of them. So we need to display both error and success messages. Do you think we should simplify it ?

Resources/public/js/views/tabs/ez-locationviewversionstabview.js
yannickroger: We also use it when the user clicks on the cancel button in the confirm box. So it would also refresh that behavior. Cool with that ?

dpobel: good remark, indeed in that case, the refresh is clearly not necessary

yannickroger:
Ready for final review @dpobel @StephaneDiot and others

StephaneDiot:
+1

dpobel:
+1 when Travis agrees (besides behat)

PR #642, opened 2016-07-05
EZP-26011: Content on the Fly embeds unreliable with the Solr Bundle

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26011
Requires https://github.com/ezsystems/content-on-the-fly-prototype-bundle/pull/15

# Description

To render the embed elements and the images in the RichText editor a search is done to load the corresponding Content(Info). Right after creating a Content, the search index might not be up to date yet, as a result the embed or the image added based on a Content that was just created with the Content on the Fly feature can not be done.

To avoid that, this patch changes the way the embed/image tag is rendered right after the usage of the Universal Discovery Widget to use the selection provided by the UDW instead of searching for the Content(Info).

# Tests

manual tests + unit tests

### DIFF
--- a/Resources/public/js/alloyeditor/buttons/embed.js
+++ b/Resources/public/js/alloyeditor/buttons/embed.js
@@ -59,7 +59,7 @@ YUI.add('ez-alloyeditor-button-embed', function (Y) {
             this.execCommand();
             this._setContentInfo(contentInfo);
             widget = this._getWidget().setWidgetContent('');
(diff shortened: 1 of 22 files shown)


yannickroger:
+1

andrerom:
+1  _Note: We still need to find a way to omit using search for embeds for cases where user clicks directly publish before it's been indexed, and for cases where embeds might not have a location. Ref being able to do multi GET against content info api in REST and PHP._

PR #643, opened 2016-07-06
Fix EZP-26027: Pressing enter with a focused embed creates a new empty embed

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26027
New behavior caused by https://github.com/ezsystems/PlatformUIBundle/pull/596

# Description

This patch makes sure that using Enter with a focused embed or double clicking on an embed element does not add a new embed element.
Note: the solution is the one suggested in CKEditor doc http://docs.ckeditor.com/#!/api/CKEDITOR.plugins.widget-event-doubleclick

# Tests

manual tests + unit tests


andrerom:
+1

yannickroger:
+1

StephaneDiot:
+1

PR #647, opened 2016-07-08
EZP-25462: Delete an archived in Version tab

yannickroger:
Link: https://jira.ez.no/browse/EZP-25462

## Description
Adds a delete button allowing deleting archived versions.

delete_archived

## Test
Manual and unit tests

### DIFF
--- a/Resources/public/js/views/services/plugins/ez-versionsplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-versionsplugin.js
@@ -20,7 +20,7 @@ YUI.add('ez-versionsplugin', function (Y) {
      * `content` containing the eZ.Content object for which you want to load versions.
      *
      * Deleting versions:
-     * In order to use it you need to fire `deleteVersionDraft`
+     * In order to use it you need to fire `deleteVersion`
      *
      * Creating draft based on an archived version:
      * In order to use it you need to fire 'createDraft' with corresponding
@@ -36,7 +36,7 @@ YUI.add('ez-versionsplugin', function (Y) {
         initializer: function () {
             this.onHostEvent('*:loadVersions', this._loadVersions);
             this.onHostEvent('*:createDraft', this._createDraftFromArchivedVersion);
-            this.onHostEvent('*:deleteVersionDraft', this._deleteDraftConfirm);
+            this.onHostEvent('*:deleteVersion', this._confirmDeleteVersion);
         },
 
         /**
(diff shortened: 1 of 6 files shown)


yannickroger:
Ready for review @dpobel @StephaneDiot and others

crevillo:
don't really know if this part of this pull request, but wondering what it is expected to happen when in the "Archived Versions" part you select more than one and click in the `New draft based on Selected version` button. Will it start from the more recent one?

Resources/public/js/views/services/plugins/ez-versionsplugin.js
```diff
@@ -136,19 +136,19 @@ YUI.add('ez-versionsplugin', function (Y) {
...
          * are going to be deleted
          *
-         * @method _deleteDraftConfirm
+         * @method _deleteVersionConfirm
          * @private
-         * @param {EventFacade} e deleteVersionDraft event facade
+         * @param {EventFacade} e deleteVersion event facade
          */
-        _deleteDraftConfirm: function (e) {
+        _deleteVersionConfirm: function (e) {
```
dpobel: I missed it in the previous review but it should probably be called `_confirmDeleteVersion` instead

Resources/public/js/views/tabs/ez-locationviewversionstabview.js
```diff
@@ -202,26 +204,65 @@ YUI.add('ez-locationviewversionstabview', function (Y) {
         },
 
         /**
+         * Returns the selected archived versions as VersionInfo
+         *
+         * @method _getSelectedArchived
+         * @protected
+         * @return {Array} of eZ.VersionInfo
+         */
+        _getSelectedArchived: function () {
```
dpobel: it's almost the same as `_getSelectedDrafts`, you can probably reorganize a bit the code to use the same method

dpobel:
@crevillo I think this is handled in https://github.com/ezsystems/PlatformUIBundle/pull/647/files#diff-0cd9f2e49a272c207aecf6b781c2e5a3R126

yannickroger:
@crevillo the button should be disabled if you select more than one `New Draft based on selected version`. If you have suicidal tendencies and use dev tool to enable it, it will just not to anything :smile:

dpobel:
besides the 2 inline comments +1 (when Travis agrees on unit tests)

crevillo:
cool :smile: i should read the code before ask :relaxed:

PR #646, opened 2016-07-08
EZP-26049: Create Content from the Dashboard

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26049
Requires https://github.com/ezsystems/PlatformUIBundle/pull/648

# Description

This patch adds the ability to create a Content item from the dashboard. To do that, there's a new component called the Content Creation Wizard that allows to pick a Content Type and a parent Location for the future Content item. Then the user is redirected to the route implemented in #648 to create the corresponding Content item.

Screencast:

[](http://www.youtube.com/watch?v=T117_wQf_bg "Click to play on Youtube.com")

## Tasks

* [x] Add the create button on the dashboard
* [x] Add the Content Creation Wizard side view
* [x] Allow to pick a Content Type from the list
* [x] Allow to pick a parent Location for the future Content item
* [x] ~~Tweak/Add a new route to create Content~~ see https://github.com/ezsystems/PlatformUIBundle/pull/648
* [x] Redirect the user to this route

# Tests

manual + unit tests

PR #648, opened 2016-07-15
EZP-26049: Change the route used to create a Content item to accept parameters

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26049
Required by https://github.com/ezsystems/PlatformUIBundle/pull/646

# Description

This patch introduces a new route called `createContentUnder` to replace `createContent`. This route expects the Content Type id and the parent Location id to be passed in the URI. This makes the content creation form easier to reach and is a requirement to allow the content creation from the dashboard (see #646). The `createContent` route still works but several deprecation warning will be logged to the console when using it.

## Tasks

* [x] Deprecate the `createContent` route
* [x] Add a new route with the parameters
* [x] Make sure to not use any deprecated method/attribute/route

# Tests

unit tests + manual tests

### DIFF
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -941,6 +941,13 @@ YUI.add('ez-platformuiapp', function (Y) {
                     view: 'contentEditView',
                     callbacks: DEFAULT_ROUTE_CALLBACKS,
                 }, {
+                    name: "createContentUnder",
+                    path: '/create/:contentTypeId/:parentLocationId',
+                    service: Y.eZ.ContentCreateViewService,
+                    sideViews: {'navigationHub': false, 'discoveryBar': false},
+                    view: 'contentEditView',
+                    callbacks: DEFAULT_ROUTE_CALLBACKS,
+                }, {
                     name: "createContent",
                     path: '/create',
                     service: Y.eZ.ContentCreateViewService,
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -13,6 +13,7 @@ YUI.add('ez-platformuiapp', function (Y) {
     Y.namespace('eZ');
 
     var L = Y.Lang,
+        DEFAULT_ROUTE_CALLBACKS = ['open', 'checkUser', 'handleSideViews', 'handleMainView'],
         APP_OPEN = 'is-app-open',
         APP_LOADING = 'is-app-loading';
 
@@ -831,6 +832,15 @@ YUI.add('ez-platformuiapp', function (Y) {
             }
         },
     }, {
+        /**
+         * The default route callbacks
+         *
+         * @static
+         * @property DEFAULT_ROUTE_CALLBACKS
+         * @type {Array}
+         */
+        DEFAULT_ROUTE_CALLBACKS: DEFAULT_ROUTE_CALLBACKS,
+
         ATTRS: {
             /**
              * Stores the available routes for the application.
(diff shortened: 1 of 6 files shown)


dpobel:

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -489,6 +490,23 @@ YUI.add('ez-platformuiapp', function (Y) {
...
+         *
+         * @method logDeprecatedRoute
+         * @param {Object} req
+         * @param {Object} res
+         * @param {Function} next
+         */
+        logDeprecatedRoute: function (req, res, next) {
+            var route = req.route;
+
+            console.log('[DEPRECATED] The route "' + route.name + '" (path "' + route.path + '") is deprecated');
```
sunpietro: Should't it be `console.warn`?

Resources/public/js/views/services/ez-contentcreateviewservice.js
```diff
@@ -26,7 +26,95 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
...
+
+            this._loadParentLocation(request.params.parentLocationId, tasks.add(Y.bind(function (err) {
+                if ( err ) {
+                    hasError = true;
+                    this._error("Could not load the parent Location and Content");
+                }
+            }, this)));
+
+            tasks.done(Y.bind(function () {
+                if ( !hasError ) {
```
sunpietro: What happens if there's any error?

dpobel: the error event is fired in case of loading error (5 and 12 lines above) and the navigation is stopped (ie the callback is not called).

Resources/public/js/apps/ez-platformuiapp.js
dpobel: well, the console object lacks a `deprecated` log level. To me `warn` is more for potential issues close to error (a deprecation is not really in that category). Maybe we could use `notice` but we should probably have a component responsible for logging the deprecations first because such console.log calls are spread all over the place.

Resources/public/js/views/services/ez-contentcreateviewservice.js
sunpietro: Thanks for explanation. Won't it display a blank page, will it? What do you think about adding a redirection to previously visited view or dashboard when error occurs?

Resources/public/js/views/services/plugins/ez-contentcreateplugin.js
```diff
@@ -122,6 +131,24 @@ YUI.add('ez-contentcreateplugin', function (Y) {
             });
             app.navigate(app.routeUri('createContent'));
```
yannickroger: You should add a deprecate log in this method

Tests/js/views/services/assets/ez-contentcreateviewservice-tests.js
```diff
@@ -3,13 +3,367 @@
...
+                app: this.app,
+                capi: this.capi,
+                request: this.request,
+            });
+        },
+
+        tearDown: function () {
+            this.service.destroy();
+            delete this.service;
+            Y.eZ.Location = this.origLocationModel;
```
yannickroger: You should do the same for `eZ.Content`

yannickroger:
+1 with feedback taken into account

Resources/public/js/views/services/ez-contentcreateviewservice.js
dpobel: normally, you stay on the same page if an error happens at this stage. We could probably improve the error handling but that's a bit outside of the scope of this story/PR but feel free to open a improvement in JIRA with ideas and expected behavior.

dpobel:
merged to feature branch `feature/content-from-dashboard`

PR #649, opened 2016-07-15
EZP-26081: [IE] ae-toolbars not displayed on Microsoft Edge

iosolinf:
I know that no one in his/her perfect mind should be using this so called "browser", but I have a customer that uses it and he's now unable to format embed images or paragraphs in rich text.
The issue seems to be related with MS not fully supporting outline-offset.

With this simple patch, a fallout value is used when block.getComputedStyle('outline-offset') returns NaN

### DIFF
--- a/Resources/public/js/alloyeditor/toolbars/config/block-base.js
+++ b/Resources/public/js/alloyeditor/toolbars/config/block-base.js
@@ -16,9 +16,11 @@ YUI.add('ez-alloyeditor-toolbar-config-block-base', function (Y) {
     var ReactDOM = Y.eZ.AlloyEditor.ReactDOM;
 
     function outlineTotalWidth(block) {
-        var outlineOffset = parseInt(block.getComputedStyle('outline-offset'), 10);
-        var outlineWidth = parseInt(block.getComputedStyle('outline-width'), 10);
+        var outlineOffset = parseInt(block.getComputedStyle('outline-offset'), 10),
+            outlineWidth = parseInt(block.getComputedStyle('outline-width'), 10);
         if ( isNaN(outlineOffset) ) {
+            // Edge does not support offset-offset yet
+            // 1 comes from the stylesheet, see theme/alloyeditor/content.css
             outlineOffset = 1;
         }
         return outlineOffset + outlineWidth;
--- a/Resources/public/js/alloyeditor/toolbars/config/block-base.js
+++ b/Resources/public/js/alloyeditor/toolbars/config/block-base.js
@@ -18,8 +18,9 @@ YUI.add('ez-alloyeditor-toolbar-config-block-base', function (Y) {
     function outlineTotalWidth(block) {
         var outlineOffset = parseInt(block.getComputedStyle('outline-offset'), 10);
         var outlineWidth = parseInt(block.getComputedStyle('outline-width'), 10);
-        if (isNaN(outlineOffset))
-          outlineOffset = 1;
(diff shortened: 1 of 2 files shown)


andrerom:
+1 but if accepet this will need to go into 1.4 branch.

Resources/public/js/alloyeditor/toolbars/config/block-base.js
```diff
@@ -16,10 +16,12 @@ YUI.add('ez-alloyeditor-toolbar-config-block-base', function (Y) {
...
 
     function outlineTotalWidth(block) {
-        return (
-            parseInt(block.getComputedStyle('outline-offset'), 10) +
-            parseInt(block.getComputedStyle('outline-width'), 10)
-        );
+        var outlineOffset = parseInt(block.getComputedStyle('outline-offset'), 10);
+        var outlineWidth = parseInt(block.getComputedStyle('outline-width'), 10);
+        if ( isNaN(outlineOffset) ) {
+            outlineOffset = 1;
```
dpobel: why 1 ?

Resources/public/js/alloyeditor/toolbars/config/block-base.js
```diff
@@ -16,10 +16,12 @@ YUI.add('ez-alloyeditor-toolbar-config-block-base', function (Y) {
     var ReactDOM = Y.eZ.AlloyEditor.ReactDOM;
 
     function outlineTotalWidth(block) {
-        return (
-            parseInt(block.getComputedStyle('outline-offset'), 10) +
-            parseInt(block.getComputedStyle('outline-width'), 10)
-        );
+        var outlineOffset = parseInt(block.getComputedStyle('outline-offset'), 10);
+        var outlineWidth = parseInt(block.getComputedStyle('outline-width'), 10);
```
dpobel: one var statement please

Resources/public/js/alloyeditor/toolbars/config/block-base.js
iosolinf: 1 because that was the value I got in chrome and firefox everytime I inspected the value

Resources/public/js/alloyeditor/toolbars/config/block-base.js
iosolinf: Sure. I'll fix this as soon as I get back to my computer (next friday)

Resources/public/js/alloyeditor/toolbars/config/block-base.js
dpobel: oh yeah because of https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/css/theme/alloyeditor/content.css#L9

this deserves some comments in both place to clarify a bit that. Here, something like:

```js
if ( isNaN(outlineOffset) ) {
    // Edge does not support offset-offset yet
    // 1 comes from the stylesheet, see theme/alloyeditor/content.css
    outlineOffset = 1;
}
```

and in the corresponding CSS, I would add something like:

```
.ez-richtext-editable .is-block-focused,
.ez-richtext-editable .cke_widget_wrapper.cke_widget_focused>.cke_widget_element {
    outline: 2px dashed #aaa;
    /* when changing outline-offset remember to also change
     * the outline width computation for Edge in
     * js/alloyeditor/toolbars/config/block-base.js
     */
    outline-offset: 1px;
}
```

dpobel:
+1 but you should rebase your branch against 1.4 and the open the PR against 1.4 as well

PR #650, opened 2016-07-19
Fix EZP-26101: cross-browser field validation regex error.

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-26101

Firefox >= 46 now forcefully uses the 'u' flag for input pattern regex, which means that `\-\d+` is no longer valid, but should instead be `-\d+` (or `/-\d+/u`).

However, relying on the unicode flag could lead to the reverse issue on different browsers (safari and older browsers not yet implementing the unicode flag).

Fix by not relying on the input pattern, but rather javascript regex test (as used f.e. in the email field validation).


joaoinacio:
replacing with PR against 1.4, closing this one.

PR #651, opened 2016-07-19
EZP-26101: cross-browser field validation regex error

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-26101

Firefox >= 46 now forcefully uses the 'u' flag for input pattern regex, which means that `\-\d+` is no longer valid, but should instead be `-\d+` (or `/-\d+/u`).

However, relying on the unicode flag could lead to the reverse issue on different browsers (safari and older browsers not yet implementing the unicode flag).

Fix by not relying on the input pattern, but rather javascript regex test (as used f.e. in the email field validation).

### DIFF
--- a/Resources/public/js/views/fields/ez-float-editview.js
+++ b/Resources/public/js/views/fields/ez-float-editview.js
@@ -13,7 +13,7 @@ YUI.add('ez-float-editview', function (Y) {
 
     var L = Y.Lang,
         FIELDTYPE_IDENTIFIER = 'ezfloat',
-        FLOAT_PATTERN = "\\-?\\d*\\.?\\d+"; // WARNING: each backslash is doubled, because it is escaped on output otherwise
+        FLOAT_PATTERN = "-?\\d*\\.?\\d+"; // WARNING: each backslash is doubled, because it is escaped on output otherwise
 
     /**
      * Float edit view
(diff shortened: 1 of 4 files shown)


Resources/public/js/views/fields/ez-float-editview.js
```diff
@@ -13,7 +13,7 @@ YUI.add('ez-float-editview', function (Y) {
 
     var L = Y.Lang,
         FIELDTYPE_IDENTIFIER = 'ezfloat',
-        FLOAT_PATTERN = "\\-?\\d*\\.?\\d+"; // WARNING: each backslash is doubled, because it is escaped on output otherwise
```
dpobel: if I get the issue right, Firefox now complains when if some characters are escaped while that's not necessary. As far as I can tell, here `-` does not need to be escaped so I guess using:
```js
FLOAT_PATTERN = "-?\\d*\\.?\\d+";
```
should fix the issue (and this would result in a much small patch)

joaoinacio: Yes. The purpose of the patch was to not only fix the firefox (mis)behavior, but to provide better cross-browser compatibility¹ and making sure validation does work.

\[1\]: http://caniuse.com/#feat=input-pattern

dpobel: as far as I can tell, that's not needed, we don't rely on the form not being submitted when an input is invalid and we don't use the default message as well so unless we have other(s) issue(s), fixing the pattern is the best way to fix the issue as it's the root of it.

andrerom:
@joaoinacio any news here? Your back tomorrow?

joaoinacio:
updated

andrerom:
+1

dpobel:
+1

joaoinacio:
(added test pattern update)

andrerom:
+1

PR #653, opened 2016-07-25
[Behat] Improve CommonActions and Authentication context Css selectors

miguelcleverti:
In order to stop fetching the elements using their text value the css selectors needed some changes.

This PR adds some specific tag to the PlatformUI markup in order to better fetch the elements. Also some unused methods were cleaned up, since their were not being used anymore.


miguelcleverti:
ping @bdunogier @dpobel @andrerom @joaoinacio

andrerom:
looks like  a good direction, as for the details here @bdunogier would have more insight on bdd part and @dpobel / @yannickroger on templates parts.

Resources/public/templates/navigation/item.hbt
```diff
@@ -1 +1 @@
-<a href="{{ path route.name route.params }}" class="ez-navigation-item">{{ title }}</a>
+<a href="{{ path route.name route.params }}" class="ez-navigation-item" data-navigation="{{ title }}">{{ title }}</a>
```
dpobel: what is the point of that ? ok you are not searching by text content, but you are still searching by text which can be translated or changed for whatever reason. Actually, each navigation item has a unique identifier so we should probably expose it somewhere so that you can use it in the tests.

Resources/public/templates/partials/tree.hbt
```diff
@@ -1,6 +1,6 @@
 <ul class="ez-tree-level">
 {{#each children}}
-    <li class="ez-tree-node{{#if state.leaf}} is-tree-node-leaf{{/if}} {{#if state.open}}is-tree-node-open{{else}}is-tree-node-close{{/if}}" data-node-id="{{id}}">
+    <li class="ez-tree-node{{#if state.leaf}} is-tree-node-leaf{{/if}} {{#if state.open}}is-tree-node-open{{else}}is-tree-node-close{{/if}}" data-node-id="{{id}}" title="{{ data.contentInfo.name }}">
```
dpobel: kind of the same as https://github.com/ezsystems/PlatformUIBundle/pull/653/files#r72211802, but there's already a `data-node-id` attribute which contains the Location id, I think, you should use it instead of search by Content name.

Resources/public/templates/usermenuitemfireevent.hbt
```diff
@@ -1 +1 @@
-<a href="#" class="ez-user-menu-item">{{ title }}</a>
+<a href="#" class="ez-user-menu-item" menu-title="{{ title }}">{{ title }}</a>
```
dpobel: `menu-title` is not a valid attribute name. other than that, that's exactly the same case as in https://github.com/ezsystems/PlatformUIBundle/pull/653/files#r72211802, we should expose in the markup the event name so that you can identify the link.

Resources/public/templates/navigation/item.hbt
dpobel: Created https://jira.ez.no/browse/EZP-26120 for that.

Resources/public/templates/usermenuitemfireevent.hbt
dpobel: And https://jira.ez.no/browse/EZP-26121 for that case

Resources/public/templates/partials/tree.hbt
joaoinacio: IMO this is a good example of searching by actual text being useful / correct.

In order to use the location ids, AFAICT, there are only 2 options: 
1: Use ids instead of names in sentences (which doesn't make much sense), or 
2: Use the API to retrieve ids from contents when performing tests (`When` / `Then` steps).

The problem with the second approach is that using the API introduces additional complexity as it's not solely using/testing the application interface, and could potentially add points of failure (caching issues between API and UI, for example).

Resources/public/templates/navigation/item.hbt
miguelcleverti: @dpobel okay, I have already done this, should I include it in this PR or in a separate one?

miguelcleverti: I made it in a separate PR so it's easier to review #654

Resources/public/templates/partials/tree.hbt
miguelcleverti: ping @bdunogier @dpobel 
Like @joaoinacio said this is a case where using text simplify things. What are your opinions on this?

bdunogier: Hmmm. Interesting.

I mostly agree with @joaoinacio's analysis. But can't we keep the name as a scenario side piece of information ? If whenever we create a location for a scenario, we need to reference its name, we can map the name to the id.

Except for built-in locations that we would use, we can always do it, right ?

It would avoid having to add the name as an attribute.

miguelcleverti: @bdunogier the problem is with contents created in PlatformUI, in this case there is no easy way to get the location id.

bdunogier: Hmm, after content has been created, you are redirected to this content, are you not ? In that case, you can read the CSS class that contains the location id for it, can't you ?

miguelcleverti: The location id is in the 'Details' tab in a element that only has a generic class, so selecting this element using css selectors only in complicated too. For this we must expose something in the 'Details' tab for the location id.

bdunogier: I completely agree, let's add it :-)

miguelcleverti: Or maybe only add a new class `location-id-value`... What do you think @dpobel?

miguelcleverti:
ping @dpobel, @bdunogier. After trying some different way I found that in order to use the `node-id` in content tree elements would need a lot of code scattered by multiple contexts. I made a some tests and moving the `title` attribute to the parent element does not seem to break anything and this simplify the code greatly. What do you think?

bdunogier:
Could you link to the change you're referring to @miguelcleverti ?

Resources/public/templates/partials/tree.hbt
```diff
@@ -1,7 +1,7 @@
 <ul class="ez-tree-level">
 {{#each children}}
-    <li class="ez-tree-node{{#if state.leaf}} is-tree-node-leaf{{/if}} {{#if state.open}}is-tree-node-open{{else}}is-tree-node-close{{/if}}" data-node-id="{{id}}">
```
miguelcleverti: @bdunogier, @dpobel these changes

Features/Context/SubContext/Authentication.php
```diff
@@ -35,7 +35,8 @@ public function goToPlatformUiAndLogIn($username, $password)
         $this->waitWhileLoading();
         $this->fillFieldWithValue('username', $username);
         $this->fillFieldWithValue('password', $password);
-        $this->iClickAtButton('Login');
+        $loginButton = $this->findWithWait('.ez-loginform-button');
```
bdunogier: Nitpick, that also applies below: `$this->findWithWait('.ez-loginform-button')->click()`

bdunogier:
Looks okay to me. @dpobel ?

Resources/public/templates/partials/tree.hbt
dpobel: what's the point of moving the title from the link to the list item ? I mean you can find the link and just retrieve the parent node to get the list item. I'm saying that because with this change, you'll end up with something like:

```html
<li title="A nice Content in the tree">
    <a href="#" <!-- ... -->>A nice Content in the tree</li>
    <ul class="ez-tree-level">
        <li title="A subitm of a nice Content in the tree">
        <!-- ... -->
        </li>
    </ul>
</li>
```

then the title is also associated with the subitems (the `ul`) of a given Location which is a bit weird semantically speaking. Also the `li` represents a Location in the tree while the `a` represents the Content item so the name (`title`) should stay at the Content item level since a Location has no name.

miguelcleverti:
ping @dpobel @bdunogier @joaoinacio. Removed the previous changes to the `title` attribute in the content tree elements. Now the selecting of the elements is done using the `data-node-id`  and the node id is fetched using the search service.

andrerom:
Seems you might want to rebase the PR or something, as there are some behat failures here.

miguelcleverti:
All the Behat test failures are due to Deprecation warning. It is ready to be reviewed.

andrerom:
Master [is not failing on behat](https://travis-ci.org/ezsystems/PlatformUIBundle/builds/171030032), could you rebase on top of latests? and if you introudce deprecation issues please fix them ;)

miguelcleverti:
PR approved by QA

PR #654, opened 2016-07-26
EZP-26120: Expose the navigation items identifier in the navigation hub

miguelcleverti:
As discussed on PR #653 the navigation items identifier should be exposed in order to the improve element searching in behat tests.

jira: https://jira.ez.no/browse/EZP-26120

### DIFF
--- a/Resources/public/js/views/navigation/ez-navigationitemview.js
+++ b/Resources/public/js/views/navigation/ez-navigationitemview.js
@@ -24,7 +24,9 @@ YUI.add('ez-navigationitemview', function (Y) {
      */
     Y.eZ.NavigationItemView = Y.Base.create('navigationItemView', Y.eZ.TemplateBasedView, [], {
         initializer: function () {
-            this.containerTemplate = '<li class="' + this._generateViewClassName(this._getName()) + '"/>';
+            var dataNavigationIdentifier = 'data-navigation-item-identifier="' + this.get('identifier') + '"';
+
+            this.containerTemplate = '<li class="' + this._generateViewClassName(this._getName()) + '" ' + dataNavigationIdentifier + '/>';
             this.after('selectedChange', this._uiSelectedChange);
             this.after('routeChange', function (){
                 this.render();
--- a/Tests/js/views/navigation/assets/ez-navigationitemview-tests.js
+++ b/Tests/js/views/navigation/assets/ez-navigationitemview-tests.js
@@ -13,6 +13,7 @@ YUI.add('ez-navigationitemview-tests', function (Y) {
 
         setUp: function () {
             this.title = "Title";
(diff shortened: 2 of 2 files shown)


dpobel:
Can you please include the issue number in the commit and in the PR title and in the pull request with a link to the issue ?

Also this is not the correct approach, as mentioned in the issue description, we should add the identifier in a data attribute on the view container. This would avoid adding such logic in the template and would make this behavior more robust when extending this class (it is meant for that).

(In addition, your patch is probably breaking a unit tests and you'll have to add some unit tests as well).

miguelcleverti:
ping @dpobel if I understand correctly this should do it, if not could you provide an example.

Resources/public/js/views/navigation/ez-navigationitemview.js
```diff
@@ -51,6 +51,7 @@ YUI.add('ez-navigationitemview', function (Y) {
         render: function () {
             var container = this.get('container');
 
+            container.setAttribute('data-navigation-item-identifier', this.get('identifier'));
```
dpobel: I would move that to the initializer where the `containerTemplate` is defined. Also this needs to be unit tested.

miguelcleverti:
@dpobel unit test added

Tests/js/views/navigation/assets/ez-navigationitemview-tests.js
```diff
@@ -44,6 +46,19 @@ YUI.add('ez-navigationitemview-tests', function (Y) {
             );
         },
 
+        "The Container should have a data identifier": function () {
```
dpobel: `"Should add the navigation identifier on the container"`

Resources/public/js/views/navigation/ez-navigationitemview.js
```diff
@@ -24,7 +24,8 @@ YUI.add('ez-navigationitemview', function (Y) {
      */
     Y.eZ.NavigationItemView = Y.Base.create('navigationItemView', Y.eZ.TemplateBasedView, [], {
         initializer: function () {
-            this.containerTemplate = '<li class="' + this._generateViewClassName(this._getName()) + '"/>';
+            var dataNavigationIdentifier = 'data-navigation-item-identifier="' + this.get('identifier') + '"';
+            this.containerTemplate = '<li class="' + this._generateViewClassName(this._getName()) + '" ' + dataNavigationIdentifier + '" />';
```
dpobel: nitpick: add an empty line between the variable declarations and actual code

dpobel:
besides the 2 nitpicks +1

StephaneDiot:
+1

miguelcleverti:
ping @yannickroger @glye all green

andrerom:
+1

PR #656, opened 2016-07-27
EZP-26124: Results of search in UDW overflow

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-26124

**Screen**
search-results-overflow

**Description**
Added overflow auto on results container.

**Tests**
manual tests

### DIFF
--- a/Resources/public/css/views/universaldiscovery/search.css
+++ b/Resources/public/css/views/universaldiscovery/search.css
@@ -4,6 +4,7 @@
  */
 
 .ez-view-universaldiscoverysearchview {
+    max-height: 100%;
     display: -webkit-box;
     display: -webkit-flex;
     display: -ms-flexbox;
@@ -127,4 +128,13 @@
 
 .ez-view-universaldiscoverysearchview .ez-searchresult-navigation-link[rel="prev"] {
     padding-right: 1em;
+}
+
+.ez-view-universaldiscoverysearchview .ez-searchresult-content {
+    overflow: auto;
+    height: calc(100% - 6em);
+}
+
+.ez-view-universaldiscoverysearchview .ez-ud-pane-search-result {
+    max-height: calc(100% - 9em)
 }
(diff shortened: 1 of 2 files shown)


dew326:
ping @dpobel @yannickroger

Resources/public/css/views/universaldiscovery/search.css
```diff
@@ -128,3 +129,12 @@
 .ez-view-universaldiscoverysearchview .ez-searchresult-navigation-link[rel="prev"] {
     padding-right: 1em;
 }
+
+.ez-view-universaldiscoverysearchview .ez-searchresult-content {
+    overflow: auto;
+    height: calc(100% - 6em);
```
dpobel: is there a way to avoid such *magic* value ? I understand `6em` comes from the height of `.ez-ud-pane-search-form` (which should not be set btw) but `9em` below, I don't get how you get this.

dpobel: actually, after digging a bit into the code, maybe there's a way to avoid those magic values and make the whole thing more robust against changes. I had similar issue with the Tree in the browse tab, you can do pretty much the same as https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/css/views/universaldiscovery/browse.css#L48

so basically here given the `ez-ud-pane-search-result-list` element (the parent of `ez-searchresult-content`) has `display: flex`, you can write something like:

```css
.ez-view-universaldiscoverysearchview .ez-searchresult-content {
    height: 100px;
    overflow: auto;
    flex: 1 0 auto;
}
```

as written in the linked comment, the `height` is just there to set a (initial) height on the element so that overflow has an effect, but the actual size will be computed based on the `flex` rule.
It seems to work for me both in Chrome and Firefox and has the benefit of not relying on any other pre-defined size elsewhere.

dpobel: (and with that, the others changes seem to not be necessary)

dew326:
ping @dpobel - your solution works great

StephaneDiot:
+1

dpobel:
self +1 then :)

dpobel:
@dew326 don't forget to apply the same fix to master

PR #658, opened 2016-07-29
EZP-25465: Implement a Draft conflict screen

yannickroger:
Link: https://jira.ez.no/browse/EZP-25465

## Description
On the location view, when clicking on the  edit button of a content which has drafts, a box is displayed to let you select an existing draft or create a new one.

draft_conflict

## Test
Manual and unit tests

## TODO
- [x] Feature
- [x] Test coverage
- [x] Create a follow up bug for confirbox close transition: https://jira.ez.no/browse/EZP-26192
- [x] Create a follow up for the dashboard edit buttons: https://jira.ez.no/browse/EZP-26191


Resources/public/js/views/confirmbox/ez-draftconflictview.js
```diff
@@ -0,0 +1,86 @@
...
+            return drafts;
+        },
+
+        /**
+         *  Processes the link provided by the event.
+         *
+         * @protected
+         * @method _processLink
+         */
+        _processLink: function (e) {
```
dpobel: you can tidy up `process` in the same category as `handle` ;-)
maybe you can call that method `_confirmDraftEdition` ?

Resources/public/js/views/ez-confirmboxview.js
```diff
@@ -178,6 +224,29 @@ YUI.add('ez-confirmboxview', function (Y) {
...
+            },
+
+            /**
+             * Wether on not to display the `Confirm` and `Cancel` buttons
+             *
+             * @attribute button
+             * @default true
+             * @type {Boolean}
+             */
+            button: {
```
dpobel: that's not super accurate, `renderDefaultButtons` maybe ?

Resources/public/js/views/ez-confirmboxview.js
```diff
@@ -80,6 +85,42 @@ YUI.add('ez-confirmboxview', function (Y) {
...
         /**
+         * Hides of reveal the buttons
+         *
+         * @method _uiButtonDisplay
+         * @protected
+         */
+        _uiUpdateButton: function () {
+            var container = this.get('container').one('.ez-confirmbox-tools');
+
+            if (this.get('button') === true) {
```
dpobel: no need to strictly compare to `true`

Resources/public/js/views/ez-confirmboxview.js
```diff
@@ -80,6 +85,42 @@ YUI.add('ez-confirmboxview', function (Y) {
...
+                container.show();
```
dpobel: to make that easier to test, if would actually add or remove a class on the container to show/hide the corresponding toolbar

Resources/public/js/views/services/plugins/ez-editcontentplugin.js
```diff
@@ -0,0 +1,113 @@
...
+         * editAction event handler, makes the application navigate to edit the
+         * content available in the event facade
+         *
+         * @method _editContent
+         * @protected
+         * @param {Object} e event facade of the editAction event
+         * //TODO need content
+         * //TODO need languageCode
+         */
+        _editContent: function (e) {
```
dpobel: would better name something like `_checkExistingDraft` to reflect what it actually does

Resources/public/js/views/services/plugins/ez-editcontentplugin.js
```diff
@@ -0,0 +1,113 @@
...
+            var service = this.get('host'),
+                app = service.get('app'),
+                content = e.content,
+                languageCode = e.languageCode,
+                callback = function(error, versions) {
+                    app.set('loading', false);
+                    if (error) {
+                        return;
```
dpobel: I think it's safe to navigate to the edit form without version id in that case

Resources/public/js/views/confirmbox/ez-draftconflictview.js
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * The Draft conflict confirm box view.
+     *
+     * @namespace eZ
+     * @class DraftConflict
+     * @constructor
+     * @extends eZ.TemplateBasedView
+     */
+    Y.eZ.DraftConflict = Y.Base.create('draftConflictView', Y.eZ.TemplateBasedView, [], {
```
dpobel: `Y.eZ.DraftConflictView`

Resources/public/js/views/confirmbox/ez-draftconflictview.js
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+YUI.add('ez-draftconflictview', function (Y) {
+    "use strict";
+
+    /**
+     * The Draft conflict confirm box view.
```
dpobel: you should omit *confirm box*, it can actually be used in others contexts

Resources/public/js/views/ez-confirmboxview.js
```diff
@@ -178,6 +224,29 @@ YUI.add('ez-confirmboxview', function (Y) {
...
             /**
+             * View of the confirmBox.
+             * A Y.View that can be provided to be displayed in the confirm box
+             *
+             * @attribute view
+             * @default {}
+             * @type {Object}
+             */
+            view: {
+                value: {},
```
dpobel: `null` would be a better default value

Resources/public/js/views/ez-confirmboxview.js
```diff
@@ -178,6 +224,29 @@ YUI.add('ez-confirmboxview', function (Y) {
             },
 
             /**
+             * View of the confirmBox.
+             * A Y.View that can be provided to be displayed in the confirm box
+             *
+             * @attribute view
+             * @default {}
+             * @type {Object}
```
dpobel: actually it stores a Y.View (or null given https://github.com/ezsystems/PlatformUIBundle/pull/658/files#r72792835)

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -60,20 +60,18 @@ YUI.add('ez-locationviewviewservice', function (Y) {
         /**
          * editAction event handler, makes the application navigate to edit the
          * content available in the event facade
-         *
+         * TODO
          * @method _editContent
          * @protected
          * @param {Object} e event facade of the editAction event
+         *
```
dpobel: TODO remove :p

Resources/public/js/views/services/plugins/ez-editcontentplugin.js
```diff
@@ -0,0 +1,113 @@
...
+                };
+
+            app.set('loading', true);
+
+            /**
+             * Fired when versions are about to be loaded
+             *
+             * @event loadVersions
+             */
+            service.fire('loadVersions',{
```
dpobel: while this is working, that's a bit weird. Instead, you should directly call `content.loadVersions` here

dpobel: ah actually, I realize the plugin also group the versions by status so you can not directly use `loadVersions` unless you move the grouping in the Content model (under an option passed in the first parameters of `loadVersions`)

Resources/public/templates/confirmbox.hbt
```diff
@@ -1,7 +1,10 @@
 <a href="#" class="ez-confirmbox-close-icon ez-confirmbox-close" data-icon-after="&#xe62a;"></a><h1 class="ez-confirmbox-title">{{ title }}</h1>
 
 <div class="ez-confirmbox-details">{{&details }}</div>
+
 <p class="ez-confirmbox-tools">
     <a href="#" class="ez-confirmbox-cancel ez-confirmbox-close">Cancel</a>
     <button class="ez-confirmbox-confirm ez-button pure-button">Confirm</button>
 </p>
+
+<div class="ez-confirmbox-view"></div>
```
dpobel: I would move that before the toolbar

Resources/public/js/views/services/plugins/ez-editcontentplugin.js
yannickroger: I thought about that. If transfering the sorting in the content model I was thinking of a dedicated method because the result have a different structure (array of version and array of object).
The new method `loadVersionsByStatus` could first call `loadVersions` and then do the sorting.

Resources/public/js/views/services/plugins/ez-contenteditplugin.js
```diff
@@ -0,0 +1,99 @@
...
+                            renderDefaultButtons: false,
+                            confirmHandler: function (e) {
+                                app.navigate(e.route);
+
+                                /**
+                                 * Fired after the redirect to edit has been
+                                 * done to close the confirm box.
+                                 * @event confirmBoxClose
+                                 */
+                                service.fire('confirmBoxClose');
```
dpobel: that should not be necessary. In previous, usage of the confirm box, we don't do that so if it happens to be necessary, something is wrong somewhere.

Resources/public/js/views/confirmbox/ez-draftconflictview.js
```diff
@@ -0,0 +1,177 @@
...
+    "use strict";
+
+    Y.namespace('eZ');
+
+    var CLASS_ROW_SELECTED = 'is-row-selected',
+        SELECTOR_ROW = '.ez-draft-conflict-list-row',
+        SELECTOR_CONTENT = '.ez-draft-conflict-list-content',
+        EVENTS = {};
+
+    EVENTS[SELECTOR_ROW] = {tap: '_uiShowRowOptions'};
```
dpobel: you are duplicating some code from DashboardBlockAsynchronousView, it would be nice to extract the line selection logic in a dedicated extension applied to DashboardBlockAsynchronousView and DraftConflictView

Resources/public/js/views/services/plugins/ez-contenteditplugin.js
```diff
@@ -0,0 +1,99 @@
...
+                            languageCode: languageCode
+                        })
+                    );
+                }
+            });
+        },
+    }, {
+        NS: 'contentEdit',
+
+        ATTRS: {
```
dpobel: should be removed

Resources/public/js/views/services/plugins/ez-versionsplugin.js
```diff
@@ -60,36 +60,14 @@ YUI.add('ez-versionsplugin', function (Y) {
...
-        /**
-         * Sorts an array of version info by status
-         *
-         * @method _sortVersions
-         * @protected
-         * @param versions {Array} of eZ.VersionInfo
-         * @return {Object} of sorted versions by status:
-         *              struct.<status_name>: {Array} of eZ.VersionInfo
-         */
-        _sortVersions: function(versions) {
```
dpobel: I think this protected method was introduced in 1.4, so you can not removed it, you should instead add a deprecation warning and the `@deprecated` tag so that it's removed in 2.0

Resources/public/templates/confirmbox/draftconflict.hbt
```diff
@@ -0,0 +1,54 @@
...
+        ({{ translate_property contentType.names }})
+    </div>
+
+    <div class="ez-draft-conflict-header-button">
+        <a class="ez-draft-conflict-link pure-button ez-button"
+            href="{{ path 'editContent'
+                    id=content.id
+                    languageCode=languageCode
+                }}">
+            + New Draft
```
dpobel: the `+` should not be in the anchor text but added with some CSS

Resources/public/templates/confirmbox/draftconflict.hbt
```diff
@@ -0,0 +1,54 @@
+<div class="ez-draft-conflict-header">
+    <div class="ez-draft-conflict-header-subtitle">
```
dpobel: usually when you start adding a class with `header` it's a good sign that you should use the `header` tag and/or a `hX`

Resources/public/templates/confirmbox/draftconflict.hbt
```diff
@@ -0,0 +1,54 @@
...
+        {{ lookup content.currentVersion.names languageCode }}
+        ({{ translate_property contentType.names }})
+    </div>
+
+    <div class="ez-draft-conflict-header-button">
```
dpobel: it's more a toolbar than a header

Resources/public/templates/confirmbox/draftconflict.hbt
```diff
@@ -0,0 +1,54 @@
...
+            </td>
+            <td class="ez-draft-conflict-list-cell">
+                {{languageCodes}}
+            </td>
+            <td class="ez-draft-conflict-list-cell">
+                {{resources.Creator}}
+            </td>
+            <td class="ez-draft-conflict-list-cell ez-draft-conflict-list-cell-option">
+                {{modificationDate}}
+                <div class="ez-draft-conflict-list-row-options">
```
dpobel: in the initial specification, the button is supposed to be inside the table (this works better here since we only have one table) and that's actually easier to do, you just have to add a new column in the table and the `.ez-draft-conflict-link` elements are hidden (`visibility: hidden`) until the row is selected.

yannickroger:
Ready for testing

Resources/public/js/views/ez-confirmboxview.js
```diff
@@ -80,6 +83,45 @@ YUI.add('ez-confirmboxview', function (Y) {
...
+
+        /**
+         * Updates the view in the rendered confirmBox
+         *
+         * @method _uiUpdateView
+         * @protected
+         */
+        _uiUpdateView: function () {
+            var view = this.get('view');
+            if (view instanceof Y.View) {
```
StephaneDiot: You should add an empty line after variable declaration (and maybe also delete the empty line after the 'if')

Resources/public/js/views/ez-confirmboxview.js
```diff
@@ -178,6 +228,29 @@ YUI.add('ez-confirmboxview', function (Y) {
...
+             * @attribute view
+             * @default null
+             * @type {Y.View|null}
+             */
+            view: {
+                value: null,
+            },
+
+            /**
+             * Wether on not to display the `Confirm` and `Cancel` buttons
```
StephaneDiot: typo

StephaneDiot:
+1

dpobel:
+1

PR #439, opened 2015-11-19
Fix EZP-25072: Deletion of "Role Assignment" redirects to wrong location

glye:
> Fixes https://jira.ez.no/browse/EZP-25072
> Status: Ready for review
> Depends on https://github.com/ezsystems/PlatformUIBundle/pull/438

This uses the tab loading feature I added in https://github.com/ezsystems/PlatformUIBundle/pull/438, so it can only be merged after that.
Behat test fail is irrelevant.

PR #660, opened 2016-08-05
EZP-25825: Deleting a user role assignment switches to the role tab

iosolinf:
Stick to Assignments Tab after assignment removal


yannickroger:
As stated in https://github.com/ezsystems/PlatformUIBundle/pull/439 this will be dealt with in a later Epic

PR #662, opened 2016-08-09
[Travis]  Run using Trusty image & docker for Behat

andrerom:
Moves to Travis Trusty images, as well as move behat testing itself to use docker images. Both changes provide some more speed, and is far more stable then what is the current case.

Tests to get running:
- [x] Unit tests
- [x] Behat running more stable *(besides faster images, fixes some blocking behat issues)*
- [x] JS tests. *fails with `/bin/sh: 1: grover: not found`*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,33 +1,33 @@
-sudo: false
+# Use trusty for better performance (and avoiding mysql/postgres/solr gone issues on precise and container infra)
+dist: trusty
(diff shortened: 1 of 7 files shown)


andrerom:
Behat running, input on how to get node-js tests running on trusty env wanted, see description above. //cc @dpobel @yannickroger @StephaneDiot

PR #663, opened 2016-08-16
EZP-28164: Set max size of OE image previews to avoid breaking right menu

libertjeremy:
> Issue: https://jira.ez.no/browse/EZP-28164

This is a PR to fix a bug when inserting a large image in the content editor.

Before :

ez-bug-img

After : 

ez-bug-img-fixed

Jérémy.

### DIFF
--- a/Resources/public/css/modules/embed.css
+++ b/Resources/public/css/modules/embed.css
@@ -14,6 +14,7 @@
 
 [data-ezelement="ezembed"] .ez-embed-content {
     margin: 0;
+    max-width: 100%;
 }
 
 [data-ezelement="ezembed"] .ez-embed-content:before {


andrerom:
+1

But I think @dpobel had in mind we should have some sort of visual clue when we resize the image, so Editor knows it is bigger.

oxeron:
Could we add a small overlay icon on the bottom left corner indicating that the image was resized in the editor only ? And when the user clicks on it, it displays with the original size (breaking the editor window but showing the real size when the content will be published) ?

And when clicking again on the image, it is resized again.

andrerom:
@oxeron as the resizing is done in css, so atm that is not easily possible.

It is actually a good thing that it is handled by css, *if* we can also find some way to target the elements being resized to apply some icon, then the whole logic you are suggesting can be done in pure css. Otherwise it will have to be done in javascript.

oxeron:
What do you think @libertjeremy ?

andrerom:
@dpobel / @bdunogier2 / @bdunogier  Should we perhaps accept this as is for now as it at least makes sure UI isn't broken? And it's not like the editor aims to be wysiwyg, so in that sense this is ok even if it should provide hint to editor at some point.

bdunogier:
Looks good to me, with my limited frontend knowledge.

@dpobel ?

P.S. I got my regular account back, @andrerom :)

oxeron:
Thanks Bertrand ! Long time no see ;)

dpobel:
@libertjeremy @oxeron  sorry I completely forgot this pull request :(

https://github.com/ezsystems/PlatformUIBundle/pull/663#issuecomment-240134452 is a good idea but this needs quite some work, if you really want that please create the corresponding improvement in https://jira.ez.no/browse/EZP (under PlatformUI component). Maybe an alternative would be to keep the size of the image in the editor but make sure this does not cover the right toolbar and maybe have an horizontal scrollbar in the editor area to see the image (but the rest of the editor content should not be scrollable).

@libertjeremy If you don't have to time to work on that these days, can you just create a bug in https://jira.ez.no/browse/EZP (under PlatformUI component), we'll try to fix that in the coming weeks (we are about to start a stabilization phase for the next release)

andrerom:
Looking into this again given we would all benefit from this.

Here is what we can do:
1. The css proposed here is just fine as first stage
2. On top of this we can add javascript to iterate over images and compare [naturalwidth](http://caniuse.com/#feat=img-naturalwidth-naturalheight) with the width of the element and add a second css class to it if it is detected to normally be larger. This can have some additional styling to it. _(this script will need to be re ran on screen resize)_

@dpobel Could we consider accepting this (1.) with an issue, and propose the followup (2.) as story to PM?

andrerom:
Thanks @libertjeremy, took a long time to get feedback on this, but we opted to include it so thank you. If you feel like improving this don't hesitate to suggest something, either like proposed above or otherwise :)

PR #664, opened 2016-08-25
EZP-26145: Unable to undo in the RichText editor

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26145

# Description

In Online Editor, some operations can not be undone, especially the removal of some text in the middle of a paragraph. This is happening because while doing that, we are also cleaning up some auto-generated ids and this seems to confuse the *undo manager*. To avoid that, before cleaning up the ids, the undo manager is locked so that the cleanup is not part of the undo stack.

# Tests

manual tests + unit tests

### DIFF
--- a/Resources/public/js/alloyeditor/plugins/yui3.js
+++ b/Resources/public/js/alloyeditor/plugins/yui3.js
@@ -11,12 +11,14 @@ YUI.add('ez-alloyeditor-plugin-yui3', function (Y) {
     }
 
     function cleanUpIds(editor) {
+        editor.undoManager.lock();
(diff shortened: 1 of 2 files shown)


lserwatka:
@dpobel Would it be possible to backport it to 1.3 series? Project that requires this fix does not run on 1.4 series yet.

dpobel:
@lserwatka of course, but it would be easier if the affect version in the JIRA issue was right in the first place.

andrerom:
+1

yannickroger:
+1

StephaneDiot:
+1

PR #667, opened 2016-08-29
Fix EZP-26117: Required richtext not validated as filled with single embed

yannickroger:
Link: https://jira.ez.no/browse/EZP-26117

## Description
When editing a content with a mandatory richtext field (such as article). If only an single embed is present, the field is still considered empty.

## Test
Manual and unit tests

### DIFF
--- a/Resources/public/js/views/fields/ez-richtext-editview.js
+++ b/Resources/public/js/views/fields/ez-richtext-editview.js
@@ -241,9 +241,12 @@ YUI.add('ez-richtext-editview', function (Y) {
                     return element.get('children').some(function (node) {
                         return L.trim(node.get('text')) !== '';
                     });
+                },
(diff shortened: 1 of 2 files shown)


andrerom:
+1

Resources/public/js/views/fields/ez-richtext-editview.js
```diff
@@ -241,9 +241,12 @@ YUI.add('ez-richtext-editview', function (Y) {
                     return element.get('children').some(function (node) {
                         return L.trim(node.get('text')) !== '';
                     });
+                },
+                hasChildWithEmbed = function(element) {
```
dpobel: I would call it `hasEmbedElement`

dpobel:
beside the nitpick on the name, +1

PR #668, opened 2016-08-30
EZP-26179: UI Stability issues due to session loss

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26179

# Description

This patch improves PlatformUI to better check the user session and react to an expired session.
The app change makes sure we always check the user session on "page loading", this adds a request on each page but it's the only way to have the status of the session.
In addition to that, a check is added to the error notification to verify if the error happens because of an expired session. This is a cheap way of doing that will only be in 1.4 and 1.5. In master (1.6), this will be made a bit differently with a dedicated error API to decouple the UI notification from the actual error handling.

## Tasks

* [x] Make sure to always check the session server side in the `checkUser` middleware to redirect the user to the login if the session has expired
* [x] Add a check against the session after a failing request
* [x] Make sure the login form allows to login after the session has expired (CSRF Token issue...), fixed in https://github.com/ezsystems/ezpublish-kernel/pull/1769

### DIFF
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -458,8 +458,8 @@ YUI.add('ez-platformuiapp', function (Y) {
                 anonymousUserId = this.get('anonymousUserId'),
                 userId = this.get('user').get('id');
 
-            if ( userId ) {
-                callback(userId === anonymousUserId);
+            if ( userId === anonymousUserId ) {
+                callback(true, false);
                 return;
             }
             capi.isLoggedIn(function (error, response) {
--- a/Resources/public/js/apps/plugins/ez-notificationhubplugin.js
+++ b/Resources/public/js/apps/plugins/ez-notificationhubplugin.js
@@ -29,9 +29,65 @@ YUI.add('ez-notificationhubplugin', function (Y) {
                 app.showSideView('notificationHub');
             });
 
-            app.on('*:notify', function (e) {
-                app.showSideView('notificationHub', {notification: e.notification});
+            app.on('*:notify', Y.bind(function (e) {
+                var notification = e.notification;
+
+                if ( notification.state === 'error' ) {
+                    this._checkIsLoggedIn(
+                        this._logOut,
+                        Y.bind(this._showNotification, this, e.notification)
+                    );
+                } else {
+                    this._showNotification(e.notification);
+                }
+            }, this));
+        },
+
+        /**
+         * Checks if the user is logged in and call the corresponding callback.
+         *
+         * @private
+         * @method _checkIsLoggedIn
+         * @param {Function} notLoggedInCallback
+         * @param {Function} loggedInCallback
+         */
+        _checkIsLoggedIn: function (notLoggedInCallback, loggedInCallback) {
+            var app = this.get('host');
+
+            app.set('loading', true);
+            app.isLoggedIn(Y.bind(function (notLoggedIn) {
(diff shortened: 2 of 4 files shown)


dpobel:
for the last point, I'm blocked by the behavior of the REST API. Let me explain a bit what's happening.

Reminder: from JavaScript, I can not manipulate the session cookie so if the browser has that cookie, it is send when doing a request without any option not do so.

So with the current patch, if a request fails, we do a refresh session call (POST to `/user/sessions/<sessionId>/refresh`) to check if the session is still alive. This call is specified in: https://github.com/ezsystems/ezpublish-kernel/blob/master/doc/specifications/rest/REST-API-V2.rst#refresh-session-get-session-s-user-information
If the session does not exist, it is supposed to answer a 404 but that's impossible because if the session does not exist, the CSRF token can not be valid, so it will answer with a 401. At the moment, 401 or 404, that does not make any difference, PlatformUI/the JS REST client considers, you are not logged in so afterward we try to delete the session to get rid of the session cookie with a DELETE request to `/user/sessions/<sessionId>` specified in https://github.com/ezsystems/ezpublish-kernel/blob/master/doc/specifications/rest/REST-API-V2.rst#delete-session-logout-a-user. Here, we have exactly the same problem, the session has expired, the CSRF is then invalid, so the response status will be a 401 (not the specified 404) and the browser still has the session cookie.

At this point, the user is redirected to the login form but he can not login again because the browser still has the session cookie so the API expects a valid CSRF that does not exist since the corresponding session does not exist.

Now the question, is how to fix that ? At some point, the browser should receive a header telling that the session cookie should be removed otherwise this can not work.
ping @bdunogier @andrerom

andrerom:
So at least `DELETE /user/sessions/<sessionId>` should be adjusted to delete a invalid session cookie if session is invalid, meaning CSRF is only taken into account if session actually exists. But  maybe I'm overlooking something.

dpobel:
@andrerom yes that's one option. This logic could also be moved or be "duplicated" in `POST /user/sessions/<sessionId>/refresh`. Or removing the session cookie could be done more globally by the application in any response because a session cookie referencing an expired session does not bring much (unless I'm overlooking something ?)

bdunogier:
I'll study the session part to see how we can improve that. Thank you for the thorough analysis, @dpobel.

andrerom:
@dpobel true, maybe the logic I talked about above should partly be the case for refresh as well so it is cable of saying 404. And yes, we need to check how we in the past, and other applications behave on expired invalid cookies, I would think either a session reset or re-creation of new empty session on with the existing session id might be the case.

bdunogier:
> At this point, the user is redirected to the login form but he can not login again because the browser still has the session cookie so the API expects a valid CSRF that does not exist since the corresponding session does not exist.

About this, it sounds a bit counter-productive to check the csrf token when logging in. By definition, if you're logging in, you do not have a session, and you're supposedly providing credentials to create a new session.

I'm not sure how yet, but we should disable csrf protection on the login route. Do we agree ?

andrerom:
> I'm not sure how yet, but we should disable csrf protection on the login route. Do we agree ?

yes


> Credentials aren't expected by this route, and if the session has expired, well, it can't validate anything. Do we also agree ?

yes, it sounds like we can safly ignore missing csrf tokens here and on delete of session. IF session is invalid in the first place, *and we are certain session id in url is same as user has*. Open question is if php/symfony/us in some cases will reset session and generate new session id when session has expired.

dpobel:

PR #669, opened 2016-08-30
Fix EZP-26130: Search for content with many locations

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-26130

## Description

The searchPlugin was not made to be able to search content with multiple location. This is now fixed

## Tests

- Manual & unit tests

### DIFF
--- a/Resources/public/js/views/services/plugins/ez-searchplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-searchplugin.js
@@ -374,7 +374,12 @@ YUI.add('ez-searchplugin', function (Y) {
             contentIds = Y.Array.reduce(locationStructArr, '', function (previousId, struct, index) {
                 var contentId = struct.location.get('contentInfo').get('contentId');
 
-                contentIdsLocationIndexMap[contentId] = index;
+                if (!contentIdsLocationIndexMap[contentId]) {
+                    contentIdsLocationIndexMap[contentId] = [index];
+                } else {
+                    contentIdsLocationIndexMap[contentId].push(index);
+                }
+
                 previousId = previousId ? previousId + ',' : previousId;
                 return previousId + contentId;
             });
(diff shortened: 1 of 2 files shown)


StephaneDiot:
ping @dpobel @yannickroger

Resources/public/js/views/services/plugins/ez-searchplugin.js
```diff
@@ -394,7 +399,9 @@ YUI.add('ez-searchplugin', function (Y) {
                     var content = this._createContent(hit),
                         locationIndex = contentIdsLocationIndexMap[content.get('contentId')];
```
sunpietro: What do you think about renaming it to something like `locationIndexes`? It may contain more than one index.

Resources/public/js/views/services/plugins/ez-searchplugin.js
```diff
@@ -394,7 +399,9 @@ YUI.add('ez-searchplugin', function (Y) {
...
 
-                    locationStructArr[locationIndex].content = content;
+                    Y.Array.each(locationIndex, function(index) {
```
sunpietro: `locationIndexes.forEach(function (locationIndex) {`

yannickroger:
+1 with @sunpietro 's feedback taken into account

Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js
```diff
@@ -926,8 +931,15 @@ YUI.add('ez-searchplugin-tests', function (Y) {
                     if (query === that.locationQuery) {
                         cb(false, that.locationResponse);
                     } else if (query === that.contentQuery) {
+                        var contentIds = [];
+
+                        Y.Array.each(
```
dpobel: it would be a bit more elegant to use use `map` in such case. Also there are small CS issues (`);` is badly placed, a space is missing after `function`)

dpobel:
+1 when the inline comments are fixed

PR #671, opened 2016-09-01
Fix EZP-26236: Events are lost if container is created in initializer

yannickroger:
Link: https://jira.ez.no/browse/EZP-26236

## Description
This is a very technical YUI related bug.

When the container is created in the initializer, events added later on (like the initializer of a view inheriting from the first one) are not attached to the view.

This patch makes sure to attach the events to the DOM if a container already exists.

## Tests
Unit tests and basic sanity.

### DIFF
--- a/Resources/public/js/views/ez-view.js
+++ b/Resources/public/js/views/ez-view.js
@@ -44,6 +44,25 @@ YUI.add('ez-view', function (Y) {
         },
 
         /**
+         * Adds DOM event handlers on the current view. It makes sure
+         * event subscriptions are really taken into account which might
+         * not be the case if the view container has already been initialized.
+         * see https://jira.ez.no/browse/EZP-26236
+         *
+         * @method _addDOMEventHandlers
+         * @protected
+         * @param {Object} events
+         */
+        _addDOMEventHandlers: function (events) {
+            this.events = Y.merge(this.events, events);
+
+            if ( this._container ) {
+                this.attachEvents();
+            }
+        },
+
+
+        /**
          * Sets the active attribute of the sub views stored in attributes to
          * the same value as the current view
          *
--- a/Tests/js/views/assets/ez-view-tests.js
+++ b/Tests/js/views/assets/ez-view-tests.js
@@ -3,7 +3,7 @@
  * For full copyright and license information view LICENSE file distributed with this source code.
  */
 YUI.add('ez-view-tests', function (Y) {
-    var viewTest, pluginTests, Assert = Y.Assert;
+    var viewTest, pluginTests, addEventsTest, Assert = Y.Assert;
 
     viewTest = new Y.Test.Case({
         name: "eZ View view tests",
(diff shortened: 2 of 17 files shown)


sunpietro:
@yannickroger when does such a situation happen with losing DOM events? I've never met that case.

yannickroger:
@sunpietro Let me explain how I ran into that when working on https://jira.ez.no/browse/EZP-26191.

I am adding events to https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/dashboard/ez-dashboardblockmycontentview.js#L27

I couldn't get those event to work. After investigating with @dpobel we found that this was happening because the container was created in the parent class's initializer(): https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/dashboard/ez-dashboardblockasynchronousview.js#L45

So when I was adding my events, they were not attached because this is donne while creating the container.

Usually the container is created in the render() method after all initializer() have been processed and that's why it's the first time we are running into this issue.

sunpietro:
@yannickroger thanks for the explanation. That's strange, because I thought the DOM events are actually attached when the view is rendered, not when getting the container from the view settings.
Good job!

Resources/public/js/views/ez-view.js
```diff
@@ -44,6 +44,22 @@ YUI.add('ez-view', function (Y) {
         },
 
         /**
+         * Add a list of events to the DOM event handler
```
dpobel: this is a bit inaccurate and deserves to be a bit more detailed. For instance,
```js
/**
 * Adds DOM event handlers on the current view. It makes sure
 * event subscriptions are really taken into account which might
 * not be the case if the view container has already been initialized.
 * see https://jira.ez.no/browse/EZP-26236
```

Tests/js/views/assets/ez-view-tests.js
```diff
@@ -36,7 +36,56 @@ YUI.add('ez-view-tests', function (Y) {
...
+            this.event2 = {'.whatever': {
+                'tap': '_something'
+            }};
+            this.view = new this.TestView({
+                events: this.event1,
+            });
+        },
+
+        _testAddEventsInDom: function () {
+            this.view._addDOMEventHandlers(this.event2);
```
dpobel: it's a protected method, you can not call it directly. Instead, you should define one or several view contructor(s) (like `this.TestView`) which use `_addDOMEventHandlers` in different cases (after the container has been initialized or in the case where the container is not initialized)

Tests/js/views/assets/ez-view-tests.js
```diff
@@ -36,7 +36,56 @@ YUI.add('ez-view-tests', function (Y) {
...
+            this._testAddEventsInDom();
+        },
+
+        "Should add events to the DOM event handler and attach events": function () {
+            this.view.set('container', '.container');
+            this._testAddEventsInDom();
+
+            Assert.areSame(
+                2,
+                this.view._attachedViewEvents.length,
```
dpobel: this is a protected protected, you can not use it directly. Instead your TestView should subscribe to a real event (like click on something) and you should check that the subscription works as expected.

yannickroger:

Tests/js/views/ez-view.html
```diff
@@ -24,7 +29,7 @@
         filter: loaderFilter,
         modules: {
             "ez-view": {
-                requires: ['view', 'base-pluginhost', 'ez-pluginregistry'],
+                requires: ['view', 'base-pluginhost', 'ez-pluginregistry', 'event-tap'],
```
dpobel: that's not the right place to add `event-tap`, the test depends on it not the implementation so it should be added at the end of ez-view-tests.js

Tests/js/views/assets/ez-view-tests.js
```diff
@@ -36,7 +36,92 @@ YUI.add('ez-view-tests', function (Y) {
...
+
+    addEventsTest = new Y.Test.Case({
+        name: "eZ View addEvent tests",
+
+        _testEventsInDom: function () {
+            var link1 = this.view.get('container').one('.event1'),
+                link2 = this.view.get('container').one('.event2');
+
+
+            link1.simulateGesture('tap', Y.bind(function () {
```
dpobel: choosing `tap` makes the whole thing a bit complicated since you are forced to have asynchronous test (wait / resume), it would be easier to use a regular DOM event like click. You could then write:

```js
link1.simulate('click');
Assert.isTrue(
    this.view.get('event1Fired'),
    "click on .event1 should have been caught"
);
```

Tests/js/views/assets/ez-view-tests.js
```diff
@@ -36,7 +36,92 @@ YUI.add('ez-view-tests', function (Y) {
...
+                this.resume(Y.bind(function (e) {
+                    Assert.isTrue(
+                        this.view.get('event1Fired'),
+                        "The event1 should have been fired"
```
dpobel: that's wrong, you test that the event was caught

Tests/js/views/assets/ez-view-tests.js
```diff
@@ -36,7 +36,92 @@ YUI.add('ez-view-tests', function (Y) {
...
+            return new this.TestView({
+                events: {'.event1': {
+                            'tap': '_event1'
+                        }},
+                container: '.container',
+            });
+        },
+
+        "Should add events to the DOM event handler": function () {
+            this.view = this._getTestView(Y.eZ.View);
```
dpobel: this.view is never destroyed, you should add a `tearDown` method for that

Tests/js/views/assets/ez-view-tests.js
```diff
@@ -36,7 +36,92 @@ YUI.add('ez-view-tests', function (Y) {
...
+            }, this));
+            this.wait();
+        },
+
+        _getTestView: function (parent) {
+
+            this.TestView = Y.Base.create('TestView', parent, [], {
+                initializer: function() {
+                    this._addDOMEventHandlers(
+                        {'.event2': {
```
dpobel: CS issue

Tests/js/views/assets/ez-view-tests.js
```diff
@@ -36,7 +36,92 @@ YUI.add('ez-view-tests', function (Y) {
...
+                        value: false
+                    },
+                    event2Fired: {
+                        value: false
+                    },
+                }
+            });
+
+            return new this.TestView({
+                events: {'.event1': {
```
dpobel: CS

Tests/js/views/assets/ez-view-tests.js
yannickroger: done

yannickroger:
The return of "Feedback taken into account @dpobel "

dpobel:
+1

yannickroger:
Merging with only one +1 as it is very technical.

PR #520, opened 2016-02-29
Update Behat login functions to match the new user profile

micszo:
After https://jira.ez.no/browse/EZP-25446 has been merged ( #513 ) the Behat functions iLogout and iShouldBeLoggedIn became outdated causing scenarios to fail at start.
This PR adds an extra click in the user profile area and changes the logout link identification.

PR #675, opened 2016-09-11
[Behat] Increase spin time to try to avoid failure on first step

andrerom:
On slow machine / when debug is enabled, and on empty cache.

Also cleanup iShouldBeLoggedIn to not do nested spin after #520 _(findWithWait uses spin underneath, having two nested does not increase the timeout)_.

### DIFF
--- a/Features/Context/PlatformUI.php
+++ b/Features/Context/PlatformUI.php
@@ -26,7 +26,7 @@ class PlatformUI extends Context
     /**
      * Max. time to use when trying to access html DOM elements (seconds).
      */
-    const SPIN_TIMEOUT = 5;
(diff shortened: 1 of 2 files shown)


andrerom:
review ping @bdunogier

miguelcleverti:
Increasing timeouts is a bit of a risk, because he might end up delaying some test cases(the one the require that the timeout occurs). I don't think I have encountered this issue so far and are not the test running on the docker machine? When I ran the test on the docker the application was so fast no timeouts ever occur. Anyway it looks good. I'm +1.

andrerom:
The timeout happens on Travis, which sometimes have lots of load.

bdunogier:
+1

PR #677, opened 2016-09-15
EZP-26317: Add cancelable timeout on loading screen hide

sunpietro:
https://jira.ez.no/browse/EZP-26317

**TO DO:**
- [x] implemented cancelable loading screen hiding timeout,
- [x] updated JS unit tests

### DIFF
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -14,7 +14,8 @@ YUI.add('ez-platformuiapp', function (Y) {
 
     var L = Y.Lang,
         APP_OPEN = 'is-app-open',
-        APP_LOADING = 'is-app-loading';
+        APP_LOADING = 'is-app-loading',
+        LOADING_TIMEOUT = 500;
 
     /**
      * PlatformUI Application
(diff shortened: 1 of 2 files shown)


sunpietro:
ping @dpobel @yannickroger @StephaneDiot @dew326

dpobel:
if I get your issue and patch correctly, you want to delay the removal of the loading state because in some cases when the view becomes active in the app some elements are still loading so this extra delay allows you to set `loading` to true again before the effect of setting it to false is taken into account.

That seems a valid improvement but the current patch applies the delay on every single "page load" while that's only necessary in StudioUI. Also, if we could avoid yet another async code execution for such a simple thing, that would be great.

So how to proceed ? I think the fact that you don't want the `loading` attribute to be set to false even if the view is about to be displayed should be communicated by the view itself (In your case, the view is generating an iframe so you want to wait for that iframe to be loaded).
The `loading` attribute is set to false in https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/apps/ez-platformuiapp.js#L814 which is actually an `activeViewChange` handler so you have access to the new active View. So to not add a BC break and only apply this process, you could write something like:

```js
if ( !e.newVal.get('keepAppLoadingAfterActive') ) {
    this.set('loading', false);
}
```
That way, when you want this special behavior, you just have to define the `keepAppLoadingAfterActive` to a thruthy value.
(I'm not super inspired for the name so feel free to propose something else if you have a better idea)

sunpietro:
@dpobel thank you for your comment. I'll check if that work for me. So far the only solution I've found working was hiding the loading screen after timeout with an option to cancel the timeout and start it again. It can apply to many things, not only the loading of preview.

PR #678, opened 2016-09-17
Ran translation:update command

yannickroger:
Working on #525 I had to run `php app/console translation:update` on PlatformUI.
I had a huge "backlog" diff that was not part of what I am working on and would mess up my diff, so I am opening this PR to deal with it.

Command ran : `php app/console translation:update --dump-messages --force --output-format=xlf --no-backup en eZPlatformUIBundle`

### DIFF
--- a/Resources/translations/ezrepoforms_content_type.en.xlf
+++ b/Resources/translations/ezrepoforms_content_type.en.xlf
@@ -5,9 +5,9 @@
       <tool tool-id="symfony" tool-name="Symfony"/>
     </header>
     <body>
-      <trans-unit id="8f43b15f9faa5a91f5834c78a47cb864" resname="dashboard.title">
-        <source>dashboard.title</source>
-        <target>Administration dashboard</target>
+      <trans-unit id="972127730f863287531a351da529a802" resname="content_type.edit_title">
+        <source>content_type.edit_title</source>
+        <target>__content_type.edit_title</target>
       </trans-unit>
     </body>
   </file>
--- a/Resources/translations/role.en.xlf
+++ b/Resources/translations/role.en.xlf
@@ -1,212 +1,215 @@
 <?xml version="1.0" encoding="utf-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" version="1.2">
   <file source-language="en" target-language="en" datatype="plaintext" original="file.ext">
+    <header>
+      <tool tool-id="symfony" tool-name="Symfony"/>
+    </header>
     <body>
-      <trans-unit id="b51b764f4a4a4454ce7ae9aa09052a93" resname="role.dashboard_title">
+      <trans-unit id="fbefa01af4762e0ece92415ab7d1783f" resname="role.dashboard_title">
         <source>role.dashboard_title</source>
         <target>Roles</target>
       </trans-unit>
-      <trans-unit id="116fe107715e06c5f0b283003ff9f23f" resname="role">
+      <trans-unit id="29a7e96467b69a9f5a93332e29e9b0de" resname="role">
         <source>role</source>
         <target>Role</target>
       </trans-unit>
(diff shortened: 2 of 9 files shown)


yannickroger:
ping @andrerom @glye @bdunogier

andrerom:
+1

Resources/translations/ezrepoforms_content_type.en.xlf
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" version="1.2">
+  <file source-language="en" target-language="en" datatype="plaintext" original="file.ext">
+    <header>
+      <tool tool-id="symfony" tool-name="Symfony"/>
+    </header>
+    <body>
+      <trans-unit id="972127730f863287531a351da529a802" resname="content_type.edit_title">
+        <source>content_type.edit_title</source>
+        <target>__content_type.edit_title</target>
```
glye: This should be an actual translation message, shouldn't it? "Edit content type" or something like that.

Resources/translations/role.en.xlf
```diff
@@ -1,212 +1,215 @@
 <?xml version="1.0" encoding="utf-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" version="1.2">
   <file source-language="en" target-language="en" datatype="plaintext" original="file.ext">
+    <header>
+      <tool tool-id="symfony" tool-name="Symfony"/>
+    </header>
     <body>
-      <trans-unit id="b51b764f4a4a4454ce7ae9aa09052a93" resname="role.dashboard_title">
+      <trans-unit id="fbefa01af4762e0ece92415ab7d1783f" resname="role.dashboard_title">
```
glye: Why are the hashes changed? Won't this cause a bit of trouble for those who have already made translations? (if anyone) I realise it's the script that's doing this, but would be nice to know why.

Resources/translations/ezrepoforms_content_type.en.xlf
glye: More like: `Edit content type <%contentTypeName%>`

ref: https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/views/ContentType/update_content_type.html.twig#L6

glye: Actually, looking at my own installation, it is just: `Edit <%contentTypeName%>`
Don't know why it shows up as a placeholder... do you?

yannickroger: Actually this is not displayed. It is falling back on the translation provided by the repository-form: https://github.com/ezsystems/repository-forms/blob/master/bundle/Resources/translations/ezrepoforms_content_type.en.xlf#L58

I guess the generator adds it here for reference but I can't find a documentation confirming this.

glye: Right. Then it may be ok to keep it as is, so translators don't waste their time. Ideally we shouldn't have duplicate entries like this, but if the script generates them, it's easier to just keep them...

Resources/translations/role.en.xlf
glye: +1 when this is fixed and/or it is clear why it happens :)
(Seems I can't edit my change request text)

yannickroger: I have no clue. It happens on a couple of files. Maybe the version that generated them in the first place had an issue.
Regarding people who made translation it is the `target` that is being used.

glye: Mysteeeeerious :) Ok, I'll approve. Crossing fingers it won't somehow bite us later.

Resources/translations/ezrepoforms_content_type.en.xlf
yannickroger: It is also a way to tell people which keys are actually used by the given bundle and for us to know we can customize these keys at the bundle level if needed.

PR #680, opened 2016-09-23
Fix EZP-26357: RelationList field displays related objects for all translations

yannickroger:
Link: https://jira.ez.no/browse/EZP-26357

### Relation list
In the field(edit)view of relation list we would load every relation attached to the object instead of the one mentioned in the field.

### Relation
For the relation field, we would load the relation from the content relations instead of the field relation.

### Note
Unfortunately, the content ids stored in the field are API ids and not REST ids, this should be fixed in [EZP-23000](https://jira.ez.no/browse/EZP-23000) but for 1.5 we need to generate the REST id ourselves.

## Tests
Manual and unit tests

### DIFF
--- a/Resources/public/js/models/ez-contentmodel.js
+++ b/Resources/public/js/models/ez-contentmodel.js
@@ -212,19 +212,45 @@ YUI.add('ez-contentmodel', function (Y) {
                 return this.get('relations');
             }
 
+            if (type === "ATTRIBUTE" && fieldDefinitionIdentifier) {
+                return this._getRelationsFromField(fieldDefinitionIdentifier);
+            }
+
             relations = Y.Array.filter(this.get('relations'), function (relation) {
-                if (
-                    fieldDefFilter
-                    && type === relation.type
-                    && fieldDefinitionIdentifier === relation.fieldDefinitionIdentifier
-                ) {
-                    return true;
-                } else if ( !fieldDefFilter && type === relation.type ) {
+                if ( !fieldDefFilter && type === relation.type ) {
                     return true;
                 }
                 return false;
             });
             return relations;
+        },
+
+        /**
+         * Returns the relations ids from the field in the shape of an Array of Object
+         *
+         * @protected
+         * @method _getRelationsFromField
+         *
+         * @param {String} fieldDefinitionIdentifier
+         * @return {Array}
+         */
+        _getRelationsFromField: function(fieldDefinitionIdentifier) {
+            var fields = this.get('fields'),
+                destinationContentIds,
+                fieldValue = fields[fieldDefinitionIdentifier].fieldValue;
+
+            if (fieldValue.destinationContentIds) {
+                // relation list
+                destinationContentIds = fieldValue.destinationContentIds;
+            } else {
+                // relation
+                destinationContentIds = [fieldValue.destinationContentId];
+            }
+
+            return destinationContentIds.map(function (contentId) {
+                // Will use a cleaner approach on 1.6 when EZP-23000 is fixed and we have REST ID available
+                return {destination: "/api/ezp/v2/content/objects/" + contentId};
+            });
         },
 
         /**
--- a/Tests/js/models/assets/ez-contentmodel-tests.js
+++ b/Tests/js/models/assets/ez-contentmodel-tests.js
@@ -416,8 +416,8 @@ YUI.add('ez-contentmodel-tests', function (Y) {
                 var elt = expected[i];
 
                 Y.Assert.areEqual(
-                    elt.id,
-                    relation.id,
+                    elt.destination,
+                    relation.destination,
                     "The element " + i + " should be " + elt.id
                 );
             });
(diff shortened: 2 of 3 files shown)


yannickroger:
I added a fix for relation list which is also affected by the same problem.

Resources/public/js/views/fields/ez-relationlist-editview.js
```diff
@@ -93,14 +93,14 @@ YUI.add('ez-relationlist-editview', function (Y) {
...
-         * Fire the `loadObjectRelations` event
+         * Fire the `loadFieldRelations` event
          *
          * @method _fireLoadObjectRelations
          * @protected
          */
         _fireLoadObjectRelations: function () {
             if ( !this._isFieldEmpty() ) {
-                this.fire('loadObjectRelations', {
+                this.fire('loadFieldRelations', {
```
dpobel: you should not change the event name first because we should avoid to do that in a minor release second because the info that you want the relation in a given field is already expressed in the event parameters.

Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
```diff
@@ -79,6 +80,60 @@ YUI.add('ez-objectrelationsloadplugin', function (Y) {
...
+                end = stack.add(function (error, struct) {
+                    if (error) {
+                        e.target.set("loadingError", true);
+                        loadingError = true;
+                    } else {
+                        relatedContentListArray.push(struct);
+                    }
+                });
+
+            Y.Array.each(destinationContentIds, function (contentId) {
```
dpobel: actually the problem lies in eZ.Content#relations where the field value should be used when we want the relations of type `ATTRIBUTE`

yannickroger:
Feedback taken into account and test coverage completed. ping @dpobel @StephaneDiot

Resources/public/js/models/ez-contentmodel.js
```diff
@@ -212,6 +212,10 @@ YUI.add('ez-contentmodel', function (Y) {
                 return this.get('relations');
             }
 
+            if (type === "ATTRIBUTE") {
```
dpobel: actually you should check if a field definition identifier was provided because this implies `ATTRIBUTE` type and this would not break the case `content.relations('ATTRIBUTE')`

dpobel: Also that means the first condition in the filter function below becomes useless

Resources/public/js/models/ez-contentmodel.js
```diff
@@ -228,6 +232,32 @@ YUI.add('ez-contentmodel', function (Y) {
...
+         *
+         * @protected
+         * @method _getRelationsFromField
+         *
+         * @param {String} [fieldDefinitionIdentifier]
+         * @return {Array}
+         */
+        _getRelationsFromField: function(fieldDefinitionIdentifier) {
+            var fields = this.get('fields'),
+            destinationContentIds,
```
dpobel: indentation issue

Resources/public/js/models/ez-contentmodel.js
```diff
@@ -228,6 +232,32 @@ YUI.add('ez-contentmodel', function (Y) {
...
+            relations = [];
+
+            if (fieldDefinitionIdentifier) {
+                destinationContentIds = fields[fieldDefinitionIdentifier].fieldValue.destinationContentIds;
+
+                destinationContentIds.forEach(function (contentId) {
```
dpobel: nitpick: this is typical case where you should use `map`

Resources/public/js/views/services/plugins/ez-objectrelationloadplugin.js
```diff
@@ -36,11 +36,12 @@ YUI.add('ez-objectrelationloadplugin', function (Y) {
         _loadFieldRelatedContent: function (e) {
             var loadOptions = {api: this.get('host').get('capi')},
                 relatedContent = this.get('relatedContent'),
-                contentDestination = this.get('host').get('content').relations(
-                    'ATTRIBUTE', e.fieldDefinitionIdentifier
-                ).shift();
+                fields = this.get('host').get('content').get('fields'),
+                destinationContentId = fields[e.fieldDefinitionIdentifier].fieldValue.destinationContentId,
```
dpobel: you are kind of duplicating the `relations` method here, so you should still use it (that means `relations` has to be able to handle both `ezobjectrelation` and `ezobjectrelationlist` fields)

yannickroger:

Tests/js/models/assets/ez-contentmodel-tests.js
```diff
@@ -427,8 +427,38 @@ YUI.add('ez-contentmodel-tests', function (Y) {
...
         },
 
-        "Should filter ATTRIBUTE relations with a field identifier": function () {
-            this._testRelations(this.relationsAttribute2, "ATTRIBUTE", "attr2");
+        "Should return ATTRIBUTE relation list with a field identifier": function () {
+            var fieldDefinitionIdentifier = 'attr2',
+                fields = {},
+                destinationContentId = 42;
+
+            fields[fieldDefinitionIdentifier] = {fieldValue:
```
dpobel: CS
```js
fields[fieldDefinitionIdentifier] = {
    fieldValue: {destinationContentIds: [destinationContentId]},
};
```

Tests/js/models/assets/ez-contentmodel-tests.js
```diff
@@ -427,8 +427,38 @@ YUI.add('ez-contentmodel-tests', function (Y) {
...
+                fieldDefinitionIdentifier
+            );
+        },
+
+        "Should return ATTRIBUTE relation with a field identifier": function () {
+            var fieldDefinitionIdentifier = 'attr2',
+                fields = {},
+                destinationContentId = 42;
+
+            fields[fieldDefinitionIdentifier] = {fieldValue:
```
dpobel: CS

Resources/public/js/models/ez-contentmodel.js
```diff
@@ -228,6 +226,34 @@ YUI.add('ez-contentmodel', function (Y) {
         },
 
         /**
+         * Returns the relations ids from the field in the shape of an Array of Object
+         *
+         * @protected
+         * @method _getRelationsFromField
+         *
+         * @param {String} [fieldDefinitionIdentifier]
```
dpobel: `fieldDefinitionIdentifier` is mandatory now

PR #681, opened 2016-09-29
[travis] Run behat without interaction

bdunogier:


bdunogier:
Travis agreed. Could one of you approve, for the sake of it ? :)

Ping @yannickroger.

andrerom:
+1 ( and on mobile so cant use new approval system, just like reactions it is missing :.( )

PR #682, opened 2016-09-30
Fix EZP-26371: "Reset password" link does nothing.

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-26371

Removes currently unused link, which should be reintroduced once the feature is done/available.


joaoinacio:
@andrerom : I'm guessing this should be on 1.4 as well?

andrerom:
No 1.5 and up is fine, last 1.4 patch had been made.

dpobel:
+1

pcardiga:
PR verified by QA - +1

PR #683, opened 2016-10-05
EZP-26135: Line breaks added in richtext editor when editing content twice

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26135

# Description

This is yet again a problem about XML vs. HTML. Basically in PlatformUI, we use the `xhtml5edit` representation of the RichText fields. This `xhtml5edit` is an XML dialect closed to XHTML and then HTML, but some details confuse browser parsers mainly the auto-closed tag like `<br/>` and `<div/>` (for embeds). So this patch introduces a better transformation from `xhtml5edit` to HTML.

# Tests

manual + unit tests


dpobel:

PR #684, opened 2016-10-06
Proof of concept of bundle generator

bdunogier:
`php app/console generate:ez:platform-ui-plugin`

Only supported format is `yml` at the moment.

### TODO
- [ ] support other configuration formats
- [ ] Fill in the TODO list
- [ ] Check if the generated code actually works as expected

### DIFF
--- a/Command/GeneratePlatformUISkeletonBundleCommand.php
+++ b/Command/GeneratePlatformUISkeletonBundleCommand.php
@@ -1,0 +1,38 @@
+<?php
+namespace EzSystems\PlatformUIBundle\Command;
+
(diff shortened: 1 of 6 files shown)


bdunogier:
I expected https://github.com/ezsystems/PlatformUIBundle/pull/684/commits/e56aa10c6c965375f59ab9164e58aa885ab09231 to fix php-cs, but it doesn't seem so...

dpobel:
awesome :) I'll try to test that shortly

yannickroger:
Great idea :smiley:

PR #687, opened 2016-10-12
[Prototype] EZP-26309: Prototype the usage of REST Embedding in PlatformUI

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26309

# Description

This is a prototype to use REST Embeddding (implemented in https://github.com/ezsystems/ezpublish-kernel/pull/1741) in PlatformUI. This is done for 2 parts in PlatformUI:

1. on the "My Drafts" block on the dashboard, this avoids having to load the ContentInfo for each version individually.
1. on the "Version" tab when viewing a Content, this allows to get the Creator of a Version without a new requests.

# Outcome

I see 2 issues in the current implementation:

1. it seems like the ContentInfo has been removed from the Location response. This is breaking some parts of PlatformUI.
1. there's a recursion protection which prevents the system from embedding the same object several times. For instance 2 versions can have the same creator (or you can have 2 drafts for the same Content) and this is something very complicated to handle in the PlatformUI... :)

Other than that, this is working well. The performance improvement on the dashboard block is really interesting and we could so the same in a lots of places!

### DIFF
--- a/Resources/public/js/models/extensions/ez-contentinfo-base.js
+++ b/Resources/public/js/models/extensions/ez-contentinfo-base.js
@@ -25,26 +25,36 @@ YUI.add('ez-contentinfo-base', function (Y) {
          * @method loadVersions
          * @param {Object} options
          * @param {Object} options.api (required) the JS REST client instance
+         * @param {Boolean} [options.preloadCreator]
          * @param {Function} callback
          */
         loadVersions: function (options, callback) {
             var versions = [],
-                contentService = options.api.getContentService();
+                contentService = options.api.getContentService(),
+                cb = function (error, response) {
+                    if (error) {
+                        callback(error, response);
+                        return;
+                    }
 
-            contentService.loadVersions(this.get('id'), function (error, response) {
-                if (error) {
-                    callback(error, response);
-                    return;
+                    Y.Array.each(response.document.VersionList.VersionItem, function (versionItemHash) {
+                        var versionInfo = new Y.eZ.VersionInfo();
+                        versionInfo.loadFromHash(versionItemHash);
+                        versions.push(versionInfo);
+                    });
+
+                    callback(error, versions);
+                };
+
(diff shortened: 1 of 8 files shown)


Resources/public/js/models/extensions/ez-contentinfo-base.js
```diff
@@ -25,26 +25,36 @@ YUI.add('ez-contentinfo-base', function (Y) {
...
+                        callback(error, response);
+                        return;
+                    }
 
-            contentService.loadVersions(this.get('id'), function (error, response) {
-                if (error) {
-                    callback(error, response);
-                    return;
-                }
+                    Y.Array.each(response.document.VersionList.VersionItem, function (versionItemHash) {
```
sunpietro: Nitpick: `response.document.VersionList.VersionItem.forEach()` is a better, native way of doing it

Resources/public/js/models/ez-restmodel.js
```diff
@@ -134,6 +134,17 @@ YUI.add('ez-restmodel', function (Y) {
                 }
             });
             Y.Array.each(this.constructor.LINKS_MAP, function (item) {
+                var attrName, linkName;
+
+                if ( L.isObject(item) ) {
+                    linkName = Y.Object.keys(item)[0];
```
sunpietro: Nitpick: `Object.keys()` works natively everywhere

Resources/public/js/models/extensions/ez-contentinfo-base.js
dpobel: thanks for the nitpicking, but what about the core feature/implementation ? (In case, you missed it, that's only a **Prototype**)

sunpietro: @dpobel I'm going to check it ASAP. Very interesting approach.

PR #689, opened 2016-10-13
EZP-25428 translate changes (Damien)

dpobel:
my part of this boring work... :)

Some strings are not translated yet because we need support parameters, the following templates have such strings:

* searchlistview.hbt
* universaldiscovery/search.hbt
* universaldiscovery/confirmedlist.hbt
* contentedit.hbt
* subitem/grid.hbt
* subitem/listmore.hbt

PR #690, opened 2016-10-14
EZP-25428 Adding translation keys (1/3)

yannickroger:
## Description
We split the fun of adding the keys in template in 3 with @dpobel (#689) and @StephaneDiot (#692)

This is my part of the work.

## Test
Basic sanity on the most complex templates

## TODO
- [x] Squash commits by domains
- [x] Sanity test

### DIFF
--- a/Resources/public/templates/fields/edit/binaryfile.hbt
+++ b/Resources/public/templates/fields/edit/binaryfile.hbt
@@ -8,33 +8,47 @@
             <p class="ez-field-description is-showing-description">
                 {{ translate_property fieldDefinition.descriptions }}
             </p>
-            <p class="ez-binarybase-help">Drag and drop your file here or simply click on the buttons and select your file manually.</p>
+            <p class="ez-binarybase-help">
+                {{translate "binaryfile.help" "fieldedit"}}
+            </p>
         </label>
     </div>
     <div class="pure-u ez-editfield-input-area">
         <div class="ez-editfield-input ez-binarybase-drop-area"><div class="ez-binaryfile-input-ui">
             <div class="ez-binaryfile-content ez-font-icon ez-binarybase-content">
                 <ul class="ez-binaryfile-properties">
-                    <li><b>File name:</b> <span class="ez-binaryfile-properties-name">{{ binaryfile.name }}</span></li>
-                    <li><b>Type:</b> <span class="ez-binaryfile-properties-type">{{ binaryfile.type }}</span></li>
-                    <li><b>File size:</b> <span class="ez-binaryfile-properties-size">{{ binaryfile.size }}</span>&nbsp;bytes</li>
-                    <li class="ez-binaryfile-properties-download"><a href="{{ binaryfile.uri }}" class="ez-binaryfile-download" target="_blank">Download the file</a></li>
+                    <li>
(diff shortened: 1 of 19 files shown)


Resources/public/templates/fields/edit/binaryfile.hbt
```diff
@@ -8,33 +8,47 @@
...
+                        <span class="ez-binaryfile-properties-name">{{ binaryfile.name }}</span>
+                    </li>
+                    <li>
+                        <b>{{translate "binaryfile.type" "fieldedit"}}</b>
+                        <span class="ez-binaryfile-properties-type">{{ binaryfile.type }}</span>
+                    </li>
+                    <li>
+                        <b>{{translate "binaryfile.size" "fieldedit"}}</b>
+                        <span class="ez-binaryfile-properties-size">{{ binaryfile.size }}</span>&nbsp;
+                        {{translate "binaryfile.bites" "fieldedit"}}
```
dpobel: bytes ;)

Resources/public/templates/fields/edit/date.hbt
```diff
@@ -21,7 +21,7 @@
             >
```
dpobel: the placeholder should be translated I think

Resources/public/templates/fields/edit/image.hbt
```diff
@@ -8,54 +8,66 @@
...
+            <p class="ez-binarybase-help">
+                {{translate "image.help" "fieldedit"}}
+            </p>
         </label>
     </div>
     <div class="pure-u ez-editfield-input-area">
         <div class="ez-editfield-input ez-binarybase-drop-area"><div class="ez-image-input-ui">
             <div class="ez-image-editpreview ez-binarybase-content">
                 <div class="ez-image-editpreview-image">
                     <img alt="Image preview" class="ez-image-preview" {{#if image.displayUri}}src="{{ image.displayUri }}"{{/if}} alt="Image preview">
```
dpobel: missing translation on alt value

Resources/public/templates/navigationhub.hbt
```diff
@@ -11,7 +11,7 @@
     <ul class="ez-navigation-{{ @key }} is-navigation-hidden">
```
dpobel: the alt above should also be translated

Resources/translations/confirm.en.xlf
```diff
@@ -0,0 +1,34 @@
...
+        <source>draftconflict.newdraft</source>
+        <target>New Draft</target>
+      </trans-unit>
+      <trans-unit id="36595b085e28a9a9c678bd8a368f1507" resname="draftconflict.language">
+        <source>draftconflict.language</source>
+        <target>Language</target>
+      </trans-unit>
+      <trans-unit id="ff9f19d1379eaae89ed53ef575cdcee8" resname="draftconflict.owner">
+        <source>draftconflict.owner</source>
+        <target>_Owner</target>
```
dpobel: _ can be removed

Resources/public/templates/fields/edit/binaryfile.hbt
dpobel: actually we should probably have a string with a parameter for the size, so until we have that, we should let the original string

PR #691, opened 2016-10-14
Fix EZP-26336: Inconsistent behaviour of the search UDW method

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26336

# Description

This patch fixes an issue in the search UDW method where the first search never returns anything even if the REST API response contains some results. Then following searches have a wrong result count. This is happening because the rendering of the search result is triggered by the change of the search result list attribute but the result count attribute is updated afterwards.

# Tests

manual test + unit tests


dpobel:

PR #693, opened 2016-10-14
Fix EZP-26036: Double content type icon in the grid view

yannickroger:
Link: https://jira.ez.no/browse/EZP-26036

## Description
Removed static Icon as it is not needed.

## Tests
Basic sanity on grid


miguelcleverti:
PR approved by QA

PR #686, opened 2016-10-10
EZP-26315: Improve UDW selected view

StephaneDiot:
jira : https://jira.ez.no/browse/EZP-26315

## Description

Goal here was to improve the selected view of the UDW with infos and image.

## Screenshot

image

## Tests

unit and manually tested

PR #695, opened 2016-10-17
Fix EZP-26433: Impossible to select a Content in the UDW search method if the UDW is configured to allow a multiple selection

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26433

# Description

This is a regression caused by https://github.com/ezsystems/PlatformUIBundle/pull/686. This is happening because in some cases the view attributes are reset, this process includes the `multiple` attribute which then contains its default value (`false`) instead of the configured value. By fixing that, I also figured out that the Content item not always available (while it should if we want to display an image field when there's one). This is also fixed by forcing  `loadContent` to be true. (This parameter is now useless, it'll be deprecated shortly in https://jira.ez.no/browse/EZP-26434).

# Tests

manual tests + unit tests

### DIFF
--- a/Resources/public/js/views/ez-universaldiscoveryview.js
+++ b/Resources/public/js/views/ez-universaldiscoveryview.js
@@ -283,7 +283,7 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
 
                 method.setAttrs({
                     'multiple': this.get('multiple'),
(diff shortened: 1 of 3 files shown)


dpobel:

miguelcleverti:
PR approved by QA

PR #697, opened 2016-10-18
Fix EZP-26241: JS error when giving the focus to the editor by clicking on an embed

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26241

# Description

This is happening only in Firefox and the JavaScript error is due to the fact that the editor and the widget do not have the focus yet. So this patch makes sure the editor and widget have the focus.

# Tests

Unit tests + manual tests in Firefox and Chrome


dpobel:

PR #698, opened 2016-10-18
Fix EZP-26246: Wrong visibility status after refresh

yannickroger:
Link: https://jira.ez.no/browse/EZP-26246

## Description
When hiding/revealing a location, the `hidden` was updated in the PlatformUI but not the `invisible`. In order not to duplicate the business logic (that is more complicated for `invisible`) we just update the PlatformUI in memory `location` with the result of the REST response.

## Test
Manual test and updated unit tests.

### DIFF
--- a/Tests/js/models/assets/ez-locationmodel-tests.js
+++ b/Tests/js/models/assets/ez-locationmodel-tests.js
@@ -558,12 +558,21 @@ YUI.add('ez-locationmodel-tests', function (Y) {
             delete this.model;
         },
 
-        _configureUpdateLocationMock: function (cbError) {
+        _configureUpdateLocationMock: function (cbError, invisible, hidden) {
+            var response = {
+                    document: {
+                        Location: {
+                            invisible: invisible,
+                            hidden: hidden,
+                        },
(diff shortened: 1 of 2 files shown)


yannickroger:
ping @dpobel @StephaneDiot

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -558,20 +558,25 @@ YUI.add('ez-locationmodel-tests', function (Y) {
             delete this.model;
         },
 
-        _configureUpdateLocationMock: function (cbError) {
+        _configureUpdateLocationMock: function (cbError, invisible, hidden) {
+            var response = { document: { Location: {
```
dpobel:
```js
var response = {
        document: {
            Location: {
                invisible: invisible,
                hidden: hidden,
            },
        },
    };
```

or everything on one line.

yannickroger:
Feedback by @dpobel done. ping @StephaneDiot

miguelcleverti:
PR approved by QA

PR #699, opened 2016-10-18
Fix EZP-26016: JavaScript error when removing an embed if it's the last element in the editor

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26016

# Description

This patch just makes sure we don't try to give the focus to an *internal* element generated by the widget CKEditor plugin which results in a JavaScript error because this element is later removed from the document.

# Tests

manual tests + unit tests


dpobel:

miguelcleverti:
PR approved by QA

PR #700, opened 2016-10-18
Fix EZP-26384: Fail to save a Content item with a RichText field containing an anchor

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26384

# Description

This patch makes sure the anchors (`a` element without `href` attribute) are removed from the RichText field value since we don't support anchors (yet).

# Tests

manual tests + unit tests.


dpobel:
ping @yannickroger @StephaneDiot @andrerom

dpobel:
@andrerom +1 on your last comment, hence why this patch introduces a processor we can easily disable/deprecate as soon as we support anchors.

miguelcleverti:
PR approved by QA

PR #702, opened 2016-10-19
Fix EZP-26020: (+) sign disappears after aligning an embed/image

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26020

# Description

After aligning an embed or an image, the (+) button disappears. This patch fixes that by making sure the `editorInteraction` event is fired when changing the alignment of a widget so that AlloyEditor knows on which element the user was acting and then can react accordingly.

# Tests

unit tests + manual tests


dpobel:
ping @stephanediot @yannickroger

pcardiga:
PR Verified by QA

PR #703, opened 2016-10-20
Fix EZP-26355: Cannot paste an URL in a RichText field

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26355

# Description

When pasting an URL in Online Editor, the editor has a weird behavior. This is happening because AlloyEditor dev recently added a plugin (`ae_embed`) that implements [the oEmbed API](http://oembed.com/) so when pasting an URL the editor tries to create a widget which does not play well with our own customizations. So for now the solution is to disable `ae_embed`. (We'll maybe re-introduce that when we have custom tags)

# Tests

manual tests + unit tests


dpobel:
ping @stephanediot @yannickroger

PR #704, opened 2016-10-21
EZP-26390: Fail to save Content item after pasting

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26390

# Description

Before this patch, the paste behavior was not configured so the AlloyEditor/CKEditor was applying [the `semantic-content` filter](http://docs.ckeditor.com/#!/guide/dev_advanced_content_filter-section-filtering-pasted-and-dropped-content) in Webkit/Blink based browser when pasting unless the code was detected as coming from the Word.

This patch configures and tweaks the paste behaviour:

* it adds a custom paste filter to match what we actually support server side and in the RichText editor (so this filter is also applied with others browsers)
* it applies the previous paste filter even if the pasted code is detected as coming from Word (the `pastefromword` plugin keeps a lots of div, that's exactly the customer issue)
* it adds an additional cleanup to avoid having paragraph inside list items.

# Tests

unit tests + manual tests

### DIFF
--- a/Tests/js/alloyeditor/plugins/assets/ez-alloyeditor-plugin-paste-tests.js
+++ b/Tests/js/alloyeditor/plugins/assets/ez-alloyeditor-plugin-paste-tests.js
@@ -1,0 +1,251 @@
+/*
+ * Copyright (C) eZ Systems AS. All rights reserved.
+ * For full copyright and license information view LICENSE file distributed with this source code.
+ */
+/* global CKEDITOR, AlloyEditor */
+YUI.add('ez-alloyeditor-plugin-paste-tests', function (Y) {
+    var definePluginTest, pasteTest,
+        Assert = Y.Assert, Mock = Y.Mock;
+
+    definePluginTest = new Y.Test.Case({
+        name: "eZ AlloyEditor paste plugin define test",
+
+        setUp: function () {
+            this.editor = new Mock();
+            Mock.expect(this.editor, {
+                method: 'on',
+                args: ['paste', Mock.Value.Function],
+            });
+        },
+
+        tearDown: function () {
+            delete this.editor;
+        },
+
+        "Should define the paste plugin": function () {
+            var plugin = CKEDITOR.plugins.get('ezpaste');
+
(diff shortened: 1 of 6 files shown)


dpobel:

joaoinacio:
The failure seems to be caused by https://github.com/ezsystems/ezplatform-solr-search-engine/commit/a82329e4ca951cce in ezplatform-solr-search-engine v1.1.2 , without the matching impl in kernel 6.5.1.1 (see https://jira.ez.no/browse/EZP-26219)

Tests/js/alloyeditor/plugins/assets/ez-alloyeditor-plugin-paste-tests.js
```diff
@@ -0,0 +1,251 @@
...
+                expected = '<p>Led Zeppelin</p>';
+
+            this._testPaste(code, expected, "The image should have been removed");
+        },
+
+        "Should remove videos": function () {
+            var code = '<p>Led Zeppelin<video src="#"></video></p>',
+                expected = '<p>Led Zeppelin</p>';
+
+            this._testPaste(code, expected, "The image should have been removed");
```
StephaneDiot: niptick : "The video should ...."

Tests/js/alloyeditor/plugins/assets/ez-alloyeditor-plugin-paste-tests.js
```diff
@@ -0,0 +1,251 @@
...
+        "Should keep the strong element": function () {
+            var code = '<p>Led Zeppelin - <strong>Over the hill and far away</strong></p>';
+
+            this._testPaste(code, code, "The strong element should be kept");
+        },
+
+        "Should keep the b element": function () {
+            var code = '<p>Led Zeppelin - <b>Over the hill and far away</b></p>';
+
+            this._testPaste(code, code, "The belement should be kept");
```
StephaneDiot: niptick : "The b element ..."

miguelcleverti:
PR Approved by QA

PR #705, opened 2016-10-24
Fix EZP-26492: Impossible to edit a Content with a RichText field in Edge

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26492

# Description

That's a regression from https://github.com/ezsystems/PlatformUIBundle/pull/683 because in Edge (happens also in IE11), `document.implementation.createHTMLDocument` expects a parameter while in others browser, this parameter is not mandatory.
The fix just removes this call as it's not necessary since `document` is already an HTML document.

# Tests

manual tests + unit tests


dpobel:

pcardiga:
Verified by QA.
Tested on Windows 10 with IE11 and Edge. 
Also searched for possible side effects on Firefox, Chrome and Opera and found none.

PR #706, opened 2016-10-24
Fix EZP-26486: In Firefox, media view doesn't update the video

yannickroger:
Link: https://jira.ez.no/browse/EZP-26486

## Description
Firefox doesn't update the video if the link doesn't change. So we are adding the id of the video in order to force the refresh.

Example of generated URL: `http://ezp.local/content/download/84/322?id=video/aa5ef523f7825634e10dbacb065c1507.webm`

## Tests
Manual and unit tests

### DIFF
--- a/Resources/public/templates/fields/view/media.hbt
+++ b/Resources/public/templates/fields/view/media.hbt
@@ -10,10 +10,10 @@
         <div class="ez-media-view">
             <div class="ez-media-view-player">
             {{#if isAudio}}
-                <audio src="{{ value.uri }}" controls="controls" preload="auto" class="ez-media-player ez-media-audio-player"></audio>
+                <audio src="{{ value.uri }}?id={{ value.id }}" controls="controls" preload="auto" class="ez-media-player ez-media-audio-player"></audio>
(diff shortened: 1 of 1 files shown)


Resources/public/js/views/fields/ez-media-view.js
```diff
@@ -32,10 +32,25 @@ YUI.add('ez-media-view', function (Y) {
...
+         *
+         * @method _generateUriWithTimestamp
+         * @return {String} The URI with a timestamp parameter or "" if field is empty
+         */
+        _generateUriWithTimestamp: function() {
+            if (this._isFieldEmpty()) {
+                return "";
+            }
+
+            return this.get('field').fieldValue.uri + "?t=" + Date.now();
```
andrerom: no version number we can add in url instead of always refreshing?

yannickroger: The version number is not available in this view. But looking for it I found an id that does the trick :smile: 

Following @dpobel 's advice I modified the URL in the template only.

yannickroger:
Ready for review @dpobel @StephaneDiot @andrerom

miguelcleverti:
PR approved by QA

PR #709, opened 2016-10-28
Fixed embed unit tests to not fail with AlloyEditor 1.2.4

dpobel:
# Description

The upgrade to AlloyEditor 1.2.4 broke the unit tests (caused by a change in the "placeholder" plugin). This patch removes the failing tests as the place where the HTML code is inserted is handled by another plugin, instead we just make sure the insertion of a new embed calls the API exposed by this plugin.

# Tests

TravisCI seems to be happy :champagne:

### DIFF
--- a/Tests/js/alloyeditor/plugins/assets/ez-alloyeditor-plugin-embed-tests.js
+++ b/Tests/js/alloyeditor/plugins/assets/ez-alloyeditor-plugin-embed-tests.js
@@ -756,9 +756,21 @@ YUI.add('ez-alloyeditor-plugin-embed-tests', function (Y) {
                     },
                 }
             );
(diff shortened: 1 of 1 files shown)


andrerom:
```
✖ [eZ AlloyEditor embed plugin tests]: Passed: 29 Failed: 1 Total: 30 (ignored 0) (3.288 seconds)
    Should insert the widget at the beginning of the document
       Unexpected error: 'undefined' is not an object (evaluating 'b.checkReadOnly')
```

dpobel:
ok I'm reproducing it locally now. It seems it's caused by the last update of AlloyEditor (1.2.3 to 1.2.4)

dpobel:

PR #710, opened 2016-11-02
EZP-26431: Allow strings with parameters to be translated

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26431

# Description

This patch brings the ability to translate strings containing some parameters. The `translate` Handlebars helper now accepts named parameters that can be used in the strings. The Handlebars dumper has been changed to correctly handle those extra parameters and the corresponding strings have been added to the translation files.

* [x] Change the Handlebars helpers to accept the parameters
* [x] Make sure the dumper works with the new syntax
* [x] Add the missing translations

# Tests

unit tests + manual tests


dpobel:

PR #711, opened 2016-11-02
Fix EZP-26469: Items hidden from barview after scrolling

yannickroger:
Link: https://jira.ez.no/browse/EZP-26469

## Description
Scrolling was not taken into account in the calculation to hide items from the left/right menus.

## Tests
Manual and unit tests


pcardiga:
Verified by QA

PR #712, opened 2016-11-03
EZP-26325: Use query and filter in PlatformUI

StephaneDiot:
Jira: https://jira.ez.no/browse/EZP-26325

## Description

The goal here was to be able to use query and filter Query properties instead of criteria. By using this users will now be able, while using Solr, to get search results sorted with relevancy.

BTW you need to have a ez-JS-REST-client up to date with https://github.com/ezsystems/ez-js-rest-client/pull/80 (hopefully merged soon).

## Tests
- unit tests

## Todo
- [x] Review & fixes
- [x] Bump composer.json / * requirements on rest client once that has been merged and new tag is out

### DIFF
--- a/Resources/public/js/views/services/plugins/ez-searchplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-searchplugin.js
@@ -69,6 +69,9 @@ YUI.add('ez-searchplugin', function (Y) {
             }
             if (search.criteria) {
                 query.setCriteria(search.criteria);
+                console.log('[DEPRECATED] Criteria property is deprecated');
+                console.log('[DEPRECATED] it will be removed from PlatformUI 2.0');
+                console.log('[DEPRECATED] Please use Query or Filter instead');
             }
 
             query.setLimitAndOffset(search.limit, search.offset);
--- a/Resources/public/js/views/services/plugins/ez-searchplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-searchplugin.js
@@ -46,7 +46,9 @@ YUI.add('ez-searchplugin', function (Y) {
          * @param {String} name
          * @param {String} type
          * @param {Object} search
-         * @param {Object} search.criteria
+         * @param {Object} [search.criteria]
+         * @param {Object} [search.query]
+         * @param {Object} [search.filter]
          * @param {Object} search.sortClauses
          * @param {Number} [search.limit]
          * @param {Number} [search.offset]
@@ -59,13 +61,21 @@ YUI.add('ez-searchplugin', function (Y) {
             query = contentService.newViewCreateStruct(name, type);
             // TODO ViewCreateStruct should expose an API
             // see https://jira.ez.no/browse/EZP-24808
-            query.body.ViewInput[type].Criteria = search.criteria;
-            query.body.ViewInput[type].offset = search.offset;
-            query.body.ViewInput[type].limit = search.limit;
+            if (search.query) {
+                query.setQuery(search.query);
+            }
+            if (search.filter) {
+                query.setFilter(search.filter);
+            }
+            if (search.criteria) {
+                query.setCriteria(search.criteria);
+            }
+
+            query.setLimitAndOffset(search.limit, search.offset);
             if ( search.sortClauses ) {
-                query.body.ViewInput[type].SortClauses = search.sortClauses;
+                query.setSortClauses(search.sortClauses);
             } else if ( search.sortLocation ) {
-                query.body.ViewInput[type].SortClauses = search.sortLocation.getSortClause();
+                query.setSortClauses(search.sortLocation.getSortClause());
             }
 
             return query;
@@ -125,7 +135,9 @@ YUI.add('ez-searchplugin', function (Y) {
          * @method findLocations
          * @param {Object} search
          * @param {String} search.viewName the name of the REST view to use
-         * @param {Object} search.criteria the search criteria used as Criteria in LocationQuery
+         * @param {Object} search.criteria (deprecated) the search criteria used as Criteria in LocationQuery
+         * @param {Object} search.query  the search query used as Query in LocationQuery
+         * @param {Object} search.filter the search filter used as Filter in LocationQuery
          * @param {Object} [search.sortClauses] the sort clauses
          * @param {eZ.Location} [search.sortLocation] the Location to use to
          * generate the sort clauses
@@ -180,7 +192,9 @@ YUI.add('ez-searchplugin', function (Y) {
          * @protected
          * @param {EventFacade} e
          * @param {Object} e.search
-         * @param {Object} e.search.criteria the search criteria used as Criteria in LocationQuery
+         * @param {Object} e.search.criteria (deprecated) the search criteria used as Criteria in LocationQuery
+         * @param {Object} e.search.query  the search query used as Query in LocationQuery
+         * @param {Object} e.search.filter the search filter used as Filter in LocationQuery
          * @param {Object} [e.search.sortClauses] the search sort clauses
          * @param {eZ.Location} [search.sortLocation] the Location to use to
          * generate the sort clauses
@@ -402,7 +416,7 @@ YUI.add('ez-searchplugin', function (Y) {
             });
 
             query = this._createNewCreateViewStruct('contents-loading-' + viewName, 'ContentQuery', {
-                criteria: {
+                filter: {
                     "ContentIdCriterion": contentIds
                 },
                 // In case we are asking for more then 25 items which is default limit, specify limit
--- a/Resources/public/js/views/services/plugins/ez-searchplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-searchplugin.js
@@ -69,9 +69,12 @@ YUI.add('ez-searchplugin', function (Y) {
             }
             if (search.criteria) {
                 query.setCriteria(search.criteria);
+<<<<<<< ec7f7514a95176a5aa31d3dc3e9caf68f31ad205
                 console.log('[DEPRECATED] Criteria property is deprecated');
                 console.log('[DEPRECATED] it will be removed from PlatformUI 2.0');
                 console.log('[DEPRECATED] Please use Query or Filter instead');
+=======
+>>>>>>> EZP-26325: Use query and filter in PlatformUI
             }
 
             query.setLimitAndOffset(search.limit, search.offset);
--- a/Resources/public/js/views/services/plugins/ez-searchplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-searchplugin.js
@@ -107,7 +107,39 @@ YUI.add('ez-searchplugin', function (Y) {
          * @param {Number} e.callback.count the total number of search result
          */
         _doContentSearch: function (e) {
-            var query = this._createNewCreateViewStruct(e.viewName, 'ContentQuery', e.search),
+            var search = Y.merge(e.search);
+
+            search.viewName = e.viewName;
+            search.loadContentType = e.loadContentType;
+            delete search.callback;
+            this.findContent(search, e.callback);
+        },
+
+        /**
+         * Executes a Content search based on the provided `search` object.
+         *
+         * @method findContent
+         * @param {Object} search
+         * @param {String} search.viewName the name of the REST view to use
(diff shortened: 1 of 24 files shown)


StephaneDiot:
ready for review @dpobel @yannickroger

Tests/js/models/assets/ez-contentmodel-tests.js
```diff
@@ -820,15 +820,22 @@ YUI.add('ez-contentmodel-tests', function (Y) {
...
                         }
                     }
                 }
-            };
+            });
+            Mock.expect(this.query, {
+                method: 'setFilter',
+                args: [Mock.Value.Object],
+                run: Y.bind(function (arg) {
+                    this.query.body.ViewInput.LocationQuery.Filter = arg;
```
dpobel: you should not need that. The point of defining a Mock is too avoid this (reimplementation of the object being mocked)

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -765,16 +765,22 @@ YUI.add('ez-locationmodel-tests', function (Y) {
...
                     }
                 }
-            };
-
+            });
+            Mock.expect(this.viewCreateStruct, {
+                method: 'setFilter',
+                args: [Mock.Value.Object],
+                run: Y.bind(function (arg) {
+                    this.viewCreateStruct.body.ViewInput.LocationQuery.Filter = arg;
```
dpobel: same as https://github.com/ezsystems/PlatformUIBundle/pull/712/files#diff-b1369eadf69b693372a3aedfb871e051

Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js
```diff
@@ -22,8 +37,15 @@ YUI.add('ez-searchplugin-tests', function (Y) {
             this.capi = new Mock();
             this.contentService = new Mock();
             this.viewName = 'REST-View-Name';
-            this.query = {body: {ViewInput: {ContentQuery: {}}}};
-
+            this.query = new Y.Mock({body: {ViewInput: {ContentQuery: {}}}});
```
dpobel: no need for the parameter

Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js
```diff
@@ -3,10 +3,25 @@
...
+        Mock.expect(context.query, {
+            method: queryMethodName,
+            args: [Mock.Value.Object],
+        });
+
+        Mock.expect(context.contentService, {
+            method: 'createView',
+            args: [context.query, Mock.Value.Function],
+            run: Y.bind(function () {
+                Y.Mock.verify(context.query);
```
dpobel: you should not do that here, you have no guarantee this will be called. You should do that at the end of each test.

Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js
```diff
@@ -22,8 +37,15 @@ YUI.add('ez-searchplugin-tests', function (Y) {
             this.capi = new Mock();
             this.contentService = new Mock();
             this.viewName = 'REST-View-Name';
-            this.query = {body: {ViewInput: {ContentQuery: {}}}};
-
+            this.query = new Y.Mock({body: {ViewInput: {ContentQuery: {}}}});
+            Mock.expect(this.query, {
+                method: 'setLimitAndOffset',
+                args: [Mock.Value.Any, Mock.Value.Any],
```
dpobel: this method is supposed to receive specific values not *Any*

Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js
```diff
@@ -22,8 +37,15 @@ YUI.add('ez-searchplugin-tests', function (Y) {
...
+            });
+            Mock.expect(this.query, {
+                method: 'setSortClauses',
+                args: [Mock.Value.Object],
```
dpobel: same here, you are expecting a very specific object

Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js
```diff
@@ -296,8 +316,16 @@ YUI.add('ez-searchplugin-tests', function (Y) {
             this.capi = new Mock();
             this.contentService = new Mock();
             this.viewName = 'REST-View-Name';
-            this.query = {body: {ViewInput: {LocationQuery: {}}}};
+            this.query = new Mock({body: {ViewInput: {LocationQuery: {}}}});
```
dpobel: same remarks as in `contentSearchTests`

Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js
```diff
@@ -567,8 +603,16 @@ YUI.add('ez-searchplugin-tests', function (Y) {
...
 
+            Mock.expect(this.query, {
+                method: 'setLimitAndOffset',
+                args: [Mock.Value.Any, Mock.Value.Any],
```
dpobel: and same as in the others tests

Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js
```diff
@@ -796,8 +840,8 @@ YUI.add('ez-searchplugin-tests', function (Y) {
             this.capi = new Mock();
             this.contentService = new Mock();
             this.viewName = 'REST-View-Name';
-            this.locationQuery = {body: {ViewInput: {LocationQuery: {}}}};
-            this.contentQuery = {body: {ViewInput: {ContentQuery: {}}}};
+            this.locationQuery = new Mock({body: {ViewInput: {LocationQuery: {}}}});
```
dpobel: no need for the parameter

Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js
```diff
@@ -904,6 +948,19 @@ YUI.add('ez-searchplugin-tests', function (Y) {
                     }
                 };
             };
+
+            Mock.expect(this.contentQuery, {
```
dpobel: those expectations are too weak

Resources/public/js/views/services/plugins/ez-searchplugin.js
```diff
@@ -59,13 +61,21 @@ YUI.add('ez-searchplugin', function (Y) {
...
-            query.body.ViewInput[type].offset = search.offset;
-            query.body.ViewInput[type].limit = search.limit;
+            if (search.query) {
+                query.setQuery(search.query);
+            }
+            if (search.filter) {
+                query.setFilter(search.filter);
+            }
+            if (search.criteria) {
+                query.setCriteria(search.criteria);
```
dpobel: you should also trigger a deprecated log

andrerom:
Note: Added todo's in description in order to make sure we update composer requirements once the rest client changes are merged and tagged for UI use.

StephaneDiot:
updated @dpobel

Tests/js/models/assets/ez-contentmodel-tests.js
```diff
@@ -820,15 +820,11 @@ YUI.add('ez-contentmodel-tests', function (Y) {
...
-                        'LocationQuery': {
-                            'Criteria' : ""
-                        }
-                    }
-                }
-            };
+            this.query = new Y.Mock();
+            Mock.expect(this.query, {
+                method: 'setFilter',
+                args: [Mock.Value.Object],
```
dpobel: this is too weak, you should also check that the passed object contains the key `ContentIdCriterion` with the content id as its value.

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -765,16 +765,11 @@ YUI.add('ez-locationmodel-tests', function (Y) {
...
-
-                        }
-                    }
-                }
-            };
-
+            this.viewCreateStruct = new Y.Mock();
+            Mock.expect(this.viewCreateStruct, {
+                method: 'setFilter',
+                args: [Mock.Value.Object],
```
dpobel: too weak as well, basically same thing as in ez-contentobject-tests.js

Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js
```diff
@@ -22,8 +34,23 @@ YUI.add('ez-searchplugin-tests', function (Y) {
             this.capi = new Mock();
             this.contentService = new Mock();
             this.viewName = 'REST-View-Name';
-            this.query = {body: {ViewInput: {ContentQuery: {}}}};
-
+            this.query = new Y.Mock();
+            this.sortClauses = {};
+            Mock.expect(this.query, {
+                method: 'setLimitAndOffset',
+                args: [Mock.Value.Number, Mock.Value.Number],
```
dpobel: this is still too weak, you don't expect any number but very specific values for limit and offset

Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js
```diff
@@ -22,8 +34,23 @@ YUI.add('ez-searchplugin-tests', function (Y) {
...
+            });
+            Mock.expect(this.query, {
+                method: 'setSortClauses',
+                args: [Mock.Value.Object],
```
dpobel: instead of having a `run` function with an assertion, you can directly put `this.sortClauses` here

Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js
```diff
@@ -111,7 +147,8 @@ YUI.add('ez-searchplugin-tests', function (Y) {
             this.view.fire('contentSearch', {
                 viewName: this.viewName,
                 search: {
-                    criteria: {},
+                    offset: 42,
```
dpobel: do you need to add `offset` and `limit` ? those parameters are documented as optional, so this should not be needed

Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js
```diff
@@ -296,8 +336,24 @@ YUI.add('ez-searchplugin-tests', function (Y) {
...
+            this.query = new Mock();
+            this.sortClauses = {};
 
+            Mock.expect(this.query, {
+                method: 'setLimitAndOffset',
+                args: [Mock.Value.Number, Mock.Value.Number],
+            });
+            Mock.expect(this.query, {
+                method: 'setSortClauses',
+                args: [Mock.Value.Object],
```
dpobel: same as previously, `this.sortClauses` can be used

Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js
```diff
@@ -296,8 +336,24 @@ YUI.add('ez-searchplugin-tests', function (Y) {
             this.capi = new Mock();
             this.contentService = new Mock();
             this.viewName = 'REST-View-Name';
-            this.query = {body: {ViewInput: {LocationQuery: {}}}};
+            this.query = new Mock();
+            this.sortClauses = {};
 
+            Mock.expect(this.query, {
+                method: 'setLimitAndOffset',
+                args: [Mock.Value.Number, Mock.Value.Number],
```
dpobel: too weak as well

Tests/js/views/services/plugins/assets/ez-searchplugin-tests.js
dpobel: ok I understand, with your patch offset and limit become mandatory. This is a BC break (most likely this breaks some parts of PlatformUI) so the solution is not to fix the unit tests but fix the code.

StephaneDiot: Well no, offset and limit aren't mandatory... But the query mocks now expect 2 args number now for their setLimitAndOffset method (that's why i had Y.Mock.Any before). So now I had to fill the tests with limit and offset.

dpobel: then it's completely useless and you fix your mock instead of changing each and every test

miguelcleverti:
PR approved by QA

PR #714, opened 2016-11-03
EZP-26430: Provide a shortcut to Y.Translator.trans()

yannickroger:
Link: https://jira.ez.no/browse/EZP-26430

## Description
This patch provides a shortcut `Y.eZ.trans` for `Y.eZ.Translator.trans` so it is shorter and provides a better developer experience. Note that the only supported way of adding translation strings will be to use the shortcut. Also the upcoming Translation Dumper (the other part of this story) will only detect the shortcut.

## Test
Manual and unit tests

### DIFF
--- a/Resources/public/js/external/ez-translator.js
+++ b/Resources/public/js/external/ez-translator.js
@@ -17,4 +17,15 @@ YUI.add('ez-translator', function (Y) {
      * @constructor
      */
     Y.eZ.Translator = Translator;
+
+    /**
+     * Shortcut to the Y.eZ.Translator.trans method
+     *
+     * @namespace eZ
+     * @method Y.eZ.trans
+     * @param String key message to be translated
+     * @param Object parameters to be applied to the message
+     * @param String domain where the translation can be found
+     */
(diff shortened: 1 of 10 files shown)


Resources/public/js/external/ez-translator.js
```diff
@@ -17,4 +17,12 @@ YUI.add('ez-translator', function (Y) {
      * @constructor
      */
     Y.eZ.Translator = Translator;
+
+    /**
+     * Shortcut the to Y.eZ.Translator.trans method
```
dpobel: typo

Resources/public/js/external/ez-translator.js
```diff
@@ -17,4 +17,12 @@ YUI.add('ez-translator', function (Y) {
...
+     *
+     * @namespace eZ
+     * @method Y.eZ.trans
```
dpobel: you should document the expected parameters

yannickroger:
Not sending this to QA as it is a technical cleanup.

PR #715, opened 2016-11-03
EZP-26546: ez-navigationitemsubtreeview route matcher should only match the full id

miguelcleverti:
The route matcher for ez-navigationitemsubtreeview is matching the incomplete id.
Example for location id 1/57 location 1/5 is also matched.

By adding a delimiter('/') we guarantee that only the full id is match and also the subitems.


dpobel:
also this patch should be applied to 1.5 first

miguelcleverti:
@dpobel, @yannickroger rebased against 1.5 and commit messaged updated with ezp number

PR #672, opened 2016-09-06
EZP-26240: Update editing UI with server side validation

StephaneDiot:
jira : https://jira.ez.no/browse/EZP-26240
## Description

The goal here was to be able to give the field edit views the error messages coming from the server side. Currently to see those "server side error messages" you must overpass the application validation so you have to disable it to see the feature.

## ScreenCast

https://youtu.be/8sVKlcpDBcY

## Tests

Manual and unit tests

PR #717, opened 2016-11-04
EZP-26543: Infinite loading when publishing/saving/previewing an unallowed content

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26543

# Description

This patch fixes a regression introduced in [EZP-26240](https://jira.ez.no/browse/EZP-26240) / https://github.com/ezsystems/PlatformUIBundle/pull/672. Basically when a draft saving fails, the failure does not necessary come from a validation error on fields, as a result the error response might not contain any field error detail so this patch just adds a check to avoid a JavaScript error.

# Tests

manual tests + unit tests


dpobel:

PR #719, opened 2016-11-04
Implemented EZP-26428: Extensions can register their translation domains

bdunogier:
> Implements [EZP-26428](http://jira.ez.no/browse/EZP-26428)

Any bundle that extends PlatformUI and wants to register custom translation domains can expose the translation domains it uses in templates.

To do so, one must make the the Symfony Container Extension from a bundle implement `EzSystems\PlatformUIBundle\DependencyInjection\PlatformUIExtension`. Then implement the `getTranslationDomains` method in the extension, making it return an array of domains (ex: "return ['domain_one', 'domain_two']).

```php
// src/MyBundle/DependencyInjection/MyExtension.php
namespace MyBundle\DependencyInjection;

use EzSystems\PlatformUIBundle\DependencyInjection\PlatformUIExtension;

class MyExtension extends Extension implements PlatformUIExtension
{
    public function getTranslationDomains()
    {
        return ['mydomain', 'myotherdomain']
    }
}
```

The list of domains will be collected, in the bundle's registration order, merged into one list, and passed on to PlatformUI using an application config provider.

### Testing done
Compiler pass is unit tested. Templating / rendering is manually tested.

### DIFF
--- a/DependencyInjection/Compiler/TranslationDomainsExtensionsPass.php
+++ b/DependencyInjection/Compiler/TranslationDomainsExtensionsPass.php
@@ -1,0 +1,24 @@
+<?php
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\PlatformUIBundle\DependencyInjection\Compiler;
+
+use EzSystems\PlatformUIBundle\DependencyInjection\PlatformUIExtension;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class TranslationDomainsExtensionsPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        $domains = [];
+        foreach ($container->getExtensions() as $extension) {
+            if ($extension instanceof PlatformUIExtension) {
+                $domains = array_merge($domains, $extension->getTranslationDomains());
+            }
+        }
+        $container->setParameter('ez_platformui.translation_domains', $domains);
+    }
+}
--- a/DependencyInjection/EzPlatformUIExtension.php
+++ b/DependencyInjection/EzPlatformUIExtension.php
@@ -18,7 +18,7 @@ use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 use Symfony\Component\DependencyInjection\Loader;
 use Symfony\Component\Yaml\Yaml;
 
-class EzPlatformUIExtension extends Extension implements PrependExtensionInterface
+class EzPlatformUIExtension extends Extension implements PrependExtensionInterface, PlatformUIExtension
 {
     public function getAlias()
     {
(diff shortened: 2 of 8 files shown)


composer.json
```diff
@@ -13,11 +13,12 @@
         "ezsystems/ezpublish-kernel": "^6.6",
         "ezsystems/ez-support-tools": "~0.1.0",
         "zetacomponents/system-information": "^1.1",
-        "willdurand/js-translation-bundle": "^2.5"
+        "willdurand/js-translation-bundle": "^2.5",
+        "mockery/mockery": "^0.9.5"
```
dpobel: shouldn't this be in require-dev ?

DependencyInjection/Compiler/TranslationDomainsExtensionsPass.php
```diff
@@ -0,0 +1,27 @@
...
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class TranslationDomainsExtensionsPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        $domains = [];
+        foreach ($container->getExtensions() as $extension) {
+            if ($extension instanceof PlatformUIExtension) {
+                if (isset($hasDuplicates)) {
```
dpobel: seems to be a leftover of a previous version, `$hasDuplicates` is never set

bdunogier:
Ping @nicolas-bastien.

composer.json
bdunogier: Yep.

DependencyInjection/Compiler/TranslationDomainsExtensionsPass.php
bdunogier: It was actually a (kind of) todo. Do you think we should have duplicates handling ? If we don't want to, we can just ditch the `isset`.

bdunogier:
> Could you also update shell.html.twig to use the new translationDomains parameter ?

Done. The only thing left is the duplicate check: do we want/need to check for duplicate domain declarations or not ?

DependencyInjection/Compiler/TranslationDomainsExtensionsPass.php
dpobel: I think for now we don't need to handle duplicates. In EZP-26556 we'll rename our own domains with a prefix and we should document clearly that extension developers should do the same. Hopefully this should be enough to avoid duplicated domains.

bdunogier:
Removed the duplicate check let's see if Travis agrees :)

bdunogier:
And it's green.

Ready to send to QA @dpobel ?

DependencyInjection/EzPlatformUIExtension.php
```diff
@@ -73,4 +73,17 @@ private function prependImageVariations(ContainerBuilder $container)
...
     }
+
+    /**
+     * Returns the translation domains used by the extension.
+     * @return array An array of extensions
+     */
+    public function getTranslationDomains()
+    {
+        return [
+            'bar', 'confirm', 'contentedit', 'contenttypeselector', 'dashboardblocks', 'editorial',
```
yannickroger: Could you add these. In a yml file ?

bdunogier: Consensus: document how these could be loaded from a YAML in the translatable extensions documentation.

PR #720, opened 2016-11-07
EZP-26552: Use the languages provided by the browser to translate the UI

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26552

# Description

This patch changes the way we pick the language(s) to translate the UI. With these changes, we'll try to translate the interface in the languages provided by the user's browser. This affects only the part generated in the browser. The server side generated content will need to adopt a similar strategy.

## Tasks

* [x] Application config provider for the languages
* [x] Change Y.eZ(.Translator).trans to handle the fallback
* [x] Load translations and initialize the Translator

# Tests

manual tests and unit tests


dpobel:

Resources/public/js/services/ez-translator.js
```diff
@@ -0,0 +1,69 @@
...
+     */
+    Y.eZ.Translator.setPreferredLanguages = function (languages) {
+        Translator.fallback = ''; // disable the Translator fallback strategy
+        preferredLanguages = languages;
+    };
+
+    /**
+     * Translates the given key from the domain with the given parameters.
+     *
+     * ** /!\ Use the shortcut `Y.eZ.trans` so that you translation keys are
```
yannickroger: Nitpick : you

PR #722, opened 2016-11-08
EZP-25552: Finder method view

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-25552

## Description

First step of the UDW explorer : define a finder view having a  selected view.

## Tests

Units and manual tests

### DIFF
--- a/Resources/config/yui.yml
+++ b/Resources/config/yui.yml
@@ -529,8 +529,8 @@ system:
                         - 'ez-tabs'
                         - 'ez-templatebasedview'
                         - 'universaldiscoveryview-ez-template'
-                        - 'ez-universaldiscoverybrowseview'
                         - 'ez-universaldiscoverysearchview'
+                        - 'ez-universaldiscoveryfinderview'
                         - 'ez-universaldiscoveryconfirmedlistview'
                         - 'array-extras'
                         - 'event-tap'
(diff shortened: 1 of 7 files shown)


Resources/public/js/views/universaldiscovery/ez-universaldiscoverybrowseview.js
```diff
@@ -22,6 +22,7 @@ YUI.add('ez-universaldiscoverybrowseview', function (Y) {
      */
     Y.eZ.UniversalDiscoveryBrowseView = Y.Base.create('universalDiscoveryBrowseView', Y.eZ.UniversalDiscoveryMethodBaseView, [], {
         initializer: function () {
+            console.warn('UniversalDiscoveryBrowseView is deprecated');
```
dpobel: this is not the *standard* message syntax for the deprecated notice (we also use `console.log`). Also, since the complete component is deprecated, this should probably be triggered directly when the module is loaded.

Resources/config/yui.yml
```diff
@@ -531,6 +531,7 @@ system:
                         - 'universaldiscoveryview-ez-template'
                         - 'ez-universaldiscoverybrowseview'
```
dpobel: this one should be removed

Resources/public/js/views/ez-universaldiscoveryview.js
```diff
@@ -122,7 +122,7 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
         },
 
         /**
-         * Notifies the browse method views that a content is removed from the
+         * Notifies the finder method views that a content is removed from the
```
dpobel: actually, the method notifies all method views

dpobel:
(don't forget to merge the commits :))

Resources/public/js/views/universaldiscovery/ez-universaldiscoveryfinderview.js
```diff
@@ -0,0 +1,161 @@
...
+            });
+        },
+    }, {
+        ATTRS: {
+            /**
+             * @attribute title
+             * @default 'Browse'
+             */
+            title: {
+                value: 'Browse',
```
dpobel: this should be translated with Y.eZ.trans

PR #723, opened 2016-11-08
EZP-24695: Increase tree view limit to 100 & avoid loading same types in parallel

andrerom:
> issue: https://jira.ez.no/browse/EZP-24695

1. As agreed with PM, we will the limit of current tree menu to 100 for now until the real
paging support is added sometime next year. Thus reduce likelihood this is hitting
people.

2. However with limit increased, the content type always re-loading issue gets amplified, so make sure code takes care about only loading each and every content type value once. _This should solve the content type loading issue ok for now, and we can rather improve uppon it with more caching etc later._ e.g. The extreme case is reduction of backend requests from 25 to 2 on Projects folder in tree menu with platform demo data.

Todo:
- [x] Verify tests
- [x] depending on complexity consider rebasing for 1.5 branch

### DIFF
--- a/Resources/public/js/views/services/plugins/ez-searchplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-searchplugin.js
@@ -326,32 +326,49 @@ YUI.add('ez-searchplugin', function (Y) {
          *
          * @method _loadContentTypeForLocations
          * @protected
-         * @param {Array|Null} locationStructArr
+         * @param {Array|Null} structArray
+         * @param {Function} getContentTypeIdFunc
          * @param {Function} callback
          */
         _loadContentTypeForStruct: function (structArray, getContentTypeIdFunc, callback) {
             var tasks = new Y.Parallel(),
-                loadContentTypeError = false;
+                loadContentTypeError = false,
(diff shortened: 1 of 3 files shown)


andrerom:
Tests seems to be passing, ready for review @dpobel @yannickroger @StephaneDiot

Resources/public/js/views/services/plugins/ez-searchplugin.js
```diff
@@ -326,32 +326,48 @@ YUI.add('ez-searchplugin', function (Y) {
...
          * @param {Function} callback
          */
         _loadContentTypeForStruct: function (structArray, getContentTypeIdFunc, callback) {
             var tasks = new Y.Parallel(),
-                loadContentTypeError = false;
+                loadContentTypeError = false,
+                contentTypList = [];
+
+            // Get list of content types ids
+            var contentTypeIds = Y.Array.map(structArray, function (struct, i) {
```
dpobel: CS: the variable should declared in the `var` block above (one `var` block per function)

andrerom:
_Sent to QA_

PR #724, opened 2016-11-16
EZP-26322: Provide a "preview" link in the UDW selected view

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26322

# Description

Screencast:
[](http://www.youtube.com/watch?v=PbYbJfylW6s "Click to play on Youtube.com")

This patch adds the ability to view the fields of a Content that is selected in the Universal Discovery Widget. This is done with a side view call "Content Peek" that uses the RawContentView to display the fields.

To keep a reasonable patch size, it's not yet possible to from one language to another but that will be implemented in [EZP-26613](https://jira.ez.no/browse/EZP-26613). Also, at the moment, the Relation and Relations field are not correctly rendered in such context, this will be fixed in [EZP-26614](https://jira.ez.no/browse/EZP-26614).

## Tasks

* [x] Improve the side view system so that a side view can be rerendered each time it becomes active
* [x] Change RawContentView and LanguageSwitcherView to not require any parameter
* [x] Implement a ContentPeek side view able to display the fields of a Content
* [x] ~~Allow to switch language in the ContentPeekView~~ Postponed to [EZP-26613](https://jira.ez.no/browse/EZP-26613)
* [x] Make sure all fields are correctly rendered (Relation(s) not working [EZP-26614](https://jira.ez.no/browse/EZP-26614))

# Tests

unit tests + manual tests

PR #725, opened 2016-11-17
EZP-26614: Relation and Relations fields not rendered in the Content Peek View

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26614

# Description

Fixes the issue found in https://github.com/ezsystems/PlatformUIBundle/pull/724 about Relation and Relations field rendering in the new Content Peek view.

Basically, this patch changes the events used to load related Content items to now require a source Content item instead of relying on the object referenced by the `content` attribute of the view service. This allows the plugins handling those events to work with view services not having the `content` attribute. This also implies injecting the Content item to the field views and while at it I also inject the Content Type  to be consistent with the field edit views.

# Tests

unit test + manual tests

### DIFF
--- a/Resources/public/js/views/ez-fieldview.js
+++ b/Resources/public/js/views/ez-fieldview.js
@@ -120,6 +120,22 @@ YUI.add('ez-fieldview', function (Y) {
              * @type {}
              */
             field: {},
+
+            /**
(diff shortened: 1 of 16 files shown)


dpobel:

PR #726, opened 2016-11-18
EZP-26613: Allow to switch language in the Content Peek View

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26613

# Description

This story is about providing a way for the user to switch from one language to another while viewing the fields of a Content from the Universal Discovery Widget.

While working on that, I realized I did a design mistake regarding the representation of Content items and Versions when they are translated. I chose to have one Content model object (or one Version model object) per language with the corresponding fields. That's why, when loading a Content item or a Version you can pass a language code so that the REST API only returns the fields in the given language. But you can't pass such option when searching for Content, as a result if in such case the REST response contains the fields in every languages in which the Content item is translated, so the REST model parsing is completely unreliable and you might have the fields in any languages (depending on their order in the REST Response). So most of this patch (the 17 first commits) is about changing that so one Content model object (and one Version model object) represents the Content item (or the Version) in its available translations unless it was loaded with a language code restriction (which still works).

The rest of the patch is really about having the ability to change switch from one language to the other. By default, the fields in the *main language* are displayed.

Screencast: 
[](http://www.youtube.com/watch?v=6Gg50SYX8uk "Click to play on Youtube.com")

## Tasks

* [x] Make `Content.get('fields')` a shortcut to `Content.get('currentVersion').get('fields')`
* [x] Change `Content#getField` to call `Version#getField`
* [x] Stores the fields in several languages
* [x] Add a `language` parameter to `Content#getField` to call `Version#getField`
* [x] Deprecate `Content.get(fields')` and `Version.get('fields')`
* [x] Change `Content#getFieldsOfType`
* [x] Change existing code to use new fields handling
* [ ] ~~Remove the `languageCode` parameter when loading a Content or a Version~~ Postponed to [EZP-26656](https://jira.ez.no/browse/EZP-26656)
* [x] Switch language in the Content Peek View


dpobel:

@sunpietro given this changes the core Content and Version models, could you test this patch with Studio, normally this should not break anything but most likely you'll get some deprecation log in the console.

sunpietro:
@dpobel I'll check it

sunpietro:
@dpobel I get error like the following:
zrzut ekranu 2016-11-22 o 14 31 51
The error doesn't appear on master.
I'm not sure what is it related to. As debugger information leads to YUI files that are concatenated.

dpobel:
@sunpietro without the stack trace, it's hard to guess what is going on, could you please provide it ?

sunpietro:
@dpobel I have just finished quick PR re-testing to get the issues I mentioned earlier, but they're gone. 
So 👍 from me :)

PR #721, opened 2016-11-07
EZP-26430: Translation dumper to detect strings to translate in .js

yannickroger:
Link: https://jira.ez.no/browse/EZP-26430

## Description
Some translations strings are located in JS files. This PR provides a script to automatically detect them and add them to the symfony catalog.

This extractor relies on nodejs to parse the javascript files.

Note: Translation strings using variables parameters are not detected at the moment.

## Tests
Manual and unit tests.

PR #727, opened 2016-11-18
Made PlatformUI js & hbt extractors usable on any bundle

bdunogier:
Changes the extractors so that they use the bundle provided as an argument to the update script. Makes files from any bundle extractible.

Note: requires that the bundle has a `Resources/views` directory, even if it is empty.

Also changes the Javascript extractor to execute node using the Sf Process component.


bdunogier:
Ping @nicolas-bastien :)

.travis.yml
```diff
@@ -30,6 +30,7 @@ branches:
     only:
         - master
         - /^\d.\d+$/
+        - ezp-26430-js_translator_dumper-improvements
```
nicolas-bastien: I think we can add an expression here like : /^ezp-.*$/

yannickroger: This commit will not be merged, it's just because this PR is not based on master but on a custom branch.

Tests/fixtures/extractor/with_translation.js
```diff
@@ -1,6 +0,0 @@
-//Somme dummy javascript code with some translation in it.
-
-var result1 = Y.eZ.trans('test.translation.result1', {}, 'testdomain'),
```
yannickroger: Did you forget to add the new folder while commiting?

bdunogier: Little bit.

.travis.yml
bdunogier: Temporary commit, meant to be removed.

Translation/PlatformUIExtractor.php
```diff
@@ -0,0 +1,48 @@
...
+    protected function updateResource(&$resource, $path)
+    {
+        if (is_string($resource) && strpos($resource, $this->pathSuffix) !== null) {
+            $resource = $this->replacePath($resource, $path);
+        } elseif (is_array($resource)) {
+            $resource = array_map(
+                function ($value) use ($path) {
+                    if (is_string($value) && strpos($value, $this->pathSuffix) !== null) {
+                        return $this->replacePath($value, $path);
+                    } else {
```
nicolas-bastien: here else is not really needed

Translation/PlatformUIExtractor.php
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Updates the resource by replacing the path to templates (Resource/views)
+     * with the path handled by the extractor.
+     *
+     * @param string $resource
+     * @param string $path The path 'Resources/views' should be replaced with
+     */
+    protected function updateResource(&$resource, $path)
+    {
+        if (is_string($resource) && strpos($resource, $this->pathSuffix) !== null) {
```
nicolas-bastien: Here I would have centralized the test in replacePath so this code would be easier.

```
protected function updateResource(&$resource, $path)
{
    if (is_array($resource)) {
        $resource = array_map(
            function ($value) use ($path) {
               return $this->replacePath($value, $path);
           },
           $resource
        );

    $resource = $this->replacePath($resource, $path);
}
```


```php
private function replacePath(&$in, $path)
{
    if (!is_string($value) || strpos($value, $this->pathSuffix) === null) {
         return $resource;
    }

    return str_replace($this->pathSuffix, $path, $in);
}
```

bdunogier: I'd keep the `is_array`/`is_string` test in `updateResource`, but move the `strpos` one to `replacePath`. Then we would have the best of both worlds, I think. I'll write it down and see how it looks like (moving this to #721)

PR #729, opened 2016-11-24
EZP-26669: Align the JavaScript translate strategy on the Symfony one

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26669

# Description

With this patch, only the first language provided by the browser is considered and if it's something like `language_COUNTRY`, the `language` alone is also added to the list before the configured default locale.

# Tests

manual tests + unit tests


dpobel:
ping @bdunogier @yannickroger

dpobel:
@nicolas-bastien as well :)

PR #730, opened 2016-11-24
EZP-26600: Use the request's languages to translate pjax

bdunogier:
> Implements [EZP-26600](http://jira.ez.no/browse/EZP-26600)

Adds a Request Subscriber that runs before our / symfony's LocaleListener, and sets the request's locale for PJAX requests.

### TODO
- [x] Unit test

### DIFF
--- a/EventListener/PjaxBrowserLanguageSubscriber.php
+++ b/EventListener/PjaxBrowserLanguageSubscriber.php
@@ -1,0 +1,38 @@
+<?php
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\PlatformUIBundle\EventListener;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * On PJAX requests, sets the request's locale from the browser's accept-language header.
+ */
+class PjaxBrowserLanguageSubscriber implements EventSubscriberInterface
+{
+    public static function getSubscribedEvents()
+    {
+        return [
+            KernelEvents::REQUEST => [
+                // priority needs to be higher than the CoreBundle's LocaleListener
+                ['setPjaxRequestLocale', 50],
+            ],
+        ];
+    }
+
+    public function setPjaxRequestLocale(GetResponseEvent $event)
+    {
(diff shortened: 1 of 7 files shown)


EventListener/PjaxBrowserLanguageSubscriber.php
```diff
@@ -0,0 +1,38 @@
...
+                ['setPjaxRequestLocale', 50],
+            ],
+        ];
+    }
+
+    public function setPjaxRequestLocale(GetResponseEvent $event)
+    {
+        $request = $event->getRequest();
+
+        if (!$event->isMasterRequest() || !$request->headers->has('x-pjax')) {
```
nicolas-bastien: x-pjax is lowercase here and uppercase in your test, don't we have constant for that ?

EventListener/PjaxBrowserLanguageSubscriber.php
```diff
@@ -0,0 +1,38 @@
...
+namespace EzSystems\PlatformUIBundle\EventListener;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * On PJAX requests, sets the request's locale from the browser's accept-language header.
+ */
+class PjaxBrowserLanguageSubscriber implements EventSubscriberInterface
```
nicolas-bastien: In general I would remove the Pjax to something more generic like LocalSubscriber so it would work even if it's not pjax.

And if in the future we have to has a condition based on user settings then it's not only brower languages

bdunogier: The first name I chose was actually based on Locale something.

But given that this implementation/class is meant to act only on Pjax requests, and tests for it, I'd actually keep this name.

I don't _disagree_, so I'm gonna cut the apple in half: `RequestLocaleSubscriber::setPjaxRequestLocale()`. In addition, I have added a `PjaxRequestMatcher`, that has the pjax request matching logic.

EventListener/PjaxBrowserLanguageSubscriber.php
bdunogier: Nope, we do not. As said, while I don't disagree with the usage of constants for this, I don't have a good place for it at the moment. It is also used in another listener, and therefore can't go into this listener. The PJAX controllers do not share any common class that we could use for this.

Would you create an interface or class for this ? I'd suggest that we postpone this for when we will work further on PJAX, and find a suitable place for the constant at this point. What do you think ?

PR #731, opened 2016-11-25
EZP-25348: Allow to transform paragraph into heading and heading into paragraph

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25348 (+ fix for https://jira.ez.no/browse/EZP-26667)

# Description

Screencast:

[](http://www.youtube.com/watch?v=e22rng7PoaA "Click to play on Youtube.com")

This allows the editor to transform a heading into a paragraph and a paragraph into a heading as well. While working on that, I also fixed EZP-26667 which happens to be larger than expected as it does not only affect the *add toolbar* but also several buttons (hence the size of the patch).

# Tests

manual test + unit test


andrerom:
👌

dpobel:

PR #728, opened 2016-11-23
EZP-26662: Implemented ContentType view markup

bdunogier:
> [EZP-26662](https://jira.ez.no/browse/EZP-26662)
> Work in progress

PR #732, opened 2016-11-25
EZP-26662: Content type view data and design

nicolas-bastien:
Against PR : #728

Depends on : https://github.com/ezsystems/ezpublish-kernel/pull/1832

https://jira.ez.no/browse/EZP-26662

Here is what it looks like : 
ezp26662-content-view

### DIFF
--- a/Resources/public/css/theme/views/serverside.css
+++ b/Resources/public/css/theme/views/serverside.css
@@ -2,7 +2,6 @@
  * Copyright (C) eZ Systems AS. All rights reserved.
  * For full copyright and license information view LICENSE file distributed with this source code.
  */
-
 .ez-view-serversideview {
     background: #F5F4F2;
 }
@@ -13,4 +12,57 @@
 
 .ez-view-serversideview .ez-tabs.ez-serverside-content {
     border: 0;
+}
+
+.ez-view-serversideview .ez-contenttype-view {
+    background: #FAFAFA;
+}
+
+.ez-view-serversideview .ez-contenttype-view .ez-contenttype-title,
+.ez-view-serversideview .ez-contenttype-view .ez-contenttype-subtitle {
+    background: transparent;
+    color: #2b84b1;
+    border: none;
+    border-bottom: 1px solid #2b84b1;
+}
+
+.ez-view-serversideview .ez-contenttype-view .ez-contenttype-metadata {
+    background: #f3f3f3;
+}
+
+.ez-view-serversideview .ez-contenttype-view .ez-contenttype-metadata-name {
+    background: #ededed;
+    font-weight: bold;
+}
+
+.ez-view-serversideview .ez-contenttype-view .ez-contenttype-metadata-help {
+    font-weight: normal;
+    font-style: italic;
+    font-size: 0.7em;
+}
+
+.ez-view-serversideview .ez-contenttype-view .ez-field-definition-header {
+    background: #c2c2c2;
+}
+
+.ez-view-serversideview .ez-contenttype-view .ez-field-definition-main-properties {
+    border-top: 1px solid #c2c2c2;
+}
+
+.ez-view-serversideview .ez-contenttype-view .ez-field-definition-property-name {
+    background: #ededed;
+    font-weight: bold;
+}
+
+.ez-view-serversideview .ez-contenttype-view .ez-field-definition-property-value {
+    background: #f3f3f3;
+}
+
+.ez-view-serversideview .ez-contenttype-view .ez-fielddefinition-setting-name {
+    background: #ededed;
+    font-weight: bold;
+}
+
+.ez-view-serversideview .ez-contenttype-view .ez-fielddefinition-setting-value {
+    background: #f3f3f3;
 }
--- a/Resources/public/css/views/serverside.css
+++ b/Resources/public/css/views/serverside.css
@@ -2,7 +2,102 @@
  * Copyright (C) eZ Systems AS. All rights reserved.
  * For full copyright and license information view LICENSE file distributed with this source code.
  */
-
 .ez-view-serversideview .ez-platform-version {
     margin-top: 0;
+}
+
+.ez-view-serversideview .ez-contenttype-icon:before {
+    padding-right: 0.3em;
+}
+
+.ez-view-serversideview .ez-serverside-content {
+    padding-top: 0;
+}
+
+.ez-view-serversideview .ez-contenttype-view {
+    padding: 0.5em 1em;
+}
+
+.ez-view-serversideview .ez-field-definition-group {
+    padding-left: 2.5em;
+}
+
+.ez-view-serversideview .ez-contenttype-title {
+    padding: 0.8em 0;
+}
+
+.ez-view-serversideview .ez-contenttype-subtitle {
+    padding: 0.5em 0;
+}
+
+.ez-view-serversideview .ez-contenttype-metadata {
+    display: flex;
+}
+
+.ez-view-serversideview .ez-contenttype-metadata-name,
+.ez-view-serversideview .ez-contenttype-metadata-value {
+    padding: 0.8em 1em;
+}
+
+.ez-view-serversideview .ez-contenttype-metadata-name {
+    width: 250px;
+}
+
+.ez-view-serversideview .ez-field-definitions {
+    width: 100%;
+    overflow: hidden;
+}
+
+.ez-view-serversideview .ez-field-definition-header {
+    padding: 1em;
+    text-align: left;
+}
+
+.ez-view-serversideview .ez-field-definition-main-properties {
+    padding: 0.8em 1em;
+    text-align: left;
+}
+
+.ez-view-serversideview .ez-field-definition-properties {
+    width: 100%;
+    list-style-type: none;
+    padding-left: 0;
+}
+
+.ez-view-serversideview .ez-field-definition-property {
+    display: flex;
+}
+
+.ez-view-serversideview .ez-field-definition-property-name {
+    width: 200px;
+    padding: 0.8em 1em;
+}
+
+.ez-view-serversideview .ez-field-definition-property-value {
+    padding: 0.8em 1em;
+    width: 100%;
+}
+
+.ez-view-serversideview .ez-fielddefinition-settings {
+    width: 100%;
+    list-style-type: none;
+    padding-left: 0;
+}
+
+.ez-view-serversideview .ez-fielddefinition-setting {
+    display: flex;
+}
+
+.ez-view-serversideview .ez-fielddefinition-setting-name {
+    width: 200px;
+    padding: 0.8em 1em;
+}
+
+.ez-view-serversideview .ez-fielddefinition-setting-value {
+    padding: 0.8em 1em;
+    width: 100%;
+}
+
+.ez-view-serversideview .ez-toolbar .ez-toolbar-action {
+    display: inline;
 }
--- a/Resources/translations/content_type.en.xlf
+++ b/Resources/translations/content_type.en.xlf
@@ -1,28 +1,22 @@
 <?xml version="1.0" encoding="utf-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:jms="urn:jms:translation" version="1.2">
-  <file date="2016-11-29T11:49:06Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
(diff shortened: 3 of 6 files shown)


nicolas-bastien:
@dpobel what do you think about the style ?

Resources/public/css/views/serverside.css
```diff
@@ -2,7 +2,97 @@
  * Copyright (C) eZ Systems AS. All rights reserved.
  * For full copyright and license information view LICENSE file distributed with this source code.
  */
-
 .ez-view-serversideview .ez-platform-version {
     margin-top: 0;
 }
+
+.ez-view-serversideview .ez-serverside-content {
+    padding-top: 0px;
```
dpobel: you can omit the unit

Resources/public/css/views/serverside.css
```diff
@@ -2,7 +2,97 @@
...
+}
+
+.ez-view-serversideview .ez-serverside-content-inner {
+    padding: 0.5em 1em;
+}
+
+.ez-view-serversideview .ez-serverside-content-inner > section > section {
```
dpobel: a tad too specific (in addition to the inner class name already mentioned below), so you should add a class on the section you want to style (ez-field-definition-group ?), so this would give:
```css
.ez-view-serversideview .ez-field-definition-group {
    /* ... */
}
```
.ez-serverside-content-inner is not needed if the class name is specific enough (hence why it's important to respect the domain)

Resources/public/css/views/serverside.css
```diff
@@ -2,7 +2,97 @@
...
+
+.ez-view-serversideview .ez-serverside-content-inner .ez-section-subtitle-name {
+    padding: 0.5em 0;
+}
+
+.ez-view-serversideview .ez-serverside-content-inner .ez-contenttype-metadata {
+    display: flex;
+}
+
+.ez-view-serversideview .ez-serverside-content-inner .ez-contenttype-metadata > * {
```
dpobel: we try to avoid the `*` selector, since ez-contenttype-metadata is supposed to contain ez-contenttype-metadata-name and ez-contenttype-metadata-value, this one is easy to expand :)

Resources/views/ContentType/view_content_type.html.twig
```diff
@@ -19,130 +19,143 @@
 {% endblock %}
 
 {% block header_title %}
-    <h1 class="ez-page-header-name ez-contenttype-icon ez-contenttype-icon-{{ content_type.identifier }}">{{ content_type_name }}</h1>
+    <h1 class="ez-page-header-name ez-contenttype-icon ez-contenttype-icon-{{ content_type.identifier }}" data-icon="">{{ content_type_name }}</h1>
```
nicolas-bastien: here I did that at the beginnig to have the margin but I can move that on ez-contenttype-icon ?

Resources/views/ContentType/view_content_type.html.twig
```diff
@@ -19,130 +19,143 @@
...
+                {{
+                    "content_type.last_modified"|trans({
+                        "%date%": content_type.modificationDate|localizeddate(locale=app.request.locale),
+                        "%modifier%": modifier.login
+                    })
+                }}
+            </p>
+
+            <section>
+                <h2 class="ez-section-title-name">{{ 'content_type.metadata.title'|trans}}</h2>
```
bdunogier: `ez-contenttype-title` ?

Resources/views/ContentType/view_content_type.html.twig
```diff
@@ -19,130 +19,143 @@
...
+                                            <li class="field-definition-data">
+                                                <div class="field-definition-data-name">{{ 'field_definition.searchable'|trans }}</div>
+                                                <div class="field-definition-data-value">{{ 'yes_no'|transchoice(field_definition.isSearchable) }}</div>
+                                            </li>
+                                            <li class="field-definition-data">
+                                                <div class="field-definition-data-name">{{ 'field_definition.translatable'|trans }}</div>
+                                                <div class="field-definition-data-value">{{ 'yes_no'|transchoice(field_definition.isTranslatable) }}</div>
+                                            </li>
+                                        </ul>
+                                        <span class="field-definition-data-title">{{ 'field_definition.specific_data_list.title'|trans }}</span>
```
bdunogier: I personally wouldn't bother with a title here (told it to Supriya already).

Resources/translations/content_type.en.xlf
```diff
@@ -145,6 +145,70 @@
         <source>content_type.is_in_use</source>
         <target>The content type cannot be deleted because it has content items, or because you don't have permission.</target>
       </trans-unit>
+      <trans-unit id="c06173c2f83c7d3b1db6b93751eac959" resname="content_type.metadata.title">
+        <source>content_type.metadata.title</source>
+        <target>Global properties</target>
```
bdunogier: We should use the "desc" function from the JMS translation bundle to set the english strings, and use the extractor to generate the xlf file. Keep 'em like this for now, but this part interacts with the work on the crowdin integration.

Resources/views/ContentType/view_content_type.html.twig
```diff
@@ -19,130 +19,143 @@
...
+                    <div class="ez-contenttype-metafata-value">{{ 'content_type.default_availability.value'|transchoice(content_type.defaultAlwaysAvailable) }}</div>
+                </div>
+            </section>
+
+            <section>
+                <h2 class="ez-section-title-name">{{ 'content_type.content.title'|trans }}</h2>
+
+                {% for group, field_definitions in fielddefinitions_by_group %}
+                    <section>
+                        <h3 class="ez-section-subtitle-name">{{ group|trans([], 'ezrepoforms_content_type') }}</h3>
```
bdunogier: Yep, you need to use the context @dpobel gave you.

nicolas-bastien:
@dpobel @bdunogier 
So I reworked all to fit ez standard I guess.
I removed the field settings titles + rename the section class

Resources/views/ContentType/view_content_type.html.twig
nicolas-bastien: ok + for subtitles

Resources/views/ContentType/view_content_type.html.twig
nicolas-bastien: don't fell so confortable to modify your code here I guess it's ok like that ?

Resources/public/css/theme/views/serverside.css
```diff
@@ -14,3 +13,56 @@
...
+
+.ez-view-serversideview .ez-contenttype-view .ez-field-definitions .ez-field-definition-header {
+    background: #c2c2c2;
+}
+
+.ez-view-serversideview .ez-contenttype-view .ez-field-definitions .ez-field-definition-main-properties {
+    border-top: 1px solid #c2c2c2;
+}
+
+.ez-view-serversideview .ez-contenttype-view .ez-field-definitions .ez-field-definition-properties .ez-field-definition-property .ez-field-definition-property-name {
```
dpobel: you don't need to be that specific, I think you can omit `.ez-field-definitions .ez-field-definition-properties .ez-field-definition-property` (and more or less the same can be applied to others selector)

Resources/translations/content_type.en.xlf
```diff
@@ -123,7 +123,7 @@
       </trans-unit>
       <trans-unit id="9297b47a2cead7403adaab0f0ee6fdb2" resname="content_type.default_children_sorting">
         <source>content_type.default_children_sorting</source>
-        <target>Default children sorting</target>
+        <target>Default field for sorting children</target>
```
dpobel: *field* is wrong here, *property* maybe ?

Resources/views/ContentType/view_content_type.html.twig
dpobel: let's talk about that :)

Resources/translations/content_type.en.xlf
nicolas-bastien: synchronize this with edit form

nicolas-bastien:
@dpobel done

nicolas-bastien:
@dpobel squash done travis green

nicolas-bastien:
@bdunogier sync with jms translations

bdunogier:
Looks good to me.

Resources/views/ContentType/view_content_type.html.twig
```diff
@@ -10,99 +10,151 @@
...
+                    <div class="ez-contenttype-metadata-name">{{ "content_type.identifier"|trans|desc("Identifier") }}</div>
+                    <div class="ez-contenttype-metadata-value">{{ content_type.identifier }}</div>
                 </div>
-
-                <div>
-                    <h6>{{ "content_type.description"|trans }}:</h6>
-                    {{ content_type_description }}
+                <div class="ez-contenttype-metadata">
+                    <div class="ez-contenttype-metadata-name">{{ "content_type.description"|trans|desc("Description") }}</div>
+                    <div class="ez-contenttype-metadata-value">{{ content_type_description }}</div>
```
bdunogier: Even if it was done like that before, maybe we should get rid of the `{{ set }}` calls above. In my opinion, the code will be, I think, more readable with the actual property / method next to the markup.

Furthermore, given that we now use many more variables in the template, there are no real reasons why we'd assign individual variables to some strings, are there ?

bdunogier: We need to cover the bug with the `description` fields.

I suggest we simply drop that line for the time being, and add it when it fixed.

andrerom:
Nice work :)

Side: with very long content type it will be a bit annoying with edit button at the bottom, or?

bdunogier:
> Side: with very long content type it will be a bit annoying with edit button at the bottom, or?

I agree.
I have asked @dpobel a few days ago if we could use the action bar for this, and I think it was possible, but needed some work. That would be the best option, in my opinion.

andrerom:
yes, fully agree.

nicolas-bastien:
@andrerom @bdunogier what about this ?

ezp26662-content-type-action-top

andrerom:
I'm ok with that.

nicolas-bastien:
@andrerom if we just display the code like this, I just have to commit, no automatic scroll just action display on top and on bottom.

Do you want me to commit that or not ?

andrerom:
@bdunogier WDYT? for me I can go both ways, it's only until next release.

Resources/views/ContentType/view_content_type.html.twig
nicolas-bastien: @bdunogier ok for the direct code but for this is content type description, the bug is on content field description and it's already in comment.

bdunogier:
I would keep it for a follow-up, it was specified that way.

Resources/views/ContentType/view_content_type.html.twig
bdunogier: Okay, my bad, I got confused. I'll remove that comment :)

bdunogier:
Now we just need QA's approval :)

pcardiga:
QA is on it :)

Resources/views/ContentType/view_content_type.html.twig
```diff
@@ -6,103 +6,154 @@
...
+                                    <td class="ez-field-definition-main-properties">{{ field_definition.identifier }}</td>
+                                    <td class="ez-field-definition-main-properties">{{ field_definition.fieldTypeIdentifier }}</td>
+                                    <td class="ez-field-definition-main-properties"></td>
+                                </tr>
+                                <tr>
+                                    <td colspan="4">
+                                        <ul class="ez-field-definition-properties">
+                                            <li class="ez-field-definition-property">
+                                                <div class="ez-field-definition-property-name">{{ 'field_definition.description'|trans|desc("Description") }}</div>
+                                                <div class="ez-field-definition-property-value">{#{ field_definition.description(language_code) }#}</div>
```
andrerom: Seems to be disabled

Maybe it just needs a default like content type description?
`field_definition.descriptions[language_code]|default('')`

ramzi-arfaoui: Yes that solve the problem, already tested

andrerom: @bdunogier Was there any other reason why it was commented out or should proposed solution above work to your knowledge as well?

andrerom: Ahh, I see it is pending: https://github.com/ezsystems/ezpublish-kernel/pull/1876

PR #738, opened 2016-11-29
EZP-26704: JMS Translation Bundle integration

bdunogier:
> [EZP-26704](http://jira.ez.no/browse/EZP-26704)

Sets up `jms/translation-bundle` as the extractor for translation strings.

The HandleBar and Javascript extractors are adapted to the JMS interfaces (based on the existing implementations). An extractor is added for PlatformUI PHP notifications (calls to `notify()` and `notifyError()`.

After regenerating the strings, it seems that the only lost strings are not found in the source code (the list ca be found at https://github.com/ezsystems/PlatformUIBundle/commit/46b98cc027db08bc94ef2bd81d534fc43dad9353).


Resources/translations/bar.en.xlf
```diff
@@ -1,145 +1,150 @@
...
+      <tool tool-id="JMSTranslationBundle" tool-name="JMSTranslationBundle" tool-version="1.1.0-DEV"/>
+      <note>The source node in most cases contains the sample message as written by the developer. If it looks like a dot-delimitted string such as "form.label.firstname", then the developer has not provided a default message.</note>
     </header>
     <body>
-      <trans-unit id="efd1900ac313302929fcb09d0b7bc056" resname="translateaction.existing.translations">
-        <source>translateaction.existing.translations</source>
-        <target>Existing Translations</target>
+      <trans-unit id="b89182df324a10cfccb8dac5f047355c3ba043ca" resname="actionbar.copy">
+        <source>actionbar.copy</source>
+        <target state="new">actionbar.copy</target>
```
dpobel: it would be nice to keep the translations :)

bdunogier: Hmm, it was supposed to. Oh, I know. Frak. I'll fix it :)

bdunogier: It should be okay now @dpobel :)

Translation/JavascriptFileVisitor.php
```diff
@@ -0,0 +1,73 @@
...
+namespace EzSystems\PlatformUIBundle\Translation;
+
+use JMS\TranslationBundle\Model\Message;
+use JMS\TranslationBundle\Model\MessageCatalogue;
+use JMS\TranslationBundle\Translation\Extractor\FileVisitorInterface;
+use Symfony\Component\Process\Process;
+
+class JavascriptFileVisitor implements FileVisitorInterface
+{
+    private $translationDumperPath;
```
nicolas-bastien: phpdoc

Resources/config/services.yml
```diff
@@ -271,12 +271,29 @@ services:
...
+        class: EzSystems\PlatformUIBundle\Translation\JavascriptFileVisitor
+        arguments:
+            - "@=service('kernel').locateResource('@eZPlatformUIBundle/bin/Translation/translation_dumper.js')"
+        tags:
+            - { name: jms_translation.file_visitor, alias: ez_javascript }
+
+    ezsystems.platformui.pjax.request_matcher:
+        class: "%ezsystems.platformui.pjax.request_matcher.class%"
+
+    ezsystems.platformui.translation.file_visitor.php_notifications:
```
nicolas-bastien: you should remove the last (s) here to be align on class and alias namming

Resources/translations/content_type.en.xlf
```diff
@@ -1,149 +1,231 @@
...
     </header>
     <body>
-      <trans-unit id="4f1cfd17c4ea63fdcfb4a10fda0755ff" resname="content_type.dashboard_title">
-        <source>content_type.dashboard_title</source>
-        <target>Content types</target>
+      <trans-unit id="d6c5835e7dba84e22f3f6cfd985dbff2ba5f7494" resname="content_type.always_available">
+        <source>Available</source>
+        <target>Available</target>
+        <note>key: content_type.always_available</note>
+        <jms:reference-file line="73">/../../../../.././Resources/views/ContentType/view_content_type.html.twig</jms:reference-file>
```
dpobel: that's a lot of `..` (that's a detail)

bdunogier: Yeah, I know. The different extractors don't behave the same way, and it produces this. It's either a lot of dots, or `vendor/...` :)

I could maybe handle that in the custom file-writer, but it does qualify as a bugfix, does it not ? ;)

PR #740, opened 2016-11-29
EZP-26520: add content group link in content edition breadcrumb

nicolas-bastien:
https://jira.ez.no/browse/EZP-26520


yannickroger:
@nicolas-bastien @dpobel it looks like travis is using master as the base branch. So maybe it's a bug from changing the base branch.

Resources/less/less.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
dpobel: this should not be there :)

nicolas-bastien:
@dpobel ready to merge

@yannickroger thanks

PR #744, opened 2016-11-29
EZP-26475: Infinite loader if embed Content item is unreachable

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26475

# Description

This patch adds some checks when loading the embedded Content so that the embed does not keep on displaying "Loading" when a Content cannot be loaded. This can happen in different scenarios:

* the user does not have right to read the embedded Content item
* the Content item was removed (even if in this case, the RichText content should have been updated see https://jira.ez.no/browse/EZP-23540)
* the Content item is in the Trash
* the embed represents an Image but the Content item image field is empty
* the embed represents an Image but there was an error loading the image variation

Screencast:

[](http://www.youtube.com/watch?v=WVRN39iD9iM "Click to play on Youtube.com")

## Tasks

* [x] Handle regular content
* [x] Handle images
* [x] Make embed state class are removed

# Tests

manual tests + unit tests

### DIFF
--- a/Resources/public/js/views/fields/richtext/ez-richtext-resolveimage.js
+++ b/Resources/public/js/views/fields/richtext/ez-richtext-resolveimage.js
@@ -19,7 +19,9 @@ YUI.add('ez-richtext-resolveimage', function (Y) {
      * @extends eZ.RichTextResolveEmbed
      * @constructor
      */
-    var ResolveImage = function () { };
+    var ResolveImage = function () {
+            this._notLoadedMessage = "Image could not be loaded";
+        };
 
     Y.extend(ResolveImage, Y.eZ.RichTextResolveEmbed);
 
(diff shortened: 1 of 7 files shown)


dpobel:

Resources/public/js/views/fields/richtext/ez-richtext-resolveimage.js
```diff
@@ -68,14 +70,38 @@ YUI.add('ez-richtext-resolveimage', function (Y) {
...
             var content = struct.content,
+                contentId = content.get('contentId'),
                 contentType = struct.contentType,
                 imageField;
 
-            imageField = content.get('fields')[contentType.getFieldDefinitionIdentifiers('ezimage')[0]];
-            mapNode[content.get('contentId')].forEach(Y.bind(this._loadVariation, this, view, content, imageField));
+            imageField = content.getField(
+                contentType.getFieldDefinitionIdentifiers('ezimage')[0],
+                content.get('mainLanguageCode')
```
dpobel: self note, in 1.6 `getField` has only one parameter

dpobel:
self note: when porting that to 1.7, remember to check `getField` call to pass the main language.

PR #747, opened 2016-11-30
EZP-25318: Having the (+) button while selecting text is misleading

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25318

# Description

This patch makes sure the (+) button is not available when there's a text selection in Online Editor. This is to avoid the user to think the selection will be used to create a new element while the selection is just replaced by what is created.

# Tests

manual + unit tests


dpobel:

miguelcleverti:
PR approved by QA

PR #750, opened 2016-12-01
EZP-24994: "Default Location" is not being honored when using the UDW for "Relations List" and "Relation" fields

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24994
To fully work, this patch requires https://github.com/ezsystems/ezpublish-kernel/pull/1833 (without it, it has no effect)

# Description

This patch makes sure the UDW is run with the correct `startingLocationId` parameter when filling a Relations or Relation field with a default Location/selection root.

# Tests

manual test + unit tests

### DIFF
--- a/Resources/public/js/views/fields/ez-relation-editview.js
+++ b/Resources/public/js/views/fields/ez-relation-editview.js
@@ -137,6 +137,7 @@ YUI.add('ez-relation-editview', function (Y) {
                 config: {
                     contentDiscoveredHandler: Y.bind(this._selectRelation, this),
                     cancelDiscoverHandler: Y.bind(this.validate, this),
+                    startingLocationId: this.get('fieldDefinition').fieldSettings.selectionRootHref,
                 },
             });
         },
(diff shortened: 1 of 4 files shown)


dpobel:

bdunogier:
Note that this should probably bump the kernel requirement to ^6.7.0, since this property will only exist in that version.

bdunogier:
We might wanna bump the kernel requirement, since it depends on this new property. It would be ^6.6.7 for 1.6, and ^6.7.0 for master/1.7. Could you do it in this PR @dpobel ?

dpobel:
@bdunogier I'll do it. For the record, if the corresponding kernel patch is not there, this patch has no effect (and the bug is still there of course)

miguelcleverti:
PR approved by QA

bdunogier:
https://github.com/ezsystems/ezpublish-kernel/pull/1833 has been merged, you can move on. Remember to update the requirement maybe @dpobel ?

composer.json
```diff
@@ -10,7 +10,7 @@
     "require": {
         "ezsystems/platform-ui-assets-bundle": "^3.0.0",
         "ezsystems/repository-forms": "^1.2",
-        "ezsystems/ezpublish-kernel": "^6.6",
+        "ezsystems/ezpublish-kernel": "^6.6.7",
```
andrerom: 6.7? :)

bdunogier: Nope, or it's like forcing an update to 1.7, is it not ?

Anyway, according to Damien, the change is backward compatible here as well, meaning no requirement changes.

dpobel:
@andrerom actually for 1.6, the fix will really be in 6.6.7 but given what I said in my last comment (and in the description), 6.6.7 is not strictly required will keep the composer.json as it is.

PR #751, opened 2016-12-01
[travis] added strings check

bdunogier:
> [EZP-26731](http://jira.ez.no/browse/EZP-26731)

Runs the `bin/updatestrings.sh` (also part of this PR) script in dry-run mode, from `bin/travis/checkstrings.sh`, and fails ifthe pull-request add nor remove strings.

### DIFF
--- a/bin/travis/checkstrings.sh
+++ b/bin/travis/checkstrings.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env sh
+
+OUTPUT=$(./bin/updatestrings.sh --dry-run)
+
+echo "> Output from updatestrings.sh"
+echo $OUTPUT
+
+echo "> Strings check result"
+
+ADDED_MESSAGES=$(echo "$OUTPUT" | awk '/Added Messages/ { print $3 }')
+DELETED_MESSAGES=$(echo "$OUTPUT" | awk '/Deleted Messages/ { print $3 }')
+
+RETURN=0
+
+if test "$ADDED_MESSAGES" != "0"; then
+    echo "$ADDED_MESSAGES strings are added";
+    RETURN=1
+fi
+
+if test "$DELETED_MESSAGES" != "0"; then
+    echo "$DELETED_MESSAGES strings are deleted";
+    RETURN=1
+fi
+
+if test "$RETURN" != 0; then
+    echo "Failure: strings need to be regenerated"
+fi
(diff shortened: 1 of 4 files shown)


bin/travis/checkstrings.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env sh
+OUTPUT=$(sh ./bin/updatestrings.sh --dry-run)
+ADDED_MESSAGES=$(echo "$OUTPUT" | awk '/Added Messages/ { print $3 }')
+DELETED_MESSAGES=$(echo "$OUTPUT" | awk '/Deleted Messages/ { print $3 }')
+
+RETURN=0
+
+if test "$ADDED_MESSAGES" != 0; then
```
dpobel: usually this is written with `[]` instead of `test` and also to avoid any ambiguity, I would compare against `"0"` not `0`, so something like:

```sh
if [ "$ADDED_MESSAGES" != "0" ] ; then
```

bin/travis/checkstrings.sh
```diff
@@ -0,0 +1,22 @@
...
+if test "$DELETED_MESSAGES" != 0; then
+    echo "$DELETED_MESSAGES strings are deleted";
+    RETURN=1
+fi
+
+if test "$RETURN" != 0; then
+    echo "Failure: strings need to be regenerated"
+fi
+
+exit "$RETURN";
```
dpobel: nitpick: `exit $RETURN` is enough :)

dpobel (changes requested):

bin/updatestrings.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
+../../../app/console translation:extract en -v \
```
andrerom: maybe change to run from project root as they seems to depend on that anyway? _(and check if in root while at it, this is placed in public folder, aka no `.<folder>` like `.travis`, so should probably do some sanity checks before running)_

andrerom: maybe change to run from project root as they seems to depend on that anyway? _(and check if in root while at it, this is placed in public folder, aka no `.<folder>` like `.travis`, so should probably do some sanity checks before running)_

andrerom (changes requested):

PR #752, opened 2016-12-01
Fix EZP-26515: Unable to select an old draft to edit if there are too many

yannickroger:
Link: https://jira.ez.no/browse/EZP-26515

## Description
When to many draft are present, the bottom of the list is hidden bellow the screen limit. This patch adds a scrollbar when the list is too long.

long-draft-list

## Tests
Manual tests on firefox and chrome

### DIFF
--- a/Resources/public/css/views/confirmbox.css
+++ b/Resources/public/css/views/confirmbox.css
@@ -26,6 +26,9 @@
     display: inline-block;
     margin: auto;
     padding: 1.5em 2em;
+    max-height: 100%;
+    overflow: auto;
+    box-sizing: border-box;
 }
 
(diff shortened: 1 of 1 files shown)


Resources/public/css/views/draftconflict.css
```diff
@@ -4,6 +4,11 @@
  */
 
 
+.ez-view-draftconflictview {
+    overflow: auto;
+    max-height: 30em;
```
dpobel: 30em ? what is the reasoning behind this value ?

yannickroger:
@dpobel The solution you gave works fine. PR updated.

PR #756, opened 2016-12-02
Fix the behavior with long content name

yannickroger:
Link: https://jira.ez.no/browse/EZP-26592

## Description
When using a long content name. It is often truncated or worse it pushes some buttons out of the screen. This PR tries to fix most of these behaviors.

## Tests
Manual tests on firefox and chrome

### DIFF
--- a/Resources/public/css/views/dashboard-blocks.css
+++ b/Resources/public/css/views/dashboard-blocks.css
@@ -22,6 +22,13 @@
     margin-top: 1em;
 }
 
+.ez-view-dashboardblocksview .ez-dashboard-content-name {
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    overflow: hidden;
+    max-width: 15em;
+}
+
 @media (min-width: 1024px) {
     .ez-view-dashboardblocksview {
         padding: 2% 6%;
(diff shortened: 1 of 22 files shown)


dpobel:
some screenshots or a screencast would be useful for such a fix

Resources/public/css/views/dashboard-blocks.css
```diff
@@ -22,6 +22,13 @@
     margin-top: 1em;
 }
 
+.ez-view-dashboardblocksview .ez-view-dashboard-content-name {
```
dpobel: `.ez-view-dashboard-content-name` ? `ez-view-` prefix is reserved for view container class, I guess you meant `ez-dashboard-content-name`

Resources/public/templates/tabs/locations.hbt
```diff
@@ -32,11 +32,17 @@
                                     <ul class="ez-breadcrumbs-list">
                                         {{#each this.path}}
                                         <li class="ez-breadcrumbs-item">
-                                            <a href="{{ path "viewLocation" id=id languageCode=contentInfo.mainLanguageCode }}">{{ contentInfo.name }}</a>
+                                            <a  href="{{ path "viewLocation" id=id languageCode=contentInfo.mainLanguageCode }}"
```
dpobel: nitpick: why 2 spaces after `<a` ?

PR #759, opened 2016-12-06
EZP-26750: When combine: true is not defined in any extensions yui.yml file default becomes false reducing performance

sunpietro:
https://jira.ez.no/browse/EZP-26750

Updated the PlatformUI configuration validation by setting the default value to `combine` property from `yui.yml` files.
After this fix, all the `yui.yml` files that add new JS modules or override existing ones will have the `combine: true` value even when they are omitted.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -63,6 +63,7 @@ class Configuration extends SiteAccessConfiguration
                         ->info("Filter to apply to module urls. This filter will modify the default path for all modules.\nPossible values are 'raw', 'min' or 'debug''")
                     ->end()
                     ->booleanNode('combine')
+                        ->defaultTrue()
                         ->info('If true, YUI combo loader will be used.')
                     ->end()
                     ->arrayNode('modules')
--- a/Resources/config/yui.yml
+++ b/Resources/config/yui.yml
@@ -2,7 +2,6 @@ system:
     default:
         yui:
             filter: min
-            combine: true
             modules:
                 capi:
                     path: "%ez_platformui.external_assets_public_dir%/vendors/ez-js-rest-client/dist/CAPI.js"


sunpietro:
ping @andrerom @kamilmusial @lserwatka @dpobel @bdunogier

andrerom:
Would like at least one approval or otherwise from PUI team, @dpobel @yannickroger @StephaneDiot

bdunogier:
We are on it. My only concern is that this would happen the day before the RC1. 

Does anyone know why it wasn't made true by default ?

kamilmusial:
@bdunogier maybe wrong assumption that bundles below PlatformUI will inherit that value? (which would make sense imo)

sunpietro:
Well, it was true for PlatformUI. For PlatformUI extensions it was not always true. We found this issue when dealing with SSL issues and we found out that the app is trying to load YUI3 from Yahoo CDNs.

bdunogier:
One minor thing: could you please shorten the commit message ? It should be understandable in one line. Feel free to use the next lines to elaborate, it's always useful.

sunpietro:
@bdunogier I hope you like it now :)

andrerom:
> We found this issue when dealing with SSL issues

@bdunogier that makes this an escalated issue btw :/

yannickroger:
Could you explain why it tries to load from CDN. How does it trigger that side effect.

andrerom:
> But I'm gonna block it until we have clear steps to reproduce so that QA can test it.

simple, extend platform ui, or use studio ui without hardcoding combined: true and you end up with non combined and slower performance.

> -1 for having this in 1.6. It is too late for 1.6.2 and maybe a bit dangerous for BC

Don't see the BC, but as 1.6.2 is already in rc we can definitely change this to master and make it part of 1.7.0-rc1 which will be tagged tomorrow. Support can provide this as optional patch to 1.6 customers.

bdunogier:
> simple, extend platform ui, or use studio ui without hardcoding combined

It's not for us, it's for QA. They will ask for these steps.

I'm fine with an optional patch from support on this one.

yannickroger:
One the 1.6 topic. Use master and have an optional patch for user needing it is good for me.

andrerom:
@bdunogier / @yannickroger base branch changed, ok now?

yannickroger:
Yes. I am still concerned about people not being able to explain how this patch fixes the side effect.

lserwatka:
If you create a new bundle that loads YUI module, or code that is depending on YUI, and you won't explicitly set `combine: true`, then setting from PlatformUIBundle will be omitted and YUI Loader will load dependencies from CDN. This has something to do with Symfony config merging, but I'm not an expert.

yannickroger:
I understand the problem.
I had a quick look at the combo loader code that I don't really know.

My main concern is that YUI, if I get it right, is loaded here https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/views/PlatformUI/shell.html.twig#L50

So it doesn't seem to be affected by the combo loader. As it is shipped by assetic.

My guess would be that if that hasn't been somehow loaded before, YUI modules must try to fallback on Yahoo's CDN. But that would need a more serious investigation.

andrerom:
@sunpietro Can you (or @kamilmusial ) maybe split this issue into two?

One for this PR for the fact that you need to manually enable combine or else you'll end up with performance degradation as the setting defined in yml isn't correctly merged in. And separate one on the issue when the YUI loading code falls back to CDN loading which breaks when using SSL? We will need more details on that one as rightfully implied by @yannickroger.

@yannickroger would that work for you?

yannickroger:
It sounds good to me 😄

andrerom:
Issue and title here updated, commit title will be adjusted on squash commit.
All good @yannickroger ?

PR #764, opened 2016-12-07
EZP-26749: Error in UDW when displaying 25 or more objects per container

andrerom:
> issue: https://jira.ez.no/browse/EZP-26749
> Regression caused by #723

Not detected on normal tree view as content is not loaded there, in UDW they are, and missing limit on search for content items means it is limited to 25 even if more locations where previously loaded.

Todo:
- [x] Test assertion for logic change
- [x] Swap base for 1.6 as this breaks UDW for 1.6.2 when more than 25 items

### DIFF
--- a/Resources/public/js/views/services/plugins/ez-searchplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-searchplugin.js
@@ -404,7 +404,9 @@ YUI.add('ez-searchplugin', function (Y) {
             query = this._createNewCreateViewStruct('contents-loading-' + viewName, 'ContentQuery', {
                 criteria: {
                     "ContentIdCriterion": contentIds
-                }
+                },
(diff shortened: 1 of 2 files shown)


andrerom:
review ping @yannickroger @StephaneDiot /cc @bdunogier

andrerom:
passing, review ping @yannickroger @StephaneDiot /cc @bdunogier

andrerom:
Take 2, now with test assertion and adjustment to logic to be more correct.

Review ping @yannickroger @StephaneDiot

yannickroger:
Don't forget to send to QA :)

pcardiga:
PR verified by QA and seems to fix the issue

andrerom:
Thanks @pcardiga

PR #767, opened 2016-12-13
Initial localisation work for handlebar templates: DateTime

andrerom:
Scope:
- Get dates to use a human readable format as opposed to a computer readable one

Not scope:
- setting right localization, aka picking right locale _(depends on language picking in UI)_
- number formatting
- polyfill: not needed, Intl is supported by all browsers we support now


composer.json
```diff
@@ -8,7 +8,7 @@
         }
     },
     "require": {
-        "ezsystems/platform-ui-assets-bundle": "^3.0.0",
+        "ezsystems/platform-ui-assets-bundle": "^3.1.0@alpha",
```
andrerom: tmp, remove stability flag as soon as PR is approved and tag is made there.

andrerom:
@bdunogier Does this look ok to you? If so I would like to send it to QA post release.

bdunogier:
It looks okay, but my look isn't very educated there.

I'd really like one of the senior UI team members to have a look before we consider merging this.

andrerom:
ok

andrerom:
> I'd really like one of the senior UI team members to have a look before we consider merging this.

Yes, specifically we would need input from @dpobel if the addition of `handlebars-helper-intl` is ok, or if he would prefer something else. 

> It looks okay, but my look isn't very educated there.

`handlebars-helper-intl` wraps "EcmaScript Intl" which as of Safari 10 is supported by all major browsers now. The api in question used is the one already used in `ez-subitemlistitemview.js`, and the reason why I also did a change there was to align them so:
- `en` locale is not hard coded, all use browser locale instead
- use only `toLocaleTimeString` which is same api used by `formatTime` _(for dateTime)_, instead of custom contacting of the string as was before
- don't force 12 hour clock, let locale pick it

bdunogier:
Also, I know that @dpobel and @yannickroger had been discussing this over the last year. I'll try to get some input.

andrerom:
rebased

Had a second look at FormatJS, and form what I can see it does not add anything api wise for dates themselves, so what is missing here and moved to issue mentioned below is something ala passing object with current locale around so right one is used once we allow language/locale switching in UI.

So think this is all good here now.

> Note that https://jira.ez.no/browse/EZP-23429 already mentioned dates. If we keep the new one you have created, @andrerom, I think that we should edit the other one to make it about numbers only (since it is what would be left).

done :)

yannickroger:
+0,5 👿

bower.json
```diff
@@ -16,6 +16,7 @@
     "yui3": "http://yui.zenfs.com/releases/yui3/yui_3.18.1.zip",
     "alloyeditor": "~1.2.3",
     "widget": "http://download.ckeditor.com/widget/releases/widget_4.5.9.zip",
-    "lineutils": "http://download.ckeditor.com/lineutils/releases/lineutils_4.5.9.zip"
+    "lineutils": "http://download.ckeditor.com/lineutils/releases/lineutils_4.5.9.zip",
+    "handlebars-helper-intl": "~1.1.2"
```
andrerom: ok

andrerom:
> you also need to update Resources/public/templates/tabs/versions.hbt

ok

> And what about Date and DateAndTime field view templates ?

good point

andrerom:
@dpobel I have a feeling we should move formatting to templates now for versions and ez-subitemlistitemview.js, it's view logic and feels quite wrong to tests this.

dpobel:
@andrerom on subitem why not (this involves changing the definition of `lastModificationDate` property in https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/subitem/ez-subitemlistitemview.js#L484 to use a custom template and deprecating the `_formatDate` method).
However, on Date and DateAndTime views, I disagree because the date formatting is the main responsibility of such view and with the `useSeconds` setting, the logic is not 100% straightforward (this would also mean to add a custom template for date)

Resources/public/js/views/fields/ez-dateandtime-view.js
```diff
@@ -48,24 +48,44 @@ YUI.add('ez-dateandtime-view', function (Y) {
...
          *
-         * @method _formatDate
+         * @method _formatTime
          * @protected
          * @param {Date} date
          * @return String
          */
         _formatTime: function (date) {
-            var options;
+            var options = {hour: '2-digit', minute: '2-digit'};
```
dpobel: if I'm not mistaken, this method is not used anywhere since it is overridden in ez-time-view but this one is much better so instead of changing it there, you should just deprecate it and move the code based on `toLocaleTimeString` to ez-time-view

Resources/public/js/views/fields/ez-dateandtime-view.js
```diff
@@ -48,24 +48,44 @@ YUI.add('ez-dateandtime-view', function (Y) {
          * @return String
          */
         _formatDate: function (date) {
-            return date.toLocaleDateString();
```
dpobel: that the same case as for `_formatTime`, this one should be deprecated and ez-date-view should redefine a `_formatDate` method.

andrerom:
Updated _(skipped the section discussed above on subitem and field views)_

Resources/public/js/views/fields/ez-dateandtime-view.js
```diff
@@ -42,6 +42,7 @@ YUI.add('ez-dateandtime-view', function (Y) {
          * Formats the date part of the date object according to the locale
          * settings of the browser
          *
+         * @deprecated Since 1.7.1
```
dpobel: you should add the console.log calls in the method ([example](https://github.com/ezsystems/PlatformUIBundle/blob/bd887d42fe34896fde6ba8b04508d5e046995860/Resources/public/js/views/ez-languageswitcherview.js#L136)) so that the notice is displayed in the console.

Resources/public/js/views/fields/ez-dateandtime-view.js
```diff
@@ -52,11 +53,31 @@ YUI.add('ez-dateandtime-view', function (Y) {
...
+        },
+
+        /**
          * Formats the time part of the date object according to the locale
          * settings of the browser and to the `useSeconds` field definition
          * settings.
          *
-         * @method _formatDate
+         * @deprecated Since 1.7.1
+         * @method _formatTime
```
dpobel: same as _formatDate, console.log calls are missing

sunpietro:
@andrerom what do you think about creating a JS view helper/extension that provides methods to convert dates and times to a correct standard? A thing like `ez-expandable`. It could be easily reused in many places.

andrerom:
@sunpietro a bit out of scope here, and then what about handlebars? Ideally we should have an object with the default settings for date, datetime and time formatting options, and get the handlebars-inlt end points to use that by default as well.

andrerom:
> @andrerom unless I'm mistaken, if we start using Handlebars-intl then we should start passing either unix timestamps or date objects into our templates and allow the helper to translate it, am I right?

correct, but if you already use `date.toLocaleDateString` you don't have to. And if you use something else underneath make sure to change to ES Intl. Contact me on a different means then this PR if there is something.

PR #769, opened 2016-12-16
Fix EZP-26550: Infinite loading when editing unallowed content

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-26550

## Description

Goal here was to remove JS error when trying to edit an unallowed content. This was produce because _sortVersions method couldn't handle the case with no versions. This is solved using Y.array.each instead of Object.forEach ( because Y.array.each is smarter and check if array is empty before iterating )

While solving this bug, another one was revealed : version model wasn't returning response on error while creating or updating... This is now fixed.

## Screencast

https://youtu.be/8YapcakQ_NI

## Tests 

unit and manually tested

### DIFF
--- a/Resources/public/js/models/extensions/ez-contentinfo-base.js
+++ b/Resources/public/js/models/extensions/ez-contentinfo-base.js
@@ -73,7 +73,7 @@ YUI.add('ez-contentinfo-base', function (Y) {
         _sortVersions: function(versions) {
             var versionsByStatus = {};
 
-            versions.forEach(function (version) {
+            Y.Array.each(versions, function (version) {
                 if ( !versionsByStatus[version.get('status')]) {
                     versionsByStatus[version.get('status')] = [];
                 }
--- a/Resources/public/js/views/services/plugins/extensions/ez-draftserversidevalidation.js
+++ b/Resources/public/js/views/services/plugins/extensions/ez-draftserversidevalidation.js
@@ -32,10 +32,18 @@ YUI.add('ez-draftserversidevalidation', function (Y) {
          */
         _parseServerFieldsErrors: function (response, serverSideErrorCallback) {
             var serverSideFieldsError = [],
-                error = response.document.ErrorMessage,
-                fieldsError = error.errorDetails ?  error.errorDetails.fields : [];
+                error,
+                fieldsError;
+
+            if ( !response.document && serverSideErrorCallback) {
+                serverSideErrorCallback(serverSideFieldsError);
+                return;
+            }
 
             if (serverSideErrorCallback) {
+                error = response.document.ErrorMessage;
+                fieldsError = error.errorDetails ?  error.errorDetails.fields : [];
+
                 fieldsError.forEach(function (field) {
                     field.errors.forEach(function (error) {
                         var serverSideFieldErrorDetails;
@@ -45,7 +53,6 @@ YUI.add('ez-draftserversidevalidation', function (Y) {
                         serverSideFieldsError.push(serverSideFieldErrorDetails);
                     });
                 });
-
                 serverSideErrorCallback(serverSideFieldsError);
             }
         },
--- a/Tests/js/extensions/assets/ez-draftserversidevalidation-tests.js
+++ b/Tests/js/extensions/assets/ez-draftserversidevalidation-tests.js
@@ -45,11 +45,9 @@ YUI.add('ez-draftserversidevalidation-tests', function (Y) {
             Assert.isTrue(serverSideErrorCallbackCalled, "serverSideErrorCallback should have been called");
         },
 
-        // regression test for https://jira.ez.no/browse/EZP-26543
-        "Should handle error without details": function () {
+        _handleError: function () {
             var serverSideErrorCallbackCalled = false;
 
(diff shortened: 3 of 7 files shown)


Resources/public/js/views/services/plugins/extensions/ez-draftserversidevalidation.js
```diff
@@ -32,22 +32,26 @@ YUI.add('ez-draftserversidevalidation', function (Y) {
...
+                            serverSideFieldErrorDetails.parse(error);
+                            serverSideFieldsError.push(serverSideFieldErrorDetails);
+                        });
                     });
-                });
-
+                }
                 serverSideErrorCallback(serverSideFieldsError);
             }
+
```
yannickroger: nitpick: empty line

Resources/public/js/models/extensions/ez-contentinfo-base.js
```diff
@@ -73,7 +73,7 @@ YUI.add('ez-contentinfo-base', function (Y) {
         _sortVersions: function(versions) {
             var versionsByStatus = {};
 
-            versions.forEach(function (version) {
+            Y.Array.each(versions, function (version) {
```
yannickroger: Why not use the same thing as for `fieldsError` in the other file.

Resources/public/js/views/services/plugins/extensions/ez-draftserversidevalidation.js
```diff
@@ -32,22 +32,26 @@ YUI.add('ez-draftserversidevalidation', function (Y) {
...
-                error = response.document.ErrorMessage,
-                fieldsError = error.errorDetails ?  error.errorDetails.fields : [];
+                error,
+                fieldsError;
 
             if (serverSideErrorCallback) {
-                fieldsError.forEach(function (field) {
-                    field.errors.forEach(function (error) {
-                        var serverSideFieldErrorDetails;
+                if ( response ) {
```
dpobel: how is that possible ? `response` is documented as required actually calling this method without a response is odd since the point of this method is to parse the response.

Tests/js/models/extensions/assets/ez-contentinfo-base-tests.js
```diff
@@ -167,6 +167,29 @@ YUI.add('ez-contentinfo-base-tests', function (Y) {
 
             Assert.isTrue(callbackCalled, 'Should call callback function');
         },
+
+        'Should handle load versions by status without versions': function () {
```
dpobel: that's not clear, actually you fixed the error handling, so the test name should probably contain 'error handling' :)

Resources/public/js/views/services/plugins/ez-publishdraftplugin.js
```diff
@@ -99,7 +99,9 @@ YUI.add('ez-publishdraftplugin', function (Y) {
                 notificationIdentifier = this._buildNotificationIdentifier(content.get('id'));
 
             if ( error ) {
-                this._parseServerFieldsErrors(response, this._serverSideErrorCallback);
+                if (response) {
```
dpobel: you just moved the question raised in https://github.com/ezsystems/PlatformUIBundle/pull/769#discussion_r94311013 here but the question remains. a `response` is supposed to be always available here, so unless I'm missing something the fix is not to check is the response is there but to make sure it's always there so you need to fix the case(s) where the response is not provided.

StephaneDiot:
@dpobel updated

StephaneDiot:
ping @dpobel

Tests/js/extensions/assets/ez-draftserversidevalidation-tests.js
```diff
@@ -67,5 +65,17 @@ YUI.add('ez-draftserversidevalidation-tests', function (Y) {
             });
             Assert.isTrue(serverSideErrorCallbackCalled, "serverSideErrorCallback should have been called");
         },
+        
+        // regression test for https://jira.ez.no/browse/EZP-26543
+        "Should handle error without details": function () {
+                  delete this.errorResponse.document;
```
dpobel: cs

Resources/public/js/views/services/plugins/extensions/ez-draftserversidevalidation.js
```diff
@@ -32,20 +32,24 @@ YUI.add('ez-draftserversidevalidation', function (Y) {
...
-                error = response.document.ErrorMessage,
-                fieldsError = error.errorDetails ?  error.errorDetails.fields : [];
+                error,
+                fieldsError;
 
             if (serverSideErrorCallback) {
-                fieldsError.forEach(function (field) {
-                    field.errors.forEach(function (error) {
-                        var serverSideFieldErrorDetails;
+                if (response.document) {
```
dpobel: it's a bit weird to check the response after the callback, why not write it like:
```js
var serverSideFieldsError = [],
    error,
    fieldsError;

if ( !response.document ) {
    return;
}
error = response.document.ErrorMessage;
fieldsError = error.errorDetails ?  error.errorDetails.fields : [];
// ...
```

StephaneDiot:
updated @dpobel

dpobel:
@StephaneDiot don't forget to clean up the commits before merging

PR #771, opened 2016-12-22
EZP-26523: Sorting UI in Object Relation List field

janit:
https://jira.ez.no/browse/EZP-26523

The Relation List field is not sortable currently. This PR adds sorting to the relation list UI and the correct order is also stored to the backend.

What is not included is loading relations in the order defined by the backend. This means that in the admin UI relation sorting will not seem to be working reliably.


andrerom:
Seems like the remote repo has been removed, so if anyone is willing to continue on this and open new PR please do so.

PR #772, opened 2017-01-03
EZP-26516: As an editor, I want to be able to add/edit tabular data in RichText editor

andrerom:
> Issue: https://jira.ez.no/browse/EZP-26516

Todo:
- [x] Add Tests
- [x] Translation string
- [x] Disable Embed and Image inside table for now _(logic for inline toolbar to cover several elements at the same time like table+embed is for another larger story later)_
- [x] Table styling
- [x] Verify if table heading work in all options exposed by alloy _(works, even scope attribute for screen readers seems to work)_

#### Screenshots (As of April)
screen shot 2017-04-29 at 15 58 28
screen shot 2017-04-29 at 15 58 47
screen shot 2017-04-29 at 16 19 03
screen shot 2017-04-29 at 16 19 16
screen shot 2017-04-29 at 16 13 34

### DIFF
--- a/Resources/public/js/alloyeditor/buttons/embed.jsx
+++ b/Resources/public/js/alloyeditor/buttons/embed.jsx
@@ -40,6 +40,18 @@ YUI.add('ez-alloyeditor-button-embed', function (Y) {
         },
 
         /**
+         * Lifecycle. Invoked once before the component is mounted.
+         * The return value will be used as the initial value of this.state.
+         *
+         * @method getInitialState
+         */
+        getInitialState: function() {
+            return {
+                isDisabled: false,
+            };
+        },
+
+        /**
          * Returns the UDW title to pick a Content to embed.
          *
          * @method _getUDWTitle
@@ -48,6 +60,23 @@ YUI.add('ez-alloyeditor-button-embed', function (Y) {
          */
         _getUDWTitle: function () {
             return Y.eZ.trans('select.a.content.to.embed', {}, 'onlineeditor');
+        },
+
+        /**
+         * Checks if the command is disabled in the current selection.
+         *
+         * @method isDisabled
+         * @return {Boolean} True if the command is disabled, false otherwise.
+         */
+        isDisabled: function () {
+            var path = this.props.editor.get('nativeEditor').elementPath();
+
+            // http://docs.ckeditor.com/#!/api/CKEDITOR.dom.elementPath
+            // There is also isContextFor( tag ), so if there is a way to specify where embeds
+            // are valid that would potentially be cleaner
+            this.state.isDisabled = path && path.contains('table', true) !== null;
+
+            return this.state.isDisabled;
         },
 
         /**
@@ -70,10 +99,10 @@ YUI.add('ez-alloyeditor-button-embed', function (Y) {
         },
 
         render: function () {
-            var css = "ae-button ez-ae-labeled-button" + this.getStateClasses();
+            var css = "ae-button ez-ae-labeled-button" + this.getStateClasses(), disabled = this.state.isDisabled;
 
             return (
-                <button className={css} onClick={this._chooseContent} tabIndex={this.props.tabIndex}>
+                <button className={css} disabled={disabled} onClick={this._chooseContent} tabIndex={this.props.tabIndex}>
                     <span className="ez-ae-icon ez-ae-icon-embed ez-font-icon"></span>
                     <p className="ez-ae-label">{Y.eZ.trans('embed', {}, 'onlineeditor')}</p>
                 </button>
--- a/Resources/public/js/alloyeditor/buttons/embed.jsx
+++ b/Resources/public/js/alloyeditor/buttons/embed.jsx
@@ -57,11 +57,7 @@ YUI.add('ez-alloyeditor-button-embed', function (Y) {
          * @return {Boolean} True if the command is disabled, false otherwise.
          */
         isDisabled: function () {
-            var path = this.props.editor.get('nativeEditor').elementPath();
-
-            // http://docs.ckeditor.com/#!/api/CKEDITOR.dom.elementPath
-            // There is also isContextFor( tag ) which potential could have been used.
-            return path && path.contains('table', true) !== null;
+            return !this.props.editor.get('nativeEditor').ezembed.canBeAdded();
         },
 
         /**
--- a/Resources/public/js/alloyeditor/buttons/embed.jsx
+++ b/Resources/public/js/alloyeditor/buttons/embed.jsx
@@ -40,18 +40,6 @@ YUI.add('ez-alloyeditor-button-embed', function (Y) {
         },
 
         /**
-         * Lifecycle. Invoked once before the component is mounted.
-         * The return value will be used as the initial value of this.state.
-         *
-         * @method getInitialState
-         */
-        getInitialState: function() {
-            return {
-                isDisabled: false,
-            };
-        },
-
-        /**
          * Returns the UDW title to pick a Content to embed.
          *
          * @method _getUDWTitle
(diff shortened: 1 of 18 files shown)


rolandbenedetti:
From my perspective, it's good to keep it like this for now.

andrerom:
Ok, but I'd still like to adjust the style of the table.


@dpobel anything else you'd expect here? And if you agree with adjusting color of borders on tables, where should that be placed?

dpobel:
@andrerom indeed the table deserves some styling. You should probably do the same thing as for embed ie add a CSS module `richtext-table.css` (so 2 CSS file, one for layout, one for theme). But unlike for embed, you can not just target `table` elements in this CSS, so I guess you should make sure both the richtext field view and field edit views generate an element with the same class (I can see that RichText field view has a `div.ez-richtext-content`, maybe the field edit view should have the class `ez-richtext-content` class added somewhere and then you can style the table based on this class).

Code wise, that seems ok (besides minor CS problems) but by quickly testing I'm more worried about what's happening when you put some content in the table. For instance, if you add an embed in a cell, you don't have the toolbar to manage it, so you can not do anything (you can not even remove it!). Also on embed in table cell, it seems like there's an additional paragraph in the cell that can not be removed (for others elements it seems to be possible)

bdunogier:
`andrerom wants to merge 1 commit into *1.7* from eztable`

New features in a stable branch ? :)

andrerom:
> New features in a stable branch ? :)

It's a minor change in current form, as such I would call it an improvement, it's just exposing what is already there. But given input above this will probably grow way out of what I'm able to work on, so will most likely push back to backlog.

andrerom:
PR updated, with changes to disable embed and image button while inside table as suggested by @rolandbenedetti.

@dpobel Ref @ezrobot I'm unsure how I can limit the width lines of the generated js files.

@dpobel / @supriyabhargava / @SylvainGuittard / @rolandbenedetti See updated images above for what current status is for the optimal use cases. When it comes to remaining not so optimal use case be aware that it is sometimes possible to insert headings in the table _(normally it is created after the table, but in some cases inside)_:
screen shot 2017-04-29 at 16 29 57

But from the looks of the [online AlloyEditor Demo](http://alloyeditor.com/demo/), it seems they have added a toolbar to be able to set it back to normal:
screen shot 2017-04-29 at 16 33 24

So should we aim to disable heading button also, or aim to update alloy as a followup? wdyt @dpobel?

rolandbenedetti:
Hello André, 

Congrats, honestly may be not perfect, but that would be a great move forward in my opinion.
It's great to have the Alloy table's headers dropdown supported.

About the new styling dropdown, I think it would be great to also be able to use it. It gets me a bit confused though. On one end, it allows to choose the whole style of a table (I guess under _Object styles_) which I think can be very useful especially if developers can easily define such styles, on the other it acts on one single cell, allowing to update the style of a cell's content to normal. Once done, I see no way to move back to the heading style. But beside that, I'd support adding this as well.

+1 for me

andrerom:
> About the new styling dropdown, I think it would be great to also be able to use it. It gets me a bit confused though. ...

As said I won't be able to add that here, so we can deal with that in followup on how it should behave. We most likely don't support the given style choices, so only thing I found clearly relevant without needs for additional backend features _(potentially thinning out our content vs design argument)_, is the option to normalize style.

So, not relevant here, lets discuss those concepts separately as part of effort top update alloy form 1.2.5 to 1.4.0 _(or higher once we do it)_.

SylvainGuittard:
Impressive @andrerom. Good job.

> About the new styling dropdown, I think it would be great to also be able to use it. It gets me a bit confused though. On one end, it allows to choose the whole style of a table (I guess under Object styles) which I think can be very useful especially if developers can easily define such styles, on the other it acts on one single cell, allowing to update the style of a cell's content to normal. Once done, I see no way to move back to the heading style. But beside that, I'd support adding this as well.

I totally agree. `Normal` is very confusing. I would love to have this styling option and maybe hide the `Normal` option.

rolandbenedetti:
> lets discuss those concepts separately as part of effort top update alloy from 1.2.5 to 1.4.0 (or higher once we do it). It's a different story.

Absolutely !

Resources/public/js/alloyeditor/buttons/embed.jsx
```diff
@@ -51,6 +63,23 @@ YUI.add('ez-alloyeditor-button-embed', function (Y) {
...
+         * @method isDisabled
+         * @return {Boolean} True if the command is disabled, false otherwise.
+         */
+        isDisabled: function () {
+            var path = this.props.editor.get('nativeEditor').elementPath();
+
+            // http://docs.ckeditor.com/#!/api/CKEDITOR.dom.elementPath
+            // There is also isContextFor( tag ), so if there is a way to specify where embeds
+            // are valid that would potentially be cleaner
+            this.state.isDisabled = path && path.contains('table', true) !== null;
```
dpobel: this is forbidden https://facebook.github.io/react/docs/state-and-lifecycle.html#do-not-modify-state-directly ;)
also as you wrote in the comment above, the *is disabled* logic should not be here but rather in the embed plugin. That said, I'm unsure how you can access the embed plugin for here, that's a bit far away in my head.

andrerom: Ok, then I should not use state at all, this is already called on render(), no point causing recursive rendering.

What I was trying to avoid was `isDisabled()` having to be called two times by design, from `getStateClasses()` and `render()`.

Input on how I can avoid welcome, but for next commit I'll get rid of this.

andrerom: > also as you wrote in the comment above, the is disabled logic should not be here but rather in the embed plugin.

I wasn't saying embed plugin, but maybe indeed that is where logic for isContextFor would have to live. But for now I'll cleanup the comment.

Resources/public/js/alloyeditor/buttons/embed.jsx
```diff
@@ -51,6 +51,20 @@ YUI.add('ez-alloyeditor-button-embed', function (Y) {
         },
 
         /**
+         * Checks if the command is disabled in the current selection.
+         *
+         * @method isDisabled
+         * @return {Boolean} True if the command is disabled, false otherwise.
+         */
+        isDisabled: function () {
```
dpobel: it should be protected (ie starts with a `_` and be documented as protected)

Resources/public/js/alloyeditor/buttons/embed.jsx
```diff
@@ -51,6 +51,20 @@ YUI.add('ez-alloyeditor-button-embed', function (Y) {
...
+            var path = this.props.editor.get('nativeEditor').elementPath();
+
+            // http://docs.ckeditor.com/#!/api/CKEDITOR.dom.elementPath
+            // There is also isContextFor( tag ) which potential could have been used.
+            return path && path.contains('table', true) !== null;
```
dpobel: quickly tried to put that in a re-usable API to avoid duplication and here is what can be done:
in Resources/public/js/alloyeditor/plugins/embed.js you can write something like:

```js
CKEDITOR.plugins.add('ezembed', {                                           
    requires: 'widget,ezaddcontent',                                                                                                                                                                                                       
                                                                                                                                                                                                                                               
    init: function (editor) {
        editor.ezembed = {
            canBeAdded: function () { // feel to suggest a better name
                var path = editor.elementPath();

                return path && path.contains('table', true) !== null;
            }
        }

        editor.widgets.add('ezembed', {
        // ...
    },
```

and here then you can write:

```js
_isDisabled: function () {
    return this.props.editor.get('nativeEditor').ezembed.canBeAdded();
}
```
```

Resources/public/js/alloyeditor/buttons/embed.jsx
andrerom: could do that, but then it won't work in `getStateClasses()` as expected by `ButtonStateClasses`

ref: https://github.com/liferay/alloy-editor/blob/25151898f9385f09096f05ab7cf30aeea281330d/src/ui/react/src/components/base/button-state-classes.js

dpobel: good point :)

Resources/public/js/alloyeditor/buttons/embed.jsx
andrerom: Done in 4681fe9 _(with adjusted logic as mentioned on skype)_

andrerom:
Merging as QA is off today, they'll have to bombard this as part of the release tarball as of tomorrow.

PR #773, opened 2017-01-04
Crowdin in-context translation integration

bdunogier:
> [EZP-26828](http://jira.ez.no/browse/EZP-26828)

Adds the javascript code that integrates [Crowdin's in-context translation UI](https://crowdin.com/page/in-context-localization).

In-context translation of PlatformUI

### Usage
To trigger translations, the ach-UG translation, specific to crowdin, must be installed. It can be done by running `composer require ezplatform-i18n/ezplatform-i18n_ach_ug`. The requirement will be added to ezsystems/ezplatform by default.

To trigger the in-context UI, https://github.com/ezsystems/ezpublish-kernel/pull/1889 is required.

### TODO
- [x] BazingaTranslationBundle patch (https://github.com/willdurand/BazingaJsTranslationBundle/pull/191)
- [x] Clarify where the ach-UG translation files should be taken from / hosted (added instructions + downloadable file)
- [ ] Behat scenario ?
- [x] Unit test the subscriber
- [x] Consider changing the subscriber's folder (talk to Nicolas)
- [ ] Make the name of the crowdin project configurable (or follow-up ?).
      Would allow any project to use in-context translation.

### DIFF
--- a/Resources/views/PlatformUI/shell.html.twig
+++ b/Resources/views/PlatformUI/shell.html.twig
@@ -5,6 +5,14 @@
     <title>eZ Platform UI</title>
     <meta name="viewport" content="width=device-width"/>
     <meta charset="utf-8"/>
+
+    {% if parameters.interfaceLanguages[0] == 'ach_UG' %}
+        <script type="text/javascript">
+            var _jipt = [];
+            _jipt.push(['project', 'ezplatform']);
+        </script>
+        <script type="text/javascript" src="//cdn.crowdin.com/jipt/jipt.js"></script>
+    {% endif %}
 {% stylesheets '$css.files;ez_platformui$' filter='cssrewrite' %}
     <link rel="stylesheet" href="{{ asset_url }}"/>
 {% endstylesheets %}


nicolas-bastien:
@bdunogier for that point : 

> Clarify where the ach-UG translation files should be taken from / hosted (added instructions + downloadable file)

I guess it will be distribute in https://github.com/ezplatform-i18n/ezplatform-i18n-ach_UG. ?

bdunogier:
> I guess it will be distribute in https://github.com/ezplatform-i18n/ezplatform-i18n-ach_UG. ?

Well, that would be one way, yes.
But my issue with that is that you would have to install _something else_ in order to get working a feature that is supposed to work. In-context will just not work without it. I'd rather have it shipped with the product, so that anybody can translate.

bdunogier:
About the open question above, my plan turns out to be a little more complex:

Given that in-context has been proven to work here, we could:
- move the cookie listener to the CoreBundle.
- keep the `shell.html.twig` changes in this pull-request and merge them. They're totally harmless unless you trigger the custom language
- add `ezplatform-i18n/ezplatform-i18n-ach_ug` to the `ezplatform` requirements. That way, users are free to remove it (note that we MUST tell people to add it to their own composer.json)

Any opinion @dpobel or @nicolas-bastien ?

nicolas-bastien:
@bdunogier 

- Moving the listener to core make sense cause translation collector is already there

+ I agree for the two other points

bdunogier:
Cleanup done !

bdunogier:
Ping @yannickroger @nicolas-bastien (for the cleaned up version).

PR #774, opened 2017-01-06
EZP-26816: Allow to sort the subitem list with a visible column

yannickroger:
Link: https://jira.ez.no/browse/EZP-26816

## Description
The goal of this patch is to allow sorting the column of the subitem list.

Screencast: https://drive.google.com/file/d/0B5amepf6EkXbWTdSblBkMS03OTg/view

## Tests
Manual tests on Firefox and Chrome. Updated unit test coverage.

## TODO
- [x] Confirmation from product management if arrows are supposed to be this way (pointing up for ascending)
- [x] rebase


yannickroger:
Ready for review @dpobel @StephaneDiot and others.

The only todo left is about a feedback from product management. In the specs, it was said that arrows should be pointing up when sorted ASC. I would have done the opposite.

Resources/public/css/theme/views/subitem/listmore.css
```diff
@@ -10,3 +10,21 @@
 .ez-view-subitemlistmoreview .ez-subitemlist-no-sub-items {
     font-style: italic;
 }
+
+.ez-view-subitemlistmoreview .ez-subitem-column-sortable:hover:after {
+    content: "\E926";
+    font-family: 'ez-platformui-icomoon';
```
dpobel: instead of adding `font-family` (and maybe `vertical-align`) you should add the class `ez-font-icon` on `ez-subitem-column-sortable` element

Resources/public/js/views/subitem/ez-subitemlistmoreview.js
```diff
@@ -44,6 +53,11 @@ YUI.add('ez-subitemlistmoreview', function (Y) {
                     e.location.onceAfter('priorityChange', Y.bind(this._refresh, this));
                 }
             });
+
+            this.after('activeChange', this._addSortClass);
+
+            // Sort order is modified in the details tab
```
dpobel: this comment is a bit misleading

Resources/public/js/views/subitem/ez-subitemlistmoreview.js
```diff
@@ -83,6 +97,85 @@ YUI.add('ez-subitemlistmoreview', function (Y) {
...
+            });
+        },
+
+        /**
+         * Adds the sort class according to the `sortCondition` attribute
+         *
+         * @protected
+         * @method _addSortClass
+         */
+        _addSortClass: function () {
```
dpobel: should probably be named `_addSortOrderClass` (same thing for the reset method, `_resetSortOrderClass`)

Resources/public/js/views/subitem/ez-subitemlistmoreview.js
```diff
@@ -83,6 +97,85 @@ YUI.add('ez-subitemlistmoreview', function (Y) {
...
+
+        /**
+         * Retrives the property for a given sortField
+         *
+         * @protected
+         * @method _getPropertyBySortfield
+         * @param {String} sortField
+         * @return {String} property
+         */
+        _getPropertyBySortfield: function (sortField) {
```
dpobel: `_getPropertyBySortField`

yannickroger:
Feedback taken into account, ping @StephaneDiot and others

andrerom:
@yannickroger If @rolandbenedetti or @inakijv does not give input today assume you should rather use industry standard way of visualizing ascending or descending.

andrerom:
Regarding descending / ascending, on mac I have the following:
screen shot 2017-01-23 at 11 52 25

Which is the same as on KDE right? @yannickroger unless you get feedback otherwise from @rolandbenedetti or @inakijv today, assume you should align with how most other applications present this.

yannickroger:
Kde is not the same way. Maybe that's why I am not used to the other way.
sort-order

So it's all good, sending to QA as is.

yannickroger:
Officially KDE is weird then.
On windows it's the same as mac :
sort-win

inakijv:
There is no a consistent interpretation on this, it is true (and that's why the conflict), but it's mostly accepted in they way that has been proposed :-)

PR #779, opened 2017-01-17
Fixed travis failures in RootInfoTest

bdunogier:
Symfony's AssetsHelper was changed recently. It has a backward compatibility feature that enforces
a legacy behaviour when all 3 arguments are defined. The fix is to set arguments 2 and 3
to null (their default value) in order to enter this BC layer.

### TODO
- [ ] Check if we can use the newer version of the lib instead
- [ ] See if other branches might be affected

### DIFF
--- a/ApplicationConfig/Providers/RootInfo.php
+++ b/ApplicationConfig/Providers/RootInfo.php
@@ -13,8 +13,10 @@ class RootInfo implements Provider
     /** @var \Symfony\Component\HttpFoundation\RequestStack */
     private $requestStack;
 
-    /** @var \Symfony\Component\Templating\Asset\PackageInterface */
+    /** @var \Symfony\Bundle\FrameworkBundle\Templating\Helper\AssetsHelper */
     private $assetsHelper;
+
+    private $externalAssetsDirectory;
 
     public function __construct(RequestStack $requestStack, AssetsHelper $assetsHelper, $externalAssetsDirectory)
     {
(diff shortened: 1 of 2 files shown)


bdunogier:
Okay, here's the actual reason for the failure, maybe even failures:

`  - Installing symfony/symfony (v3.2.2) Loading from cache`

Given that ezplatform itself doesn't allow 3.2, it doesn't make sense to test it here. Can we just add it to platform-ui-bundle's requirements ? @andrerom opinion, since you have been following-up on Sf 3.2 ?

bdunogier:
Changed the PR to restrict dev requirements to symfony `~2.8.0`, like eZ Platform. Reverted the previous changes. Tests should pass.

andrerom:
> @andrerom do you have an opinion, since you have been following-up on Sf 3.2 ?

limiting it for limited time is ok, but optimal case is to adapt to make sure we support both so we can make the transition to 2.x smoother _(as all these things will be blockers when we soon start to create a 2.0-dev branch for ezplatform)_.

bdunogier:
I have added a story for this one to the Sf 3 epic. And yes, we need to fix those, absolutely.

bdunogier:
Merging this until an Sf 3 compatible fix is implemented, in which case the restriction will have to be lifted anyway.

bdunogier:
Travi passes on https://travis-ci.org/ezsystems/PlatformUIBundle/builds/193984410, merging.

PR #781, opened 2017-01-20
EZP-26768 : fix EZP-26768 unit tests

nicolas-bastien:
Jira : https://jira.ez.no/browse/EZP-26868

## Description

Fix test failure introduced by EZP-26768.

Add root and comboBase default value added to fix the previous one + enable second route in router mock.

## Test

Should return travis to green

### DIFF
--- a/Tests/Twig/Fixtures/functions/ez_editorial_yui_config.test
+++ b/Tests/Twig/Fixtures/functions/ez_editorial_yui_config.test
@@ -7,6 +7,6 @@ ez_platformui_ui_config "twig" function
 --DATA--
 return array('YUIConf' => 'YUI.GlobalConfig')
 --EXPECT--
-YUI.GlobalConfig = {"filter":"min","modules":[]};
-myObj = {"filter":"min","modules":[]};
-{"filter":"min","modules":[]};
+YUI.GlobalConfig = {"filter":"min","modules":[],"root":"","comboBase":"yui_combo_loader?"};
+myObj = {"filter":"min","modules":[],"root":"","comboBase":"yui_combo_loader?"};
+{"filter":"min","modules":[],"root":"","comboBase":"yui_combo_loader?"};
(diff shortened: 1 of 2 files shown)


Tests/Twig/Fixtures/functions/ez_editorial_yui_config.test
```diff
@@ -7,6 +7,6 @@ ez_platformui_ui_config "twig" function
 --DATA--
 return array('YUIConf' => 'YUI.GlobalConfig')
 --EXPECT--
-YUI.GlobalConfig = {"filter":"min","modules":[]};
-myObj = {"filter":"min","modules":[]};
-{"filter":"min","modules":[]};
+YUI.GlobalConfig = {"filter":"min","modules":[],"root":"","comboBase":"?"};
```
nicolas-bastien: Here I don't understand really why comboBase is only '?'

nicolas-bastien:
@dpobel travis fails on RootInfoTest now, I saw that you're the last to have modified this line I guess you will know where it comes from ?

bdunogier:
I have merged the pull-request that fixes the RootInfoTest failures, you can rebase to get travis to maybe go green.

Tests/Twig/Fixtures/functions/ez_editorial_yui_config.test
bdunogier: To be frank, neither do I :)

I'm not a big fan of just fixing the fixture without understanding what's happening. Can you try to look deeper into it ?

bdunogier:
By the way, 1.7 is affected as well, can you edit the pull-request to be against it instead ?

Tests/Twig/Fixtures/functions/ez_editorial_yui_config.test
nicolas-bastien: I did that why I was asking if somebody from UI maybe has a quick answer to my question.

nicolas-bastien:
@bdunogier branch rebased + fix the comboBase question. Router mock was not the same so I centralize the creation to be consistent.

PR #782, opened 2017-01-23
EZP-26881: Update usage of AssetHelper to symfony 3.x

alongosz:
> Fixes [EZP-26881](https://jira.ez.no/browse/EZP-26881)

With Symfony `3`, the test for `RootInfo` fails. Detailed description of a probable cause of the problem and solution will be provided in a consecutive comment.

**TODO**:
- [x] Change `\EzSystems\PlatformUIBundle\Tests\ApplicationConfig\Providers\RootInfoTest` to be compatible with both Symfony `2` and `3` (94fce3d).
- [x] Bump Symfony requirements to `"~2.8|~3.2"` (7773a32).
- [x] Allow running PHPUnit tests with different versions of Symfony (66a3478).
- [x] Check if Travis agrees.
- [x] Get feedback from the others to check if we're not missing something.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,9 @@ matrix:
           env: BEFORE="./bin/travis/prepare_behat.sh" TEST_CMD="./bin/travis/runbehat.sh" AFTER_SUCCESS='echo "After success"' RUN_INSTALL=1 COMPOSE_FILE="doc/docker-compose/base-prod.yml:doc/docker-compose/selenium.yml" SYMFONY_ENV=behat SYMFONY_DEBUG=0
         - env: BEFORE="./bin/travis/setupnode.sh" TEST_CMD="./bin/travis/runnode.sh" AFTER_SUCCESS="./bin/travis/generate_apidoc.sh"
         - php: 7.0
-          env: BEFORE="./bin/travis/setupphpunit.sh" TEST_CMD="./vendor/bin/phpunit -c phpunit.xml" AFTER_SUCCESS='echo "After success"'
+          env: BEFORE="./bin/travis/setupphpunit.sh" SYMFONY_VERSION="^2.8" TEST_CMD="./vendor/bin/phpunit -c phpunit.xml" AFTER_SUCCESS='echo "After success"'
+        - php: 7.1
+          env: BEFORE="./bin/travis/setupphpunit.sh" SYMFONY_VERSION="^3.2" TEST_CMD="./vendor/bin/phpunit -c phpunit.xml" AFTER_SUCCESS='echo "After success"'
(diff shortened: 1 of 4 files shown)


alongosz:
It seems that problem with `RootInfoTest` is not with usage of `AssetHelper` in `RootInfo` but rather due to PHPUnit mock for method not being called. When a number of arguments (doesn't matter that they're default) changed, mock no longer is executed, thus returning `null`. 

Since we want to support both Sf2 and Sf3, I changed, in 94fce3d, `AssetHelper::getUrl` mock, to inner `Package::getUrl` mock which is directly called by `AssetHelper` and is the same in Sf2 and Sf3.

To check that, I've added a possibility to set specific Symfony version in TravisCI tests (66a3478).

However, right now it's not possible to run Behat with Sf3 as there are more dependencies to be updated for Sf3 to work with entire `ezplatform` package. So, it raises a question - is it enough for now or did I miss something? :)

Review ping @bdunogier @nicolas-bastien @andrerom

andrerom:
@bdunogier you want this in 1.7 as well?

bdunogier:
I don't think it's worth it, @andrerom. It just improves tests, and we won't make eZ Platform 1.7 compatible with sf3 anyway.

bdunogier:
Merged. Can you move the issue forward, @alongosz ?

alongosz:
@bdunogier Thx, done.

andrerom:
@alongosz Testing a bit on [full install with symfony 3.x](https://github.com/andrerom/ezplatform/tree/2.0) and seems there are still issues:
```
  [Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]                                                                       
  The service "ezsystems.platformui.application_config.provider.root_info" has a dependency on a non-existent service "templating.helper.assets".
```

Other Bundles seems to pick `assets.packages` service instead in container builder if the one above is not set, see: https://github.com/codepeak/ImageBundle/blob/0acaadefaf02c02dda44c81cee7dd9fc7cb60540/Services/ImageHandling.php#L137

alongosz:
> Afaik `assets.packages` was added with AssetBundle in Symfony 2.7 so maybe we should simply always use that?

That's right, trying to determine if extra refactoring is needed (tests won't show that due to mocks).
Now, after replacing this service and getting rid of `ServiceNotFoundException` I'm stuck on
```
PHP Fatal error:  Uncaught Symfony\Component\Debug\Exception\UndefinedMethodException: Attempted to call an undefined method named "setSiteAccess" of class "Symfony\Bundle\FrameworkBundle\Routing\Router"
```
when Sf is trying to compile container.

PR #786, opened 2017-01-25
EZP-26897: Display broken when subitems sorted by not displayed column

yannickroger:
Link: https://jira.ez.no/browse/EZP-26897

## Description
When the default sort order is not displayable in the subitem list (for example LocationPath). The view freezes.
This happens because we are trying to apply a CSS class on a non existing column.

## Tests
Manual tests on Chrome/Firefox and unit test.


yannickroger:
ping @dpobel @StephaneDiot

yannickroger:
Done @dpobel

PR #787, opened 2017-01-26
EZP-26835 : Fix file reference in extract translations + update files

nicolas-bastien:
Jira: https://jira.ez.no/browse/EZP-26835

## Description

Simplify sed command to clean file reference + make path relative


dpobel:
while at it, it would be nice to make it executable and to restore the working directory at the end of the script.

yannickroger:
+1 with @dpobel 's feedback

PR #788, opened 2017-01-27
EZP-26751: Value of fields not flagged as Translatable can be updated 

yannickroger:
Link: https://jira.ez.no/browse/EZP-26751

## Description
Field that can not be translated are not flagged as they should on other languages. They also can be updated. This PR fixes that.

not-translatable

## Tests
Manual test and unit test update

## TODO
- [x] Squash commits

### DIFF
--- a/Resources/public/js/views/ez-contenteditformview.js
+++ b/Resources/public/js/views/ez-contenteditformview.js
@@ -71,10 +71,13 @@ YUI.add('ez-contenteditformview', function (Y) {
 
             Y.Object.each(fieldDefinitions, function (def) {
                 var EditView, view,
-                    field = version.getField(def.identifier, languageCode);
+                    field = version.getField(def.identifier, languageCode),
+                    translating;
 
                 if (field) {
                     try {
+                        translating = content.get('mainLanguageCode') !== '' && languageCode !== content.get('mainLanguageCode');
+
                         EditView = Y.eZ.FieldEditView.getFieldEditView(def.fieldType);
 
                         view = new EditView({
@@ -85,6 +88,7 @@ YUI.add('ez-contenteditformview', function (Y) {
                             field: field,
                             config: config,
                             languageCode: languageCode,
+                            translating: translating,
                             user: user,
                         });
                         that._fieldEditViewsByDefinitionId[def.id] = view;
--- a/Resources/public/js/views/ez-fieldeditview.js
+++ b/Resources/public/js/views/ez-fieldeditview.js
@@ -77,6 +77,7 @@ YUI.add('ez-fieldeditview', function (Y) {
                     content: this.get('content').toJSON(),
                     version: this.get('version').toJSON(),
                     contentType: this.get('contentType').toJSON(),
+                    isNotTranslatable: this.get('isNotTranslatable'),
                 },
                 container = this.get('container');
 
@@ -276,6 +277,14 @@ YUI.add('ez-fieldeditview', function (Y) {
             this.after('errorStatusChange', this._errorUI);
 
             this._addDOMEventHandlers(_events);
+
+            this.after('isNotTranslatableChange', function() {
+                if (this.get('isNotTranslatable')) {
+                    this.get('container').addClass('is-not-translatable');
+                }
+            });
+
+            this._set('isNotTranslatable', this.get('translating') && !this.get('fieldDefinition').isTranslatable);
         },
 
         /**
@@ -459,7 +468,29 @@ YUI.add('ez-fieldeditview', function (Y) {
              */
             languageCode: {
                 value: null
-            }
+            },
+
+            /**
+             * True if the field is currently being translated.
+             *
+             * @attribute translating
+             * @type {Boolean}
+             * @required
+             */
+            translating: {},
+
+            /**
+             * True if the field is not translatable
+             *
+             * @attribute isNotTranslatable
+             * @readOnly
+             * @type {Boolean}
+             * @default false
+             */
+            isNotTranslatable: {
+                value: false,
+                readOnly: true,
+            },
         },
 
         /**
--- a/Resources/public/js/views/fields/ez-richtext-editview.js
+++ b/Resources/public/js/views/fields/ez-richtext-editview.js
@@ -50,7 +50,9 @@ YUI.add('ez-richtext-editview', function (Y) {
                 if ( this.get('active') ) {
                     this._initEditor();
                 } else {
-                    this.get('editor').destroy();
+                    if (this.get('editor')) {
+                        this.get('editor').destroy();
+                    }
                 }
             });
             this.after('focusModeChange', this._uiFocusMode);
@@ -157,6 +159,10 @@ YUI.add('ez-richtext-editview', function (Y) {
                     'ezaddcontent', 'widget', 'ezembed', 'ezremoveblock',
                     'ezfocusblock', 'yui3', 'ezpaste',
                 ];
+
+            if (this.get('isNotTranslatable')) {
+                return;
+            }
 
             this._registerExternalCKEditorPlugin('widget', 'widget/');
             this._registerExternalCKEditorPlugin('lineutils', 'lineutils/');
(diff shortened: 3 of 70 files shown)


Resources/public/js/views/ez-contenteditformview.js
```diff
@@ -85,6 +88,7 @@ YUI.add('ez-contenteditformview', function (Y) {
                             field: field,
                             config: config,
                             languageCode: languageCode,
+                            isNotTranslatable: isNotTranslatable,
```
dpobel: not super convinced by the name but I don't have a better idea at the moment

Resources/public/templates/fields/edit/checkbox.hbt
```diff
@@ -1,4 +1,4 @@
-<div class="pure-g ez-editfield-row">
+<div class="pure-g ez-editfield-row {{#if isNotTranslatable}}ez-editfield-not-translatable{{/if}}">
```
dpobel: I would add this class on the view container instead. Also, given it represents a state, it should be name `is-not-translatable` (or something like that depending on the name of this state)

Resources/public/css/theme/views/contenteditform.css
```diff
@@ -24,3 +24,22 @@
 .ez-view-contenteditformview .is-collapsed .fieldgroup-name:after {
     content: "\E604";
 }
+
+.ez-view-contenteditformview .ez-editfield-not-translatable-msg {
```
dpobel: this should not be in contenteditform.css but in fieldedit.css as you actually style a field edit view

Resources/public/js/views/ez-contenteditformview.js
```diff
@@ -71,10 +71,13 @@ YUI.add('ez-contenteditformview', function (Y) {
 
             Y.Object.each(fieldDefinitions, function (def) {
                 var EditView, view,
-                    field = version.getField(def.identifier, languageCode);
+                    field = version.getField(def.identifier, languageCode),
+                    isNotTranslatable;
 
                 if (field) {
                     try {
+                        isNotTranslatable = !(def.isTranslatable || languageCode === content.get('mainLanguageCode'));
```
dpobel: did you check that when creating a Content item ?

Resources/public/js/views/ez-fieldeditview.js
```diff
@@ -276,6 +277,14 @@ YUI.add('ez-fieldeditview', function (Y) {
             this.after('errorStatusChange', this._errorUI);
 
             this._addDOMEventHandlers(_events);
+
+            this.set('isNotTranslatable', this.get('translating') && !this.get('fieldDefinition').isTranslatable);
+
+            this.after('activeChange', Y.bind(function() {
```
dpobel: you should subscribe to `isNotTranslatableChange` event instead

Resources/public/js/views/ez-fieldeditview.js
```diff
@@ -276,6 +277,14 @@ YUI.add('ez-fieldeditview', function (Y) {
             this.after('errorStatusChange', this._errorUI);
 
             this._addDOMEventHandlers(_events);
+
+            this.set('isNotTranslatable', this.get('translating') && !this.get('fieldDefinition').isTranslatable);
```
dpobel: `isNotTranslatable` should be marked as read only since it is a computed value so here you should use `_set` and the attribute should have the read only flag

Resources/public/js/views/fields/ez-richtext-editview.js
```diff
@@ -351,7 +357,11 @@ YUI.add('ez-richtext-editview', function (Y) {
          * @return {Object}
          */
         _getFieldValue: function () {
-            var value = this._getXHTML5EditValue();
+            var value = "";
```
dpobel: most likely you should init it with the field value

yannickroger:
Feedback taken into account @dpobel and PR is complete fo review.

andrerom:
As this is support issue this should be against 1.7 branch right?

yannickroger:
@andrerom I was considering it at first. But the diff is quite big and it kind of changes the behavior so it is more than a bugfix.
@dpobel often referred this as an improvement as it is a border line bug (everything works without this patch, the interface is just misleading).

andrerom:
It is border line, but the size of a change is not what we choose to call it a bug or not, with test coverage, own manual testing, QA story testing, patch release certification, you should be able to be confident in what you are proposing and decide depending on if it can provide customer value when the version they have is clearly missing pieces. Customers are who we work for in the end after all.

But if you truly want to say to support and customers that this is not something we want to help them in with the LTS version they are on, then by all means do so. But then be clear with support directly, and also change issue type so this is no longer on thirdline _(which is the case here)_ ;)

yannickroger:
I just explained why I did this fix on master and not on 1.7. It is only arguments that I am more that happy to discuss with anyone and I am always happy to change my mind (I think I have proven that overtime).

I know it is unit tested and QA. Unit test was the hardest part of the PR.

Regarding public personal attacks (quality of my work, involvement with customers) I won't reply to it as I think it was unfair in the first place.

bdunogier:
The fact remains that this patch is much more of an improvement, or even a new feature, than a bugfix. The feature fixes a bug for a user, but it is still a feature, and we do not backport features as per semver.

Can we get a limited version of the feature into 1.7, that would unblock the user ?

yannickroger:
We can provide a patch that applies on 1.7 for users wanting it, but they would have to reapply it after each update.
Or split the fix so it could be applied only on non translatable fields that they use to have a smaller codeprint.

andrerom:
providing patch but not applying it to 1.7 won't give any value here, my comment was about being clear and decide on bug / improvement, I have updated the comment to more clearly reflect that.

Resources/public/js/views/ez-fieldeditview.js
```diff
@@ -459,7 +468,31 @@ YUI.add('ez-fieldeditview', function (Y) {
...
                 value: null
-            }
+            },
+
+            /**
+             * True if the field is currently being translated.
+             *
+             * @attribute translating
+             * @type {Boolean}
+             * @default null
```
dpobel: you should remove the default value and mark the attribute as required

Resources/public/js/views/ez-fieldeditview.js
```diff
@@ -459,7 +468,31 @@ YUI.add('ez-fieldeditview', function (Y) {
...
+             */
+            translating: {
+                value: null
+            },
+
+            /**
+             * True if the field is not translatable
+             *
+             * @attribute isNotTranslatable
+             * @readonly
```
dpobel: `@readOnly` (unsure if yuidoc understands the `@readonly`)

Resources/public/js/views/ez-fieldeditview.js
```diff
@@ -459,7 +468,31 @@ YUI.add('ez-fieldeditview', function (Y) {
...
+             * @type {Boolean}
+             * @default null
+             */
+            isNotTranslatable: {
+                value: null,
```
dpobel: I would put false as the default value

Resources/public/css/views/fieldedit.css
```diff
@@ -90,3 +90,22 @@
...
+    font-size: 0.8em;
+    display:block;
+}
+
+.is-not-translatable .ez-fielddefinition-name {
+    opacity: 0.3;
+}
+
+.is-not-translatable .ez-editfield-not-translatable:before {
+    content: "\e61b";
```
dpobel: this should be in the theme css

Resources/public/css/views/fields/edit/author.css
```diff
@@ -32,3 +32,7 @@
     display: block;
     margin-top: 3em;
 }
+
+.ez-view-authoreditview.is-not-translatable .ez-field-sublabel {
+    opacity: 0.3;
```
dpobel: all the opacity changes (in this CSS file and the others) should be moved in the theme CSS

yannickroger:
Feedback by dpobel taken into account. Base branch is now 1.7.

sunpietro:
Looks good 👍 it should not break anything in Studio

PR #791, opened 2017-01-27
Replace deprecated templating.helper.assets with assets.packages service

alongosz:
> Follow-up to [EZP-26881](https://jira.ez.no/browse/EZP-26881) and #782 

It turns out there's more to AssetsHelper issue than just unit test. I expected that but couldn't find any relevant info until I tried to compile Symfony 3 container for eZ Platform 2.0.

The `@templating.helper.assets` service has been deprecated and got removed in Symfony `3.x`. It needs to be replaced with `@assets.packages` service.

**TODO**:
- [x] Replace usage of deprecated `@templating.helper.assets` service with `@assets.packages` service (31aef5a).
- [x] Quote `%parameters%` in Yaml as keeping them unquoted is deprecated since Symfony `3.x` (e702486). // small fix, so I included it in the scope of this PR

### DIFF
--- a/ApplicationConfig/Providers/RootInfo.php
+++ b/ApplicationConfig/Providers/RootInfo.php
@@ -5,7 +5,7 @@
 namespace EzSystems\PlatformUIBundle\ApplicationConfig\Providers;
 
 use EzSystems\PlatformUIBundle\ApplicationConfig\Provider;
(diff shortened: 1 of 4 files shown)


andrerom:
Merging for testing with Sf 3.2, failure is ezrobot caused by upgrade of php-cs-fixer.

PR #793, opened 2017-02-02
Fix EZP-26755: Content type edit title is not displayed

bdunogier:
> [EZP-26755](https://jira.ez.no/browse/EZP-26755)

This particular translation string is defined in platform-ui-bundle, but the translation was in repository-forms. It was lost during the regeneration of strings. This copies the translations to platform-ui.


bdunogier:
Sent to QA.

PR #795, opened 2017-02-02
Feature EZP-25552: UDW finder content explorer

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-25793


bdunogier:
Note: it would be really better if Travis could pass before merging those.

PR #796, opened 2017-02-02
EZP-26899: Handle internal links when displaying RichText fields

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26899

# Description

This patch makes sure internal links (`ezlocation://`) links are usable in PlatformUI. If an internal link points to a removed Location or a Location the user is not able to see, the link is just disabled.

Screencast:

[](http://www.youtube.com/watch?v=DsnYWMh2pgs "Click to play on Youtube.com")

# Tests

unit test + manual tests

### DIFF
--- a/Resources/public/js/views/services/plugins/ez-searchplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-searchplugin.js
@@ -311,7 +311,8 @@ YUI.add('ez-searchplugin', function (Y) {
          * generate the sort clauses
          * @param {Integer} e.search.offset the offset for the search result
          * @param {Integer} e.search.limit number of records returned
-         * @param {String} e.resultAttribute the name of attribute that will by updated with search results
+         * @param {Function} [e.callback] callback to call when search is done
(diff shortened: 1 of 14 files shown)


Resources/public/js/views/services/plugins/ez-searchplugin.js
```diff
@@ -324,21 +325,26 @@ YUI.add('ez-searchplugin', function (Y) {
         _doLocationSearch: function (e) {
             var search = Y.merge(e.search),
                 listView = e.target,
-                attrs = {'loadingError': false};
+                attrs = {'loadingError': false},
+                callback = function (error, result, resultCount) {
+                    attrs.loadingError = error ? true : false;
```
dew326: What do you think about writing it like this: `attrs.loadingError = !!error;`?

dpobel: hum it's more a matter of taste than anything but the ternary version is maybe a tiny bit more readable :)

PR #799, opened 2017-02-03
WIP EZP-25552: Starting location

StephaneDiot:
WIP

### DIFF
--- a/Resources/public/js/views/ez-universaldiscoveryview.js
+++ b/Resources/public/js/views/ez-universaldiscoveryview.js
@@ -268,8 +268,8 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
          */
         _updateMethods: function () {
             var visibleMethod = this.get('visibleMethod'),
-                startingLocationId = this.get('startingLocationId');
-
+                startingLocationId = this.get('startingLocationId'),
+                startingLocation = this.get('startingLocation');
             /**
              * Stores a reference to the visible method view
              *
@@ -285,6 +285,7 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
                     'multiple': this.get('multiple'),
                     'loadContent': true,
                     'startingLocationId': startingLocationId,
+                    'startingLocation': startingLocation,
                     'visible': visible,
                     'isSelectable': Y.bind(this.get('isSelectable'), this),
                     'active': this.get('active'),
@@ -502,13 +503,24 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
             },
 
             /**
-             * The location id that the UDW tree will select on start
+             * The location id that the UDW explorer will select on start
              * @attribute startingLocationId
              * @type {String}
              * @default false if there is no starting location
              *
              */
             startingLocationId: {
+                value: false,
+            },
+
+            /**
+             * The location that the UDW will select on start on.
+             * @attribute startingLocation
+             * @type {String}
+             * @default false if there is no starting location
+             *
+             */
+            startingLocation: {
                 value: false,
             },
 
@@ -563,7 +575,8 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
                             multiple: this.get('multiple'),
                             loadContent: true,
                             isAlreadySelected: Y.bind(this._isAlreadySelected, this),
-                            startingLocationId: this.get('startingLocationId')
+                            startingLocationId: this.get('startingLocationId'),
+                            startingLocation: this.get('startingLocation'),
                         }),
                         new Y.eZ.UniversalDiscoverySearchView({
                             bubbleTargets: this,
@@ -571,7 +584,8 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
                             multiple: this.get('multiple'),
                             loadContent: true,
                             isAlreadySelected: Y.bind(this._isAlreadySelected, this),
-                            startingLocationId: this.get('startingLocationId')
+                            startingLocationId: this.get('startingLocationId'),
+                            startingLocation: this.get('startingLocation'),
                         }),
                     ];
                 },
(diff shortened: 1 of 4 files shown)


Resources/public/js/views/services/ez-universaldiscoveryviewservice.js
```diff
@@ -22,6 +22,51 @@ YUI.add('ez-universaldiscoveryviewservice', function (Y) {
...
+         * @param {Function} callback
+         */
+        _load: function (callback) {
+            var startingLocation = new Y.eZ.Location(),
+                path = [];
+
+            if (this.get('parameters').startingLocationId) {
+                startingLocation.set('id', this.get('parameters').startingLocationId);
+                this._loadStartingLocationPath(startingLocation, Y.bind(function(locationPath) {
+                    startingLocation.get('path').unshift(this.get('virtualRootLocation'));
```
dpobel: you can not do that. It seems you also need in addition the virtual root, so you should make sure it's passed to the UDW view alongside the startingLocation

Resources/public/js/views/services/ez-universaldiscoveryviewservice.js
```diff
@@ -22,6 +22,51 @@ YUI.add('ez-universaldiscoveryviewservice', function (Y) {
...
+         * @param {Function} callback
+         */
+        _loadStartingLocationPath: function (startingLocation, callback) {
+            var options = {api: this.get('capi')};
+
+            startingLocation.load(options, function (error) {
+                if (!error) {
+                    startingLocation.loadPath(options, function (error, response) {
+                        if (!error) {
+                            callback(response);
```
dpobel: why calling the callback with `response` while it's not used anywhere

Resources/public/js/views/services/ez-universaldiscoveryviewservice.js
```diff
@@ -22,6 +22,51 @@ YUI.add('ez-universaldiscoveryviewservice', function (Y) {
...
+         * @method _loadStartingLocationPath
+         * @protected
+         * @param {eZ.Location} startingLocation
+         * @param {Function} callback
+         */
+        _loadStartingLocationPath: function (startingLocation, callback) {
+            var options = {api: this.get('capi')};
+
+            startingLocation.load(options, function (error) {
+                if (!error) {
```
dpobel: something has to be done when an error occurs otherwise the user will be stuck in front of the app loader. In case of error while loading the startingLocation or its path, we should probably act as if the UDW was launched without a startingLocationId.

Resources/public/js/views/services/ez-universaldiscoveryviewservice.js
```diff
@@ -22,6 +22,51 @@ YUI.add('ez-universaldiscoveryviewservice', function (Y) {
...
+         * @method _load
+         * @protected
+         * @param {Function} callback
+         */
+        _load: function (callback) {
+            var startingLocation = new Y.eZ.Location(),
+                path = [];
+
+            if (this.get('parameters').startingLocationId) {
+                startingLocation.set('id', this.get('parameters').startingLocationId);
```
dpobel: you should put the app in loading mode before the asynchronous loading

Resources/public/js/views/ez-universaldiscoveryview.js
```diff
@@ -268,8 +268,8 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
          */
         _updateMethods: function () {
             var visibleMethod = this.get('visibleMethod'),
-                startingLocationId = this.get('startingLocationId');
-
+                startingLocationId = this.get('startingLocationId'),
+                startingLocation = this.get('startingLocation');
```
dpobel: you should add the startingLocation attribute definition

Resources/public/js/views/ez-universaldiscoveryview.js
```diff
@@ -563,15 +564,17 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
                             multiple: this.get('multiple'),
                             loadContent: true,
                             isAlreadySelected: Y.bind(this._isAlreadySelected, this),
-                            startingLocationId: this.get('startingLocationId')
+                            startingLocationId: this.get('startingLocationId'),
+                            startingLocation: this.get('startingLocation'),
```
dpobel: same remark as in https://github.com/ezsystems/PlatformUIBundle/pull/799/files#r99341566, the UDW browse view should now have a startingLocation attribute definition

Resources/public/js/views/services/ez-universaldiscoveryviewservice.js
```diff
@@ -22,6 +22,51 @@ YUI.add('ez-universaldiscoveryviewservice', function (Y) {
...
+                this._loadStartingLocationPath(startingLocation, Y.bind(function(locationPath) {
+                    startingLocation.get('path').unshift(this.get('virtualRootLocation'));
+                    this.get('parameters').startingLocation = startingLocation;
+                    callback();
+                }, this));
+            } else {
+                this.get('parameters').startingLocation = this.get('virtualRootLocation');
```
dpobel: https://github.com/ezsystems/PlatformUIBundle/pull/799/files#r99340534 means in that case, startingLocation should be null or something like that

Resources/public/js/views/services/ez-universaldiscoveryviewservice.js
```diff
@@ -22,6 +22,53 @@ YUI.add('ez-universaldiscoveryviewservice', function (Y) {
...
+         */
+        _load: function (callback) {
+            var startingLocation = new Y.eZ.Location(),
+                path = [];
+
+            if (this.get('parameters').startingLocationId) {
+                startingLocation.set('id', this.get('parameters').startingLocationId);
+                this.get('app').set('loading', true);
+                this._loadStartingLocationPath(startingLocation, Y.bind(function() {
+                    startingLocation.get('path').unshift(this.get('virtualRootLocation'));
```
dpobel: you can not do that. `path` belongs to the Location and should not be modified from the outside. To avoid doing that, you should probably pass the virtual root Location to the UDW view in addition to the start location.

PR #805, opened 2017-02-15
Fix EZP-26996: typo in translation for locationview.details.remoteid

wizhippo:


wizhippo:
@dpobel I believe I have done as you asked.

bdunogier:
> @bdunogier / @yannickroger Does this affect the i18n packages?

Yes it does, starting from 1.8, using the new `bin/extract-translations.sh`.
The strings also need to be generated for 1.7, but the script is probably different.

bdunogier:
After talking to Damien, no need to run any script, since the english string only exists in translation files.

What will have to be done is execute `bin/synchronize-translations.sh` from ezplatform after the merge, and commit to ezsystems/ezplatform-i18n after the sync.

dpobel:
perfect @wizhippon, I'm merging and taking care of forward porting :) thanks again

PR #806, opened 2017-02-16
Fix EZP-26997: Finder gradual loading fix

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-26997

## Description

After 100 items Finder  couldn't load more items and the levelview. This was due to the fact that we weren't watching scroll event after loading the 50 first additional items. This is now fixed

## Tests

unit and manually tested


andrerom:
_(target branch should be 1.8, this is a bug afaics)_

PR #809, opened 2017-02-20
EZP-26044: Handle height change and scroll on navigation

StephaneDiot:
jira : https://jira.ez.no/browse/EZP-26044 and https://jira.ez.no/browse/EZP-26470

## Description

The first goal of this fix was to correct the bug where navigation hub was hiding upper part of the screen. After investigation the opposite effect (screen was moved down) was also found in some cases.
To fix those effects I had to correct https://jira.ez.no/browse/EZP-26470 (which is basically that navigating between diferent location view were keeping the scroll offset instead of reseting it). So this pull request is fixing both https://jira.ez.no/browse/EZP-26044 and https://jira.ez.no/browse/EZP-26470

## Tests

Unit and manually tested

### DIFF
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -736,6 +736,16 @@ YUI.add('ez-platformuiapp', function (Y) {
                         update: true,
                         render: true
                     });
+
+                    if (window && app.get('scrollToTop')) {
+                        // Scroll to the top of the page. The timeout ensures that the
+                        // scroll happens after navigation begins, so that the current
+                        // scroll position will be restored if the user clicks the back
+                        // button.
+                        setTimeout(function () {
+                            window.scroll(0, 0);
+                        }, 1);
+                    }
                 };
 
             if ( this.get('activeView') ) {
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -737,7 +737,7 @@ YUI.add('ez-platformuiapp', function (Y) {
                         render: true
                     });
 
-                    if (window && app.get('scrollToTop')) {
+                    if (app.get('scrollToTop')) {
                         // Scroll to the top of the page. The timeout ensures that the
                         // scroll happens after navigation begins, so that the current
                         // scroll position will be restored if the user clicks the back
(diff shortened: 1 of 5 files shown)


Resources/public/js/views/ez-locationviewview.js
```diff
@@ -31,6 +31,7 @@ YUI.add('ez-locationviewview', function (Y) {
                 Y.Array.each(this.get('tabs'), function (t) {
                     t.set('active', this.get('active'));
                 }, this);
+                window.scrollTo(0,0);
```
dpobel: as far as I get it, it's not the right way (tm) to fix the scroll issue. First, this should be made once and for all views not only the LocationView. Second, you should probably take example on the YUI code doing that (see https://yuilibrary.com/yui/docs/api/files/pjax_js_pjax-base.js.html#l305)

StephaneDiot:
updated @dpobel

Tests/js/apps/assets/ez-platformuiapp-tests.js
```diff
@@ -199,6 +199,37 @@ YUI.add('ez-platformuiapp-tests', function (Y) {
...
+        },
+
+        tearDown: function () {
+            
+            this.app.destroy();
+            delete this.app;
+        },
+
+        "Should scroll to top and left of the window when showing main view": function () {
+            var that = this;
```
dpobel: please no more `that` :)

Tests/js/apps/assets/ez-platformuiapp-tests.js
```diff
@@ -199,6 +199,37 @@ YUI.add('ez-platformuiapp-tests', function (Y) {
...
+
+        setUp: function () {
+            this.app = new Y.eZ.PlatformUIApp({
+                container: '.app',
+                viewContainer: '.view-container'
+            });
+        },
+
+        tearDown: function () {
+
```
dpobel: cs

Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -736,6 +736,16 @@ YUI.add('ez-platformuiapp', function (Y) {
                         update: true,
                         render: true
                     });
+
+                    if (window && app.get('scrollToTop')) {
```
dpobel: no need to test `window`, it's always there

yannickroger:
Is there a reason why base branch is not 1.7?

Tests/js/views/assets/ez-locationviewview-tests.js
```diff
@@ -1093,7 +1093,6 @@ YUI.add('ez-locationviewview-tests', function (Y) {
         },
     });
 
-
```
yannickroger: nitpick, not needed.

PR #811, opened 2017-02-22
EZP-27008: Use UDW to browse content from discovery bar

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-27008

## Description

Goal is to replace our old fashioned content tree by a brand new UDW explorer to be able to explore from the discovery bar. :call_me_hand: 

## Screencast 

https://youtu.be/eBAuB9TXMU4

## Tests

- Unit and manual tests

### DIFF
--- a/Resources/public/js/views/actions/ez-treeactionview.js
+++ b/Resources/public/js/views/actions/ez-treeactionview.js
@@ -10,6 +10,9 @@ YUI.add('ez-treeactionview', function (Y) {
      * @method ez-treeactionview
      */
     Y.namespace('eZ');
+    
+    console.log('[DEPRECRATED] eZ.TreeActionView is deprecated');
+    console.log('[DEPRECRATED] it will be removed from PlatformUI 2.0');
 
     var NAVIGATION_MAX_HEIGHT = 120;
 
@@ -36,6 +39,7 @@ YUI.add('ez-treeactionview', function (Y) {
          *
          * @method _uiExpand
          * @protected
+         * @deprecated
          * @param {Object} e event facade
          */
         _uiExpand: function (e) {
@@ -50,6 +54,7 @@ YUI.add('ez-treeactionview', function (Y) {
          *
          * @method _handleClickOutside
          * @param {Boolean} expanded state
+         * @deprecated
          * @protected
          */
         _handleClickOutside: function (expanded) {
@@ -66,6 +71,7 @@ YUI.add('ez-treeactionview', function (Y) {
          * Sets the maximum height of the expandable area of the view
          *
          * @method _uiSetMaxHeight
+         * @deprecated
          * @protected
          */
         _uiSetMaxHeight: function () {
@@ -104,6 +110,7 @@ YUI.add('ez-treeactionview', function (Y) {
          *
          * @method _uiNavigate
          * @param {EventFacade} e
+         * @deprecated
          * @protected
          */
         _uiNavigate: function (e) {
@@ -114,6 +121,7 @@ YUI.add('ez-treeactionview', function (Y) {
          * Toggles the expanded state
          *
          * @method _toggleExpanded
+         * @deprecated
          * @protected
          */
         _toggleExpanded: function () {
@@ -126,6 +134,7 @@ YUI.add('ez-treeactionview', function (Y) {
              *
              * @attribute tree
              * @type eZ.ContentTree
+             * @deprecated
              * @writeOnce
              */
             tree: {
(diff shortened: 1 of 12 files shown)


Resources/public/js/views/actions/ez-treeactionview.js
```diff
@@ -22,20 +22,13 @@ YUI.add('ez-treeactionview', function (Y) {
      * @extends eZ.ButtonActionView
      */
     Y.eZ.TreeActionView = Y.Base.create('treeActionView', Y.eZ.ButtonActionView, [Y.eZ.Expandable], {
```
dpobel: looking at your patch, I'm starting to think that it does not make any sense to not deprecate this component as well. so I think, this one should also be deprecated completely and in the discovery view, it should be replaced by a raw ButtonActionView

Resources/public/js/views/ez-discoverybarview.js
```diff
@@ -54,8 +54,8 @@ YUI.add('ez-discoverybarview', function (Y) {
                             label: Y.eZ.trans('discoverybar.search', {}, 'bar'),
                             priority: 900
                         }),
-                        new Y.eZ.TreeActionView({
-                            actionId: "tree",
+                        new Y.eZ.ButtonActionView({
+                            actionId: "find",
```
dpobel: `browse` would be more suited

Resources/public/js/views/services/ez-discoverybarviewservice.js
```diff
@@ -20,10 +20,59 @@ YUI.add('ez-discoverybarviewservice', function (Y) {
...
+        _fireContentDiscover: function () {
+            var startingLocationId;
+
+            if (this._isLocationViewDisplayed()) {
+               startingLocationId = this.get('app').get('activeView').get('location').get('id');
+            }
+
+            this.fire('contentDiscover', {
+                config: {
+                    title: 'Find location', //provisional
```
dpobel: this really needs to be translated :)

Resources/public/js/views/services/ez-discoverybarviewservice.js
```diff
@@ -20,10 +20,59 @@ YUI.add('ez-discoverybarviewservice', function (Y) {
...
+
+        /**
+         * Navigate to the selected location.
+         *
+         * @method _navigateToLocation
+         * @param {eventFacade} e
+         * @protected
+         */
+        _navigateToLocation: function (e) {
+            this.get('app').navigateTo('viewLocation',
```
dpobel: cs:

```js
this.get('app').navigateTo('viewLocation', {
    id: e.selection.location.get('id'),
    languageCode: e.selection.content.get('mainLanguageCode')
});
```

Tests/js/views/services/assets/ez-discoverybarviewservice-tests.js
```diff
@@ -88,7 +89,147 @@ YUI.add('ez-discoverybarviewservice-tests', function (Y) {
...
+
+            Y.eZ.Location = Y.Base.create('location', Y.Base, [], {}, {
+                ATTRS: {
+                    id: {
+                        value: this.locationId
+                    }
+                }
+            });
+
+            this.location = new Y.eZ.Location();
```
dpobel: you should just use Y.eZ.Base (and remove the definition of Y.eZ.Location)

Tests/js/views/services/assets/ez-discoverybarviewservice-tests.js
```diff
@@ -88,7 +89,147 @@ YUI.add('ez-discoverybarviewservice-tests', function (Y) {
...
+                    "startingLocationId should be undefined"
+                );
+            }, this));
+            this.service.fire('findAction');
+
+            Assert.isTrue(contentDiscoverFired, 'contentDiscover should be fired');
+        },
+
+        "Should navigate to the selected location when content is discovered": function () {
+            var locationMock = new Y.Mock(),
```
dpobel: using mocks here does not bring much, you should also use a Y.eZ.Base

Tests/js/views/services/assets/ez-discoverybarviewservice-tests.js
```diff
@@ -88,7 +89,147 @@ YUI.add('ez-discoverybarviewservice-tests', function (Y) {
...
+                method: 'get',
+                args: ["activeView"],
+                callCount: 2,
+                returns: this.activeView
+            });
+
+            Y.Mock.expect(locationMock, {
+                method: 'get',
+                args: ['id'],
+                returns: this.locationId
```
dpobel: you use the same location id for both the starting Location and the picked Location

StephaneDiot:
updated @dpobel

PR #812, opened 2017-02-27
EZP-26068: Fix Google Maps to inject api key

andrerom:
> Issue: https://jira.ez.no/browse/EZP-26068
> Depends on: https://github.com/ezsystems/ezpublish-kernel/pull/1820

Q:
- Is it ok to use dependency injection on method instead of setApiKey() as was originally discussed? Only value I see with setter injection is if there is a possibility to avoid views having to deal with this logic in the future, but this optional argument does not hinder that.
- Which tests should be adjusted on the views?
- And how can I avoid the jshint error below given the identifier comes from yml config _(I afaik can't inject the key only unless I create a custom application config class just for the google key)_

### DIFF
--- a/Resources/public/js/services/ez-googlemapapiloader.js
+++ b/Resources/public/js/services/ez-googlemapapiloader.js
@@ -14,7 +14,7 @@ YUI.add('ez-googlemapapiloader', function (Y) {
 
     var EVENT_MAP_API_READY = 'mapAPIReady',
         EVENT_MAP_API_FAILED = 'mapAPIFailed',
-        GMAP_JSONP_URI = 'https://maps.googleapis.com/maps/api/js?v=3.exp&sensor=false&callback={callback}';
+        GMAP_JSONP_URI = 'https://maps.googleapis.com/maps/api/js?callback={callback}';
 
     /**
      * Google Maps API loader. It fires the `mapAPIReady` event when the Google
@@ -69,8 +69,9 @@ YUI.add('ez-googlemapapiloader', function (Y) {
          * `mapAPIReady` or `mapAPIFailed` depending on results
          *
          * @method load
+         * @param String [key] to use for api calls, can be omitted but then map might not work
          */
-        load: function () {
+        load: function (key) {
             var request;
 
             if (this._isLoading) {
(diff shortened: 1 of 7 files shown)


andrerom:
@yannickroger / @StephaneDiot any input on the open questions? :)

Resources/public/js/services/ez-googlemapapiloader.js
```diff
@@ -69,8 +69,9 @@ YUI.add('ez-googlemapapiloader', function (Y) {
          * `mapAPIReady` or `mapAPIFailed` depending on results
          *
          * @method load
+         * @param String key to use for api calls, can be omitted but then map might not work
```
dpobel: should be marked as optional (`@param String [key]`)

Resources/public/js/services/ez-googlemapapiloader.js
```diff
@@ -69,8 +69,9 @@ YUI.add('ez-googlemapapiloader', function (Y) {
...
          */
-        load: function () {
+        load: function ( key ) {
```
dpobel: CS: no space around the parameter

dpobel: (same in ez-maplocation-editview.js and ez-maplocation-view.js and in -tests.js)

Resources/public/js/views/fields/ez-maplocation-editview.js
```diff
@@ -63,9 +63,10 @@ YUI.add('ez-maplocation-editview', function (Y) {
          * @method initializer
          */
         initializer: function () {
-            var mapLoader = this.get('mapAPILoader');
+            var mapLoader = this.get('mapAPILoader'),
+                config = this.get('config');
 
-            mapLoader.load();
```
dpobel: to avoid the jshint error, you can write:
```js
/*jshint camelcase: false */
mapLoader.load(config.apiKeys.google_maps || '');
/*jshint camelcase: true */
```

andrerom:
>  For the view tests:

Any hints on how I set different config to test if it is used correctly?

andrerom:
> Any hints on how I set different config to test if it is used correctly?

@dpobel / @yannickroger / @StephaneDiot ?

dpobel:
@andrerom in ez-maplocation-view-tests.js, this is quite easy, each test in `loadTest` test case creates its own view instance, so you can add a test where `Y.eZ.MapLocationView` receives a config with an API key and where you change the `loaderMock#load` expectation to receive the corresponding key.

As written above, ez-maplocation-editview-tests.js is a mess :-/ but you can do pretty much the same in `APILoadingTest` except that you'll have to deal the *global* monkey patching of `Y.eZ.GoogleMapAPILoader.prototype.load`...

andrerom:
Added use of key in some of the existing view test cases now, so that should complete the coverage with the existing `ez-googlemapapiloader-tests.js` coverage.

Also cleaned up the code a bit to for instance avoid the whole js style issue for instance.

Review ping @dpobel @yannickroger @StephaneDiot

PR #819, opened 2017-03-07
EZP-25478: Restore items whose original location has been deleted

yannickroger:
Link: https://jira.ez.no/browse/EZP-25478

## Description
When a subtree was deleted, in the trash, it was not possible to restore a subitem of this subtree because its parent had been deleted. With this patch, you can now select a new parent for these items in the trash.

Screencast: https://drive.google.com/file/d/0B5amepf6EkXbTTduR1BuaHJmWmc/view

## Tests
Manual and unit tests

### DIFF
--- a/Resources/public/js/models/ez-trashitemmodel.js
+++ b/Resources/public/js/models/ez-trashitemmodel.js
@@ -39,17 +39,22 @@ YUI.add('ez-trashitemmodel', function (Y) {
         },
 
         /**
-         * Restores the item to it's original location
+         * Restores the item to it's original location or to a given one if specified
          *
          * @method restore
          * @param {Object} options the required for the update
          * @param {Object} options.api (required) the JS REST client instance
+         * @param {Object} [options.destination] if provided, locationId under which the item will be restored.
          * @param {Function} callback a callback executed when the operation is finished
          */
         restore: function (options, callback) {
             var contentService = options.api.getContentService();
 
-            contentService.recover(this.get('id'), callback);
+            if (options.destination) {
+                contentService.recover(this.get('id'), options.destination, callback);
+            } else {
+                contentService.recover(this.get('id'), callback);
+            }
         },
 
     }, {
--- a/Resources/public/js/views/ez-trashview.js
+++ b/Resources/public/js/views/ez-trashview.js
@@ -26,6 +26,9 @@ YUI.add('ez-trashview', function (Y) {
             '.ez-trashitem-box': {
                 'change': '_updateTrashBarButtons'
             },
+            '.ez-trashitem-restore': {
+                'tap': '_restoreSingleTrashItem'
+            },
         },
 
         initializer: function () {
@@ -96,7 +99,48 @@ YUI.add('ez-trashview', function (Y) {
         },
 
         /**
-         * Restores the selected trash items
+         * Finds a trash item in the item list and returns it
+         *
+         * @private
+         * @method _findTrashItems
+         * @param {String} trashItemId
+         * @return {Array} of TrashItem
+         */
+        _findTrashItems: function (trashItemId) {
+            var resultArray = [];
+            Y.Array.some(this.get('trashItems'), function (trashItem) {
+                if (trashItemId === trashItem.item.get('id')) {
+                    resultArray.push(trashItem.item);
+                    return true;
+                }
+            });
+
+            return resultArray;
+        },
+
+        /**
+         * Fires the `restoreItems` event
+         *
+         * @private
+         * @method _fireRestoreItems
+         * @param {Array} trashItems Trash items to be restored
+         * @param {String} [destination] if provided, Location Id where it will be restored
+         */
+        _fireRestoreItems: function (trashItems, destination) {
+            /**
+            * Fired to restore the selected items
+            * @event restoreItems
+            * @param {Array} trashItems Trash items to be restored
+            * @param {String} [destination] if provided, Location Id where it will be restored
+            */
+            this.fire('restoreItems', {
+                trashItems: trashItems,
+                destination: destination,
+            });
+        },
+
+        /**
+         * Restores selected trash items
          *
          * @private
          * @method _restoreTrashItems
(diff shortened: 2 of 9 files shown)


Resources/public/js/views/ez-trashview.js
```diff
@@ -122,6 +125,47 @@ YUI.add('ez-trashview', function (Y) {
...
+
+            e.preventDefault();
+            this.fire('contentDiscover', {
+                config: {
+                    title: Y.eZ.trans('trash.ancestor.to.select', {}, 'trash'),
+                    multiple: false,
+                    contentDiscoveredHandler: Y.bind(function (e) {
+                        var trashItems = [];
+
+                        Y.Array.some(this.get('trashItems'), function (trashItem) {
```
dpobel: I would put that in a dedicated method to avoid duplicating the code here (compared to `_restoreTrashItems`)

Resources/public/js/models/ez-trashitemmodel.js
```diff
@@ -39,17 +39,22 @@ YUI.add('ez-trashitemmodel', function (Y) {
         },
 
         /**
-         * Restores the item to it's original location
+         * Restores the item to it's original location or to a given one if specified
          *
          * @method restore
          * @param {Object} options the required for the update
          * @param {Object} options.api (required) the JS REST client instance
+         * @param {Object} options.destination (optional) locationId under which the item will be restored.
```
dpobel: instead of writing `@param {Object} options.destination (optional)`, write `@param {Object} [options.destination]`

Resources/public/js/views/ez-trashview.js
```diff
@@ -122,6 +125,47 @@ YUI.add('ez-trashview', function (Y) {
...
+                            if (trashItemId === trashItem.item.get('id')) {
+                                trashItems.push(trashItem.item);
+                                return true;
+                            }
+                        });
+
+                        /**
+                         * Fired to restore the selected items
+                         * @event restoreItems
```
dpobel: please document the event parameters.

Tests/js/models/assets/ez-trashitemmodel-tests.js
```diff
@@ -257,6 +257,38 @@ YUI.add('ez-trashitemmodel-tests', function (Y) {
 
             Mock.verify(this.contentServiceMock);
         },
+
+        "Should call the api on restore with destination provided": function () {
+            var restoreCallback = function () {},
+                destination = {};
+
+            this.options = {api: this.apiMock, destination: destination};
```
dpobel: why not just using a local variable ?

Tests/js/models/assets/ez-trashitemmodel-tests.js
```diff
@@ -257,6 +257,38 @@ YUI.add('ez-trashitemmodel-tests', function (Y) {
...
+
+            Mock.expect(this.contentServiceMock, {
+                method: "recover",
+                args: [this.id, destination, Mock.Value.Function],
+                run: Y.bind(function (id, options, callback) {
+                    Assert.areSame(
```
dpobel: missing assertions on options

Tests/js/views/assets/ez-trashview-tests.js
```diff
@@ -269,6 +270,65 @@ YUI.add('ez-trashview-tests', function (Y) {
 
             this.view.fire('whatever:restoreTrashItemsAction');
         },
+
+        "Should fire `restoreItems` with destination when restoring a single item": function () {
```
dpobel: this test should be split in 3 tests:

* run the UDW
* allow to pick only container
* fire restoreItems with the expected parameters

Tests/js/models/assets/ez-trashitemmodel-tests.js
yannickroger: actually, the mistake here is that `options` should be called `destination` and it is already asserted automatically in the `args`.

yannickroger:

Resources/public/js/views/ez-trashview.js
```diff
@@ -96,6 +99,44 @@ YUI.add('ez-trashview', function (Y) {
...
 
         /**
+         * Finds a trash item in the item list and adds it to a result array
+         *
+         * @private
+         * @method _findTrashItems
+         * @param {String} trashItemId
+         * @param {Array} resultArray Item will be added to this array if it is found
+         */
+        _findTrashItems: function (trashItemId, resultArray) {
```
dpobel: it's weird to have the result in a parameter of such method, why not returning the result ?

yannickroger:

Resources/public/js/views/ez-trashview.js
```diff
@@ -106,19 +147,40 @@ YUI.add('ez-trashview', function (Y) {
...
-            /**
-             * Fired to restore the selected items
-             * @event restoreItems
-             */
-            this.fire('restoreItems', {trashItems: trashItems});
+            this._fireRestoreItems(trashItems);
+        },
+
+        /**
+         * Restores the selected (using UDW) trash items
```
StephaneDiot: 'trash item" as your UDW is configured with `multiple: false`

Resources/public/js/views/ez-trashview.js
```diff
@@ -106,19 +147,40 @@ YUI.add('ez-trashview', function (Y) {
...
+                config: {
+                    title: Y.eZ.trans('trash.ancestor.to.select', {}, 'trash'),
+                    multiple: false,
+                    contentDiscoveredHandler: Y.bind(function (e) {
+                        var trashItems = [];
+
+                        this._findTrashItems(trashItemId, trashItems);
+
+                        this._fireRestoreItems(trashItems, e.selection.location.get('id'));
+
```
StephaneDiot: Not sure you really need this empty line (maybe same above)

Resources/public/templates/trash.hbt
```diff
@@ -43,6 +43,11 @@
                             </ul>
                             {{else}}
                             <span class="ez-trashview-info-message">{{translate 'trash.ancestors' 'trash'}}</span>
+                            <button
+                                data-trash-item-id="{{item.id}}"
+                                class="ez-trashitem-restore pure-button ez-font-icon ez-button">
```
StephaneDiot: I'm not sure you need ez-font-icon here.

yannickroger:
Feedback by @StephaneDiot taken into account, just waiting for your go @dpobel

PR #823, opened 2017-03-20
EZP-27050: Draft is kept after editing and publishing user

yannickroger:
Link: https://jira.ez.no/browse/EZP-27050

## Description
When editing and publishing a new draft, a draft is left behind.
See the comment in the code for details.

Only drawback of this workaround is that the version number is bumped by 2 instead of two.

The cleaner option (or a follow up?) would be to improve the API to be able to update a given version of a user. It should not be too complicated there is more BC concern.

## Tests
Manual test

### DIFF
--- a/Controller/Rest/ContentController.php
+++ b/Controller/Rest/ContentController.php
@@ -74,6 +74,15 @@ class ContentController extends Content
             $this->mapRequestToUserUpdateStruct($request)
         );
 
+        /*
+         * EZP-27050: Deleting the draft automatically created when entering the proxy.
+         * When updating the user with the API a new draft is created and used.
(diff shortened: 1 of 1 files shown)


yannickroger:
ezrobot is fixed.
Ready for review.

bdunogier:
Feedback after looking into it with Yannick: it should instead be fixed by proxying the REST call that creates a user draft (POST to `.../currentversion`). This call for user content items should be intercepted, and respond with the currently published version id.

The REST call that updates the user will ignore the version number (it already does), and the user will be updated & published without an extra draft.

yannickroger:
closed in favor of #825

PR #825, opened 2017-03-21
Fix EZP-27050: Draft is kept after editing and publishing user

yannickroger:
Link: https://jira.ez.no/browse/EZP-27050

## Description
When editing and publishing a new draft, a draft is left behind.

This patch makes sure a draft is not created when we edit a user.

## Tests
Manual test


pcardiga:
From the QA and functional point of view, this patch seems to fix the problem. It has also the advantage over the previous and [closed PR #823](https://github.com/ezsystems/PlatformUIBundle/pull/823) of not bumping versions 2 by 2.

PR #826, opened 2017-03-23
EZP-27113: Add edit button to sub-item list

yannickroger:
Link: https://jira.ez.no/browse/EZP-27113

## Description
Adds the possibility to edit a content from the sub item list. If a draft is present, the draft conflict screen will be displayed.

Screencast: https://drive.google.com/file/d/0B5amepf6EkXbeUpQNVltZk9hQm8/view

## Tests
Manual test and updated unit test

### DIFF
--- a/Tests/js/views/ez-searchlistview.html
+++ b/Tests/js/views/ez-searchlistview.html
@@ -60,12 +60,16 @@
                 fullpath: "../../../../Resources/public/js/extensions/ez-loadmorepagination.js"
             },
             "ez-subitemlistitemview": {
-                requires: ['ez-templatebasedview', 'ez-translateproperty'],
+                requires: ['ez-templatebasedview', 'ez-translateproperty', 'ez-draftconflict'],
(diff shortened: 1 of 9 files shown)


Tests/js/views/ez-searchlistview.html
```diff
@@ -60,13 +60,17 @@
                 fullpath: "../../../../Resources/public/js/extensions/ez-loadmorepagination.js"
             },
             "ez-subitemlistitemview": {
-                requires: ['ez-templatebasedview', 'ez-translateproperty'],
+                requires: ['ez-templatebasedview', 'ez-translateproperty', 'ez-draftconflict'],
```
dpobel: why this change ?

yannickroger: It won't load the "ez-subitemlistitemview"  without the dependency on the extension (added in this PR) and the test will fail.

dpobel: ah I misread the previous line :) Need more coffee syndrome it seems ;)

PR #828, opened 2017-03-29
Fix EZP-27147: Not translatable binary fields can be updated using drag&drop

yannickroger:
Link: https://jira.ez.no/browse/EZP-27147

## Description
When fixing the fields in #788 I forgot to prevent from drag and dropping on binary fields. So goal of this PR is to fix that.

## Tests
Manual test and completed unit tests

### DIFF
--- a/Resources/public/js/views/fields/ez-binarybase-editview.js
+++ b/Resources/public/js/views/fields/ez-binarybase-editview.js
@@ -240,7 +240,7 @@ YUI.add('ez-binarybase-editview', function (Y) {
          */
         _base64ToFileStruct: function (file, e) {
             var base64 = e.target.result.replace(/^.*;base64,/, '');
-            
+
             this._set('file', this._createFileStruct(file, base64));
             e.target.onload = undefined;
         },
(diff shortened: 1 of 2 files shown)


Resources/public/js/views/fields/ez-binarybase-editview.js
```diff
@@ -357,6 +357,9 @@ YUI.add('ez-binarybase-editview', function (Y) {
          */
         _prepareDrop: function (e) {
             e.preventDefault();
+            if (this.get('isNotTranslatable')) {
```
dpobel: I would do this check before preventing the default behavior and not calling `e.preventDefault()` if drag and drop is not handled

Resources/public/js/views/fields/ez-binarybase-editview.js
```diff
@@ -394,6 +397,10 @@ YUI.add('ez-binarybase-editview', function (Y) {
             var files = e._event.dataTransfer.files;
 
             e.preventDefault();
+            if (this.get('isNotTranslatable')) {
```
dpobel: same as above

Resources/public/js/views/fields/ez-binarybase-editview.js
yannickroger: If I do that, the browsers reads the image as a file (instead of the page) `file://myfile.png`

dpobel: that's the default browser's behavior when you drop an image in a page that does not handle the drop event. And indeed, you are right, it makes sense to avoid it, so you can keep the code as it is.

PR #827, opened 2017-03-27
How to make a clean Pull Request

StephaneDiot:
## Description

As a punishment for being exceptionally late in one my ~~'slave'~~ ~~'junior'~~ 'lead printer' tasks, this is my 2 cents of contributing documentation on how to contribute and make a clean and pretty pull request.

PR #830, opened 2017-03-31
Fix EZP-27174: RichText "This field is required" message not translated

dpobel:
JIRA: https://jira.ez.no/browse/EZP-27174

# Description

This message was simply forgotten...

# Test

manual test


yannickroger:
You should update the unit test and review #827 👿

PR #833, opened 2017-04-04
Fix EZP-27144: Current location not viewable when content browsing

StephaneDiot:
jira : https://jira.ez.no/browse/EZP-27144

## Description

When a user wanted to browse with the finder the current location where he currently was this was resulting in an infinite loading. Now the user can only select another location than the one he is on ( or cancel to stay on it obviously )

## Screencast

https://youtu.be/w5M2Wfygc_8
## Tests

Manually and unit tested

### DIFF
--- a/Resources/public/js/views/services/ez-discoverybarviewservice.js
+++ b/Resources/public/js/views/services/ez-discoverybarviewservice.js
@@ -60,6 +60,9 @@ YUI.add('ez-discoverybarviewservice', function (Y) {
                     startingLocationId: startingLocationId,
                     minDiscoverDepth: rootDepth,
                     confirmLabel: Y.eZ.trans('view.content.label', {}, 'bar'),
+                    isSelectable: function (contentStruct) {
+                        return startingLocationId !== contentStruct.location.get('id');
+                    }
                 },
             });
         },
--- a/Tests/js/views/services/assets/ez-discoverybarviewservice-tests.js
+++ b/Tests/js/views/services/assets/ez-discoverybarviewservice-tests.js
@@ -131,10 +131,14 @@ YUI.add('ez-discoverybarviewservice-tests', function (Y) {
         },
         
         "Should fire contentDiscover with a starting location on browseAction event": function () {
-            var contentDiscoverFired = false;
+            var contentDiscoverFired = false,
+                otherLocation = new Y.Base();
 
+            otherLocation.set('id', 'AnyOtherId');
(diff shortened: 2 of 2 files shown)


Resources/public/js/views/services/ez-discoverybarviewservice.js
```diff
@@ -60,6 +60,12 @@ YUI.add('ez-discoverybarviewservice', function (Y) {
                     startingLocationId: startingLocationId,
                     minDiscoverDepth: rootDepth,
                     confirmLabel: Y.eZ.trans('view.content.label', {}, 'bar'),
+                    isSelectable: function(contentStruct) {
+                        if (startingLocationId === contentStruct.location.get('id')) {
```
dpobel: can be simplified (and cs issue, missing space after `function`):

```js
isSelectable: function (contentStruct) {
    return startingLocationId !== contentStruct.location.get('id');
}
```

Tests/js/views/services/assets/ez-discoverybarviewservice-tests.js
```diff
@@ -131,11 +131,24 @@ YUI.add('ez-discoverybarviewservice-tests', function (Y) {
...
+                otherLocation = new Y.Base();
 
+            otherLocation.set('id', 'AnyOtherId');
             this.service.on('contentDiscover', Y.bind(function (e) {
                 contentDiscoverFired = true;
 
+                Assert.isFunction(e.config.isSelectable, "config should have a function named isSelectable");
+
+                Y.Assert.isTrue(
+                    e.config.isSelectable({location: otherLocation}),
```
dpobel: please add 2 new test cases that would call the `isSelectable` function and make sure the starting Location can not be selected and any other Location is selectable.

StephaneDiot:
Updated ping @dpobel @yannickroger

Resources/public/js/views/services/ez-discoverybarviewservice.js
```diff
@@ -60,6 +60,9 @@ YUI.add('ez-discoverybarviewservice', function (Y) {
                     startingLocationId: startingLocationId,
                     minDiscoverDepth: rootDepth,
                     confirmLabel: Y.eZ.trans('view.content.label', {}, 'bar'),
+                    isSelectable: function(contentStruct) {
```
dpobel: space still missing after `function` ...

PR #836, opened 2017-04-07
EZP-26951: Duplicated messages in error messages

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26951

# Description

2 different strings are used for the exact same purpose. This patch makes sure, we only use one.

Open question: is it OK BC wise to remove a translation or should we still keep it in the translation file ?

# Tests

unit tests


yannickroger:
If you want to be on the safe side with BC, have the delete commit only on 1.9

dpobel:
@bdunogier any opinion on the open question ?

bdunogier:
Better not to remove it. We need to study that better, so I'll play it safe.

PR #837, opened 2017-04-07
Fix EZP-27059: UDW should work without correct visible method

StephaneDiot:
JIRA: https://jira.ez.no/browse/EZP-27059

## Description

We can now launch the UDW with an 'old' and/or 'wrong' visible method set. This will now use the defaultVisibleMethod (currently 'finder')

## Tests

Unit and manually tested

### DIFF
--- a/Resources/public/js/views/ez-universaldiscoveryview.js
+++ b/Resources/public/js/views/ez-universaldiscoveryview.js
@@ -279,38 +279,58 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
          * @protected
          */
         _updateMethods: function () {
-            var visibleMethod = this.get('visibleMethod'),
-                startingLocationId = this.get('startingLocationId'),
+            var startingLocationId = this.get('startingLocationId'),
                 startingLocation = this.get('startingLocation'),
                 minDiscoverDepth = this.get('minDiscoverDepth'),
                 virtualRootLocation = this.get('virtualRootLocation');
 
-            /**
-             * Stores a reference to the visible method view
-             *
-             * @property _visibleMethodView
-             * @protected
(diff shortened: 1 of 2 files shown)


Resources/public/js/views/ez-universaldiscoveryview.js
```diff
@@ -659,6 +679,17 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
             },
 
             /**
+             * The identifier of the default visible method
+             *
+             * @attribute defaultVisibleMethod
+             * @type String
+             * @default 'finder'
+             */
+            defaultVisibleMethod: {
```
dpobel: you should not introduce an attribute for that. First because, we don't want that to be settable externally and second because the default value is already expressed in `visibleMethod` attribute

Resources/public/js/views/ez-universaldiscoveryview.js
```diff
@@ -279,38 +279,58 @@ YUI.add('ez-universaldiscoveryview', function (Y) {
...
+            }
+        },
+
+        /**
+         * Return the visible method, if a wrong or unknown one is setted, it will return the default one.
+         *
+         * @method _getCorrectVisibleMethod
+         * @protected
+         */
+        _getCorrectVisibleMethod: function () {
```
dpobel: this name does not feel right. Also, on the approach, instead of the double iteration on the available methods, why don't you post check that we found a visible method, and if none was found, you take the first one (or the one indicated in the default value of the `visibleMethod` attribute) ?

dpobel:
(and the PR should be created against 1.7 as the bug is also valid in this version)

StephaneDiot:
ping @dpobel

StephaneDiot:
closed in favor of https://github.com/ezsystems/PlatformUIBundle/pull/839

PR #839, opened 2017-04-10
Fix EZP-27059: UDW should work without correct visible method

StephaneDiot:
JIRA: https://jira.ez.no/browse/EZP-27059

## Description

We can now launch the UDW with an 'old' and/or 'wrong' visible method set. This will now use the defaultVisibleMethod (currently 'finder')

## Tests

Unit and manually tested


yannickroger:
You should fix conflicts

PR #840, opened 2017-04-10
Fix EZP-27143 Fixed custom UDW to work with content on the fly.

StephaneDiot:
jira : https://jira.ez.no/browse/EZP-27143

## Description

This PR fixes 2 things:
- You can now choose the destination of the content you want to create without raising an error. This is done by using the mainLanguageCode from the contentInfo instead of the content ( which wasn't setted in COF )
- The parameters of the UDW are now correctly reseted when launching a new instance of the UDW when we select the location where we want to create the content. Basically this allows to not have the mininimum depth root or a custom label when selecting the location.

## Tests

Manually and unit tested

### DIFF
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -618,6 +618,8 @@ YUI.add('ez-platformuiapp', function (Y) {
             });
             if ( createView ) {
                 viewInfo.instance = new viewInfo.type();
+            } else {
+                viewInfo.instance.reset();
             }
             view = viewInfo.instance;
             view.addTarget(service);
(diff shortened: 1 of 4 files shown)


Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -618,6 +618,8 @@ YUI.add('ez-platformuiapp', function (Y) {
             });
             if ( createView ) {
                 viewInfo.instance = new viewInfo.type();
+            } else {
+                viewInfo.instance.reset();
```
dpobel: this change affects each and every side view, did you check what happen with others side views ?

StephaneDiot: Yeah did a lot of manual tests and every side views seems to work correctly.

yannickroger:
Back after being reverted https://github.com/ezsystems/PlatformUIBundle/commit/463ba98840f64933432da1ae2ab61c9fc4f17ab7

PR #841, opened 2017-04-11
Fix EZP-27233: Unable to select finder items after changing tab

StephaneDiot:
JIRA: https://jira.ez.no/browse/EZP-27233

## Description

The selected content wasn't correctly resetted while changing tab from finder, it now works as expected

## Tests

Unit and manually tested

### DIFF
--- a/Tests/js/views/universaldiscovery/assets/ez-universaldiscoveryfinderexplorerview-tests.js
+++ b/Tests/js/views/universaldiscovery/assets/ez-universaldiscoveryfinderexplorerview-tests.js
@@ -153,6 +153,16 @@ YUI.add('ez-universaldiscoveryfinderexplorerview-tests', function (Y) {
                 );
             }, this));
         },
+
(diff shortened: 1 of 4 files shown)


Tests/js/views/universaldiscovery/assets/ez-universaldiscoveryfinderexplorerview-tests.js
```diff
@@ -153,6 +153,17 @@ YUI.add('ez-universaldiscoveryfinderexplorerview-tests', function (Y) {
                 );
             }, this));
         },
+
+        "Should set ownSelectedItem attribute to false when the explorer view wakeUp": function () {
+            Y.Array.each(this.levelViews, Y.bind(function (levelView) {
```
dpobel: that's a bit weird. I think you should just call `this.view.wakeUp()` and check the `ownSelectedItem` attribute value is `false`

StephaneDiot:
updated @dpobel

PR #842, opened 2017-04-12
How should we display Date and Time FieldTypes?

yannickroger:
## Description
I have been working on several bugs on timezone handling in the PlatformUI and I wanted to confirm the expected behavior from a functional perspective.

For date, I am quite confident with the behavior. I don't think somebody entering a date wants this date to be different for a reader in a different timezone.

For fieldtypes specifying also Time, if it represent an appointment happening every day (Time) or a one shot meeting/event (DateAndTime). I, as a developer, would like the expected behavior the be very clear.

This is the goal of this PR, clear things out in order to be more confident while debugging. Please read the scenario in the PR and give feedback.

## Example
I'll use Time as it is more simple.
- From my computer, configured on a `Paris/France` timezone I enter a Time: `19:00`
- When someone else views this Time (in PlatformUI) from a computer configured on `New York/USA` timezone, should one see: `19:00` (original timezone) or `13:00` (converted to their timezone)

## TODO
- [ ] Confirm behavior on Date
- [ ] Answer question for Time
- [ ] Answer question for Date and Time
- [ ] If needed, create follow up stories and bugs (on PlatformUI and kernel)

### DIFF
--- a/Features/Stories/FieldTypes/DateTimeFieldTypes.feature
+++ b/Features/Stories/FieldTypes/DateTimeFieldTypes.feature
@@ -1,0 +1,41 @@
+Feature: Storing and displaying Date, DateTime and Time in eZPlatform
+
+    Scenario: Storing and displaying Dates
+        Given my computer's locale is set to my "Paris/France"
+        And I create Content with a Date Field
+
+        When I view this Content
+        Then the Date is displayed with the same value
+
+        When my computer's locale is set to "America/New-York"
+        And I view this Content
+        Then the Date is displayed with the same value
+
+
+    Scenario: Storing and displaying Time
+        Given my computer's locale is set to my "Paris/France"
+        And I create Content with a Time Field
+
+        When I view this Content
+        Then the time is displayed in the "Paris/France" timezone
+
+        When my computer's locale is set to "America/New-York"
+        And I view this Content
+        #1?
+        Then the time is displayed in the "America/New-York" timezone
+        #or 2?
+        Then the time is displayed in the "Paris/France" timezone
+
+    Scenario: Storing and displaying DateTime
+        Given my computer's locale is set to my "Paris/France"
+        And I create Content with a DateTime Field
+
+        When I view this Content
+        Then the date and time is displayed in the "Paris/France" timezone
+
+        When my computer's locale is set to "America/New-York"
+        And I view this Content
+        #1?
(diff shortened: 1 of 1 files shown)


yannickroger:
ping @rolandbenedetti as I can't seem to be able to require a review from him in github.

Features/Stories/FieldTypes/DateTimeFieldTypes.feature
```diff
@@ -0,0 +1,41 @@
...
+    Scenario: Storing and displaying DateTime
+        Given my computer's locale is set to my "Paris/France"
+        And I create Content with a DateTime Field
+
+        When I view this Content
+        Then the date and time is displayed in the "Paris/France" timezone
+
+        When my computer's locale is set to "America/New-York"
+        And I view this Content
+        #1?
```
SylvainGuittard: Same comment as before. 
Especially if it's not the same day.  
Example: 2AM in Paris so 8PM (the day before) in NYC.

yannickroger:
Thanks @SylvainGuittard I can now add @rolandbenedetti as reviewer 😄

SylvainGuittard:
@yannickroger you're welcome ;)

andrerom:
👍  on what sylvain says, at least until we have featurer requestt to let editor pick timezone for whatever reason.

rolandbenedetti:
I would also go for the #1 choice for now, but ultimately, the real date and time is the one from the server for now, right? 
For future improvement, should this be visible/shown to the user?
We could have a toggle "Your time - Server time", it might be more meaningful than selecting time zones, time zones can be very confusing, especially when daylight saving kicks in...
Just cc. @inakijv @supriyabhargava  because they are 
motogp time selection
probably interested too.

inakijv:
Thank you @rolandbenedetti I'd suggest option #1 too. It's the one more aligned to the UX perspective. 
Providing a toggle option for selecting time zones it's really valuable from the editorial perspective :-)

yannickroger:
I am glad you are exited about this topic so we can plan cool stuff for V2 on the matter.

Thanks for your feedback on the # 1 approach. I also think it is the best for now.

Regarding what the server does, it stores the time in UTC not considering the server's local configuration, so the client (frontend or backend) is able to display it in any timezone to the end user. That's why my question was about how we create and view on the PlatformUI v1.x.

PR #844, opened 2017-04-12
EZP-23925: Add placeholder in DateAndTime for browser not supporting date/time input

yannickroger:
Link: https://jira.ez.no/browse/EZP-23925

## Description
On browser not supporting html5 inputs `date` and `time`, such as Firefox, the expected format of date was impossible to guess. This PR adds a placeholder to explain it (see screenshot). This place holder is a translatable string so the letters can be changed as long as the format remains the same.

date-time-ff-fixed

## BC
I propose this PR, which is technically an improvement, on the stable branch as it can also been seen as a bug and is safe to "backport".

## Tests
Manual test on master in firefox and chrome

### DIFF
--- a/Resources/translations/fieldedit.en.xlf
+++ b/Resources/translations/fieldedit.en.xlf
@@ -145,6 +145,18 @@
         <note>key: date.time.required</note>
         <jms:reference-file>./Resources/public/js/views/fields/ez-dateandtime-editview.js</jms:reference-file>
       </trans-unit>
+      <trans-unit id="a26a9e8cceea9e78f5ed30052cd1e0aa592498ca" resname="dateandtime.placeholder.date">
+        <source>YYYY-MM-DD</source>
+        <target>YYYY-MM-DD</target>
+        <note>key: dateandtime.placeholder.date</note>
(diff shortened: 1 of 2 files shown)


Resources/translations/fieldedit.en.xlf
```diff
@@ -145,6 +145,18 @@
...
       </trans-unit>
+      <trans-unit id="a26a9e8cceea9e78f5ed30052cd1e0aa592498ca" resname="dateandtime.placeholder.date">
+        <source>YYYY-MM-DD</source>
+        <target>YYYY-MM-DD</target>
+        <note>key: dateandtime.placeholder.date</note>
+        <jms:reference-file>Resources/public/templates/fields/edit/dateandtime.hbt</jms:reference-file>
+      </trans-unit>
+      <trans-unit id="6cb8a97a772599f1562670c41d57373303d2c3e0" resname="dateandtime.placeholder.time">
+        <source>hh:mm:ss</source>
+        <target>hh:mm:ss</target>
```
dpobel: having `:ss` in the placeholder actually depends on the `useSeconds` field definition, so you need to different placeholder strings.

yannickroger:
Feedback by @dpobel taken into account (and screenshot updated)

PR #846, opened 2017-04-14
Fix EZP-27252: Wrong message when deleting trash content

adamwojs:
JIRA issue: https://jira.ez.no/browse/EZP-27252

# Description

This PR fixes missing `%` for count parameter in source/target translation of `items.permanently.deleted` key in `/Resources/translations/trash.en.xlf`

# Steps to reproduce:

> * On the admin interface, go to trash and click on "Empty the trash";
> * After trash is deleted, the following message is displayed at the bottom: 
> %count items were permanently deleted.
> ...which does not reflect the number of contents deleted - see attached screenshot.


yannickroger:
You need to fix it in 1.7 (oldest supported branch), it is also happening there: https://github.com/ezsystems/PlatformUIBundle/blob/1.7/Resources/translations/trash.en.xlf#L52

yannickroger:
Thanks @adamwojs

adamwojs:
You welcome @yannickroger!

PR #848, opened 2017-04-18
Revert "Fix EZP-27143 Fixed custom UDW to work with content on the fly."

StephaneDiot:
Reverts ezsystems/PlatformUIBundle#840


yannickroger:
It should not have been reverted so it is back https://github.com/ezsystems/PlatformUIBundle/commit/463ba98840f64933432da1ae2ab61c9fc4f17ab7

PR #850, opened 2017-04-19
Fix EZP-27265: Missing translation for notification when user try to …

adamwojs:
JIRA issue: https://jira.ez.no/browse/EZP-27265

# Description

This PR fix missing english translation for notification displayed when user try to submit invalid data in content type editor.

# Steps to reproduce

> 1. On the admin UI go to content type edit form 
> 2. Change identifier of any field to something incorrect e.g. foo-1 
> 3. Save changes 
> 4. Message displayed on the bottom contains form.validation_error instead of human readable text - see attached screenshot

# Tests
Manual

### DIFF
--- a/Resources/translations/messages.en.xlf
+++ b/Resources/translations/messages.en.xlf
@@ -7,8 +7,8 @@
     </header>
     <body>
       <trans-unit id="07ec12f4689ee90ade282338b25b99b9677b0d08" resname="form.validation_error">
-        <source>form.validation_error</source>
-        <target>form.validation_error</target>
+        <source>Invalid form data</source>
+        <target>Invalid form data</target>
         <note>key: form.validation_error</note>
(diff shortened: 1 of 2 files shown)


Resources/translations/messages.en.xlf
```diff
@@ -8,7 +8,7 @@
     <body>
       <trans-unit id="07ec12f4689ee90ade282338b25b99b9677b0d08" resname="form.validation_error">
         <source>form.validation_error</source>
-        <target>form.validation_error</target>
```
yannickroger: You can also update the `<source>`. Here is an example on another file: https://github.com/ezsystems/PlatformUIBundle/blob/1.7/Resources/translations/fieldview.en.xlf#L10

adamwojs: Done ;-)

PR #851, opened 2017-04-24
Fix ezp 27288 create button is not visible in dashboard title bar

kmadejski:
**Jira ticket: [https://jira.ez.no/browse/EZP-27288](https://jira.ez.no/browse/EZP-27288)**

### Description
These two lines fix the bug. 

In eZPlatform Enterprise Edition button was visible because these two lines were added in studio-ui-bundle. Maybe now should be removed from the studio? 
`ezsystems/studio-ui-bundle/Resources/public/css/views/general.css:46`

### DIFF
--- a/Resources/public/css/theme/app.css
+++ b/Resources/public/css/theme/app.css
@@ -49,6 +49,7 @@ a {
 
 .ez-view-container {
     transition: opacity 0.2s ease-in-out;
+    position: relative;
 }
 
 .is-app-loading .ez-view-container {


yannickroger:
You should update your commit message.

kmadejski:
@yannickroger message changed.

kmadejski:
@yannickroger I've done tests locally and I can confirm that the same tests fail without these changes.

andrerom:
@kmadejski Do you see a way to move this fix to Content on the Fly bundle?

kmadejski:
@andrerom I've talked with @dpobel and I think that this PR should be closed. I will open the new one in COF repository.

PR #739, opened 2016-11-29
EZP-25002: Added buttons to add/remove options in Selection settings

dpobel:
JIRA: https://jira.ez.no/browse/EZP-25002
Requires https://github.com/ezsystems/repository-forms/pull/99

# Description

This PR (with the one on repository-forms) adds a UI to add or remove options in Selection field definitions. This is done in a way that also allows to do it only with the keyboard.
(Correct terms in the button will be added later in https://github.com/ezsystems/repository-forms/pull/99 once the translation system is stabilized)

Screencast:

[](http://www.youtube.com/watch?v=5FOZT0hZG5w "Click to play on Youtube.com")

# Tests

manual tests + unit tests

PR #853, opened 2017-04-27
Fix EZP-27316: Error publishing with a filled ezselection field

dpobel:
JIRA: https://jira.ez.no/browse/EZP-27316

# Description

This is a regression introduced in EZP-25002. This is happening because the patch in #739 makes the selection options be indexed starting from 1 while the Selection field type expects the options to be indexed starting from 0.

Note: This PR is created against 1.7 even if the bug does not affect eZ Platform 1.7 because the code responsible for the bug is in PlatformUI 1.7 but it is only really used with repository forms 1.6 which comes with eZ Platform 1.9 (and 1.8).

Note 2:  For 1.9, the exact same patch can be used

# Tests

unit tests + manual tests

### DIFF
--- a/Resources/public/js/views/serverside/ez-contenttypeeditserversideview.js
+++ b/Resources/public/js/views/serverside/ez-contenttypeeditserversideview.js
@@ -112,8 +112,8 @@ YUI.add('ez-contenttypeeditserversideview', function (Y) {
                 inputNameTpl = tplNode.getData('ezselection-option-input-name');
 
             settingsNode.all('.ezselection-settings-option-value input[type=text]').each(function (input, i) {
-                input.set('name', inputNameTpl.replace(PROTOTYPE_PLACEHOLDER, i+1));
-                input.set('id', inputIdTpl.replace(PROTOTYPE_PLACEHOLDER, i+1));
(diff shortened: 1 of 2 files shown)


bleakgadfly:
Just tried this patch on a brand new 1.9 installation after hitting "Option with index 0 does not exist in the field definition" going through the [Bicycle Route tutorial](https://doc.ez.no/display/DEVELOPER/Step+2+-+Create+your+content+model). For me the issue still exists.

dpobel:
Thanks @bleakgadfly for the feedback. Did you create the Content Type before applying the patch ? If so, you need to edit the Content Type and for instance add a new option, then remove it and finally save the Content Type. Normally, that should fix the issue.

galvorn56:
I just tried what @dpobel suggested and it worked, also with a brand new 1.9 installation and going through the Bicycle Route tutorial.
Editing the content type then removing and re-adding the options before saving the content type did the trick :)

bdunogier:
Would it somehow be possible to detect that the options are indexed from 1 instead of 0, and fix it automatically ?

Besides that, I'm okay with a microrelease for this

PR #858, opened 2017-05-10
Fix displaying google map in eZ Platform UI when there is a key specified in configuration

darinda:
Currently google maps won't show up in eZ Platform UI because the configuration is not read correctly. Fixed by correcting typo.


andrerom:
Hmm, unsure how I get it working, might have been I made that typo in later iteration of the code.

So thanks @darinda, could you by any chance rebase this on 1.7?

PR #859, opened 2017-05-10
Fix displaying google map in eZ Platform UI when there is a key specified in configuration

darinda:
Currently google maps won't show up in eZ Platform UI because the configuration is not read correctly. Fixed by correcting typo.

Rebased on 1.7 branch


andrerom:
Thanks @darinda, nice first contribution 🎉  😃

PR #862, opened 2017-05-15
EZP-27282: Updated PjaxRequestmatcherTest

bdunogier:
Updates the test for the change to `PjaxRequestMatcher` from 50dcfa6d8061d1a46e331f9e353ab9ca84754d41.


bdunogier:
> We should maybe just change the URI to not start with /pjax.

It's probably best, yes. Then it means we need to move it out of this file, as the `/pjax` prefix is common to all routes imported from it.

PR #863, opened 2017-05-15
Fix EZP-27330: Not possible to see the full name of content item in UDW

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-27330

# Description

This Patch expands the last selection level in UDW, which allows a user to see the full name of the content item in case of long names. Preview:
 
| Before | https://youtu.be/pEpJFmuLIrA |
|--------|------------------------------|
| **After**  | https://youtu.be/K-J0IyM-hqo |

### DIFF
--- a/Resources/public/css/theme/views/universaldiscovery/explorerlevel.css
+++ b/Resources/public/css/theme/views/universaldiscovery/explorerlevel.css
@@ -7,6 +7,8 @@
     content: "\E605";
     color: #7A7A7A;
     font-size: 140%;
+    position: absolute;
+    right: 0;
 }
 
 .ez-view-universaldiscoveryfinderexplorerlevelview .ez-explorer-level-list {
@@ -18,6 +20,7 @@
 .ez-view-universaldiscoveryfinderexplorerlevelview .ez-explorer-level-list-item {
     padding: 0.2em 0 0.2em 0;
     display: block;
+    position: relative;
 }
 
 .ez-view-universaldiscoveryfinderexplorerlevelview .ez-explorer-level-item {
--- a/Resources/public/css/views/universaldiscovery/explorerlevel.css
+++ b/Resources/public/css/views/universaldiscovery/explorerlevel.css
@@ -6,6 +6,15 @@
     min-width: 200px;
 }
 
+.ez-view-universaldiscoveryfinderexplorerlevelview:last-child {
+    width: 100%;
+    min-width: 600px;
+}
+
+.ez-view-universaldiscoveryfinderexplorerlevelview:last-child .ez-explorer-level-item {
+    width: 92.5%;
+}
+
 .ez-view-universaldiscoveryfinderexplorerlevelview .ez-ud-finder-explorerlevel-anchor {
     float: right;
     visibility: hidden;
--- a/Resources/public/css/views/universaldiscovery/selected.css
+++ b/Resources/public/css/views/universaldiscovery/selected.css
@@ -100,8 +100,8 @@
 }
 
 .ez-view-universaldiscoveryselectedview .ez-ud-selected-name {
-    text-overflow: ellipsis;
-    white-space: nowrap;
-    overflow: hidden;
+    font-size: 100%;
+    margin-top: 5px;
     min-height: 1em;
+    white-space: normal;
 }


andrerom:
_@adamwojs 1.8 is not supported anymore (no normal patches as of 1.9, as is the case [with FT releases](https://doc.ez.no/roadmap)), and 1.9 has had it's last scheduled patch release, so either target 1.7 or master._

andrerom:
As for this solution, I'm unsure what PM had in mind here.

PM review / input ping @rolandbenedetti @SylvainGuittard @inakijv

SylvainGuittard:
What PM proposed in the jira issue was only for the preview box: https://jira.ez.no/browse/EZP-27330?focusedCommentId=210411&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-210411.

Expanding the column is not a bad suggestion. I would prefer to have a resizable column. We did not include this in the jira issue in order to provide a fix to the customer asap.

SylvainGuittard:
> so either target 1.7 (LTS) or master (currently 1.10).
Definitely master

adamwojs:
> I would prefer to have a resizable column. We did not include this in the jira issue in order to provide a fix to the customer asap.

@SylvainGuittard Yes, resizable columns will be the best solution. I've experimented a little bit with CSS `resize` property that we could use to provide this functionality (here is demo: https://jsfiddle.net/ojdw90Lh/1/), but unfortunately IE and Edge not support it yet.

andrerom:
> @SylvainGuittard Yes, resizable columns will be the best solution. I've experimented a little bit with CSS resize property that we could use to provide this functionality (here is demo: https://jsfiddle.net/ojdw90Lh/1/), but unfortunately IE and Edge not support it yet.

That might be an option, Edge will eventually support it so if that can be used somehow in a way PM approves it would be the easiest and most future proof.

rolandbenedetti:
> That might be an option, Edge will eventually support it so if that can be used somehow in a way PM approves, it would be the easiest and most future proof choice.
I also support that option (which was in initial Jince design actually, we just simplified), it's also much better UX for sure than expanding by default the last level, which to me is looking a bit weird.
For the one having a browser not supporting this, well they can get the full title looking at the preview box.

inakijv:
Besides resizable columns, we would like to have the resizer/handle more visible for the user
https://codepen.io/eZ_UI/pen/wdYqVr

SylvainGuittard:
Thanks @inakijv

adamwojs:
I agree that resizer/handler should be more visible for the users, but I'm afraid that now, this solution is compatible with only Webkit-based browsers. This generates two potential problems:

1. In **Firefox** the default resizer/handler is still visible and its overlapping our resizer/handler (what does not look good in my opinion):

zrzut ekranu 2017-05-29 o 11 57 03

2. In **Edge**, column resize not working at all, but the resize handler is visible, what can be confusing for users. 

In conclusion: I think that for now, the cleanest solution is to stay with the default browser resizer/handler, as long as we don't plan to engage javascript in handling column resize:

resizable-uwd-columns

But of course, the choice belongs to PM. Ping @rolandbenedetti,  @inakijv, @SylvainGuittard

SylvainGuittard:
Thanks @adamwojs for testing this solution.
I agree with you this approach is not great for the user experience. 
How much work is it to do it in javascript and make it compatible with all supported browsers?

adamwojs:
@SylvainGuittard I estimate time cost for 1-2 days with an assumption that solution will be UDW specific. I'm not very familiar with YUI which will be the most time-consuming factor in this case, but according to this snippet:  https://jsfiddle.net/RainStudios/mw786v1w/ it looks quite simple to implement.

SylvainGuittard:
@adamwojs Thanks for the feedback and the snippet. 
I don't know if it makes sense to use YUI (even if UDW is built with YUI). If we implement a solution for UDW in javascript (not YUI), we might reuse it for the v2.
What do you think @dpobel ?

andrerom:
> How much work is it to do it in javascript and make it compatible with all supported browsers?
I think this is overdoing it.

@adamwojs Can you open a seperate smaller PR that just fixes the preview box so it shows the full name, and add full name on hover of the link. And then we can let this resizing logic discussion mature a bit and not block the customer requested usability fixes here.

adamwojs:
@andrerom Okay, I'll do it.

andrerom:
@SylvainGuittard This needs a story now, the issue reported by customer was solved in #873

@adamwojs this needs rebase and updated issue number once that is created.

SylvainGuittard:
@andrerom Done!
https://jira.ez.no/browse/EZP-27461

bdunogier:
Given that it is not prioritized, I suggest that we close this PR. It is linked from the issue Sylvain has created, and can be reopened if needed. No disagreements ?

andrerom:
Nope

PR #864, opened 2017-05-16
Remove form processing done response

bdunogier:
> [EZP-27282](http://jira.ez.no/browse/EZP-27282)
> Discussion: https://github.com/ezsystems/hybrid-platform-ui/pull/3#discussion_r116499545.

It was required by PJAX in Platform UI 1.x to handle redirections.
The hybrid platform UI, that uses the fetch function, does not require it.


Controller/Controller.php
```diff
@@ -35,11 +35,11 @@ public function performAccessChecks()
      *
      * @param string $url
      *
-     * @return FormProcessingDoneResponse
+     * @return RedirectResponse
```
dpobel: you also need to update the description above

PR #869, opened 2017-05-24
EZP-24826-checking-infinity-values

tomaszmadeyski:
This is a fix for https://jira.ez.no/browse/EZP-24826

## Description 
Numbers in float and integer fields are parsed by `parseInt` and `parseFloat` which, when the number is big enough, returns `Infinity` or `-Infinity`. Up to now it was validated as a proper number but it was submitted as null to the API.

After these changes javascript part will raise error when number is `Infinity` or `-Infinity`


tomaszmadeyski:
ping @dpobel @StephaneDiot - can you please take a look at it? (I have no permissions for adding reviewers)

andrerom:
@tomaszmadeyski Be aware you changed chmod of files, besides fixing it here if your IDE does that make sure to disable it ;)

tomaszmadeyski:
@andrerom good catch, thanks. Fixed

Resources/translations/fieldedit.en.xlf
```diff
@@ -857,6 +857,20 @@
         <note>key: value.should.be.valid.integer</note>
         <jms:reference-file>./Resources/public/js/views/fields/ez-integer-editview.js</jms:reference-file>
       </trans-unit>
+      <trans-unit id="d85e23987dc9ef6f162a6173ed81f647d9ff9fe1" resname="value.should.be.less.than.infinity">
+        <source>The value should be less than Infinity</source>
```
andrerom: As an editor I don't know what this error message is trying to tell me.

Unsure how to make it more clear while being specific, maybe simply "The value is to large, please use a valid number"


Same below

tomaszmadeyski:
@andrerom I think displaying Ifinity's numeric representation might be a bit overwhelming for user since it's something about 2^53. I changed it to generic value is too large/too small.

andrerom:
As for failure, seems we have another Twig 2.x issue here. @tomaszmadeyski / @natanael89 One of you available to take a look at that? :)

tomaszmadeyski:
@andrerom I created separate task for failing tests in Twig 2.x - https://jira.ez.no/browse/EZP-27438 and I will handle it in separate task as this is not related to this specific PR

andrerom:
Behat passes again now, so :shipit:

PR #870, opened 2017-05-26
EZP-27399: Prevent password hash being returned in by rest

Nattfarinn:
**TIcket:** https://jira.ez.no/browse/EZP-27399

This change is a bit redundant ( https://github.com/ezsystems/ezpublish-kernel/pull/1995 ), but still required due to fact there can only be one field type Processor at the time and this one overwrites Kernel one.


andrerom:
@Nattfarinn This breaks a BDD case for generating User using content api:
screen shot 2017-06-04 at 15 56 54

PR #871, opened 2017-05-30
EZP-27421: ez:yui-app:ready event is fired too early

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-27421

**Description:**
The `ez:yui-app:ready` was fired too early, before all plugins were loaded. In my case, the event was fired before handlebars helpers were registered and I couldn't use them in handlebars templates.
In this approach, the event is fired after `initialized` attribute has changed what means the whole app is loaded.

Unit tests are passing, and doesn't need to be updated or should I create a test for the alternative path when the `appReady` event is not fired?

### DIFF
--- a/Resources/public/js/apps/ez-platformuiapp.js
+++ b/Resources/public/js/apps/ez-platformuiapp.js
@@ -31,7 +31,8 @@ YUI.add('ez-platformuiapp', function (Y) {
             Y.eZ.Translator.setPreferredLanguages(this.get('interfaceLanguages'));
 
             this._setGlobals();
-            this._fireAppReadyEvent();
+
+            this.on('initializedChange', this._fireAppReadyEvent, this);
         },
 
         /**
--- a/Tests/js/apps/assets/ez-platformuiapp-tests.js
+++ b/Tests/js/apps/assets/ez-platformuiapp-tests.js
@@ -297,16 +297,17 @@ YUI.add('ez-platformuiapp-tests', function (Y) {
 
         "Should dispatch the `ez:yui-app:ready` event on the document": function () {
             var app,
-                dispatchedEvent = false;
+                dispatchedEvent = false,
+                eventHandler = function (e) {
+                    dispatchedEvent = true;
 
-            document.addEventListener('ez:yui-app:ready', function (e) {
-                dispatchedEvent = true;
+                    Assert.areSame(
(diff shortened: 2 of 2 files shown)


Resources/public/js/apps/ez-platformuiapp.js
```diff
@@ -149,10 +150,15 @@ YUI.add('ez-platformuiapp', function (Y) {
          *
          * @method _fireAppReadyEvent
          * @private
+         * @param {Object} event facade
          */
-        _fireAppReadyEvent: function () {
+        _fireAppReadyEvent: function (event) {
             var evt = new CustomEvent('ez:yui-app:ready');
 
+            if (!event.newVal) {
```
dpobel: is that really needed ? as far as I can tell, `initialized` is never set to something else than `true`

dew326:
ping @dpobel 
Removed unnecessary `if` and added the regression test.

Tests/js/apps/assets/ez-platformuiapp-tests.js
```diff
@@ -314,7 +314,30 @@ YUI.add('ez-platformuiapp-tests', function (Y) {
...
+        },
+
+        // Regression test for https://jira.ez.no/browse/EZP-27421
+        "Should dispatch the `ez:yui-app:ready` event after plugins are initialized": function () {
+            var app,
+                pluginNS = 'myPlugin',
+                dispatchedEvent = false,
+                Plugin = Y.Base.create('myPlugin', Y.Plugin.Base, [], {
+                    initializer: function () {
+                        this.on('initializedChange', function () {
```
dpobel: the way you test that seems a bit complicated. For readability, I would do the following:
* here, you should set a boolean variable (`pluginInitialized` for instance) to true
* the event listener for `ez:yui-app:ready` should be set right before the app instantiation and in it you assert that the boolean variable is true

no need for the `dispatchedEvent` variable, the event is checked in the previous test.

dpobel: also, I realize that in the previous test and in this one, the event listener for `ez:yui-app:ready` is never removed which at some point will cause weird bugs. Can you add a call to `document.removeEventListener` in both test to clean up that ?

dew326:
ping @dpobel

PR #872, opened 2017-06-01
EZP-27423: UDW view service should accept location Id

StephaneDiot:
JIRA: https://jira.ez.no/browse/EZP-27423

## Description

Universal discovery view service should be able to handle location id instead of REST id to load the starting location.

## Tests

Manually and unit tested

### DIFF
--- a/Resources/public/js/views/services/ez-universaldiscoveryviewservice.js
+++ b/Resources/public/js/views/services/ez-universaldiscoveryviewservice.js
@@ -30,20 +30,31 @@ YUI.add('ez-universaldiscoveryviewservice', function (Y) {
          * @param {Function} callback
          */
         _load: function (callback) {
-            var startingLocation,
-                parameters = this.get('parameters'),
+            var parameters = this.get('parameters'),
                 app = this.get('app');
 
             if ( parameters.startingLocationId ) {
-                startingLocation = new Y.eZ.Location();
-                startingLocation.set('id', parameters.startingLocationId);
-                app.set('loading', true);
-
-                this._loadStartingLocationPath(startingLocation, Y.bind(function(startingLoc) {
-                    this.set('startingLocation', startingLoc);
-                    app.set('loading', false);
-
-                    callback();
+                this.search.findLocations({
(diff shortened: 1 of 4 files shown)


Resources/public/js/views/services/ez-universaldiscoveryviewservice.js
```diff
@@ -30,20 +30,30 @@ YUI.add('ez-universaldiscoveryviewservice', function (Y) {
...
-                startingLocation = new Y.eZ.Location();
-                startingLocation.set('id', parameters.startingLocationId);
-                app.set('loading', true);
-
-                this._loadStartingLocationPath(startingLocation, Y.bind(function(startingLoc) {
-                    this.set('startingLocation', startingLoc);
-                    app.set('loading', false);
-
-                    callback();
+                this.search.findContent({
```
dpobel: given you are loading a Location, why not using `findLocations` ?

Resources/public/js/views/services/ez-universaldiscoveryviewservice.js
```diff
@@ -30,20 +30,30 @@ YUI.add('ez-universaldiscoveryviewservice', function (Y) {
...
+                    offset: 0
+                }, Y.bind(function (error, result) {
+                    if (!error && result.length) {
+                        this._loadStartingLocationPath(result[0].location, Y.bind(function(startingLoc) {
+                            this.set('startingLocation', startingLoc);
+                            app.set('loading', false);
+                            callback();
+                        }, this));
+                    } else {
+                        this.set('startingLocation', false);
```
dpobel: it's ok to ignore loading error, but you should log something in the Console

Resources/public/js/views/services/ez-universaldiscoveryviewservice.js
```diff
@@ -30,20 +30,31 @@ YUI.add('ez-universaldiscoveryviewservice', function (Y) {
...
+                        console.log('Could not load the startingLocation');
```
dpobel: `console.warn` would be better suited

PR #874, opened 2017-06-02
EZP-27438 fixing Twig 2.x compatibility issues in tests

tomaszmadeyski:
This fixes https://jira.ez.no/browse/EZP-27438

## Description

Together with PHP 7.x Twig is coming in 2.x version which is a BC break comparing to Twig 1.x in PHP versions lower than 7.x. 

This is a fix for tests: mocks were created in wrong way

Before this fix Travis reported errors in PHP 7.x version:

https://travis-ci.org/ezsystems/PlatformUIBundle/jobs/237106321#L474


andrerom:
Thanks @tomaszmadeyski !

PR #875, opened 2017-06-09
EZEE-1517: Add new scenarios for the content-on-the-fly-prototype-bundle bundle

mnocon:
JIRA Ticket: https://jira.ez.no/browse/EZEE-1517

Corresponding PRs:
Content on the Fly (main one) - https://github.com/ezsystems/content-on-the-fly-prototype-bundle/pull/34
StudioUI - https://github.com/ezsystems/StudioUIBundle/pull/782

Small changes here, required by "Content on the fly" PR: 
* new step (to switch UDW tab)
* changed visibility of some methods
* added two small improvements.

### DIFF
--- a/Features/Context/PlatformUI.php
+++ b/Features/Context/PlatformUI.php
@@ -198,7 +198,7 @@ class PlatformUI extends Context
      *  if an exception is thrown or the result is false wait and retry
      *  until a max timeout is reached.
      */
-    protected function spin($lambda)
+    public function spin($lambda)
     {
         $e = null;
         $timeLimit = time() + self::SPIN_TIMEOUT;
@@ -291,7 +291,7 @@ class PlatformUI extends Context
      * @param int       $iteration      Iteration number, used to control number of executions
      * @return array
      */
-    protected function getElementByText($text, $selector, $textSelector = null, $baseElement = null)
+    public function getElementByText($text, $selector, $textSelector = null, $baseElement = null)
     {
         if ($baseElement == null) {
             $baseElement = $this->getSession()->getPage();
@@ -299,7 +299,11 @@ class PlatformUI extends Context
         $elements = $this->findAllWithWait($selector, $baseElement);
         foreach ($elements as $element) {
             if ($textSelector != null) {
-                $elementText = $this->findWithWait($textSelector, $element)->getText();
+                try {
+                    $elementText = $this->findWithWait($textSelector, $element)->getText();
+                } catch (\Exception $e) {
+                    continue;
+                }
             } else {
                 $elementText = $element->getText();
             }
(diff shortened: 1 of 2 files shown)


Features/Context/SubContext/CommonActions.php
```diff
@@ -226,7 +226,12 @@ public function clickActionBar($button)
      */
     public function clickEditActionBar($button)
     {
-        $this->clickElementByText($button, '.ez-editactionbar-container .ez-action', '.action-label');
+        $actionBarElements = $this->findAllWithWait('.ez-editactionbar-container .ez-action .action-label');
```
mnocon: This change has been made because in some cases the previous approach was not efficient:
If elements have been found using the main selector and weren't visible, the spin function used waited the whole timeout for them to become visible (until going to the next item in the array). This resulted in a huge execution time of this step (clicking the button took ca. 60 seconds, now it's ca. 3 seconds).

Features/Context/PlatformUI.php
```diff
@@ -291,15 +291,19 @@ function () use ($selector, $baseElement, $checkVisibility) {
...
     {
         if ($baseElement == null) {
             $baseElement = $this->getSession()->getPage();
         }
         $elements = $this->findAllWithWait($selector, $baseElement);
         foreach ($elements as $element) {
             if ($textSelector != null) {
-                $elementText = $this->findWithWait($textSelector, $element)->getText();
+                try {
+                    $elementText = $this->findWithWait($textSelector, $element)->getText();
```
mnocon: findWithWait can throw Exceptions if element does not exist or is not visible - in this case IMHO next elements from the list should be processed (if no matching element is found this function returns "false" anyway).

Features/Context/SubContext/CommonActions.php
```diff
@@ -226,8 +226,17 @@ public function clickActionBar($button)
      */
     public function clickEditActionBar($button)
     {
-        $this->clickElementByText($button, '.ez-editactionbar-container .ez-action', '.action-label');
-        $this->waitWhileLoading();
+        $actionBarElements = $this->findAllWithWait('.ez-editactionbar-container .ez-action .action-label');
```
mnocon: This change has been made because in some cases the previous approach was not efficient:
If elements have been found using the main selector and weren't visible, the spin function waited the whole timeout for them to become visible (and only then processing the next item in array). This resulted in a huge execution time of this step (clicking the button took ca. 60 seconds, now it's ca. 3 seconds).

mnocon:
@andrerom Currently the PlatformUI Context must be reused, as it contains all the useful methods (findWithWait, clickElementByClass etc.), it would be nice to extract them to a separate Context (and in this bundle leave only those that interact with the Platform part). Discussion about the layout of behat code looks like a good idea.

andrerom:
merged, @mnocon please proceed on getting approval on the others.

mnocon:
Thanks, will do that

PR #876, opened 2017-06-12
EZP-27352: Implement <ez-universal-discovery> 

StephaneDiot:
jira: https://jira.ez.no/browse/EZP-27352

## Description

This is the base implementation of the universal discovery widget custom element : <ez-universal-discovery>

Goes with: https://github.com/ezsystems/hybrid-platform-ui/pull/16 and https://github.com/ezsystems/hybrid-platform-ui-core-components/pull/16
## Tests

Unit and manually tested

### DIFF
--- a/Resources/public/webcomponents/ez-universal-discovery.html
+++ b/Resources/public/webcomponents/ez-universal-discovery.html
@@ -1,19 +1,11 @@
 <link rel="import" href="../../../assets/ezplatform/polymer/polymer-element.html">
 <link rel="import" href="../../../webcomponents/ez-yui-app.html">
-<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery.css">
-<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery/finder.css">
-<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery/explorer.css">
-<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery/explorerlevel.css">
-<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery/search.css">
-<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery/selected.css">
 <link rel="import" href="mixins/ez-sandbox-yui-component.html">
 
 <style>
     ez-universal-discovery {
         display: block;
         background: rgba(0, 0, 0, 0.4);
-
-        /* taken from universaldiscovery.css .ez-universaldiscovery-container */
         position: fixed;
         width: 100%;
         height: 100%;
@@ -33,15 +25,13 @@
     ez-universal-discovery a {
         cursor: pointer;
     }
-
-    /* missing from modules/tabs.css (and coming for theme/modules/tabs.css) */
-    ez-universal-discovery .ez-tabs-panel {
-        display: none;
-    }
-
-    ez-universal-discovery .ez-tabs-panel.is-tab-selected {
-        display: block;
-    }
 </style>
+
+<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery.css">
+<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery/finder.css">
+<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery/explorer.css">
+<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery/explorerlevel.css">
+<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery/search.css">
+<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery/selected.css">
 
 <script src="js/ez-universal-discovery.js"></script>
--- a/Resources/public/webcomponents/ez-universal-discovery.html
+++ b/Resources/public/webcomponents/ez-universal-discovery.html
@@ -1,0 +1,47 @@
+<link rel="import" href="../../../assets/ezplatform/polymer/polymer-element.html">
+<link rel="import" href="../../../webcomponents/ez-yui-app.html">
+<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery.css">
+<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery/finder.css">
+<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery/explorer.css">
+<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery/explorerlevel.css">
+<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery/search.css">
+<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery/selected.css">
+<link rel="import" href="mixins/ez-sandbox-yui-component.html">
+
+<style>
+    ez-universal-discovery {
+        display: block;
+        background: rgba(0, 0, 0, 0.4);
+
+        /* taken from universaldiscovery.css .ez-universaldiscovery-container */
+        position: fixed;
+        width: 100%;
+        height: 100%;
+        box-sizing: border-box;
+        padding: 2% 3%;
+        top: 0;
+        left: 0;
+    }
+
+    ez-universal-discovery .ez-view-universaldiscoveryview {
+        background: #fff;
+        width: 100%;
+        height: 100%;
+        overflow: auto;
+    }
+
+    ez-universal-discovery a {
+        cursor: pointer;
+    }
+
+    /* missing from modules/tabs.css (and coming for theme/modules/tabs.css) */
+    ez-universal-discovery .ez-tabs-panel {
+        display: none;
+    }
+
+    ez-universal-discovery .ez-tabs-panel.is-tab-selected {
+        display: block;
+    }
+</style>
+
+<script src="js/ez-universal-discovery.js"></script>
--- a/Resources/public/webcomponents/js/ez-universal-discovery.js
+++ b/Resources/public/webcomponents/js/ez-universal-discovery.js
@@ -1,0 +1,142 @@
+(function () {
+
+    /**
+     * `<ez-universal-discovery>` generates the universal discovery widget.
+     *
+     * @demo demo/ez-universal-discovery.html
+     * @polymerElement
+     */
+    class UniversalDiscoveryWidget extends eZ.SandboxYUIComponentMixin(Polymer.Element) {
+        static get is() {
+            return 'ez-universal-discovery';
+        }
+        static get properties() {
+            return {
+                /**
+                 * Title of the universal discovery widget.
+                 */
+                title: {
+                    type: String,
+                },
+
+                /**
+                 * Flag indicating whether the user is able to select several
+                 * Content items.
+                 */
+                multiple: {
+                    type: Boolean,
+                    value: false,
+                },
+
+                /**
+                 * The Location that should be considered as the starting point when
+                 * discovering Content.
+                 */
+                startingLocationId: {
+                    type: String,
+                },
+
+                /**
+                 * Defines the label of the confirm button.
+                 */
+                confirmLabel: {
+                    type: String,
+                },
+
+                /**
+                * The depth of the root where we start discovering content.
+                * The UDW needs a starting location having a greater depth than the min discover depth to work.
+                 */
+                minDiscoverDepth: {
+                    type: Number,
+                },
+
+                /**
+                 * The identifier of the visible method by default.
+                 */
+                visibleMethod: {
+                    type: String,
+                },
+            };
+        }
+
+        connectedCallback() {
+            super.connectedCallback();
+            this._trackYUIAppReady(this._createUDWComponent.bind(this));
+        }
+
+        /**
+         * Converts a selection items to a YUI free representation by calling toObject on each model object.
+         *
+         * @param {Object} item
+         */
+        _convertItem(item) {
+            return {
+                content: item.content.toObject(),
+                location: item.location.toObject(),
+                contentType: item.contentType.toObject(),
+            };
+        }
+
+        /**
+         * Converts the selection to a YUI free representation.
+         *
+         * @param {Array|Object} selection
+         */
+        _convertSelection(selection) {
+            if ( this.multiple ) {
+                return selection.map(this._convertItem);
+            }
+            return this._convertItem(selection);
+        }
+
+        /**
+         * Creates the universal discovery widget component and renders the UD side view using the YUI bridge app.
+         */
+        _createUDWComponent() {
+            const {app, Y} = eZ.YUI;
+            const params = {
+                title: this.title,
+                multiple: this.multiple,
+                startingLocationId: this.startingLocationId,
+                minDiscoverDepth: this.minDiscoverDepth,
+                visibleMethod: this.visibleMethod,
+                contentDiscoveredHandler: (e) => {
+                    this.dispatchEvent(new CustomEvent('ez:confirm', {
+                        detail: {
+                            selection: this._convertSelection(e.selection),
+                        },
+                        bubbles: false,
+                    }));
+                },
+                cancelDiscoverHandler: (e) => {
+                    this.dispatchEvent(new CustomEvent('ez:cancel', {
+                        bubbles: false,
+                    }));
+                },
+                isSelectable: (selection) => {
+                    return this.dispatchEvent(new CustomEvent('ez:select', {
+                        detail: {
+                            selection: this._convertItem(selection),
+                        },
+                        bubbles: false,
+                        cancelable: true,
+                    }));
+                },
+            };
+
+            if ( this.confirmLabel ) {
+                params.confirmLabel = this.confirmLabel;
+            }
+
+            app.renderSideView(
+                Y.eZ.UniversalDiscoveryView,
+                Y.eZ.UniversalDiscoveryViewService,
+                params,
+                this._attachView.bind(this)
+            );
+        }
+    }
+
+    customElements.define(UniversalDiscoveryWidget.is, UniversalDiscoveryWidget);
+})();
--- a/Resources/public/webcomponents/js/ez-universal-discovery.js
+++ b/Resources/public/webcomponents/js/ez-universal-discovery.js
@@ -1,5 +1,4 @@
 (function () {
-
     /**
      * `<ez-universal-discovery>` generates the universal discovery widget.
      *
--- a/Resources/public/webcomponents/mixins/js/ez-sandbox-yui-component.js
+++ b/Resources/public/webcomponents/mixins/js/ez-sandbox-yui-component.js
@@ -1,0 +1,64 @@
+window.eZ = window.eZ || {};
+(function (eZ) {
+    /**
+     * Sandbox YUI component mixin support into a class extending `superClass`.
+     * The resulting class should be responsible for a component using some YUI views.
+     * It will be granted methods to check if the YUI bridge app is loaded and manage when to create
+     * the involved component.
+     * The class is also granted method to attach a YUI view to the involved custom element.
+     * And it also provides a `disconnectedCallback` method which removes the YUI view from the component
+     * but also removes the event listnener responsible for creating the component.
+     *
+     * @param {Function} superClass
+     * @return {Function}
+     */
+    eZ.SandboxYUIComponentMixin = function (superClass) {
+
+        return class extends superClass {
+            /**
+             * Attaches the YUI view as a child the custom element.
+             * It also handles loading error.
+             *
+             * @param {false|Error} err
+             * @param {eZ.ViewService} viewService
+             * @param {eZ.View} view
+             */
+            _attachView(err, viewService, view) {
+                if ( err ) {
+                    this.innerHTML = err.message;
+                    return;
+                }
+                this._view = view;
+                this.innerHTML = '';
+                this.appendChild(view.get('container').getDOMNode());
+                this.classList.add('ez-js-standard-form');
+                view.set('active', true);
+            }
+
+            disconnectedCallback() {
+                if ( this._view ) {
+                    this._view.set('active', false);
+                    this._view.destroy({remove: true});
+                }
+                this.ownerDocument.removeEventListener('ez:yui-app:ready', this._readyHandler);
+                super.disconnectedCallback();
+            }
+
+            /**
+             * Checks whether the YUI bridge App is already loaded. If it is the
+             * `createComponent` is directly called, otherwise, it subcribes to
+             * `ez:yui-app:ready` to call `createComponent` when the app is ready.
+             *
+             * @param {Function} createComponent
+             */
+            _trackYUIAppReady(createComponent) {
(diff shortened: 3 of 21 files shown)


Tests/webcomponents/js/ez-universaldiscovery.js
```diff
@@ -0,0 +1,339 @@
...
+                done(false, 'whatever', view);
+            });
+            document.dispatchEvent(new CustomEvent('ez:yui-app:ready'));
+            assert.equal(
+                container.parentNode,
+                universalDiscovery
+            );
+        });
+
+        it('should have a cancelDiscoverHandler', function () {
```
dpobel: the 3 following tests try to test too many things at the same time. For each of them, you test both the fact that a `cancelDiscoverHandler`, `contentDiscoveredHandler` or a `isSelectable` function is passed and what those functions are doing.

In a suite called ~~`UDW init`~~ `YUI Universal Discovery Widget`, you should limit this test to check that those function are passed, so the tests should be nameed:

* `should have received a contentDiscoveredHandler function`
* `should have received a cancelDiscoveredHandler function`
* `should have received a isSelectable function`

And in addition to that, there should be a top level `events` test suite containing a test suite for each event. Actually, these are the most important tests to have since it's the main public API of this component (besides the properties), that would give something like:

```js
describe('events', function () {
    describe('`ez:cancel`', function () {
        it('should be dispatched');
    });

    describe('`ez:contentDiscovered`', function () {
        it('should be dispatched with a simple selection');

        it('should be dispatched with a multiple selection');
    });

    describe('`ez:select`', function () {
        it('should be dispatched when selecting an item');

        it('should allow to prevent the selection');
    });
});
```

Tests/webcomponents/js/ez-universaldiscovery.js
```diff
@@ -0,0 +1,339 @@
...
+            universalDiscovery.addEventListener('ez:cancel', function (e) {
+                assert.isFalse(e.bubbles);
```
dpobel: Note: that's incomplete. You only check that the event is not bubbling but you never check that the event is actually dispatched, if for some reason it's not the test won't fail. In any case, that will be moved to `events > ez:cancel > should be dispatched` but that also needs to be improved.

Tests/webcomponents/js/ez-universaldiscovery.js
```diff
@@ -0,0 +1,339 @@
...
+                universalDiscovery = fixture('BasicTestFixture');
+            });
+
+            it('should render the universal discovery right away', function () {
+                assertViewRendered(renderViewSpy, window.eZ.YUI.Y);
+            });
+        });
+    });
+
+    describe('UDW init', function () {
```
dpobel: should probably be named `YUI Universal Discovery Widget`

Tests/webcomponents/js/ez-universaldiscovery.js
```diff
@@ -0,0 +1,339 @@
...
+            universalDiscovery = fixture('BasicTestFixture');
+            defineGlobals();
+        });
+
+        afterEach(function () {
+            delete window.eZ;
+            renderViewStub.restore();
+        });
+
+        it('should append the YUI view container', function () {
```
dpobel: actually most of the assertions are about the parameters. You should split that into 2 tests:
* `should append the YUI view container`
* `should have received parameters from the properties`

Resources/public/webcomponents/ez-universaldiscovery.html
```diff
@@ -0,0 +1,39 @@
...
+        width: 100%;
+        height: 100%;
+        overflow: auto;
+    }
+
+    ez-universal-discovery a {
+        cursor: pointer;
+    }
+
+    /* missing from modules/tabs.css (and coming for theme/modules/tabs.css) */
```
dpobel: but where are the CSS included in https://github.com/ezsystems/hybrid-platform-ui/blob/master/src/bundle/Resources/public/webcomponents/ez-universaldiscoverywidget.html#L31 ?

bdunogier:
Please rebase it on top of the last 2.0, the unit test error is fixed.

StephaneDiot:
Travis still failling with a up-to-date 2.0 base

Resources/public/webcomponents/ez-universal-discovery.html
```diff
@@ -0,0 +1,47 @@
+<link rel="import" href="../../../assets/ezplatform/polymer/polymer-element.html">
+<link rel="import" href="../../../webcomponents/ez-yui-app.html">
+<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery.css">
```
dpobel: for readability, please do not mix stylesheets and import. Also, stylesheets should be imported after the `<style>` element

Resources/public/webcomponents/ez-universal-discovery.html
```diff
@@ -0,0 +1,47 @@
...
+<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery/search.css">
+<link rel="stylesheet" href="../../../bundles/ezplatformui/css/views/universaldiscovery/selected.css">
+<link rel="import" href="mixins/ez-sandbox-yui-component.html">
+
+<style>
+    ez-universal-discovery {
+        display: block;
+        background: rgba(0, 0, 0, 0.4);
+
+        /* taken from universaldiscovery.css .ez-universaldiscovery-container */
```
dpobel: you should remove that comment and clean up universaldiscovery.css for unncessary element (.ez-universaldiscovery-container is gone for instance)

Resources/public/webcomponents/ez-universal-discovery.html
```diff
@@ -0,0 +1,47 @@
...
+        width: 100%;
+        height: 100%;
+        overflow: auto;
+    }
+
+    ez-universal-discovery a {
+        cursor: pointer;
+    }
+
+    /* missing from modules/tabs.css (and coming for theme/modules/tabs.css) */
```
dpobel: I already commented about that in https://github.com/ezsystems/PlatformUIBundle/pull/876#discussion_r121621177

Resources/public/webcomponents/js/ez-universal-discovery.js
```diff
@@ -0,0 +1,142 @@
+(function () {
+
```
dpobel: https://github.com/ezsystems/PlatformUIBundle/pull/876#discussion_r121621632 still useless

Resources/public/webcomponents/mixins/js/ez-sandbox-yui-component.js
```diff
@@ -0,0 +1,57 @@
+window.eZ = window.eZ || {};
+(function (eZ) {
+    /**
+     *
```
dpobel: doc ?

Tests/js/models/assets/ez-contentmodel-tests.js
```diff
@@ -1548,6 +1548,46 @@ YUI.add('ez-contentmodel-tests', function (Y) {
         },
     });
 
+    toObjectTest = new Y.Test.Case({
+        name: "eZ Content Model toObject test",
+
+        setUp: function () {
+            this.fieldsByLanguage = {fields: 'paracétamole'};
+            this.currentVersion = new Y.Mock({Version: {VersionInfo: {},Fields: {field :[]}}});
```
dpobel: why passing this object to `Y.Mock` constructor ?

Tests/js/models/assets/ez-contentmodel-tests.js
```diff
@@ -1548,6 +1548,46 @@ YUI.add('ez-contentmodel-tests', function (Y) {
...
+
+            Assert.areEqual(
+                object.id, this.content.get('contentId'),
+                "The object should have an id"
+            );
+            Assert.areEqual(
+                object.name, this.content.get('name'),
+                "The object should have a name"
+            );
+            Assert.areEqual(
```
dpobel: you should use `areSame` since you are comparing objects

Tests/js/models/assets/ez-contenttypemodel-tests.js
```diff
@@ -587,10 +587,39 @@ YUI.add('ez-contenttypemodel-tests', function (Y) {
...
+        },
+
+        tearDown: function () {
+            this.contentType.destroy();
+        },
+
+        "Should return an object based on the content": function () {
+            var object = this.contentType.toObject();
+
+            Assert.areEqual(
```
dpobel: `areSame` here and below (comparing objects)

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -1123,6 +1123,45 @@ YUI.add('ez-locationmodel-tests', function (Y) {
         },
     });
 
+    toObjectTest = new Y.Test.Case({
+        name: "eZ location Model toObject test",
+
+        setUp: function () {
+            this.contentInfoObject = {id: 42, name: 'aspirine'};
+            this.contentInfo = new Y.Mock({Content: {alwaysAvailable: false}});
```
dpobel: same as previously, why this constructor parameter?

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -1123,6 +1123,45 @@ YUI.add('ez-locationmodel-tests', function (Y) {
...
+        },
+
+        "Should return an object based on the content": function () {
+            var object = this.location.toObject();
+
+            Assert.areEqual(
+                object.id, this.location.get('locationId'),
+                "The object should have an id"
+            );
+            Assert.areSame(
```
dpobel: you can replace those 2 assertions by one comparing `object.contentInfo` and `this.contentInfoObject`

Tests/webcomponents/ez-subitem.html
```diff
@@ -10,6 +10,7 @@
         <script src="../../../web-component-tester/browser.js"></script>
 
         <link rel="import" href="/Resources/public/webcomponents/ez-subitem.html">
+        <link rel="import" href="/Resources/public/webcomponents/mixins/ez-sandbox-yui-component.html">
```
dpobel: no, that should no be there, the import in Resources/public/webcomponents/ez-subitem.html should be enough

Tests/webcomponents/ez-universal-discovery.html
```diff
@@ -0,0 +1,30 @@
...
+        <meta charset="utf-8">
+        <meta name="viewport" content="width=device-width, minimum-scale=1, initial-scale=1, user-scalable=yes">
+
+        <title>ez-universal-discovery test</title>
+
+        <script src="../../../webcomponentsjs/webcomponents-lite.js"></script>
+        <script src="../../../web-component-tester/browser.js"></script>
+
+        <link rel="import" href="/Resources/public/webcomponents/ez-universal-discovery.html">
+        <link rel="import" href="/Resources/public/webcomponents/mixins/ez-sandbox-yui-component.html">
```
dpobel: same, normally this imported by ez-universal-discovery.html

Tests/webcomponents/mixins/ez-sandbox-yui-component.html
```diff
@@ -0,0 +1,34 @@
+<!doctype html>
+<html lang="en">
+    <head>
+        <meta charset="utf-8">
+        <meta name="viewport" content="width=device-width, minimum-scale=1, initial-scale=1, user-scalable=yes">
+
+        <title>ez-subitem test</title>
```
dpobel: nope

Tests/webcomponents/mixins/ez-sandbox-yui-component.html
```diff
@@ -0,0 +1,34 @@
...
+            }
+            window.customElements.define(TestElement.is, TestElement);
+        });
+        </script>
+    </head>
+    <body>
+
+        <test-fixture id="BasicTestFixture">
+            <template>
+                <ez-test-yui-componentt></ez-test-yui-component>
```
dpobel: typo, if tests are passing with that, the HTML parser is definitely too kind :)

dpobel: actually, it's because this is not needed... and then TestElement is not needed as well

Tests/webcomponents/mixins/js/ez-sandbox-yui-component.js
```diff
@@ -0,0 +1,11 @@
+describe('ez-mixin-yui-component', function () {
+    let element;
+
+    beforeEach(function () {
+        element = fixture('BasicTestFixture');
```
dpobel: not needed, element is not used anywhere

Tests/webcomponents/js/ez-universal-discovery.js
```diff
@@ -0,0 +1,409 @@
...
+        });
+
+        it('should have a `visibleMethod` property', function () {
+            assert.equal(
+                universalDiscovery.getAttribute('visible-method'),
+                universalDiscovery.visibleMethod
+            );
+        });
+        describe('`multiple` property', function () {
+            it('should have a `multiple` property', function () {
```
dpobel:
```
it('should be defined', function () {
```

Tests/webcomponents/js/ez-universal-discovery.js
```diff
@@ -0,0 +1,409 @@
...
+                assert.equal(
```
dpobel: assert.equal on Booleans is a bit weird, why not just:
```js
assert.isTrue(universalDiscoveryMultiple.multiple);
```

Tests/webcomponents/js/ez-universal-discovery.js
```diff
@@ -0,0 +1,409 @@
...
+                    universalDiscoveryMultiple.hasAttribute('multiple'),
+                    universalDiscoveryMultiple.multiple
+                );
+            });
+
+            it('`multiple` property should be false by default', function () {
```
dpobel:
```
it('should be false by default', function () {
```

Tests/webcomponents/js/ez-universal-discovery.js
```diff
@@ -0,0 +1,409 @@
...
+        });
+
+        it('should append the YUI view container', function () {
+            const view = getViewMock(container);
+
+            renderViewStub = sinon.stub(window.eZ.YUI.app, 'renderSideView', function (View, Service, params, done) {
+                done(false, 'whatever', view);
+            });
+            document.dispatchEvent(new CustomEvent('ez:yui-app:ready'));
+            assert.equal(
```
dpobel: strictEqual

Tests/webcomponents/js/ez-universal-discovery.js
```diff
@@ -0,0 +1,409 @@
...
+
+            renderViewStub = sinon.stub(window.eZ.YUI.app, 'renderSideView', function (View, Service, params, done) {
+                done(false, 'whatever', view);
+            });
+            document.dispatchEvent(new CustomEvent('ez:yui-app:ready'));
+            assert.isTrue(universalDiscovery.classList.contains('ez-js-standard-form'));
+        });
+
+        describe('events', function () {
+
```
dpobel: cs

Tests/webcomponents/js/ez-universal-discovery.js
```diff
@@ -0,0 +1,409 @@
...
+        const Model = function () {
+            this.toObject = function () {
+                return modelObject;
+            };
+        };
+
+        return new Model();
+    }
+
+    function dispatchEvent(handlerName, selection) {
```
dpobel: that's a very confusing name. OK it dispatches an event (`ez:yui-app:ready`) but it's not really the point of it. It should rahter be called `runUDWEventHandler` or something like that.

Tests/webcomponents/js/ez-universal-discovery.js
```diff
@@ -0,0 +1,409 @@
...
+
+            describe('`ez:cancel`', function () {
+                it('should be dispatched', function() {
+                    let isDispatched = false;
+
+                    universalDiscovery.addEventListener('ez:cancel', function (e) {
+                        assert.isFalse(e.bubbles);
+                        isDispatched = true;
+                    });
+                    dispatchEvent('cancelDiscoverHandler', {});
```
dpobel: the cancel discover handler does not accept any selection so the `{}` is a bit confusing, you should just omit this parameter

Tests/webcomponents/js/ez-universal-discovery.js
```diff
@@ -0,0 +1,409 @@
...
+                    });
+                    dispatchEvent('isSelectable', selection);
+                    assert.isTrue(isDispatched);
+                });
+
+                it('should allow to prevent the selection', function() {
+                    const selection = {location: location, content: content, contentType: contentType};
+
+                    universalDiscovery.addEventListener('ez:select', function (e) {
+                        assert.isTrue(e.cancelable);
```
dpobel: that's no enough. you should called `e.preventDefault()` and check that `isSelectable` returns false in that case.

Tests/webcomponents/js/ez-universal-discovery.js
```diff
@@ -0,0 +1,409 @@
...
+                    assert.isTrue(isDispatched);
+                });
+
+                it('should be dispatched with a multiple selection', function() {
+                    const fakeEventFacade = {selection : [{location: location, content: content, contentType: contentType}]};
+                    let isDispatched = false;
+
+                    universalDiscovery.multiple = true;
+                    universalDiscovery.addEventListener('ez:confirm', function (e) {
+                        assert.equal(e.detail.selection[0].content, modelObject);
```
dpobel: strictEqual (same below)

Tests/webcomponents/js/ez-universal-discovery.js
```diff
@@ -0,0 +1,409 @@
...
+
+    describe('YUI Universal Discovery Widget', function () {
+        let renderViewStub;
+        const container = document.createElement('div');
+        const location = getModelMock();
+        const content = getModelMock();
+        const contentType = getModelMock();
+
+        function assertEventParam(e) {
+            assert.equal(e.detail.selection.content, modelObject);
```
dpobel: strictEqual
in addition, it's a bit weird to reference a global variable here and the same each and every selection property

Tests/js/models/assets/ez-contentmodel-tests.js
StephaneDiot: Well it is required for the test to run

Tests/js/models/assets/ez-locationmodel-tests.js
StephaneDiot: can't run the tests without it, I'll investigate why

Tests/js/models/assets/ez-contentmodel-tests.js
dpobel: that's right, but actually `this.currentVersion` should be the object you pass in `Y.Mock` constructor parameter. You don't need a Mock and then the expectation in the following line is also not needed (this is what confused me)

Resources/public/webcomponents/ez-universal-discovery.html
```diff
@@ -0,0 +1,41 @@
...
+        width: 100%;
+        height: 100%;
+        overflow: auto;
+    }
+
+    ez-universal-discovery a {
+        cursor: pointer;
+    }
+
+    ez-universal-discovery .ez-tabs-panel.is-tab-selected {
```
dpobel: this is also not needed, tabs are now correctly integrated

Tests/js/models/assets/ez-contentmodel-tests.js
```diff
@@ -1548,6 +1548,43 @@ YUI.add('ez-contentmodel-tests', function (Y) {
         },
     });
 
+    toObjectTest = new Y.Test.Case({
+        name: "eZ Content Model toObject test",
+
+        setUp: function () {
+            this.fieldsByLanguage = {fields: 'paracétamole'};
```
dpobel: not needed it seems

Tests/js/models/assets/ez-locationmodel-tests.js
```diff
@@ -1123,6 +1123,51 @@ YUI.add('ez-locationmodel-tests', function (Y) {
         },
     });
 
+    toObjectTest = new Y.Test.Case({
+        name: "eZ location Model toObject test",
+
+        setUp: function () {
+            this.loadResponse = {
```
dpobel: what's the point of adding a property to the test case if it's not used any where else than in this method. Just use a local variable

Tests/webcomponents/js/ez-universal-discovery.js
```diff
@@ -0,0 +1,440 @@
...
+        const container = document.createElement('div');
+        const locationObject = {};
+        const contentObject = {};
+        const contentTypeObject = {};
+
+        const location = getLocationMock();
+        const content = getContentMock();
+        const contentType = getContentTypeMock();
+
+        function getLocationMock() {
```
dpobel: those 3 functions are very close, why not defining a `getModelMock` which would receives the the object returned by `toObject`, something like:

```js
function getModelMock(object) {
    const Model = function () {
        this.toObject = function () {
            return object;
        }
    }
}
```

also it should be defined being used

Tests/webcomponents/js/ez-universal-discovery.js
```diff
@@ -0,0 +1,440 @@
...
+            assert.isTrue(universalDiscovery.classList.contains('ez-js-standard-form'));
+        });
+
+        describe('events', function () {
+            describe('`ez:cancel`', function () {
+                it('should be dispatched', function() {
+                    let isDispatched = false;
+
+                    universalDiscovery.addEventListener('ez:cancel', function (e) {
+                        assert.isFalse(e.bubbles);
```
dpobel: this assert deserves a custom error message

Tests/webcomponents/js/ez-universal-discovery.js
```diff
@@ -0,0 +1,440 @@
...
+
+                it('should allow to prevent the selection', function() {
+                    const selection = {location: location, content: content, contentType: contentType};
+                    const container = document.createElement('div');
+                    const view = getViewMock(container);
+
+                    universalDiscovery.addEventListener('ez:select', function (e) {
+                        e.preventDefault();
+                        assert.isTrue(e.cancelable);
+                        return false;
```
dpobel: that's confusing, this should not be there, `e.preventDefault()` is enough to prevent the event

Tests/webcomponents/js/ez-universal-discovery.js
```diff
@@ -0,0 +1,440 @@
...
+            const Model = function () {
+                this.toObject = function () {
+                    return contentTypeObject;
+                };
+            };
+
+            return new Model();
+        }
+
+        function assertEventParam(e) {
```
dpobel: this is a very generic name. I would call it `assertEventSelection` and move the assertion on `e.bubbles` out of it

Tests/webcomponents/mixins/ez-sandbox-yui-component.html
```diff
@@ -0,0 +1,26 @@
...
+        <meta name="viewport" content="width=device-width, minimum-scale=1, initial-scale=1, user-scalable=yes">
+
+        <title>Sandbox YUI component mixin test</title>
+
+        <script src="../../../webcomponentsjs/webcomponents-lite.js"></script>
+        <script src="../../../web-component-tester/browser.js"></script>
+
+        <link rel="import" href="/Resources/public/webcomponents/mixins/ez-sandbox-yui-component.html">
+        <script>
+        window.addEventListener('WebComponentsReady', function () {
```
dpobel: all that is completely unneeded as well. Removing `window.customElements.define` call means also removing that

StephaneDiot:
feedbacks taken into account @dpobel

PR #878, opened 2017-06-18
EZP-24800 Support allowed classes limitation for object relation fields

slaci:
This is related to this PR: https://github.com/ezsystems/ezpublish-kernel/pull/2022

This just uses the same code from the RelationList, which constraints the selectable content types based on the field setting.

### DIFF
--- a/Resources/public/js/views/fields/ez-relation-editview.js
+++ b/Resources/public/js/views/fields/ez-relation-editview.js
@@ -133,12 +133,23 @@ YUI.add('ez-relation-editview', function (Y) {
          * @param {EventFacade} e
          */
         _runUniversalDiscovery: function (e) {
+            var that = this;
+
             e.preventDefault();
             this.fire('contentDiscover', {
                 config: {
                     contentDiscoveredHandler: Y.bind(this._selectRelation, this),
                     cancelDiscoverHandler: Y.bind(this.validate, this),
                     startingLocationId: this.get('fieldDefinition').fieldSettings.selectionRootHref,
+                    isSelectable: function (contentStruct) {
+                        var selectionContentTypes = that.get('fieldDefinition').fieldSettings.selectionContentTypes,
+                            contentTypeIdentifier = contentStruct.contentType.get('identifier');
+
+                        return (
+                            (selectionContentTypes.length === 0)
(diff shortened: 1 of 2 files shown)


Resources/public/js/views/fields/ez-relation-editview.js
```diff
@@ -133,12 +133,23 @@ YUI.add('ez-relation-editview', function (Y) {
          * @param {EventFacade} e
          */
         _runUniversalDiscovery: function (e) {
+            var that = this;
```
StephaneDiot: niptick: please avoid using `that`, and bind the function given to `isSelectable`.

slaci: This is copied from the relationlist's file: https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/fields/ez-relationlist-editview.js#L184

I can rename this, but then it wont be consistent.

StephaneDiot: I agree with the consistent point, but since we wrote this code we tried to avoid using `this = that`.
I do think this would be better to write something like :
```js
isSelectable: Y.bind(function (contentStruct) {
    var selectionContentTypes = this.get('fieldDefinition').fieldSettings.selectionContentTypes,
          contentTypeIdentifier = contentStruct.contentType.get('identifier');
 
    return (
        (selectionContentTypes.length === 0)
        || (selectionContentTypes.indexOf(contentTypeIdentifier) > -1)
    );
}, this)
```

Furthermore I forgot to mention it before, but unit tests should also be updated.

andrerom:
@slaci will you be able to add unit test coverage?

See:
- https://github.com/ezsystems/PlatformUIBundle/blob/master/Tests/js/views/fields/assets/ez-relationlist-editview-tests.js
- https://github.com/ezsystems/PlatformUIBundle/blob/master/Tests/js/views/fields/assets/ez-relation-editview-tests.js

slaci:
I copied the related universal discovery test from the `ez-relationlist-editview-tests.js` and adapted for the current field. I dont really understand why do I have to use `selectionRootHref` instead of `selectionRoot` but works only this way.
Hope this is good enought, tests are green for me.

andrerom:
We'll handle improvements to the code across both fields types as a follow up, thanks.

PR #886, opened 2017-07-20
EZP-27685: Fixing LocationView Tab-Sorting when using addTabView

skonrad:
The sorting for tabs after adding a new (custom) tab is currently `DESC` instead of the intended `ASC`.

This is due to a mix-up in the usage of the parameters handed to the `sort`-function of the tabs-array.

Also created an issue regarding this problem: https://jira.ez.no/browse/EZP-27685


yannickroger:
Just a quick note before looking at the code. If 1.7 (LTS) is affected, we need to have this PR made against the 1.7 branch.
Also you can have both commits squashed together and add the issue number in it: `Fix EZP-27685: Sorting of location-view-tabs when using addTabView`

yannickroger:
If I am not mistaken, the current behavior is the right one.
For example, an item with a priority of 10 should be placed before an item with a priority of 20.

skonrad:
Hi @yannickroger,

i will rebase the branch to target 1.7 and change the PR accordingly.

Regarding the current logic, it will sort priorities from 3, 2, 1 (highest first) as soon as addTabView and the resulting sort is executed/used. Example-Screenshot contained in the mentioned ticket.

skonrad:
For clarification of the issue:

The initial array for the available tabs is not initialized using addTabView and is added directly as the default-value of the tabs-attribute in `ez-locationviewview.js`.

```js
            tabs: {
                valueFn: function () {
                    return [
                        new Y.eZ.LocationViewViewTabView({
                            ...
                            priority: 1000,
                            ...
                        }),
                        new Y.eZ.LocationViewDetailsTabView({
                            ...
                            priority: 2000,
                            ...
                        }),
                        new Y.eZ.LocationViewVersionsTabView({
                            ...
                            priority: 2500,
                            ...
                        }),
                        new Y.eZ.LocationViewLocationsTabView({
                            ...
                            priority: 3000,
                            ...
                        }),
                        new Y.eZ.LocationViewRelationsTabView({
                            ...
                            priority: 4000,
                            ...
                        }),
                    ];
                },
                writeOnce: 'initOnly',
            },
```

In a regular installation the addTabView-method is never invoked (not found an example within code by ezsystems so far) and thus the array is never subjected to sorting - hence the correct display if default installation. 

The priority is set correctly but the sorting is working reversed (desc instead of asc by priority).
Resulting image with added custom-tab:
tab-sort-error

Lines adding the new tab are as follow:
```js
host.addTabView(
    new Y.eZ.LocationViewMigrationStaticTabView({
    content: host.get('content'),
    location: host.get('location'),
    contentType: host.get('contentType'),
    config: host.get('config'),
    languageCode: host.get('languageCode'),
    priority: 5000,
    selected: false,
    bubbleTargets: host
})
```

skonrad:
bump @yannickroger what's the state for this?

dpobel:
hi @skonrad,

since the attribute is called `priority`, tabs with higher priority should appear first. So if I'm not mistaken, the `addTabView` logic is right but the default priorities of the tabs are wrong. So to me, to fix the issue, you just have to change the default priorities on tabs so that `Y.eZ.LocationViewViewTabView` has a priority of 4000, `Y.eZ.LocationViewDetailsTabView` 3000 and so on

skonrad:
@dpobel @yannickroger 

It seems there are some incompatibilities with the new kernel-version.
1.7 should probably also be limited to 1.7 to 1.10 as those worked well (afaik).

JS and Behat tests are running smoothly though.

andrerom:
@dpobel : The failure is most likley becasue of https://github.com/ezsystems/ezpublish-kernel/pull/2009 The config code will need to be adjusted like done there.

dpobel:
@andrerom not sure to understand everything, who can do that ?
btw, @skonrad is right, on the 1.7 branch, we should not install the last kernel but the one intended to work eZ Platform with 1.7 (kernel 6.7?).

skonrad:
@andrerom @dpobel what's the status here? Anything I'm required to adjust or can I move on for the time being?

andrerom:
@skonrad Thanks for the ping, no I think we are good, thanks for the PR. Merging ;)

PR #890, opened 2017-08-04
EZP-27750: Backport of EZP-25478: Restore items whose original location has been deleted

natanael89:
JIRA issue: [EZP-27750](https://jira.ez.no/browse/EZP-27750).
>After deleting the parent content, the children are also put into the trash. Even after someone restores the parent, the children still cannot be restored (information "Item's ancestors are in Trash" is still showing). This can be circumvented in eZ Platform 1.10 by using the button "Restore under a new parent", but there is no such button in eZ Platform 1.7 and this means that the children cannot be restored at all using the Platform UI.

This PR is the backport of #819. It backported almost cleanly, with one conflict in the Resources/translations/trash.en.xlf file (the date).
It allows circumventing the issue [EZP-27750](https://jira.ez.no/browse/EZP-27750) as the "Restore under a new parent" button would then be available for the children.

### DIFF
--- a/Resources/public/js/models/ez-trashitemmodel.js
+++ b/Resources/public/js/models/ez-trashitemmodel.js
@@ -39,17 +39,22 @@ YUI.add('ez-trashitemmodel', function (Y) {
         },
 
         /**
-         * Restores the item to it's original location
+         * Restores the item to it's original location or to a given one if specified
          *
          * @method restore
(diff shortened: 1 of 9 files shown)


natanael89:
Sent to the QA.

natanael89:
Verified by the QA.

PR #891, opened 2017-08-07
EZP-27751: Combining stylesheets in development environment

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-27751

## Description

This PR reduce the number of HTTP request needed to load Platform UI in dev environment (from 291 to 59) by combining stylesheets into one file as it does in prod environment, which in consequence reduce application loading time.  

CSS files are combined, not minified so they are still readable.

## Statistics 

Measured as a complete time to load `/ez` in the following environment:
- Clean eZ Platform 1.10 installation
- Server side: VM with Ubuntu Server 16.04 + Apache 2.4 + PHP 7 (without enabled ACP, XDEBUG)
- Client side: macOS Sierra, Chrome 59 + disabled browser cache  

| #       | Before [s] | After [s] | 
|---------|------------|-----------| 
| 1       | 22,04      | 5,79      | 
| 2       | 22,19      | 5,65      | 
| 3       | 21,74      | 5,71      | 
| 4       | 22,52      | 5,44      | 
| 5       | 21,94      | 5,12      | 
| 6       | 22,1       | 6,51      | 
| 7       | 22,75      | 5,22      | 
| 8       | 21,99      | 5,62      | 
| 9       | 21,74      | 5,35      | 
| 10      | 21,25      | 5,65      | 
| **AVERAGE** | **22,026**     | **5,606**     |

### DIFF
--- a/Resources/views/PlatformUI/shell.html.twig
+++ b/Resources/views/PlatformUI/shell.html.twig
@@ -13,7 +13,7 @@
         </script>
         <script type="text/javascript" src="//cdn.crowdin.com/jipt/jipt.js"></script>
     {% endif %}
-{% stylesheets '$css.files;ez_platformui$' filter='cssrewrite' %}
+{% stylesheets '$css.files;ez_platformui$' filter='cssrewrite' combine=true %}
     <link rel="stylesheet" href="{{ asset_url }}"/>
 {% endstylesheets %}
     <style>


yannickroger:
You are forcing the behavior in the template, there must be a way to do it with a setting for the dev env.

adamwojs:
> there must be a way to do it with a setting for the dev env.

The `AsseticBundle` does not give the ability to configure this. If you don't define a value of `combine` in a `stylesheet` tag, the value is resolved depending on `asstic.debug`. 

Also, the `combine` doesn't accept variable reference as value, so for now,the only one way to solve this is:

```twig
{# parameters.cssCombine is injected from new configuration node #}
{% if parameters.cssCombine %} 
    {% stylesheets '$css.files;ez_platformui$' filter='cssrewrite' combine=true %}
    <link rel="stylesheet" href="{{ asset_url }}"/>
    {% endstylesheets %}
{% else %}
    {% stylesheets '$css.files;ez_platformui$' filter='cssrewrite' combine=false %}
    <link rel="stylesheet" href="{{ asset_url }}"/>
    {% endstylesheets %}
{% endif %}
```

But I hope that i'm wrong here. What do you think @yannickroger?

andrerom:
Ideally should have been AsseticBundle config, but given it's more or less dead status that probably won't happen even if we do PR for it. so..

Would perhaps call it `{% if parameters.forceCSSCombine %} ` or something, implying otherwise we would not define it in the markup like today.

andrerom:
@adamwojs Something like this perhaps?

```diff
diff --git a/Resources/config/default_settings.yml b/Resources/config/default_settings.yml
index 3a25c522..6c01caed 100644
--- a/Resources/config/default_settings.yml
+++ b/Resources/config/default_settings.yml
@@ -13,6 +13,7 @@ parameters:
     # Type ('js' or 'template')
     # ez_platformui.<scope>.yui.modules.<module_name>.type
     ez_platformui.default.css.files: []
+    ez_platformui.default.css.combine: true
     ez_platformui.default.javascript.files: []
 
     ezsettings.global.system_info_view:
diff --git a/Resources/config/services.yml b/Resources/config/services.yml
index 427168f1..6d36773e 100644
--- a/Resources/config/services.yml
+++ b/Resources/config/services.yml
@@ -89,6 +89,12 @@ services:
         tags:
             - {name: ezsystems.platformui.application_config_provider, key: 'countriesInfo'}
 
+    ezsystems.platformui.application_config.provider.css_combine:
+        class: '%ezsystems.platformui.application_config.provider.value.class%'
+        arguments: ['%ez_platformui.default.css.combine%']
+        tags:
+            - {name: ezsystems.platformui.application_config_provider, key: 'cssCombine'}
+
     ezsystems.platformui.application_config.provider.locales_map:
         class: "%ezsystems.platformui.application_config.provider.locales_map.class%"
         arguments:
diff --git a/Resources/views/PlatformUI/shell.html.twig b/Resources/views/PlatformUI/shell.html.twig
index 64bd4baa..3198976f 100644
--- a/Resources/views/PlatformUI/shell.html.twig
+++ b/Resources/views/PlatformUI/shell.html.twig
@@ -15,7 +15,7 @@
         </script>
         <script type="text/javascript" src="//cdn.crowdin.com/jipt/jipt.js"></script>
     {% endif %}
-{% stylesheets '$css.files;ez_platformui$' filter='cssrewrite' %}
+{% stylesheets '$css.files;ez_platformui$' filter='cssrewrite' combine=parameters.cssCombine %}
     <link rel="stylesheet" href="{{ asset_url }}"/>
 {% endstylesheets %}
     <style>
```


Also any benefit setting cache headers when in prod mode? Ref:
```diff
diff --git a/Controller/PlatformUIController.php b/Controller/PlatformUIController.php
index 59736a38..4003fe56 100644
--- a/Controller/PlatformUIController.php
+++ b/Controller/PlatformUIController.php
@@ -67,7 +67,10 @@ class PlatformUIController extends Controller
         return new Response(
             $content,
             Response::HTTP_OK,
-            ['Content-Type' => $type]
+            [
+                'Content-Type' => $type,
+                'Cache-Control' => 'public, max-age=60, s-max-age=60',
+            ]
         );
     }
```

adamwojs:
Closed in favour of https://github.com/ezsystems/PlatformUIBundle/pull/922

PR #893, opened 2017-08-17
EZP-27759: As an Editor, I want to copy the whole subtree of Content Objects

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-27759

## Description

This PR adds a "Copy subtree" functionality to UI. See the related issue for more details.

zrzut ekranu 2017-08-21 o 11 24 58

zrzut ekranu 2017-10-10 o 14 30 04

zrzut ekranu 2017-10-10 o 14 17 21

## TODO

- [X] Implementation
- [x] Extract translations
- [x] Screenshot to PR description
- [x] Tests

### DIFF
--- a/Resources/config/services.yml
+++ b/Resources/config/services.yml
@@ -36,6 +36,9 @@ parameters:
     # override of the REST content controller
     ezpublish_rest.controller.content.class: EzSystems\PlatformUIBundle\Controller\Rest\ContentController
 
+    # Maximum number of content object allowed to copy at once. Possible values are:
+    # -1 to disable limiting, 0 to disable this feature, > 0 enable limiting
+    ezsystems.platformui.application_config.copy_subtree.limit: 100
 services:
     ezsystems.platformui.twig.yui_extension:
         class: "%ezsystems.platformui.twig.yui_extension.class%"
@@ -137,6 +140,13 @@ services:
             - $api_keys$
         tags:
             - {name: ezsystems.platformui.application_config_provider, key: 'apiKeys'}
+
+    ezsystems.platformui.application_config.provider.copy_subtree:
+        class: "%ezsystems.platformui.application_config.provider.value.class%"
+        arguments:
+            - {limit: '%ezsystems.platformui.application_config.copy_subtree.limit%'}
+        tags:
+            - {name: ezsystems.platformui.application_config_provider, key: 'copySubtree'}
 
     ezsystems.platformui.controller.template:
         class: "%ezsystems.platformui.controller.template.class%"
--- a/Resources/config/yui.yml
+++ b/Resources/config/yui.yml
@@ -305,6 +305,7 @@ system:
                         - 'ez-imagevariationloadplugin'
                         - 'ez-contentcreateplugin'
                         - 'ez-copycontentplugin'
+                        - 'ez-copysubtreeplugin'
                         - 'ez-versionsplugin'
                         - 'ez-locationswapplugin'
                         - 'ez-draftconflictview'
@@ -1409,6 +1410,9 @@ system:
                 ez-copycontentplugin:
                     requires: ['ez-viewservicebaseplugin', 'ez-pluginregistry', 'ez-translator']
                     path: "%ez_platformui.public_dir%/js/views/services/plugins/ez-copycontentplugin.js"
+                ez-copysubtreeplugin:
+                    requires: ['ez-viewservicebaseplugin', 'ez-pluginregistry', 'ez-translator']
+                    path: "%ez_platformui.public_dir%/js/views/services/plugins/ez-copysubtreeplugin.js"
                 ez-discarddraftplugin:
                     requires: ['ez-viewservicebaseplugin', 'ez-pluginregistry']
                     path: "%ez_platformui.public_dir%/js/views/services/plugins/ez-discarddraftplugin.js"
--- a/Resources/public/js/models/ez-locationmodel.js
+++ b/Resources/public/js/models/ez-locationmodel.js
@@ -103,6 +103,19 @@ YUI.add('ez-locationmodel', function (Y) {
         },
 
         /**
+         * Copy the location under the given parenLocationId.
+         *
+         * @method copy
+         * @param {Object} options the options for the copy.
+         * @param {Object} options.api (required) the JS REST client instance
+         * @param {String} parentLocationId the location id where we should copy the content
+         * @param {Function} callback a callback executed when the operation is finished
+         */
+        copy: function(options, parentLocationId, callback) {
+            options.api.getContentService().copySubtree(this.get('id'), parentLocationId, callback);
+        },
+
+        /**
          * Updates the sortfield and sortOrder of the location
          *
          * @method _updateSorting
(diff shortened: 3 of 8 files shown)


adamwojs:
It looks like PHPUnit results are not related to changes in this PR...

andrerom:
> Has this topic been discussed ?

Somewhat, @adamwojs will do some testing on that and give info on this ti PM, so we can see how we might warn the user, either always, or at a given subtree size. On some subtree size we may even prohibit usage and refer to how it can be done using command or api instead.

adamwojs:
1. To confirm "Copy subtree" action we could use similar dialog as for "Send to trash" action

zrzut ekranu 2017-08-25 o 11 00 21

2. We are able to get subtree size using  the following request:

```
POST /ezp/v2/views
Content-Type: application/vnd.ez.api.ViewInput+json; version=1.1
Accept: application/vnd.ez.api.View+json; version=1.1
...

{
   "ViewInput":{
      "identifier":"foo",
      "public":false,
      "LocationQuery":{
         "Criteria":{},
         "FacetBuilders":{},
         "SortClauses":{ },
         "limit":0,
         "Filter":{
            "SubtreeCriterion":"/1/2/3/4/5/6/7"
         }
      }
   }
}
```
The only thing that left to do is to define what "Very Large Subtree"  mean.

andrerom:
> The only thing that left to do is to define what "Very Large Subtree" mean.

@adamwojs for that we need some numbers. And you'r lucky, QaFoo did test excatly publishing and subtree delete this spring: https://github.com/ezsystems/ezpublish-kernel/pull/2014 /cc @alongosz

andrerom:
@adamwojs maybe you should try to reproduce their results, and see if the 6-7 sec to delete subtree of 1000 items is correct. If it is maybe we put the limit at 500 or 1000 as we can assume it will take noticeably longer on a busy install.

Also the warning you proposed should ideally have some more text inline warning user that the operation may take a while, and asking user to wait until it is done before doing anything else, or if hmm.. will UI block during this? Ideally this should be async, but we're not there yet, so need to make sure user can see it is progressing and that UI is updated once done to not show what has been deleted anymore.

SylvainGuittard:
@dpobel about very large subtree, I asked the question on jira (https://jira.ez.no/browse/EZP-27759?focusedCommentId=214725&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-214725):
> in legacy copying a large amount of sub-items via the admin interface was using a lot of the server resources and might crash the server. I don't know what eZ Platform v1.x can handle without crashing. Any clue?

andrerom:
>  I don't know what eZ Platform v1.x can handle without crashing. Any clue?

It depends on the timeout of php, vs how fast the server in question can do the deletes. See link above for numbers for subtree delete, copy subtree probably takes longer, and what they tested there is was on install w/o load. So we can try with for instance 100 as limit or something and try how long time it takes. And also see how the Ui behaves while it's executing the copy, might need to re iterate the spec/wireframe based on that.

adamwojs:
Results of profiling the "Subtree copy" action

| Number of content objects | Minimum    | 90% Percentile | Median     | Average    | Maximum    | Std. Deviation |
|---------------------------|------------|----------------|------------|------------|------------|----------------|
| 10                        | 322.70 ms  | 322.70 ms      | 327.38 ms  | 345.63 ms  | 345.63 ms  | 28.69 ms       |
| 20                        | 695.68 ms  | 724.36 ms      | 722.38 ms  | 735.22 ms  | 833.00 ms  | 47.05 ms       |
| 30                        | 955.30 ms  | 1009.34 ms     | 1002.59 ms | 1024.06 ms | 1156.52 ms | 60.00 ms       |
| 50                        | 1573.67 ms | 1573.67 ms     | 1573.67 ms | 1573.67 ms | 1573.67 ms | 50.13 ms       |
| 100                       | 3077.46 ms | 3129.43 ms     | 3118.26 ms | 3142.64 ms | 3261.57 ms | 61.27 ms       |

During the test, there was measured a copying of folder with `N` articles, where `folder` and `article` has the following definition:

| Content Type | Fields                                              | Number of translations | Avg. number of versions |
|--------------|-----------------------------------------------------|---------------------|-------------------------|
| folder       | title (`text`)                                      | 1                   | 1                       |
| article      | title (`text`), body (`richtext`), author (`autor`) | 3                   | 8                       |

Test environment: Local installation of EZP 1.11 in production mode on Mac using PHP build-in server, PHP 7.1.8, MySQL5.7.17  

By default I've set subtree copy limit (`ezsystems.platformui.application_config.copy_subtree.limit` parameter) to 100 content objects.

SylvainGuittard:
Thanks @adamwojs for those results. I think those performances are because of php7.  
Some instances have php5 with eZ Platform 1.x. Can you do the same with php5.x?

andrerom:
@SylvainGuittard normally php 5.6 is somewhere between 1.5 to 2 times slower, so worst case we are talking 10 seconds for 100 items, with high database load maybe 20-30 seconds. So limit of 100 should be fine here I think, and besides we strongly recommend people to use PHP7 now anyway.

However imo a clear followup here is to expose a command to copy subtree w/o limit, and refer to that in help text here.

SylvainGuittard:
@andrerom: Thanks. I think we can move forward on this PR.
@bdunogier: could you please create a follow up story for the `copy / subtree` command line?

adamwojs:
Feature "in action" for PHP 5.6 and PHP 7.1 (in dev mode):

* https://www.youtube.com/watch?v=57GLSfDZpgI
* https://www.youtube.com/watch?v=oufugxnvskY

It looks like I've forgotten to disable xdebug so results in the table above will be slightly better.

alongosz:
@adamwojs you probably need to rebase this PR as #911 has been merged.

SylvainGuittard:
@adamwojs Thank you very much for the screencasts. Good job

mnocon:
@adamwojs , I have a following issue:

1. Create the following structure:
a) Home/FolderLvl1
b) Home/FolderLvl1/FolderLvl2
c) Home/FolderLvl1/FolderLvl2/BlogPost
(FolderLvl1, FolderLvl2 are Folders, BlogPost is a Blog post 🙂  )
2. Copy subtree FolderLvl2 into FolderLvl1
3. Copy subtree FolderLvl2 into FolderLvl1 again

I have following results in FolderLvl1 view mode:
screen shot 2017-10-30 at 12 31 38

Clearing the cache makes the item count correct (so does adding another Content Item)

Could you take a look at this?

andrerom:
@mnocon is the same issue present when using existing copy button? If so then we can perhaps handle that as follow up.

mnocon:
@andrerom I've tried the following:
1. Structure:
- Home/FolderLvl1
- Home/FolderLvl1/FolderLvl2
2. Copy FolderLvl2 into FolderLvl1
3. Copy FolderLvl2 into FolderLvl1 again

And the counter is updated correctly, looks like it's only related to the new feature.

andrerom:
> And the counter is updated correctly, looks like it's only related to the new feature.

ok clear :)

adamwojs:
Bug discovered by @mnocon is related to HTTP Cache purging. See https://github.com/ezsystems/ezplatform-http-cache/pull/18 for more details.

andrerom:
@adamwojs can you confirm if tests passed before the lint issues appeared? Aka are we ready to merge?

PR #895, opened 2017-08-28
EZEE-1710: Change selenium version

mnocon:
> JIRA: [EZEE-1710](https://jira.ez.no/browse/EZEE-1710)

# Description

The previously used step did not use an existing user account (well known admin) but created a new one, using the following format as name: `User#uniqid()`

Essentially it works the same way, but the created username is long. So long, that in some resolutions it overlaps with the "Admin Panel" button:
error

The browsers drivers behave differently with such elements: Firefox driver used currently is able to click on Admin Panel, while Chrome driver suggested in https://github.com/ezsystems/ezplatform/pull/203 fails. With this change the username will be "Administrator User", which fits in the available space.

In addition I will create an issue that the layout does not handle very long usernames well.

### DIFF
--- a/Features/Users/users.feature
+++ b/Features/Users/users.feature
@@ -1,7 +1,7 @@
 Feature: Use the eZ Users field
 
     Background:
-        Given I am logged in as an Administrator in PlatformUI
+        Given I am logged in as admin on PlatformUI
 
     @javascript @common
     Scenario: Verify the existence of Users page and it's content


mnocon:
Not sure about the Travis build (I think I didn't break anything),will look at it.

andrerom:
afaik the issue is not your doing, it's as said in another PR regression from changes to how config is generated in 1.11. 

Maybe change this in 1.7 and up, so we can change to use selenium 3.4 also in 1.7 and up branch on ezplatform .

mnocon:
Judging by the experiences I had so far, I wouldn't change the Selenium version for 1.7 branch - it's working well there and there is no need for upgrade. I'm afraid that this change comes with temporary stability issues and it's not worth it.

Because of changes made in Studio in 1.11 there is a real need to upgrade, but I'd like to limit these changes only where it's necessary (1.11 and higher).

PR #896, opened 2017-08-31
Fix EZP-27833: Wrong parameter placeholder in notification text after hiding location

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-27833

# Description

In the translation file, there is wrong placeholder for `id` parameter in `Hiding Location "%id"` message (missing closing `%`).

# Steps to reproduce

1. Go to "Content" > "Content structure" > "Locations" Tab of any content e.g "eZ Platform"
2. "Hide" any of location 
3. There is not "%id" value parameter in the notification message text (see attached screenshot)


yannickroger:
You should make travis pass before merging.

andrerom:
> You should make travis pass before merging.

The failure has nothing to do with this change, so not really something that should be in this PR or? The change needed to do that is refactor the test to use ConfigResolver instead of hard coding assumptions on dumped internal symfony parameters structure.

PR #901, opened 2017-09-13
[WIP] Test PlatformUI PR

mnocon:
# Do not merge

> JIRA: [EZEE-1710](https://jira.ez.no/browse/EZEE-1710)

*The previously suggested changes are not needed, increasing shm size should do the trick.*:
https://github.com/ezsystems/ezplatform/pull/214

~~This is something similar to the problem that occured in https://github.com/ezsystems/PlatformUIBundle/pull/895:~~
~~When elements overlap Chrome driver reacts differently than Firefox. It tries to click the spinner and fails with UnknownError exception:~~
```
unknown error: Element <div class="ez-view-formblockview ez-view-studioblockview ezs-block--new in-workspace yui3-dd-drop" 
data-block-view="fb.FormBlockView" data-block-name="Form" data-block-type="formbuilder" 
data-block-id="b-ab96afcb-3783-3777-9c79-2e15375f9660" 
data-block-edit-text="Edit this Form block to add content" id="yui_3_18_1_1_1505199529310_13737" style="min-height: 96px;">...</div> 
is not clickable at point (544, 524). Other element would receive the click: 
<body class="ez-platformui-app-body yui3-skin-platformui" 
id="yui_3_18_1_1_1505199529310_386">...</body>
```

~~This time the issue occurs in various places and is involved with the spinner visible when application is reloading.~~
~~Screen of example failure:
example_failure~~
~~The selector to detect the spinner is still correct, I suspect Chrome driver is more prone to the spinner "blinking" (disappearing and appearing again in very short time) - it can be seen in the Studio part.~~

~~I've considered various solutions:~~
~~- Overriding the click method to catch the Exception, wait some time and then try again feels like an overkill.~~
~~- Different selectors for the spinner also did not help, as the selector is correct (and it really disappears for a moment).~~

~~My approach requires multiple checks in a row to pass in order to decide, that the spinner is indeed gone. It's a simple solution, but has some drawbacks:~~
~~- there's still a chance that a test will fail (hopefully it's low - I've chosen the $consecutiveTimesRequired by hand and it looks stable)~~
~~- every waitWhileLoading execution will execute at least 3 sleeps instead of 1 (one sleep is 250 ms).~~

~~Execution times before and after, based on StudioUI common suite:~~
~~Before; 42 min 23 sec~~
~~After: 45 min 52 sec~~

~~I think it's acceptable, at least for now.~~

### DIFF
--- a/bin/travis/prepare_behat.sh
+++ b/bin/travis/prepare_behat.sh
@@ -13,7 +13,7 @@ EZPLATFORM_BRANCH=`php -r 'echo json_decode(file_get_contents("./composer.json")
 
 cd "$HOME/build"
 
-git clone --depth 1 --single-branch --branch "$EZPLATFORM_BRANCH" https://github.com/ezsystems/ezplatform.git
+git clone --depth 1 --single-branch --branch "$EZPLATFORM_BRANCH" https://github.com/mnocon/ezplatform.git
 cd ezplatform
 
 # Install everything needed for behat testing, using our local branch of this repo
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     "extra": {
         "_ci_branch-comment_": "Keep ci branch up-to-date with master or branch if on stable. ci is never on github but convention used for ci behat testing!",
         "_ezplatform_branch_for_behat_tests_comment_": "ezplatform branch to use to run Behat tests",
-        "_ezplatform_branch_for_behat_tests": "1.11",
+        "_ezplatform_branch_for_behat_tests": "ezee-1710-increase-shm-size",
         "branch-alias": {
             "dev-master": "1.11.x-dev",
             "dev-tmp_ci_branch": "1.11.x-dev"


dpobel:
> The selector to detect the spinner is still correct, I suspect Chrome driver is more prone to the spinner "blinking" (disappearing and appearing again in very short time) - it can be seen in the Studio part.

If I understand that correctly, the spinner appears, disappears and reappears while things are still loading and you are trying to workaround that, right ? But isn't that behavior a bug ? I mean I find it weird to workaround a bug in a code that is suppose (among other things) to avoid bugs. Can't this bug be fixed ?

mnocon:
About the bug: at first I thought so too, but as far as I know this behaviour is not that easy to fix and wasn't prioritised as high to be worth the effort. It's not that the spinner behaviour got worse all of a sudden - we've changed the test driver and the new driver sometimes behaves in a different way than the previous one.

On a different note': your comment made me realise that there is also another option to try, I will try to change this solution.

About the test failure:

```
When I create a content of this Type                                                             
# EzSystems\PlatformUIBundle\Features\Context\Fields::createAContentOfThisType()
      WebDriver\Exception\UnknownError: unknown error: session deleted because of page crash
```
at first I thought that I should be able to reproduce it locally, as it the same error that enforced this change: https://github.com/ezsystems/ezplatform/blob/master/doc/docker/selenium.yml#L16
but I can't - 128m looks enough when running the tests not marked with "edge" tag.

Still, I suspect that changing it to 256m might help - I'll test it in https://github.com/ezsystems/ezplatform/pull/211

PR #903, opened 2017-09-15
[WIP] EZP-27869: Allow to restore a translation from the Version tab

dpobel:
JIRA: https://jira.ez.no/browse/EZP-27869

# Description

**WIP**

## Tasks

* [x] Expose languages of a version in a usable way
* [ ] Add the restore link
* [ ] Add method in JavaScript REST client (when corresponding API is designed)
* [ ] Design code to restore the translation (depends on API design) and use it after click


andrerom:
SIDE: This will need to bump kernel dependency for UI once it is looked at again to reflect version fo kernel that added translated related API's that will be needed for this.

PR #905, opened 2017-09-22
EZP-28030: Alloy editor add support for more builtin buttons

Plopix:
> https://jira.ez.no/browse/EZP-28030

This PR adds ` 'subscript', 'superscript', 'quote', 'strike',` and `'numbered list'`
Enjoy
screen shot 2017-09-21 at 6 49 53 pm
screen shot 2017-09-21 at 6 55 54 pm

Todo:
- [x] make sure kernel can parse/ output all the added syntax

### DIFF
--- a/Resources/public/js/alloyeditor/buttons/list.js
+++ b/Resources/public/js/alloyeditor/buttons/list.js
@@ -12,7 +12,7 @@ YUI.add('ez-alloyeditor-button-list', function (Y) {
 
     var AlloyEditor = Y.eZ.AlloyEditor,
         React = Y.eZ.React,
-        ButtonList;
+        ButtonList, ButtonListIndexed;
 
     /**
      * The ButtonList component represents a button to add an unordered list (ul).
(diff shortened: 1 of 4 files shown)


Resources/public/js/alloyeditor/buttons/list.js
```diff
@@ -58,13 +58,66 @@ YUI.add('ez-alloyeditor-button-list', function (Y) {
...
+
+    /**
+     * The ButtonListIndexed component represents a button to add an ordered list (ol).
+     *
+     * @uses AlloyEditor.ButtonCommand
+     * @uses AlloyEditor.ButtonStateClasses
+     *
+     * @class eZ.AlloyEditor.ButtonList
+     */
+    ButtonList = React.createClass({displayName: "ButtonListIndexed",
```
dew326: I think something is wrong with this generated js. I think this should be `ButtonListIndexed = ...` instead of `ButtonList = ...`

andrerom:
Fixed the wrongly generated js.

@Plopix did you test this extensively btw? was the format properly saved to richtext and back to view and edit?

@dew326 looks ok? If so we can potentially send to QA to make sure proper testing is done.
                 In parallel one of us can try to add some unit tests so this is ready for merge if it passes.

dew326:
@andrerom now it looks ok and I think we can send this to QA.
About testing, I will not be able to focus on this in the current sprint (it's hard to add something extra into the sprint since we are more to scrum). 
Let me know if you can handle this or should we try to add this to the upcoming sprint.

micszo:
As a smoke test I tried using the new numbered list feature but it looks like there is work on backend needed - XML validation error occurs:
screen shot 2017-10-06 at 10 31 50

andrerom:
Thanks @micszo, then we keep this for 1.13 release, it needs a deeper look and work on kernel.

andrerom:
@micszo Tested and only fdound issues with striked: https://github.com/ezsystems/ezpublish-kernel/pull/2103

What browser did you use?

micszo:
I used Chrome.

Plopix:
Hello,

@andrerom it is in prod for on EE project. I tested the basics, store, edit, view.
I did not test extensively on all the browser though.
++

andrerom:
Thansk @mnocon

PR #906, opened 2017-09-25
EZP-27650: + button is not visible in RichText

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-27650

**Requires:**
https://github.com/ezsystems/PlatformUIAssetsBundle/pull/20

### DIFF
--- a/Resources/config/yui.yml
+++ b/Resources/config/yui.yml
@@ -11,7 +11,7 @@ system:
                 ez-translator:
                     path: "%ez_platformui.public_dir%/js/services/ez-translator.js"
                 alloyeditor:
-                    path: "%ez_platformui.external_assets_public_dir%/vendors/alloy-editor/dist/alloy-editor/alloy-editor-all.js"
+                    path: "%ez_platformui.external_assets_public_dir%/vendors/alloy-editor/dist/alloy-editor/alloy-editor-no-react-min.js"
                 ez-alloyeditor:
                     requires: ['alloyeditor']
                     path: "%ez_platformui.public_dir%/js/external/ez-alloyeditor.js"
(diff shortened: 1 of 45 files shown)


mnocon:
Shouldn't this PR be to 1.11 branch? AFAIK master in StudioUI is currently 1.11.

dew326:
@mnocon We will introduce this fix only for master (1.12), changing the way of loading React is too big for a patch release.

mnocon:
Ok, sounds good to me - can you change fixVersion in JIRA?

dew326:
Unit tests fixed.

andrerom:
@dew326 Ready for merge? I assume the Behat failure is unrelated here btw.

dew326:
@andrerom Those PRs are all ready for merge.

The best order of merging would be:
1. Merge PRs in assets bundle (both Platform and Platform EE)
2. Tag new version of assets
3. Merge PRs for PlatformUI and StudioUI 

Also, I think Platform EE Assets should be tag as 2.0 not as next 1.x to avoid update to this changes in existing projects.
//cc @lserwatka @sunpietro

andrerom:
Sounds like a plan, who handles it? If you are busy I can optionally look into that.

lserwatka:
Yes, we need to be careful here as we are introducing React in <head> for PlatformUI. So the assets should have higher number, so versions below 1.12 won't be affected.

dew326:
If we release assets in version 4.0 and 2.0 I think we should update composer.json in PlatformUI and StudioUI. For now it's: 
`"ezsystems/platform-ui-assets-bundle": "^3.1.0",` and
`"ezsystems/ezplatform-ee-assets": "^1.0.0",`

andrerom:
yes, add commit for that so it's ready for merge once assets have been merged, built and tagged.

PR #907, opened 2017-09-27
EZP-24699: Enrich policy limitations view with human readable names

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-24699
> **Requires merge first**: https://github.com/ezsystems/repository-forms/pull/159

# Description 

This PR makes the values of policy limitations readable to humans. See https://github.com/ezsystems/repository-forms/pull/159 for more details

zrzut ekranu 2017-10-11 o 10 25 23

# TODO

- [X] Implementation
- [x] Tests

### DIFF
--- a/Resources/public/css/theme/views/serverside.css
+++ b/Resources/public/css/theme/views/serverside.css
@@ -107,3 +107,11 @@
 .ez-view-serversideview .ez-contenttype-view .ez-fielddefinition-setting-value {
     background: #f3f3f3;
 }
+
+.ez-view-serversideview .ez-policy-limitation {
+    list-style-type: none;
+}
+
+.ez-view-serversideview .ez-policy-limitation .ez-limitation-identifier {
+    font-weight: bold;
+}
--- a/Resources/public/css/views/serverside.css
+++ b/Resources/public/css/views/serverside.css
@@ -101,3 +101,7 @@
 .ez-view-serversideview .ez-toolbar .ez-toolbar-action {
     display: inline;
 }
+
+.ez-view-serversideview .ez-policy-limitation {
+    padding: 0;
+}
--- a/Resources/translations/role.en.xlf
+++ b/Resources/translations/role.en.xlf
@@ -349,6 +349,86 @@
         <note>key: selected.content.not.user.group</note>
         <jms:reference-file>Resources/public/js/views/services/ez-roleserversideviewservice.js</jms:reference-file>
       </trans-unit>
+      <trans-unit id="e71e94ff640270b4c257fedcd489a574c090cb80" resname="policy.limitation_class">
+        <source>Content Type</source>
+        <target>Content Type</target>
+        <note>key: policy.limitation_class</note>
+      </trans-unit>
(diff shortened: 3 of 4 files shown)


andrerom:
Any reason why you use `;` over `,` ? :)

andrerom:
@adamwojs what is remaining here for the `wip`?

andrerom:
Ah, the behat failures seems relevant, getting some help from QA on that?

adamwojs:
@andrerom I think that could speed up things, also I haven't fixed tests yet, because we still waiting for PM feedback. 😉

SylvainGuittard:
To make the limitations easy to read, I prefer to have `<br>` instead of `;`
Here is the screenshot of the interface for eZ Platform v2.
image
If we can have the same rendering for the limitations, it will improve the user experience (v1 -> v2).

adamwojs:
PR updated according to @SylvainGuittard suggestions.

SylvainGuittard:
Thank you so much @adamwojs :)

adamwojs:
@SylvainGuittard I don't see any problem, but we should make all limitation identifiers translatable.

andrerom:
@adamwojs For the translation change _(here or as followup, check with LS)_ could you also add something like a `title` attribute on the bold part with a string like `Limitation Identifier: %identifier%` so advanced users does not loose information by humanizing this?

SylvainGuittard:
> Did you intend for the strings to be translated btw?
We'll have to. It makes sense to have the admin interface translated. Tables and content in the tables as well.

adamwojs:
PR updated according to @SylvainGuittard and @andrerom suggestions. I've also applied suggestions about HTML structure from https://github.com/ezsystems/ezplatform-admin-ui/pull/5

lserwatka:
@adamwojs what tests are missing here to remove WIP from this PR?

andrerom:
@SylvainGuittard While I can agree on the domain names, I would disagree on things like module and function names I think, they are property values that are not translatable in the system. Would be like attempting to translate class names :P But more importantly doc is only in english, looking up what these things are for would be broken if we translate them.

adamwojs:
@lserwatka Previously the build was failing on Behat, but everything is OK after merge ezsystems/repository-forms#159 and rebase PR with current master.

lserwatka:
Thank you @adamwojs for your contribution. @SylvainGuittard we should have a follow up story on translations. I merging this one to move things forward.

SylvainGuittard:
@andrerom 
> I would disagree on things like module and function names I think

My bad. My comment was not very precise. I never thought about translating `module` and `function`.

Everything in the limitation column should be translated.
- `Content Type` should be `Type de Contenu` in french (according to the glossary: https://crowdin.com/translate/ezplatform/1140/en-fr) and same for all limitations
- Content Type names should be displayed in French is the content type has a french translation. Which brings another topic: the translation of content types...
- In am not sure yet about Path (`Subtree`)...

@lserwatka: 
> we should have a follow up story on translations. I merging this one to move things forward.

Sure. I will do it.

SylvainGuittard:
@lserwatka @adamwojs follow up story: https://jira.ez.no/browse/EZP-28072

PR #909, opened 2017-10-10
EZP-28039: Video width and height params are not saved during content creation

kmadejski:
JIRA issue: https://jira.ez.no/browse/EZP-28039

# Description
As mentioned in the JIRA issue, video width and height params are shown for the user as placeholders, but field values are still empty during content publishing.

After this change, values gathered from the file are set as field values, not as placeholders.

Some unnecessary tests were removed.

### DIFF
--- a/Resources/public/js/views/fields/ez-media-editview.js
+++ b/Resources/public/js/views/fields/ez-media-editview.js
@@ -74,7 +74,9 @@ YUI.add('ez-media-editview', function (Y) {
 
             this._attachedViewEvents.push(player.on('loadedmetadata', function () {
                 container.removeClass(IS_BEING_UPDATED);
-                that._updateWidthHeightFieldValue(player.get('videoWidth'), player.get('videoHeight'));
+                if (!that.get('width') || !that.get('height')) {
+                    that._updateWidthHeightFieldValue(player.get('videoWidth'), player.get('videoHeight'));
+                }
             }));
             this._attachedViewEvents.push(player.on('error', function () {
                 container.removeClass(IS_BEING_UPDATED);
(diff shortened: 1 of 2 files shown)


lserwatka:
@mnocon Could you have a look here today?

mnocon:
@lserwatka Yes, I will take a look at this after Alloy Editor.

mnocon:
@kmadejski The values are saved, but I don't think they are set everywhere they should be.

I'm doing following steps:
1. Create ContentType with Media field, for e.g. Video
2. Create content of type Video, select some file
3. You can see, that in preview video width and height are set correctly
4. Publish content
5. Go to the content preview

Result: 
- the dimensions are not displayed in view mode (I feel like they should be, because they values are already set?)
- when I access the content from front office, the following tag is generated
```
<video src="/content/download/56/new_ezmedia_1/test3.mov">
Your browser does not support html5 video.
</video>
```
(it has not width/height).

6. When I go to Edit mode

- When the content is loaded the width and height properties are set correctly (1280x720 in my case).

7. I set the "Width" property to 300 and publish

The width property is displayed in view mode (height is not).
The following tag is generated for front office view:
```
<video src="/content/download/56/new_ezmedia_1/test3.mov" width="300">
Your browser does not support html5 video.
</video>
```
(it has only width specified)

8. When going to Edit mode again:
- the width value is set to "1280" (not 300). Height is still 720.

Could you take a look at this?

kmadejski:
@mnocon thank you for the feedback. I've updated PR and added changes to cover mentioned use-case. Could you test it again, please?

mnocon:
@kmadejski thanks, I will take a look at this - it might take a while, we've just started certification for 1.12 release and I'll be doing mostly that now.

lserwatka:
@kmadejski Could you rebase here, so fixer will start working again?

kmadejski:
@lserwatka done :)

mnocon:
@lserwatka this PR is ok, it can be merged

andrerom:
Merged all the way up to master, nice @kmadejski!

PR #910, opened 2017-10-11
Fix Bower install on travis and php_cs_fixer

dew326:


vidarl:
Can you try to add `'yoda_style' => false,` in setRules() call in .phpcs.
Looks like yoda style is enabled by default in recent version of phpcsfix

dew326:
All green 🎉

PR #912, opened 2017-10-13
EZP-28104: As a Editor I want empty Performance tab removed

andrerom:
> Issue: https://jira.ez.no/browse/EZP-28104

As the planned feature for this tab won't happen before v2 it has been approved by PM _(slack product channel, sept 14th)_ to remove this for 1.x. And for 2.x we won't add it before it actually has functionality.

Todo:
- [x] adapt failing tests


lserwatka:
@sunpietro @dew326 could you please have a look here?

Tests/js/views/assets/ez-navigationhubview-tests.js
```diff
@@ -152,7 +148,7 @@ YUI.add('ez-navigationhubview-tests', function (Y) {
                 userProfileView: this.userProfileView,
                 container: '.container',
                 user: this.userMock,
-                studioplusNavigationItems: [
+                studioNavigationItems: [
```
sunpietro: I think, the whole property (and related tests) can be removed, unless you want to test any other tab - the studio tab in this case. Do you agree?

andrerom: removing it I think broke some of the tests, as some expects several zones, so hence the adaption here in the end result.

PR #913, opened 2017-10-17
EZP-28003: Unable to go back to the Content structure after content move to the Media library

sunpietro:
https://jira.ez.no/browse/EZP-28003

### DIFF
--- a/Resources/public/js/views/services/ez-locationviewviewservice.js
+++ b/Resources/public/js/views/services/ez-locationviewviewservice.js
@@ -2,6 +2,7 @@
  * Copyright (C) eZ Systems AS. All rights reserved.
  * For full copyright and license information view LICENSE file distributed with this source code.
  */
+/*jshint esversion: 6 */
 YUI.add('ez-locationviewviewservice', function (Y) {
     "use strict";
     /**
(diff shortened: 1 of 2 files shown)


sunpietro:
Just one comment from my side:

I've used `window.location.replace` instead of `window.history.pushState` because I wanted to remove the incorrect location of content from the browser history. `pushState` adds a new entry a last visited page, but you can still get back to the incorrect location.

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -2,6 +2,7 @@
  * Copyright (C) eZ Systems AS. All rights reserved.
  * For full copyright and license information view LICENSE file distributed with this source code.
  */
+/*jshint esversion: 6 */
```
dew326: It will be better to add this into config not per file

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -273,53 +274,89 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+        },
+
+        /**
+         * Updates the browser history by removing a reference to a previous,
+         * already non-existent location of moved content.
+         *
+         * @method _updatePrevHistory
+         * @protected
+         * @param {any} location
+         * @param {any} callback
```
dew326: I think it's not `any`.

Resources/public/js/views/services/ez-locationviewviewservice.js
```diff
@@ -273,53 +274,89 @@ YUI.add('ez-locationviewviewservice', function (Y) {
...
+         */
+        _updatePrevHistory(location, callback) {
+            const urlPart = window.decodeURIComponent(window.location.hash);
+            const language = urlPart.slice(urlPart.lastIndexOf('/'), urlPart.length);
+            const hashStart = '#/view/';
```
dew326: Can't we avoid hardcoded parts?

sunpietro:
ping @kmadejski @dew326

Resources/public/js/views/services/ez-navigationhubviewservice.js
```diff
@@ -23,6 +26,42 @@ YUI.add('ez-navigationhubviewservice', function (Y) {
...
+
+        /**
+         * Updates content link in 'Content' navigation zone for current selected content
+         *
+         * @method _resetContentNavigationItem
+         * @protected
+         * @param {Event} event
+         */
+        _resetContentNavigationItem: function () {
+            const contentItem = this.get('platformNavigationItems').find(item => {
```
dew326: I think method `getNavigationItem` do the same thing.

sunpietro: Yes, but this one is faster because I'm looking for a specific item in a specific zone. I don't have to iterate through all the zones to get that one specific item :)

dew326: I agree it's faster around 0.0002ms :D 
I think we should use dedicated methods for finding items if such method exists.

sunpietro: Ok, I can agree to this option.

lserwatka:
@kmadejski Could you help test this?

kmadejski:
@sunpietro works well. Just only one thing, when I moved content I got an exception, to be more precise I got the only red notification on the bottom. There is no error in console and no invalid request in Network tab (in developer tools).

screen shot 2017-10-20 at 14 39 26

dew326:
I think it's something from the StudioUI. Do you have Platform or Platform EE?

sunpietro:
If you have eZPlatform-EE then you definitely need the code from the PR I referenced above.
Please, update your studio-ui-bundle and use the branch from PR

kmadejski:
@sunpietro I apologize, I overlooked this reference. I patched also StudioUI bundle, and yes, the previous message doesn't appear anymore, but I've got another "red" message (no Console/Network errors). I'm getting this message right after entry to the *Content structure* tab.

screen shot 2017-10-20 at 15 17 28

sunpietro:
@kmadejski I'm on it

kmadejski:
@sunpietro after update with https://github.com/ezsystems/StudioUIBundle/pull/852 everything works well!

dew326:
Question: Isn't it a BC break if changes in PlatformUI requires additional work in StudioUI?

sunpietro:
@dew326 it's not BC break. PlatformUI didn't have any functionality to correct the history of previously visited pages after moving a content item to a new location. This fix corrects this behaviour.
The StudioUI relies on a locations previewed in the Content Structure. If the Content Structure navigation item points to an incorrect location, then moving from Content Structure to the Page Mode will cause errors. The error existed before and now it's fixed.

lserwatka:
@micszo we need a QA here + corresponding Studio PR.

micszo:
@lserwatka OK, it's top on the list now.

PR #915, opened 2017-10-25
EZP-28130: Console error when moving empty richText field

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-28130


andrerom:
should this target older branches also?

dew326:
To be honest, I don't know. The 1.7 is not affected and do we plan to release 1.11.2?

andrerom:
@dew326 if it was reported by customer, it would have been 1.11 so support can provide patch and in case we tag. But since this is not, this is edge case, and 1.12 is imminent, there is no need.

PR #916, opened 2017-11-02
[travis] fix for jshint version

damianz5:


damianz5:
It's better now but still some issues in two tests
https://travis-ci.org/ezsystems/PlatformUIBundle/jobs/297389290

bin/travis/setupnode.sh
```diff
@@ -2,6 +2,8 @@
 
 set -e
 
+npm install -g npm
```
sunpietro: Just curious. You'are asking npm to install itself globally, right?

bin/travis/setupnode.sh
```diff
@@ -13,6 +15,8 @@ npm install -g phantomjs@1.9.8
 echo "> installing global packages"
 npm install -g grunt-cli grover bower
 
+npm install -g jshint@2.9.5
```
sunpietro: I think it should install the latest jshint, not the specific version.

bin/travis/setupnode.sh
damianz5: There were some version differences, but after tests I can remove this line

dew326:
To avoid the issue in tests try to remove esversion:6 from the file and add to the jshint config:
https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/services/ez-locationviewviewservice.js#L5
https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/services/ez-navigationhubviewservice.js#L5

and add the new rule in config: `esverion: 6`
https://github.com/ezsystems/PlatformUIBundle/blob/master/jshint.json

damianz5:
@sunpietro can you check the two failed tests? [L779](https://travis-ci.org/ezsystems/PlatformUIBundle/jobs/297850932#L779) and [L785](https://travis-ci.org/ezsystems/PlatformUIBundle/jobs/297850932#L785)

damianz5:
@dew326 it's out of scope of this PR - the test environment is ready to use, please fix the tested code in separate PR

andrerom:
If we merge this, can someone take care of adapting the failing tests? From what I can see they haven't changes since 2016, so the break is caused by updated does or nod, or?

sunpietro:
@andrerom I'll correct them after merging this PR

andrerom:
Ok, @damianz5 could you change this to either 1.11 or 1.12 so we can merge?

damianz5:
@andrerom done

andrerom:
Thanks @damianz5, merged, all yours @sunpietro :D ;)

PR #917, opened 2017-11-03
EZP-28146: Unable to replace an image - UDW does not allow me to confirm an image selection

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-28146


andrerom:
Should this go in older branches also perhaps?

dew326:
1.7 is not affected and the 1.12 branch doesn't exist so I assume we keep master as 1.12 for now?

andrerom:
@dew326  Yes

lserwatka:
I created 1.12 branch, and configured master for 1.13.x-dev

PR #920, opened 2017-11-15
EZP-28078: Embedding a Content Object in the RichText Editor makes page scroll up

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-28078

## Description

This PR contains the workaround patch for unwanted page scroll to the top when user embeds a content/image or choose link location from UDW. The solution is to save and restore scroll position after command execution.    

The direct cause of this issue is `focus` call in https://github.com/ckeditor/ckeditor-dev/blob/release/4.5.x/core/editable.js#L90 (in case of embedding content).

### DIFF
--- a/Resources/public/js/alloyeditor/buttons/embed.js
+++ b/Resources/public/js/alloyeditor/buttons/embed.js
@@ -75,13 +75,16 @@ YUI.add('ez-alloyeditor-button-embed', function (Y) {
          */
         _addEmbed: function (e) {
             var contentInfo = e.selection.contentInfo,
-                scrollX = window.pageXOffset || document.documentElement.scrollLeft,
-                scrollY = window.pageYOffset || document.documentElement.scrollTop,
                 widget;
 
-            this.execCommand();
-            // Workaround for https://jira.ez.no/browse/EZP-28078
-            window.scrollTo(scrollX, scrollY);
+            if (CKEDITOR.env.chrome) {
+                // Workaround for https://jira.ez.no/browse/EZP-28078
+                var scrollY = window.pageYOffset;
+                this.execCommand();
+                window.scroll(window.pageXOffset, scrollY);
+            } else {
+                this.execCommand();
+            }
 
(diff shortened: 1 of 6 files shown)


Resources/public/js/alloyeditor/buttons/embed.js
```diff
@@ -75,9 +75,14 @@ YUI.add('ez-alloyeditor-button-embed', function (Y) {
          */
         _addEmbed: function (e) {
             var contentInfo = e.selection.contentInfo,
+                scrollX = window.pageXOffset || document.documentElement.scrollLeft,
```
dew326: As far as I know `document.documentElement.scrollLeft` and `document.documentElement.scrollTop` is for IE 8 and lower. We dropped support a long time ago for IE, so I think you can omit this.

dew326:
The same issue is on v2 and I wouldn't mind if you could create a patch for v2 ;)

dew326:
What about updating image and embed? It also fires UDW.

andrerom:
> The same issue is on v2 and I wouldn't mind if you could create a patch for v2 ;)

lets finish this one first, no customers on v2 yet :P

andrerom:
@adamwojs / @dew326 I see the issue is fixed / improved in CKEditor 4.7.x: https://github.com/ckeditor/ckeditor-dev/blob/release/4.7.x/core/editable.js#L90

Any chance we could get Alloy folks to update CKeditor? At a minimum for v2 which will use newer version right?

andrerom:
@adamwojs Can you perhaps align with the fix in CKEditor? As 1. Consider to use scrollTop, 2. only apply on chome if that is only browser affected by this

adamwojs:
> @adamwojs Can you perhaps align with the fix in CKEditor? As 1. Consider to use scrollTop, 2. only apply on chome if that is only browser affected by this

@andrerom Good idea :-)

adamwojs:
PR updated according to suggestions.

PR #922, opened 2017-11-21
EZP-27751: combining stylesheets in dev env & cache combined js in prod

andrerom:
> JIRA: https://jira.ez.no/browse/EZP-27751

Builds upon research done by @adamwojs in #891 _(so for performance numbers and context see that PR)_ to make this hopefully into a more ready state.

Todo: 
- [x] Test and Travis

### DIFF
--- a/Resources/views/PlatformUI/shell.html.twig
+++ b/Resources/views/PlatformUI/shell.html.twig
@@ -20,7 +20,7 @@
             <link rel="stylesheet" href="{{ asset_url }}"/>
         {% endstylesheets %}
     {% else %}
-        {% stylesheets '$css.files;ez_platformui$' filter='cssrewrite' combine=false %}
+        {% stylesheets '$css.files;ez_platformui$' filter='cssrewrite' %}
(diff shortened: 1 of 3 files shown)


Resources/views/PlatformUI/shell.html.twig
```diff
@@ -15,9 +15,15 @@
...
     {% endif %}
-{% stylesheets '$css.files;ez_platformui$' filter='cssrewrite' %}
-    <link rel="stylesheet" href="{{ asset_url }}"/>
-{% endstylesheets %}
+    {% if parameters.css.combine %}
+        {% stylesheets '$css.files;ez_platformui$' filter='cssrewrite' combine=true %}
+            <link rel="stylesheet" href="{{ asset_url }}"/>
+        {% endstylesheets %}
+    {% else %}
+        {% stylesheets '$css.files;ez_platformui$' filter='cssrewrite' combine=false %}
```
andrerom: self nitpick, I realize this should not be set to false but just use system default. otherwise it will be false also for prod.

PR #927, opened 2017-12-14
Fix-EZP-28398: Using Assetic versioning breaks Platform Admin UI

kmadejski:
JIRA issue: https://jira.ez.no/browse/EZP-28398

Fix for 1.7.x.

### DIFF
--- a/Resources/views/PlatformUI/shell.html.twig
+++ b/Resources/views/PlatformUI/shell.html.twig
@@ -59,7 +59,8 @@
     <script>
         {{ ez_platformui_yui_config( "YUI.GlobalConfig" ) }}
 
-        var CKEDITOR_BASEPATH = ALLOYEDITOR_BASEPATH = "{{ asset("bundles/ezplatformuiassets/vendors/alloy-editor/dist/alloy-editor/") }}";
+        var CKEDITOR_BASEPATH = ALLOYEDITOR_BASEPATH = "/bundles/ezplatformuiassets/vendors/alloy-editor/dist/alloy-editor/";
+        {#var CKEDITOR_BASEPATH = ALLOYEDITOR_BASEPATH = "{{ asset("bundles/ezplatformuiassets/vendors/alloy-editor/dist/alloy-editor/") }}";#}
 
         YUI().use('ez-platformuiapp', function (Y) {
             var app = new Y.eZ.PlatformUIApp({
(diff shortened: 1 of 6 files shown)


Resources/views/PlatformUI/shell.html.twig
```diff
@@ -59,7 +59,8 @@
     <script>
         {{ ez_platformui_yui_config( "YUI.GlobalConfig" ) }}
 
-        var CKEDITOR_BASEPATH = ALLOYEDITOR_BASEPATH = "{{ asset("bundles/ezplatformuiassets/vendors/alloy-editor/dist/alloy-editor/") }}";
+        var CKEDITOR_BASEPATH = ALLOYEDITOR_BASEPATH = "/bundles/ezplatformuiassets/vendors/alloy-editor/dist/alloy-editor/";
+        {#var CKEDITOR_BASEPATH = ALLOYEDITOR_BASEPATH = "{{ asset("bundles/ezplatformuiassets/vendors/alloy-editor/dist/alloy-editor/") }}";#}
```
sunpietro: Commented code

Resources/views/PlatformUI/shell.html.twig
```diff
@@ -59,7 +59,7 @@
     <script>
         {{ ez_platformui_yui_config( "YUI.GlobalConfig" ) }}
 
-        var CKEDITOR_BASEPATH = ALLOYEDITOR_BASEPATH = "{{ asset("bundles/ezplatformuiassets/vendors/alloy-editor/dist/alloy-editor/") }}";
+        var CKEDITOR_BASEPATH = ALLOYEDITOR_BASEPATH = "/bundles/ezplatformuiassets/vendors/alloy-editor/dist/alloy-editor/";
```
andrerom: is there some code anywhere prefixing this url with possible base path and so on or is it not needed here?

kmadejski: No, there is no code which is prefixing this patch. It is used only by AlloyEditor when it loads dependencies (like `config.js`,  `styles.js` and so on).

kmadejski:
@adamwojs your suggestion from https://github.com/ezsystems/PlatformUIBundle/pull/928#pullrequestreview-83496793 added also here.

lserwatka:
@kmadejski please merge this up to 1.12, master then master to 2.0

PR #931, opened 2017-12-19
EZP-28258: Can't publish/save content with missing Selection option

glye:
> Fix https://jira.ez.no/browse/EZP-28258
> Sent to QA

Adds a placeholder title for obsoleted selection choices. This has the positive side effect of fixing the 'x' for removal.

So when you try to publish content with a selection using options that have been removed in the type, what happens is:
- You get the same validation error as before: "Option with index -1 does not exist in the field definition." I was considering to add "The option may have been removed", but now I think that's redundant. The "-1" isn't helpful for non-devs, we could perhaps change it. What we can't do is show the title of the removed option, that info is gone.
- Obsoleted selection choices have the placeholder title "Selected option does not exist", or whatever else we translate it to.
- You can click the x to remove the invalid options.
- When viewing the field, the same placeholder titles are shown, rather than emptiness.

Not fixed:
- Error when doing the same using API calls, or outside of platform-ui, or 2.0. I don't think we should silently remove such entries. For one, that will just lead to another mysterious error if the field is required. Not sure how we should deal with it.

- [ ] Extract translations (script fails locally)

### DIFF
--- a/Resources/config/yui.yml
+++ b/Resources/config/yui.yml
@@ -986,7 +986,7 @@ system:
                     type: 'template'
                     path: "%ez_platformui.public_dir%/templates/fields/view/binaryfile.hbt"
                 ez-selection-view:
-                    requires: ['ez-fieldview', 'selectionview-ez-template']
+                    requires: ['ez-fieldview', 'selectionview-ez-template', 'ez-translator']
                     path: "%ez_platformui.public_dir%/js/views/fields/ez-selection-view.js"
                 selectionview-ez-template:
                     type: 'template'
@@ -1097,7 +1097,7 @@ system:
                     type: 'template'
                     path: "%ez_platformui.public_dir%/templates/fields/view/richtext.hbt"
                 ez-selection-editview:
-                    requires: ['ez-fieldeditview', 'ez-selectionfilterview', 'event-tap', 'event-outside', 'node-screen', 'selectioneditview-ez-template']
+                    requires: ['ez-fieldeditview', 'ez-selectionfilterview', 'event-tap', 'event-outside', 'node-screen', 'selectioneditview-ez-template', 'ez-translator']
                     path: "%ez_platformui.public_dir%/js/views/fields/ez-selection-editview.js"
                 selectioneditview-ez-template:
                     type: 'template'
--- a/Resources/public/js/views/fields/ez-selection-editview.js
+++ b/Resources/public/js/views/fields/ez-selection-editview.js
@@ -196,7 +196,11 @@ YUI.add('ez-selection-editview', function (Y) {
             }
 
             Y.Array.each(valueIndexes, function (index) {
-                res.push(options[index]);
+                if (index in options) {
+                    res.push(options[index]);
+                } else {
+                    res.push(Y.eZ.trans('select.option.does.not.exist', {}, 'fieldedit'));
+                }
             });
             return res;
         },
--- a/Resources/public/js/views/fields/ez-selection-view.js
+++ b/Resources/public/js/views/fields/ez-selection-view.js
@@ -37,10 +37,18 @@ YUI.add('ez-selection-view', function (Y) {
             if ( fieldDefinitionSettings.isMultiple ) {
                 res = [];
                 Y.Array.each(fieldValue, function (key) {
-                    res.push(fieldDefinitionSettings.options[key]);
+                    if (key in fieldDefinitionSettings.options) {
+                        res.push(fieldDefinitionSettings.options[key]);
+                    } else {
+                        res.push(Y.eZ.trans('select.option.does.not.exist', {}, 'fieldview'));
+                    }
                 });
             } else {
-                res = fieldDefinitionSettings.options[fieldValue];
+                if (fieldValue in fieldDefinitionSettings.options) {
+                    res = fieldDefinitionSettings.options[fieldValue];
+                } else {
+                    res = (Y.eZ.trans('select.option.does.not.exist', {}, 'fieldview'));
+                }
             }
 
             return res;
--- a/Tests/js/views/fields/assets/ez-selection-view-tests.js
+++ b/Tests/js/views/fields/assets/ez-selection-view-tests.js
@@ -65,7 +65,7 @@ YUI.add('ez-selection-view-tests', function (Y) {
 
             "Test empty field value": function () {
                 this._testValue(
-                    {}, undefined, "The value in the template should be undefined"
+                    [], undefined, "The value in the template should be undefined"
                 );
             },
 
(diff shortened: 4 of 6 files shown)


glye:
FYI the extract translations command `php app/console translation:extract -b eZPlatformUIBundle en --exclude-dir Features --exclude-dir DependencyInjection --exclude-dir ApplicationConfig --exclude-dir Controller --exclude-dir docs --exclude-dir grunt --exclude-dir node_modules --exclude-dir SiteAccess --exclude-dir Twig --exclude-dir vendor --exclude-dir Command --output-format xlf`

fails with `[Symfony\Component\Debug\Exception\FatalThrowableError]
Call to a member function addSource() on null`

glye:
@dew326 Are you cool with this? I'm unsure about https://github.com/ezsystems/PlatformUIBundle/pull/931/commits/d03b1f841a9c599793ccf3f55a3cfcea09a0731b - don't know why this had to be changed.

glye:
@bdunogier Do you know why the extract translations command would fail like this?
`php app/console translation:extract -b eZPlatformUIBundle en --exclude-dir Features --exclude-dir DependencyInjection --exclude-dir ApplicationConfig --exclude-dir Controller --exclude-dir docs --exclude-dir grunt --exclude-dir node_modules --exclude-dir SiteAccess --exclude-dir Twig --exclude-dir vendor --exclude-dir Command --output-format xlf`

fails with `[Symfony\Component\Debug\Exception\FatalThrowableError]
Call to a member function addSource() on null`

dew326:
This test is about preview empty selection and looks like the test is ok but the code doesn't pass it.

As I understand you can reproduce the issue in the UI by following steps:
1. Create a content type with field type selection (single, not required)
2. Create a content of this content type
3. Publish content (do not fill the selection field)
4. Check the preview.

According to test, the preview should show the message about removed option instead of empty value.

glye:
@dew326 Thanks. That's interesting. Preview of empty selection actually fails before and after my change, because the `field.value.selection` array is empty and the code assumes it won't be. Another thing to fix.
https://github.com/ezsystems/ezpublish-kernel/blob/e2d26faa54715675fc8f1eef3defbe192973bf5b/eZ/Bundle/EzPublishCoreBundle/Resources/views/content_fields.html.twig#L170

dew326:
I meant the preview in admin under the Content structure.

glye:
Oh, right. Manual test doesn't match automated test result there, I have to double check it. I'll fix the editing preview bug separately.

glye:
@dew326 Still stuck on this: I reverted the questionable commit. When manually testing, the content preview shows empty value, as it should. While the Travis test shows `select.option.does.not.exist domain=fieldview`, which makes no sense to me. Do you have any idea what's going on?

dew326:
I think I found it. In real app the fieldValue is an array `[]`, but in the test case is set to object `{}`. 
Looks like the test is wrong.

https://github.com/ezsystems/PlatformUIBundle/blob/master/Tests/js/views/fields/assets/ez-selection-view-tests.js#L68

glye:
Much thanks! Trying out modified test, and squashed & reordered PR.

glye:
Sent to QA.

bdunogier:
@glye: did you try to use `bin/extract-translations.sh` ? It is meant to be run from the `vendor/ezsystems/platform-ui-bundle` directory.

glye:
@bdunogier Thanks & sorry for late reply. In this version there is only a `bin/updatestrings.sh`. Running it fails with the same error as the `translation:extract` command, but with info about file name and exception trace. File: `Resources/public/templates/rawcontent.hbt`

```
Exception trace:
 () at /home/gl/git/ezplatform/vendor/ezsystems/platform-ui-bundle/Translation/HandleBarsFileVisitor.php:33
 EzSystems\PlatformUIBundle\Translation\HandleBarsFileVisitor->visitFile() at n/a:n/a
 call_user_func_array() at /home/gl/git/ezplatform/vendor/jms/translation-bundle/JMS/TranslationBundle/Translation/Extractor/FileExtractor.php:237
 JMS\TranslationBundle\Translation\Extractor\FileExtractor->extract() at /home/gl/git/ezplatform/vendor/jms/translation-bundle/JMS/TranslationBundle/Translation/ExtractorManager.php:137
 JMS\TranslationBundle\Translation\ExtractorManager->extract() at /home/gl/git/ezplatform/vendor/jms/translation-bundle/JMS/TranslationBundle/Translation/Updater.php:253
 JMS\TranslationBundle\Translation\Updater->setConfig() at /home/gl/git/ezplatform/vendor/jms/translation-bundle/JMS/TranslationBundle/Translation/Updater.php:143
 JMS\TranslationBundle\Translation\Updater->process() at /home/gl/git/ezplatform/vendor/jms/translation-bundle/JMS/TranslationBundle/Command/ExtractTranslationCommand.php:131
 JMS\TranslationBundle\Command\ExtractTranslationCommand->execute() at /home/gl/git/ezplatform/vendor/symfony/symfony/src/Symfony/Component/Console/Command/Command.php:242
 Symfony\Component\Console\Command\Command->run() at /home/gl/git/ezplatform/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:861
 Symfony\Component\Console\Application->doRunCommand() at /home/gl/git/ezplatform/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:193
 Symfony\Component\Console\Application->doRun() at /home/gl/git/ezplatform/vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Console/Application.php:83
 Symfony\Bundle\FrameworkBundle\Console\Application->doRun() at /home/gl/git/ezplatform/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:117
 Symfony\Component\Console\Application->run() at /home/gl/git/ezplatform/app/console:29
```

glye:
I got the translation extractor to run by adding `--exclude-name=rawcontent.hbt --exclude-name=contenteditform.hbt`, to exclude the two templates it crashed on. But the output includes a lot of stuff I have not touched (might just mean files were stale), and my new string is not included (not good).

andrerom:
> nd my new string is not included (not good).

yeh, that's one of several indications something is wrong indeed.

glye:
@dew326 @andrerom With QA now passed, are you OK with merging this, and leave the translation update as a followup to someone who has it working? (Possibly even myself, later on)

barbaragr:
There is a ticket in Jira with missing translations. I will add this one to this ticket after merge. :)

andrerom:
TBH I'm a bit unsure, is PR's supposed to _(manually or via script)_ update strings in at least `Resources/translations` with the PR itself like before, or is that not needed anymore.

PR #935, opened 2018-01-02
EZP-28565: Insert/remove link from the RichText editor makes page scroll up

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-28565

## Description 

This PR contains the workaround patch for unwanted page scroll to the top when insert/remove a link from the RichText editor. The solution is to save and restore scroll position after command execution (same as https://github.com/ezsystems/PlatformUIBundle/pull/920)


lserwatka:
@micszo we will need a round of QA.

PR #936, opened 2018-01-02
EZP-27264: New draft based on a selected version should respect relations

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-27264

## Description

This PR contains a patch for the wrong object loaded as a value of `relation`/`relationlist` field types when a user creates a new draft based on the previous version. Cause of the problem is that ID of the related object(s) was always taken from current version instead of the field value.

## Steps to reproduce

> 1. Create two folders (Folder1; Folder2)
> 2. Create one article and in the relation fieldtype, related it with "Folder1"
> 3. Publish. This will create version1
> 4. Edit the created article and update name to "Article2" and replace the relation to "Folder2"
> 5. Publish it. This will create version2
> 6. Go to "versions" tab of the article, and create a new draft based on verison1 (the one that was first related with "Folder1")
> 7. When we edit this new draft, we see that we have the related object as "Folder2", that is related to version2 (should be "Folder1")

### DIFF
--- a/Resources/public/js/views/fields/ez-relation-editview.js
+++ b/Resources/public/js/views/fields/ez-relation-editview.js
@@ -67,8 +67,7 @@ YUI.add('ez-relation-editview', function (Y) {
         _fireLoadFieldRelatedContent: function () {
             if ( !this._isFieldEmpty() ) {
                 this.fire('loadFieldRelatedContent', {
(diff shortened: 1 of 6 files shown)


adamwojs:
Target branch has been changed to 1.7.

lserwatka:
@micszo we need a round of QA here.

adamwojs:
@lserwatka I believe this PR is ready to merge 😉

PR #937, opened 2018-01-02
EZP-28577: "Last contributor" shouldn't show error when "Creator" is removed

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-28577

## Description

This PR contains patch for "An error occurred while loading the last contributor." message displayed in the "Last contributor" field on "Details" tab when in fact the "Creator" was removed. Both fields were dependent on same variable indicating loading error.
 
## Steps to reproduce

> 1. Create a new installation of eZ Platform.
> 2. In the Platform UI, create new User in the "Editors" User Group with login "test_editor".
> 3. Log out and log in again as "test_editor".
> 4. Under "Home", create new Article Content Object named "Test Article".
> 5. Log out and log in again as "admin".
> 6. Edit and publish a new version of "Test Article".
> 7. Go to the "Details" tab for "Test Article".Confirm that there is "Administrator User" in the "Last contributor" field.
> 8. Delete User "test_editor".
> 9. Go to the "Details" tab for "Test Article". You will notice the error mentioned above for the "Last contributor" field.

PR #938, opened 2018-01-08
1.13 merge into 2.0

andrerom:
This is 1.13 merged into 2.0 in order to start to merge up changes again.
All Link Manager changes where attempted to be omitted here, otherwise all changes kept.


Resources/config/services.yml
```diff
@@ -1,4 +1,5 @@
 parameters:
+    ezsystems.platformui.twig.text_extension.class: Twig_Extensions_Extension_Text
```
andrerom: @adamwojs I guess I should skip this as well here, right?

adamwojs: Yes, I've enabled this Twig extension because we need `truncate` filter for Link Manager.

andrerom: Thanks, omitted it then.

Resources/public/js/alloyeditor/buttons/linkedit.js
```diff
@@ -312,15 +312,23 @@ YUI.add('ez-alloyeditor-button-linkedit', function (Y) {
...
 
             selection.selectBookmarks(bookmarks);
 
             this.props.cancelExclusive();
 
-            editor.fire('actionPerformed', this);
+            if (navigator.userAgent.indexOf('Chrome') > -1) {
+                // Workaround for https://jira.ez.no/browse/EZP-28565
+                scrollY = window.pageYOffset;
+                editor.fire('actionPerformed', this);
```
sunpietro: I'd move this line outside the `if` clause, because it's fired in both cases.

Resources/public/js/alloyeditor/buttons/linkedit.js
```diff
@@ -353,13 +361,21 @@ YUI.add('ez-alloyeditor-button-linkedit', function (Y) {
...
 
             if (this.state.linkHref) {
                 linkAttrs.href = this.state.linkHref;
                 linkUtils.update(linkAttrs, this.state.element, modifySelection);
 
-                editor.fire('actionPerformed', this);
+                if (navigator.userAgent.indexOf('Chrome') > -1) {
+                    // Workaround for https://jira.ez.no/browse/EZP-28565
+                    scrollY = window.pageYOffset;
+                    editor.fire('actionPerformed', this);
```
sunpietro: Same here.

Resources/public/js/views/tabs/ez-locationviewdetailstabview.js
```diff
@@ -178,16 +179,21 @@ YUI.add('ez-locationviewdetailstabview', function (Y) {
...
                 attributeName: 'creator',
+                loadingErrorAttributeName: 'creatorLoadingError'
             });
 
             if (creatorId === ownerId) {
                 this.onceAfter('creatorChange', function (e) {
                     this.set('owner', this.get('creator'));
                 });
+                this.onceAfter('creatorLoadingErrorChange', function (e) {
+                    this.set('ownerLoadingError', this.get('creatorLoadingError'));
```
sunpietro: Can it be a situation like, a creator is different from an owner?

Resources/public/js/alloyeditor/buttons/linkedit.js
andrerom: This is from branch merge, so should not be fixed here, but rather by @adamwojs on original affected 1.x branch (PR was: #935)

Resources/public/js/views/tabs/ez-locationviewdetailstabview.js
andrerom: That was #937, @adamwojs?

Resources/public/js/alloyeditor/buttons/linkedit.js
andrerom: same as above :)

Resources/public/js/views/tabs/ez-locationviewdetailstabview.js
adamwojs: @sunpietro Yes, owner and creator can be different (see steps to reproduce in PR linked by @andrerom). Variables names are misleading, but actually "creator" is the author of the last version ("Last contributor") and "Owner" is the author of the initial version content. For instance:

Resources/public/js/alloyeditor/buttons/linkedit.js
adamwojs: I disagree if the assumption that we apply this workaround only for Chrome (made in https://github.com/ezsystems/PlatformUIBundle/pull/920#issuecomment-344910360) is still valid.

lserwatka:
Is it safe to merge this one @andrerom ?

andrerom:
I think so, will make it easier to maint the branches.

PR #939, opened 2018-01-19
EZEE-1845: Can't select siteaccess using More button in StudioUI

kmadejski:
JIRA issue: https://jira.ez.no/browse/EZEE-1845

# Description
This PR fixes issue, that you cant use *More* button in LandingPage editor. 
It is a bug introduced in: https://github.com/ezsystems/PlatformUIBundle/commit/fd65b441c40405e91c23a56b1a27aa7b8e8ae859

The solution is maybe not the prettiest one, but otherwise, we would have to re-build few parent elements to fix it in the cleaner way.


dew326:
I think it should go to 1.7? Besides that +1.

kmadejski:
@dew326 thanks for the review. You are right, it definitely should go to the 1.7.x.

PR #945, opened 2018-01-24
EZEE-1850: StudioUI doesn't work if symfony/webpack-encore is configu…

kmadejski:
…red to use json_manifest

JIRA issue: https://jira.ez.no/browse/EZEE-1850

# Description
As mentioned in JIRA issue, since Symfony 3.3 `symfony/webpack-encore` is core functionality. Due to that, a user should be able to use it, but the previous implementation generates warning when user configured assets to work with encore using `framework.assets: json_manifest_path` setting.

### DIFF
--- a/ApplicationConfig/Providers/RootInfo.php
+++ b/ApplicationConfig/Providers/RootInfo.php
@@ -75,6 +75,6 @@ class RootInfo implements Provider
             return $url;
         }
 
-        return preg_replace('/\?' . preg_quote($version, '/') . '/', '', $url);
+        return preg_replace('/\?' . $version . '/', '', $url);
     }
 }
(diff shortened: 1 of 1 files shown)


andrerom:
hmm, what kind of string is version here? Seems like it will contain a `?` in last condition, right?

kmadejski:
@andrerom in case of using `framework.assetic.version: some_version`, `$version` will always contain `?some_version`. In case of `framework.assets: json_manifest_path` `$version` is just `/`.

andrerom:
@kmadejski could you add a short mention of that before `$version = $this->assetsPackages->getVersion($path);` line, would make it simpler next time anyone looks at that code ;)

andrerom:
Thanks for the doc.

Q: Why do we then need the `preg_quote` change? as far as I can see the variant where `$version = '/'` will trigger the if clause as it does not contain `?` and just be returned.

kmadejski:
@andrerom indeed, it is no longer necessary, since first `if` deals with that.
 
I thought that it might be necessary in case if user set version string like for e.g. `?a1-(a1)/b` (I know, not really possible, but Symfony allows it). I tested it once again and it works without `preg_quote` so I removed it.

PR #946, opened 2018-01-31
EZP-28847: Allows user to edit directly in the language displayed

StephaneDiot:
JIRA: https://jira.ez.no/browse/EZP-28847

## Description

This allows user to edit a content directly in the languageCode selected in the language switcher of the content view (instead of mainLanguageCode)

## ScreenCast

https://youtu.be/Fcb5VGW0Wak

### DIFF
--- a/Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
+++ b/Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
@@ -313,15 +313,8 @@ YUI.add('ez-locationviewviewservice-tests', function (Y) {
         },
 
         "Should fire `editContentRequest` when receiving an editAction event": function () {
-            var languageCode = "fre-FR",
-                contentInfo = new Mock(),
+            var contentInfo = new Mock(),
                 contentType = {};
-
-            Mock.expect(contentInfo, {
-                method: 'get',
-                args: ['mainLanguageCode'],
-                returns: languageCode
-            });
 
             Mock.expect(this.locationMock, {
                 method: 'get',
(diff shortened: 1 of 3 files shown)


mnocon:
Just a quick question - isn't master equivalent to 2.1 release? Your screencast uses v1, which should be 1.13 branch IMHO.

andrerom:
@arfaram could you or customer create support ticket for this so we can track this as part of second and thirdline? /cc @kmadejski 

Also, I'm assume there are no objections on this @SylvainGuittard ?

SylvainGuittard:
@andrerom no objections :)
Thanks @StephaneDiot for this fix.

Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
```diff
@@ -315,7 +315,8 @@ YUI.add('ez-locationviewviewservice-tests', function (Y) {
         "Should fire `editContentRequest` when receiving an editAction event": function () {
             var languageCode = "fre-FR",
                 contentInfo = new Mock(),
-                contentType = {};
+                contentType = {},
+                that = this;
 
             Mock.expect(contentInfo, {
```
dew326: I think this `expect` is not needed in this test case and the mocks should be verified at the end of test

Tests/js/views/services/assets/ez-locationviewviewservice-tests.js
```diff
@@ -315,7 +315,8 @@ YUI.add('ez-locationviewviewservice-tests', function (Y) {
         "Should fire `editContentRequest` when receiving an editAction event": function () {
             var languageCode = "fre-FR",
                 contentInfo = new Mock(),
-                contentType = {};
+                contentType = {},
+                that = this;
```
dew326: Cannot we use `bind`? `that = this` is dirty to me ;)

PR #955, opened 2018-03-12
EZP-28781: Content shouldn't be visible in ezobjectrelationlist after sending it to trash

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-28781

## Description 

This PR hides items which point to content moved to trash from the view/edit of `ezobjectrelationlist`. 

## Steps to reproduce 

> 1. Create a new content type that contains ezobjectrelationlist FieldType, e.g. "menu".
> 2. Create new content using "menu" content type, name it "test" and select some contents (e.g. articles).
> 3. Publish this new content.
> 4. Select this content and go to one of the articles that is on relation list.
> 5. Send this article to the trash.
> 6. Go back to "test". Note, that deleted article still appears on the list and clicking on it doesn't redirect to this article.

### DIFF
--- a/Resources/public/js/views/fields/ez-relationlist-editview.js
+++ b/Resources/public/js/views/fields/ez-relationlist-editview.js
@@ -132,14 +132,14 @@ YUI.add('ez-relationlist-editview', function (Y) {
                 relatedContentsJSON = [];
 
             if (relatedContents !== null) {
-                relatedContentsJSON = relatedContents.reduce(function (total, value) {
+                relatedContentsJSON = relatedContents.reduce(Y.bind(function (total, value) {
                     var relatedContentJSON = value.toJSON();
                     if (this._isNewRelation(value) || relatedContentJSON.resources.MainLocation) {
                         total.push(relatedContentJSON);
                     }
 
                     return total;
-                }.bind(this), []);
+                }, this), []);
             }
 
             return {
--- a/Resources/public/js/views/fields/ez-relationlist-editview.js
+++ b/Resources/public/js/views/fields/ez-relationlist-editview.js
@@ -132,15 +132,31 @@ YUI.add('ez-relationlist-editview', function (Y) {
                 relatedContentsJSON = [];
 
             Y.Array.each(relatedContents, function (value) {
-                relatedContentsJSON.push(value.toJSON());
-            });
+                var relatedContentJSON = value.toJSON();
+
+                if (this._isNewRelation(value) || relatedContentJSON.resources.MainLocation) {
+                    relatedContentsJSON.push(relatedContentJSON);
+                }
+            }, this);
 
             return {
-                relatedContents:  relatedContentsJSON,
+                relatedContents: relatedContentsJSON,
                 loadingError: this.get('loadingError'),
                 isEmpty: this._isFieldEmpty(),
                 isRequired: this.get('fieldDefinition').isRequired,
             };
+        },
+
+        /**
+         * Check if relation to the content is new (non-existed before edit)
+         *
+         * @method _isNewRelation
+         * @protected
+         * @param {Object} object
+         * @return {boolean}
+         */
+        _isNewRelation: function(object) {
+            return object.name === 'contentInfoModel';
         },
 
         /**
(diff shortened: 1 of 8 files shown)


andrerom:
@adamwojs Seems to be some failures here.

Resources/public/js/views/fields/ez-relationlist-editview.js
```diff
@@ -129,20 +129,37 @@ YUI.add('ez-relationlist-editview', function (Y) {
...
             var relatedContents = this.get('relatedContents'),
-                relatedContentsJSON = [];
+                relatedContentsJSON = [],
+                isNewRelation = Y.bind(this._isNewRelation, this);
 
             Y.Array.each(relatedContents, function (value) {
-                relatedContentsJSON.push(value.toJSON());
+                var relatedContentJSON = value.toJSON();
+
+                if (isNewRelation(value) || relatedContentJSON.resources.MainLocation) {
```
andrerom: should this be `this._ isNewRelation(value)` ?

adamwojs: True, it could be simplified by passing `this` as the third argument of `Y.Array.each` 😉

Resources/public/js/views/fields/ez-relationlist-editview.js
```diff
@@ -132,17 +132,33 @@ YUI.add('ez-relationlist-editview', function (Y) {
                 relatedContentsJSON = [];
 
             Y.Array.each(relatedContents, function (value) {
```
sunpietro: You can do it more performant and proper way (if it was in ES6 it would be even shorter, but the unit testing setup doesn't allow ES6):
```javascript
relatedContentsJSON = relatedContentsJSON.reduce(function (total, value) {
    var relatedContentJSON = value.toJSON();

    if (this._isNewRelation(value) || relatedContentJSON.resources.MainLocation) {
        total.push(relatedContentJSON);
    }

    return total;
}.bind(this), []);
```

Resources/public/js/views/fields/ez-relationlist-editview.js
```diff
@@ -132,17 +132,33 @@ YUI.add('ez-relationlist-editview', function (Y) {
...
 
+        /**
+         * Check if relation to the content is new (non-existed before edit)
+         *
+         * @method _isNewRelation
+         * @protected
+         * @param {Object} object
+         * @return {boolean}
+         */
+        _isNewRelation: function(object) {
```
sunpietro: CS: `_isNewRelation: function (object) {`

Resources/public/js/views/fields/ez-relationlist-editview.js
```diff
@@ -132,17 +132,33 @@ YUI.add('ez-relationlist-editview', function (Y) {
...
                 isRequired: this.get('fieldDefinition').isRequired,
             };
         },
 
+        /**
+         * Check if relation to the content is new (non-existed before edit)
+         *
+         * @method _isNewRelation
+         * @protected
+         * @param {Object} object
```
sunpietro: What object? The param description is meaningless in this case. Please, name the param correctly and/or give it a meaningful description.

Resources/public/js/views/fields/ez-relationlist-view.js
```diff
@@ -66,7 +66,10 @@ YUI.add('ez-relationlist-view', function (Y) {
                 relatedContentsJSON = [];
 
             Y.Array.each(relatedContents, function (value) {
```
sunpietro: Please, follow suggestion from the comment above.

adamwojs:
@barbaragr This is already known issue: https://jira.ez.no/browse/EZP-28439

andrerom:
@adamwojs Is the failure expected? 

```
✖ [eZ Relation list Edit View tests]: Passed: 4 Failed: 17 Total: 21 (ignored 0) (0.589 seconds)
    Test required field
       Unexpected error: 'undefined' is not a function (near '...}.bind(this), []);...')


And more
```

adamwojs:
No, it isn't. I will look into this.

Resources/public/js/views/fields/ez-relationlist-editview.js
adamwojs: @sunpietro It seems that unit tests are failing on the native`bind` method, so we decided to back to `Y.bind` (probably might be missing support for it in PhantomJS <=2.1).

sunpietro: Ahh, yes. This is so limiting ...

PR #957, opened 2018-03-19
Fix EZP-28953: RichText editor's toolbox obscures the preceding paragraph's content

adamwojs:
>JIRA: https://jira.ez.no/browse/EZP-28953

## Description 

Based on solution proposed by @inakijv. The main difference is that there is not large fixed margin between two block elements. Video presenting how the patch works:  http://www.youtube.com/watch?v=NppuFBH1Dto

## Steps to reproduce

1. Create the new Article
2. Into the Intro/Body (or any other RichText field) paste some dummy text in few paragraphs
3. Click on text to get RichText editor toolbox. You will see, that toolbox obscures some parts of the text as shown in the screenshot.

### DIFF
--- a/Resources/public/css/alloyeditor/general.css
+++ b/Resources/public/css/alloyeditor/general.css
@@ -3,12 +3,44 @@
  * For full copyright and license information view LICENSE file distributed with this source code.
  */
 
+.ez-view-richtexteditview .ez-richtext-editor {
+    padding: 0 .5em;
+}
+
+.ez-view-richtexteditview .ez-richtext-editor h1.is-block-focused,
+.ez-view-richtexteditview .ez-richtext-editor h2.is-block-focused,
+.ez-view-richtexteditview .ez-richtext-editor h3.is-block-focused,
+.ez-view-richtexteditview .ez-richtext-editor h4.is-block-focused,
+.ez-view-richtexteditview .ez-richtext-editor h5.is-block-focused,
+.ez-view-richtexteditview .ez-richtext-editor h6.is-block-focused,
+.ez-view-richtexteditview .ez-richtext-editor p.is-block-focused {
+    margin-top: 3rem;
+}
+
 .ae-toolbar-add,
 .ae-toolbar,
 .ae-toolbar-styles {
     z-index: 1200;
 }
 
-.ae-ui .ae-arrow-box.ae-arrow-box-bottom.ez-ae-arrow-box-left:after {
-    left: -75%;
+.ae-ui .ae-arrow-box:after {
+    display: none;
+}
+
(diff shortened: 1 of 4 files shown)


andrerom:
Nice, maybe we can try to avoid the slightly jumpy behavior by setting aside some space in top of the editor? That said I guess we can't eliminate it completely with this approach right? Ping @inakijv / @SylvainGuittard Has some view on this?

Resources/public/css/alloyeditor/general.css
```diff
@@ -3,12 +3,49 @@
  * For full copyright and license information view LICENSE file distributed with this source code.
  */
 
+.ez-view-richtexteditview .ez-richtext-editor {
+    padding: 0 .5em;
+}
+
+.ez-view-richtexteditview .ez-richtext-editor h1.is-block-focused,
```
inakijv: Isn't possible to add the spacing without being focused, just `.ez-view-richtexteditview .ez-richtext-editor h1`?
If we add space when focused, would (probably) generate another usability issue, because text doesn't have to move, the toolbar has to adapt to the context.
Besides of this, Isn't possible to add the spacing after/below the paragraph instead of on top of it?

Resources/public/css/alloyeditor/general.css
```diff
@@ -3,12 +3,49 @@
...
 }
 
-.ae-ui .ae-arrow-box.ae-arrow-box-bottom.ez-ae-arrow-box-left:after {
-    left: -75%;
+.ae-ui .ae-arrow-box:after {
+    display: none;
+}
+
+.ae-ui .ae-toolbar, .ae-ui [class^=ae-toolbar-] {
+    padding: 0;
```
inakijv: Isn't possible to add some `margin-top` here as suggested? (Suggestion was `margin-top: 9px;`) We need the toolbar as close as possible to the dashed line, but not touching it.

adamwojs:
PR updated according to @inakijv and @dew326 suggestions. Screenshots from current version: 

zrzut ekranu 2018-03-20 o 10 51 11
zrzut ekranu 2018-03-20 o 10 51 19

adamwojs:
> @adamwojs This means there are always tons of space between each block which is what you tried to avoid in first iteration right?

@andrerom Unfortunately yes.

antoligy:
This isn't a viable solution, and attempting to give this to users is just going to offend them.  Please do not merge this.

antoligy:
I've added a suggestion to my previous comment about a way of mitigating the initial UX problem where this box disrupts a user's flow!

adamwojs:
PR reverted to the original solution 😉

antoligy:
Thanks! 👍

adamwojs:
Closed in favor of https://github.com/ezsystems/PlatformUIBundle/pull/962

PR #959, opened 2018-03-21
EZEE-1875: Wrong table headers in language/section details view.

mikadamczyk:
The translation that was duplicated in ezplatform-admin-ui has been removed.


adamwojs:
Safer would be to duplicate correct translation in both packages IMO, but since 2.0 this package is used only as a dependency for landing page editor so it should be fine.

PR #960, opened 2018-03-23
EZP-28966 Disable deleting main translation button

konradoboza:
> https://jira.ez.no/browse/EZP-28966

This PR is about disabling deleting button for main translation.


Resources/public/js/views/actions/ez-translateactionview.js
```diff
@@ -62,15 +62,35 @@ YUI.add('ez-translateactionview', function (Y) {
 
                 location: this.get('location').toJSON(),
                 content: this.get('content').toJSON(),
-                translations: translationsList,
-                canDeleteTranslations: translationsList.length > 1,
```
dew326: This is potential BC break, someone could override the template and still rely on this variables.

Resources/public/js/views/actions/ez-translateactionview.js
```diff
@@ -62,15 +62,35 @@ YUI.add('ez-translateactionview', function (Y) {
...
+         * @method _decorateTranslationsList
+         * @param {Array} translationsList list of translation that will be decorated
+         * @return {Object} object which contains language codes with visibility flags
+         * @protected
+         */
+        _decorateTranslationsList: function (translationsList) {
+            var haveMoreTranslations = translationsList.length > 1,
+                mainTranslation = this.get('content').get('mainLanguageCode');
+
+            return translationsList.map(function(translation) {
```
dew326: Change the array of string to array of objects is also a BC break.

Resources/public/js/views/actions/ez-translateactionview.js
```diff
@@ -62,15 +62,35 @@ YUI.add('ez-translateactionview', function (Y) {
...
+         * decorates the list of language codes by adding true/false flag which
+         * determines whether delete button should be visible
+         *
+         * @method _decorateTranslationsList
+         * @param {Array} translationsList list of translation that will be decorated
+         * @return {Object} object which contains language codes with visibility flags
+         * @protected
+         */
+        _decorateTranslationsList: function (translationsList) {
+            var haveMoreTranslations = translationsList.length > 1,
```
sunpietro: `hasMoreTranslations`

Resources/public/js/views/actions/ez-translateactionview.js
```diff
@@ -62,15 +62,35 @@ YUI.add('ez-translateactionview', function (Y) {
...
+        /**
+         * decorates the list of language codes by adding true/false flag which
+         * determines whether delete button should be visible
+         *
+         * @method _decorateTranslationsList
+         * @param {Array} translationsList list of translation that will be decorated
+         * @return {Object} object which contains language codes with visibility flags
+         * @protected
+         */
+        _decorateTranslationsList: function (translationsList) {
```
sunpietro: Maybe `_flagRemovableTranslations`
Also `translations` instead of `translationsList`. It's nicer to read :)

Resources/public/js/views/actions/ez-translateactionview.js
```diff
@@ -62,15 +62,35 @@ YUI.add('ez-translateactionview', function (Y) {
...
             return this;
         },
 
+        /**
+         * decorates the list of language codes by adding true/false flag which
+         * determines whether delete button should be visible
+         *
+         * @method _decorateTranslationsList
+         * @param {Array} translationsList list of translation that will be decorated
+         * @return {Object} object which contains language codes with visibility flags
```
sunpietro: What you actually return is an array of objects not an object.

Resources/public/js/views/actions/ez-translateactionview.js
sunpietro: I agree

Resources/public/templates/translateaction.hbt
```diff
@@ -18,16 +18,16 @@
...
+                    <a class="ez-contenttranslation-view-link" href="{{ path "viewLocation" id=../location.id languageCode=language }}">
+                        {{ language_name language }}
                     </a>
 
                     <span class="ez-contenttranslation-actions">
-                        <a class="ez-contenttranslation-edit-link" href="{{ path "editContent" id=../content.id languageCode=this }}">
+                        <a class="ez-contenttranslation-edit-link" href="{{ path "editContent" id=../content.id languageCode=language }}">
                             {{translate 'translateaction.edit' 'bar'}}
                         </a>
-                        {{#if ../canDeleteTranslations}}
```
dew326: I think, change this into something like: `../canDeleteTranslations && ../content.mainLanguageCode !== this` should be enough for disabling the deleting main language code.

konradoboza: That was also my first idea but unfortunately this isn't possible to achieve this way. In order to do so, I'll need to create some helpers to implement logical operators `&&` or `!==`.

dew326: The helper is fine for me. The other solution would be to do `if` inside `if`, but please don't ;)

Resources/public/js/views/actions/ez-translateactionview.js
SylvainGuittard: Not sure to understand the BC here. Could you please provide more details @dew326 ?

sunpietro: If someone had overidden the translateaction view template and relies on any variable sent to the template while rendering, then it may break the platform ui app. If you don't remove template variables then it won't break anything.

SylvainGuittard: Thanks @sunpietro

konradoboza:
I reverted my changes regarding possible BC break and created new helper `notEqual` as this is sufficient for the only condition that should be checked (if the current highlighted translation isn't main one).

Resources/public/js/views/actions/ez-translateactionview.js
```diff
@@ -64,6 +66,7 @@ YUI.add('ez-translateactionview', function (Y) {
                 content: this.get('content').toJSON(),
                 translations: translationsList,
                 canDeleteTranslations: translationsList.length > 1,
+                mainLanguageCode: mainLanguageCode,
```
dew326: seems to not used in template

Resources/public/js/views/actions/ez-translateactionview.js
```diff
@@ -45,7 +45,9 @@ YUI.add('ez-translateactionview', function (Y) {
          */
         render: function () {
             var container = this.get('container'),
-                translationsList = this.get('content').get('currentVersion').getTranslationsList(),
+                content = this.get('content'),
+                translationsList = content.get('currentVersion').getTranslationsList(),
+                mainLanguageCode = content.get('mainLanguageCode'),
```
dew326: those changes also can be reverted as you don't need `mainLanguageCode`

Resources/public/templates/translateaction.hbt
```diff
@@ -26,9 +26,9 @@
                         <a class="ez-contenttranslation-edit-link" href="{{ path "editContent" id=../content.id languageCode=this }}">
                             {{translate 'translateaction.edit' 'bar'}}
                         </a>
-                        {{#if ../canDeleteTranslations}}
-                        <a class="ez-contenttranslation-delete-link ez-font-icon" data-translation="{{.}}" href="{{ path "editContent" id=../content.id languageCode=this }}"></a>
-                        {{/if}}
+                        {{#notEqual ../content.mainLanguageCode this }}
+                            <a class="ez-contenttranslation-delete-link ez-font-icon" data-translation="{{ this }}" href="{{ path "editContent" id=../content.id languageCode=this }}"></a>
```
dew326: Is there any reason why you changed `{{ . }}` into `{{ this }}`?

konradoboza: It's better to read, as `this` is used in every other place for this value. I can revert it.

dew326: I was just curious. For me, it can stay.

Resources/public/js/views/actions/ez-translateactionview.js
konradoboza: You are right, fixed.

konradoboza:
ping @sunpietro

PR #961, opened 2018-03-27
EZEE-1876: "Confirm" event data fired by Y.eZ.DraftConflictView._confirmDraftEdition should contain selected version id and status

adamwojs:
> JIRA: https://jira.ez.no/browse/EZEE-1876

## Description 

This PR adds id and status of selected version in the `confirm` event data fired by `Y.eZ.DraftConflictView._confirmDraftEdition`

### DIFF
--- a/Resources/public/js/views/confirmbox/ez-draftconflictview.js
+++ b/Resources/public/js/views/confirmbox/ez-draftconflictview.js
@@ -139,7 +139,11 @@ YUI.add('ez-draftconflictview', function (Y) {
              * Fired to redirect to the draft edition
              * @event confirm
              */
-            this.fire('confirm', {route: e.target.getAttribute('href')});
+            this.fire('confirm', {
+                route: e.target.getAttribute('href'),
+                versionId: e.target.getAttribute('data-version-id'),
+                versionStatus: e.target.getAttribute('data-version-status')
+            });
         },
 
     }, {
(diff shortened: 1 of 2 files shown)


sunpietro:
I'm missing the description why it's needed. What problem it solves?

adamwojs:
@sunpietro I need it for https://github.com/ezsystems/StudioUIBundle/pull/890 to be able to redirect editor to selected version of the landing page. See https://github.com/ezsystems/StudioUIBundle/pull/890/commits/6fb1feab769bbb3922a3f161d7c0c65862ee072a#diff-22fe29932eeed29a81c0f1f056150054R164

sunpietro:
My proposal is that you should refactor a bit the code to be able to use a custom plugin (from the StudioUIBundle) that changes a way of redirecting a user to a proper version. If we are going to follow the rules that were previously introduced, that is no unneccesary code related to Studio inside the PlatformUI codebase, then it's the only way.

cc @andrerom @lserwatka

Nattfarinn:
A can agree that we should still avoid making unnecessary dependencies between PlatformUI and StudioUI, but it's not a dependency here. Additional code is used by Studio, thats true, but it's not related to Studio code base in any way.

I guess explanation here would be enough, because it is just exposing more information we already have in Platform and use in the event from the very same `a href` element.

It's more like an improvement on existing extensibility point.

andrerom:
Assuming @sunpietro is now ok with this, then this has enough approvals to be sent to QA here @adamwojs

adamwojs:
@micszo This issue is caused by missing changes from https://github.com/ezsystems/ezpublish-kernel/pull/2268 in your installation (Platform UI counterpart is already included in this branch). 

Please apply patch from above PR or change the "ezsystems/ezpublish-kernel" dependency to "~6.13@dev"  in your `composer.json` and everything should be fine 😉

micszo:
@adamwojs Switching to branch 6.13 in kernel helps, applying patch doesn't (hunk succeeds with offset).

PR #962, opened 2018-04-11
EZP-28953: RichText editor's toolbox obscures the preceding paragraph's content

adamwojs:
>JIRA: https://jira.ez.no/browse/EZP-28953

## Description 

Video presenting how the patch works:  https://youtu.be/BdwGLoTe6qw 

## Steps to reproduce

1. Create the new Article
2. Into the Intro/Body (or any other RichText field) paste some dummy text in few paragraphs
3. Click on text to get RichText editor toolbox. You will see, that toolbox obscures some parts of the text as shown in the screenshot.

## TODO:

- [x] Fixing unit tests for JS

PR #963, opened 2018-04-16
EZP-28953: Added debounce for updating position of the floating toolbar

adamwojs:
>JIRA: https://jira.ez.no/browse/EZP-28953

Added debounce for updating the position of the floating toolbar as @sunpietro suggested in https://github.com/ezsystems/PlatformUIBundle/pull/962#discussion_r181637986

### DIFF
--- a/Resources/public/js/alloyeditor/plugins/floatingtoolbar.js
+++ b/Resources/public/js/alloyeditor/plugins/floatingtoolbar.js
@@ -11,7 +11,9 @@ YUI.add('ez-alloyeditor-plugin-floatingtoolbar', function (Y) {
     }
 
     var FLOATING_TOOLBAR_SELECTOR = '.ae-toolbar-floating',
-        FLOATING_TOOLBAR_FIXED_CLASS = 'ae-toolbar-floating-fixed';
+        FLOATING_TOOLBAR_FIXED_CLASS = 'ae-toolbar-floating-fixed',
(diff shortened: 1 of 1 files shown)


micszo:
@adamwojs Good news first. 
It seems to work correctly on Firefox, Edge and Safari.
The bad news refers to Chrome. 
There seems to be some performance issue. 
I created an article with 15 paragraphs from lorem ipsum generator (pl.lipsum.com) and after ca. 20 clicks on paragraphs to browser becomes very slow, there's delay on scrolling, delay on setting focus to paragraph. If you play with it long enough macOS actually crushes. 
Please confirm Chrome behaviour on your side.
\cc @dew326

andrerom:
Wow, thats strange. I did not see that one, but I also did not click 20 times or more on the paragraphs. sounds like event leaking  _(more and more listeners attach themselves to the event)_ or something.

@adamwojs did you reproduce this?

adamwojs:
@andrerom @dew326 I will check it.

micszo:
It seems to be due to the _Check Spelling While Typing_ function used when there are many mistakes per paragraph (Latin text in English browser). #everydayisaschoolday

sunpietro:
@micszo @adamwojs can we avoid this kind of issues in the future?

micszo:
I will remember this till my deathbed. ;)

andrerom:
So this was false positive? aka Chrome spellchecker bug?

adamwojs:
@andrerom Exactly 😉

PR #964, opened 2018-04-24
EZP-29106: Show author name in version list / draft conflict dialog

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-29106
> **Depends on: https://github.com/ezsystems/ez-js-rest-client/pull/89, https://github.com/ezsystems/ezpublish-kernel/pull/2319** 

## Problem 

Currently, there is API path displayed only instead of full username in the "Versions" tab of location view and in the draft conflict dialog. This is raised _(understandably)_ as a large issue for editors of prominent customers.

## Solution

In general the idea is to extend the response of `GET /api/ezp/v2/content/objects/52/versions` for the Platform UI needs. This approach allow us to avoid:
- Additional REST requests
- Problems with lack of permissions to read user data by non-administrators

The customization is made by intoducting new media type `application/vnd.ez.platformui.RESOURCE+json` handled by `ezsystems.platformui.rest.output.visitor.json` service.

## Screenshots

Versions list             |  Draft conflict dialog
:-------------------------:|:-------------------------:
| screenshot-2018-4-24 home - ez studio | screenshot-2018-4-24 home - ez studio 1 |

## TODO: 

- [X]  Open PR on `ez-js-rest-client` repository which introduce ability to override `Accept` header  (https://github.com/ezsystems/ez-js-rest-client/pull/89)
- [X] Refactor  `eZ\Publish\Core\REST\Server\Output\ValueObjectVisitor\VersionInfo` to be able avoid code duplication in `EzSystems\PlatformUIBundle\Rest\ValueObjectVisitor\VersionInfo` (https://github.com/ezsystems/ezpublish-kernel/pull/2319)
- [x] Implement tests. 
- [x] Fix new code according to Coding Standards (`$ composer fix-cs`).
- [x] Ask for Code Review.
- [x]  Change composer requirements for kernel, & js rest client

### DIFF
--- a/Rest/ValueObjectVisitor/VersionInfo.php
+++ b/Rest/ValueObjectVisitor/VersionInfo.php
@@ -1,0 +1,55 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\PlatformUIBundle\Rest\ValueObjectVisitor;
+
+use eZ\Publish\API\Repository\Exceptions\NotFoundException;
+use eZ\Publish\API\Repository\Repository;
+use eZ\Publish\Core\REST\Common\Output\Generator;
+use eZ\Publish\Core\REST\Common\Output\Visitor;
+use eZ\Publish\Core\REST\Server\Output\ValueObjectVisitor\VersionInfo as BaseVersionInfo;
+use eZ\Publish\API\Repository\Values\Content\VersionInfo as APIVersionInfo;
+
+/**
+ * VersionInfo value object visitor.
+ */
+class VersionInfo extends BaseVersionInfo
+{
+    /** @var \eZ\Publish\API\Repository\Repository */
+    private $repository;
+
+    /**
+     * VersionInfo constructor.
+     *
+     * @param \eZ\Publish\API\Repository\Repository $repository
+     */
+    public function __construct(Repository $repository)
+    {
+        $this->repository = $repository;
+    }
+
+    protected function visitVersionInfoAttributes(Visitor $visitor, Generator $generator, APIVersionInfo $versionInfo)
+    {
+        parent::visitVersionInfoAttributes($visitor, $generator, $versionInfo);
+
+        $this->visitCreatorName($visitor, $generator, $versionInfo);
+    }
+
+    protected function visitCreatorName(Visitor $visitor, Generator $generator, APIVersionInfo $versionInfo)
+    {
+        /** @var \eZ\Publish\API\Repository\Values\User\User $user */
+        $user = $this->repository->sudo(function (Repository $repository) use ($versionInfo) {
+            try {
+                return $repository->getUserService()->loadUser($versionInfo->creatorId);
+            } catch (NotFoundException $e) {
+                return null;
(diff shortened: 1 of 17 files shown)


Rest/ValueObjectVisitor/VersionInfo.php
```diff
@@ -0,0 +1,113 @@
...
+            'href',
+            $this->router->generate('ezpublish_rest_loadUser',
+                array('userId' => $versionInfo->creatorId))
+        );
+        $generator->endAttribute('href');
+        $generator->endObjectElement('Creator');
+
+        /** @var \eZ\Publish\API\Repository\Values\User\User $user */
+        $user = $this->repository->sudo(function (Repository $repository) use ($versionInfo) {
+            return $repository->getUserService()->loadUser($versionInfo->creatorId);
```
andrerom: should also handle NotFound here

two options on that, either we return a fake object, legacy had a fake user for non exiting user. Or we return null, skip the user info, so that frontend code ends up displaying api path for this case.

adamwojs: Previously I delegated responsibility for handling the missing user to the UI. Ref: https://github.com/ezsystems/StudioUIBundle/pull/864

andrerom:
👍

andrerom:
@adamwojs I guess this needs final rebase on whatever you do on final kernel change? Besides this, we need to bump composer requirements for kernel, & js rest client _(packages.json, and assets in composer.json)_.

andrerom:
@adamwojs / @sunpietro Done with rest client, but for PlatformUIAssetBundle to build I needed to apply same fix as done to bower.json in this repo: https://github.com/ezsystems/PlatformUIAssetsBundle/commit/be43e2c5ae57ce9b284b6af5aef9a366863cd87c


Q: However this is src and not a build, so I assume we will need adapt our bin script in assets bundle to build and remove src manually before we can release 4.2.0. Right?

adamwojs:
> Q: However this is src and not a build, so I assume we will need adapt our bin script in assets bundle to build and remove src manually before we can release 4.2.0. Right?

@andrerom You mean additional `src` directory with wasn't there before changes intoduced in https://github.com/ezsystems/PlatformUIAssetsBundle/commit/be43e2c5ae57ce9b284b6af5aef9a366863cd87c ?  I see that in case of other dependencies (alloy-editor, handlebars-helper-intl) we are also including `src` in final tag. 

The downloaded package also contains `build` directory which seems to be similar to others releases .  Was it different in your case?

andrerom:
Found a way... (https://github.com/ezsystems/PlatformUIAssetsBundle/commit/70425ce9687f15c1cfc94b456523f23b3f5894d3)

So v4.2.0 & v5.1.0 of PlatformAssetBundle is out, @adamwojs  you can now adjust composer.json for both kernel and assets.

adamwojs:
@andrerom Done (I also squashed fixup commit)

composer.json
```diff
@@ -9,9 +9,9 @@
         }
     },
     "require": {
-        "ezsystems/platform-ui-assets-bundle": "^4.1.0",
+        "ezsystems/platform-ui-assets-bundle": "^4.1.1",
```
lserwatka: I think you need 4.2.0? There is no 4.1.1

PR #969, opened 2018-05-16
[Just For Testing] Merge up 1 7 to 1 13

mnocon:


andrerom:
⚠️ If we do this like this we need to do this as a merge commit, not as squash!

_(We have configured Github to only allow squash to remove history from PR's by default, to avoid commits like "!fixup" and so on. But branch merges needs specifically to keep merge history in order for Git and Github to know if a branch contains a commit or not, as squash changes commit sha as well)_

So this is why we usually merge directly via command line and not via PR's, any reasons why this is done on PR? Maybe solution is just to ask merger to remember to merge up?

mnocon:
What happens here:
I've merged something to 1.7 (https://github.com/ezsystems/PlatformUIBundle/pull/965) and wanted to merge it up, but there were some conflicts in JS files (that weren't mine). I've asked @adamwojs if he could help me resolve it (and thankfully he did).

I have to make some changes when merging upstream, so we agreed that Adam will push his changes to a temporal branch and I will end the merge process from there.

If I understand you correctly: everything will be all right if I close this PR and end the merge from the command line, as it's a GitHub specific setting that we want to avoid happening?

andrerom:
> If I understand you correctly: everything will be all right if I close this PR and end the merge from the command line

yes. So if the intentions was to just open PR to run travis test, just make a habit out of clearly prefixing PR summary with `[Just For Testing]` or something like that.

> as it's a GitHub specific setting that we want to avoid happening?

It's not actually github specific, what we have configured github to do is basically the same as:
git merge --squash <your_pr_branch>

This is good for dev work  on PRs where we don't need the history in the target branch in 99% of the cases. However for our own merges from one branch to another, it means the three commits here would end up as one commit with new sha and message `Merge up 1 7 to 1 13`.

PR #973, opened 2018-06-15
EZP-29239: Keeping correct order while loading RelationList items

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-29239

## Description 

Related content objects are loaded asynchronous which might lead to the wrong order of elements when rendering value of ezrelationlist field type.

### DIFF
--- a/Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
+++ b/Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
@@ -48,6 +48,7 @@ YUI.add('ez-objectrelationsloadplugin', function (Y) {
                 loadingError = false,
                 sourceContent = e.content,
                 destinationContentIds = e.destinationContentIds,
+                withLocation = e.loadLocation || e.loadLocationPath,
                 contentDestinations,
                 end = stack.add(function (error, struct) {
                     if (error) {
@@ -77,7 +78,7 @@ YUI.add('ez-objectrelationsloadplugin', function (Y) {
                 if (!loadedRelation[value]) {
                     loadedRelation[value] = true;
 
-                    if (e.loadLocation || e.loadLocationPath) {
+                    if (withLocation) {
                         this._loadContentStruct(value, e.loadLocation, e.loadLocationPath, end);
                     } else {
                         this._loadContent(value, end);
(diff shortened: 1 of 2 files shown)


Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
```diff
@@ -86,6 +86,20 @@ YUI.add('ez-objectrelationsloadplugin', function (Y) {
             }, this);
 
             stack.done(function () {
+                relatedContentListArray.sort(function (a, b) {
+                    var indexA, indexB;
+
+                    if (e.loadLocation || e.loadLocationPath) {
```
sunpietro: You should cache it before sorting to avoid computations in the sort loop

Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
```diff
@@ -86,6 +86,20 @@ YUI.add('ez-objectrelationsloadplugin', function (Y) {
             }, this);
 
             stack.done(function () {
+                relatedContentListArray.sort(function (a, b) {
+                    var indexA, indexB;
```
sunpietro: Please, each variable should be placed in a new line.

Resources/public/js/views/services/plugins/ez-objectrelationsloadplugin.js
```diff
@@ -48,6 +48,7 @@ YUI.add('ez-objectrelationsloadplugin', function (Y) {
                 loadingError = false,
                 sourceContent = e.content,
                 destinationContentIds = e.destinationContentIds,
+                withLocation = e.loadLocation || e.loadLocationPath,
```
sunpietro: It's better to use `e.hasOwnProperty('loadLocation') || e.hasOwnProperty('loadLocationPath')`

adamwojs: I'm checking here if one of those flags is TRUE. See method documentation for more details.

sunpietro: Right, good point.

lserwatka:
@micszo we must proceed with QA here.

micszo:
Retest OK on eZ Platform v1.7.7 with patch from diff.
Moving on to testing on v1.13.3.

lserwatka:
@adamwojs this one can be merged up.

adamwojs:
@lserwatka done.

PR #975, opened 2018-07-04
EZP-29254: Images embedded in RichText editor dissapear after undo

dew326:
**Jira ticket:** https://jira.ez.no/browse/EZP-29254

### DIFF
--- a/Resources/public/js/alloyeditor/plugins/embed.js
+++ b/Resources/public/js/alloyeditor/plugins/embed.js
@@ -86,6 +86,16 @@ YUI.add('ez-alloyeditor-plugin-embed', function (Y) {
                     this._getEzConfigElement();
                     this.setWidgetContent('');
                     this._cancelEditEvents();
+                    this.disableUndo();
+                },
+
+                /**
+                 * Disables the undo in CKEditor.
+                 *
+                 * @method disableUndo
+                 */
+                disableUndo: function() {
+                    editor.setKeystroke(CKEDITOR.CTRL + 90, false);
                 },
 
                 /**


lserwatka:
@dew326 could you merge it up?

konradoboza:
As I can see, this fix is about disabling undoing in RichText Editor. I think it isn't the best way to fix it, as editors rely on this feature. wdyt? @lserwatka @andrerom

lserwatka:
@dew326 what is your take on question from @konradoboza ? It this possible to fix without disabling undo feature at all?

dew326:
Well from my investigation we cannot fix it. It was correct behavior from the CKEditor perspective. 
They create snapshots and restore it on undo. The snapshot is taken as a proper string so without our additional data like images/content placeholders. When CKEditor restores the snapshot we need to load them (that's why we had this disappearing images).

BTW if there was content embed the cmd-Z didn't work at all it was just removing and loading images once again, so it's hard to tell that someone uses it.

lserwatka:
@dew326 how much effort would be to add such a place holders, so they can be restored?

dew326:
I don't think we should do something like that. This will provide inconsistency in CKEditor data.

andrerom:
@dew326 Is there a way to hook into the restore event? So we can load the images again? _(I assume we do that on rendering and when we insert elements from UDW)_

dew326:
As I understand from the ticket the reloading is also unwanted behavior.
ref: `In v2 behaviour is a bit odd: images disappear, the small rectangle shows up but immediately after that images appear once again.`

andrerom:
This is input from @konradoboza, not from customer.

So assuming we ignore that, and threat that as something to consider improving in the future: _Is it possible to hook into restore and reuse our loading logic to load images on undo?_

dew326:
Yes, this should be possible to do it like in v2.

konradoboza:
@dew326 if I understood customer's intention, the real problem is disappearing images after undoing some operation. I described behavior in v2 and give it under consideration. I think it might be a better idea to achieve a similar solution rather than disabling whole "Ctrl+Z" feature. I set v2 as affected version as in my opinion undoing operation shouldn't affect images at all.

andrerom:
@dew326 to summarize this thread, would you agree that we have two things here:
bug: v1 images lost on undo, can potentially work like v2
story: v2 improve the handling of images in OE
  - (ASSUMPTIONs:) potentially get `xhmtl5-edit` format to inline image paths and image dimensions so OE does not need to load images async over REST, as a bonus gets rid of janky images loading on initial load and on undo operations

PR #976, opened 2018-07-05
EZP-29353 UDW: unwanted highlight when clicking on content items in Firefox

mateuszbieniek:
| Question           | Answer
| ------------------ | ------------------
| **JIRA issue**     | [EZP-29353](https://jira.ez.no/browse/EZP-29353)
| **Bug/Improvement**| yes
| **Target version** | 1.13

This PR fixes unwanted highlighting of the content in UDW on Firefox 60.0.1+

### DIFF
--- a/Resources/public/js/views/universaldiscovery/ez-universaldiscoveryfinderview.js
+++ b/Resources/public/js/views/universaldiscovery/ez-universaldiscoveryfinderview.js
@@ -89,6 +89,13 @@ YUI.add('ez-universaldiscoveryfinderview', function (Y) {
          * @param {eventFacade} e.data the node data
          */
         selectContent: function (e) {
+            //fix for firefox 60.0+ selection start on click
+            if (navigator.userAgent.search("Firefox") > -1 ) {
+                if(window.getSelection) {
+                    window.getSelection().removeAllRanges();
+                }
+            }
+
             this._fireSelectContent(e.data);
             this.get('selectedView').set('contentStruct', e.data);
         },


Resources/public/js/views/universaldiscovery/ez-universaldiscoveryfinderview.js
```diff
@@ -89,6 +89,11 @@ YUI.add('ez-universaldiscoveryfinderview', function (Y) {
          * @param {eventFacade} e.data the node data
          */
         selectContent: function (e) {
+            //fix for firefox 60.0+ selection start on click
```
dew326: If only FF is affected you should check if the user uses FF.

mateuszbieniek:
Highlights are not the problem: for some reason when you click on an item in UDW in Firefox 60.0.1+ it starts selecting text (just like it would when you click and hold it down). Please see this video:

unwanted_highlight

sunpietro:
@mateuszbieniek then it looks like a browser issue.

Resources/public/js/views/universaldiscovery/ez-universaldiscoveryfinderview.js
mateuszbieniek: Done

mateuszbieniek:
Maybe, but the customer expects it to work as intended. The code in this PR fixes the issue with firefox and does not disable selection, just this unwanted effect (you can still select all you want inside UDW - if you want to).

sunpietro:
I'm not saying to disable selection of everything. I just prefer adding some styles over JS function that completely removes the possibility of selecting anything. You just want to remove colors. Please, use the right tools to fix the issue.

mateuszbieniek:
Provided code does not disable selecting. Just unwanted effect - you can still select everything, including items in UDW. I'm not sure you can fix it with CSS - if I disable selection highlights in CSS (as you mentioned earlier) it will disable it for manual selection too (when users intend to do so).

micszo:
Hi guys, are you aware of the Grover / PhantomJS test result? Is that acceptable?

mateuszbieniek:
Hi @micszo - it's not related to this PR changes.

mateuszbieniek:
ping @lserwatka - it seems PR is ready to merge.

PR #980, opened 2018-07-19
EZP-28785: Unable to publish draft after changing its translation

mateuszbieniek:
| Question           | Answer
| ------------------ | ------------------
| **JIRA issue**     | [EZP-28785](https://jira.ez.no/browse/EZP-28785)
| **Bug/Improvement**| yes
| **Target version** | 1.7
| **BC breaks**      | no
| **Tests pass**     | no
| **Doc needed**     | no

After saving new content item we are no longer "creating" content, but editing - so option to change Draft language should not be available (same behavior as editing Draft that was created before). This fix makes new content Language Dialog aware of content status, so It won't allow changing Languages after pressing "save".

**TODO**:
- [x] Implement feature / fix a bug.
- [x] Fix tests
- [x] Ask for Code Review.

### DIFF
--- a/Resources/public/js/views/services/ez-contentcreateviewservice.js
+++ b/Resources/public/js/views/services/ez-contentcreateviewservice.js
@@ -140,6 +140,9 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
          * @param {EventFacade} e
          */
         _selectLanguage: function (e) {
+            // if the content is considered new: allow changing its language from the list of available languages
+            var isChangingLanguageAllowed = this.get('content').isNew();
+
             e.preventDefault();
             this.fire('languageSelect', {
                 config: {
@@ -147,7 +150,7 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
                     languageSelectedHandler: Y.bind(this._setLanguage, this, e.target, e.fields),
                     cancelLanguageSelectionHandler: null,
                     canBaseTranslation: false,
-                    translationMode: true,
+                    translationMode: isChangingLanguageAllowed,
                     referenceLanguageList: [this.get('languageCode')]
                 },
             });
--- a/Resources/public/js/views/services/ez-contentcreateviewservice.js
+++ b/Resources/public/js/views/services/ez-contentcreateviewservice.js
@@ -140,8 +140,7 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
          * @param {EventFacade} e
          */
         _selectLanguage: function (e) {
-            // if the content is considered new: allow changing its language from the list of available languages
-            var isChangingLanguageAllowed = this.get('content').isNew();
+            var isTranslationAllowed = this.get('content').isNew();
 
             e.preventDefault();
             this.fire('languageSelect', {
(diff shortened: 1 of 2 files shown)


Resources/public/js/views/services/ez-contentcreateviewservice.js
```diff
@@ -140,14 +140,16 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
          * @param {EventFacade} e
          */
         _selectLanguage: function (e) {
+            var isNewContent = this.get('content').get('currentVersion').get('status') === '';
```
dew326: I think the `this.get('content')` has method `isNew()`.

Resources/public/js/views/services/ez-contentcreateviewservice.js
```diff
@@ -140,14 +140,16 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
          * @param {EventFacade} e
          */
         _selectLanguage: function (e) {
+            var isNewContent = this.get('content').isNew();
```
sunpietro: I'm not sure what is the relation between `isNewContent` and the `translationMode`. Maybe you should rename variable to make it more meaningful.

Resources/public/js/views/services/ez-contentcreateviewservice.js
```diff
@@ -140,14 +140,17 @@ YUI.add('ez-contentcreateviewservice', function (Y) {
          * @param {EventFacade} e
          */
         _selectLanguage: function (e) {
+            // if the content is considered new: allow changing its language from the list of available languages
+            var isChangingLanguageAllowed = this.get('content').isNew();
```
sunpietro: It would name it: `isTranslationAllowed`

mateuszbieniek: Variable renamed as suggested. Ping @sunpietro.

mateuszbieniek:
Default behavior that we can encounter in v1 is that change link is always visible, but when changing the language is not available dialog allows to only choose existing translation (for example: create draft, save it and close it, then reopen it - you will see "change" in top right corner, but clicking it won't let you change translation - only already selected will be available). So I think that removing "change" link is out of the scope of this PR and I'm not 100% sure it is what we want (if so we have to change some default behavior).

mnocon:
@mateuszbieniek I think we can merge this one

SylvainGuittard:
Maybe not too late to comment on this ticket. 
Regarding the following steps:
```
1. Create new language (e.g. pl-PL) in PlatformUI and add it to ezplatform.yml
2. Create new content (e.g. an article) in the default language (in this case en-GB).
3. Save draft.
4. Change translation of this draft.
5. Hit "Save" or "Publish".
```
How many translations does this content have on the published version?

mateuszbieniek:
@SylvainGuittard after following steps above you won't be able to publish draft at all (form will not validate) unless you will close and reopen it. Then you will be only presented with the language used to create the draft (with no option to change it - I think this is expected behavior for editing the draft) and publication will not fail (with only single translation - main Language used for the Draft).

mnocon:
@SylvainGuittard @mateuszbieniek 
It's possible to execute this scenario without this fix and not receive an error, but it's required to use a Content Type that does not have required/validated fields other than the title.

`Folder` works - validation error doesn't occur when there's no validation 🙃

SylvainGuittard:
Thanks guys for the clarification. Good for me!

mnocon:
@lserwatka I think we can merge this one

mnocon:
@lserwatka I think we can merge this one

mateuszbieniek:
ping @andrerom @lserwatka

PR #981, opened 2018-07-30
EZP-29455: Mark test as Platform only

mnocon:
All tests of PlatformUI are passing when run on ezplatform-ee, except for this one. 

I'm currently setting up running these tests here: https://github.com/ezsystems/StudioUIBundle/pull/906 and this test will be filtered out.


mnocon:
@lserwatka we can merge this one

PR #982, opened 2018-08-07
EZP-29286: As a Platform UI User, I would like my session to be extended while I still use Platform UI

sunpietro:
https://jira.ez.no/browse/EZP-29286

## Requirements

~https://github.com/ezsystems/ez-js-rest-client/pull/92~ _Merged, need to be tagged as 1.5.4 once passed by QA_

### DIFF
--- a/Resources/public/js/views/services/ez-contenteditviewservice.js
+++ b/Resources/public/js/views/services/ez-contenteditviewservice.js
@@ -36,6 +36,7 @@ YUI.add('ez-contenteditviewservice', function (Y) {
             this.after('*:closeView', this._redirectAfterClose);
             this.after('discardedDraft', this._redirectAfterDiscard);
             this.after('publishedDraft', this._redirectAfterPublish);
+            this.after('contentEditView:activeChange', this._refreshSession);
 
             this._setLanguageCode();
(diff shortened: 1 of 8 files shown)


Resources/public/js/views/services/ez-contenteditviewservice.js
```diff
@@ -36,9 +36,11 @@ YUI.add('ez-contenteditviewservice', function (Y) {
             this.after('*:closeView', this._redirectAfterClose);
             this.after('discardedDraft', this._redirectAfterDiscard);
             this.after('publishedDraft', this._redirectAfterPublish);
+            this.after('contentEditView:activeChange', this._refreshSession);
```
dew326: I think the same problem will be in `create`. Maybe it would be good to `refreshSession` in the entire application not only in edit mode?

sunpietro:
ping @dew326

andrerom:
Merging as failure is not related to fixes here.

To avoid this breaking when release the js-rrest-client change has been tagged:
- https://github.com/ezsystems/ez-js-rest-client/releases/tag/v1.5.4
- https://github.com/ezsystems/PlatformUIAssetsBundle/releases/tag/v4.2.2

PR #986, opened 2018-09-05
EZP-29234: As a developer, I want to read about customizing REST API response

konradoboza:
JIRA issue: https://jira.ez.no/browse/EZP-29234
Depends on https://github.com/ezsystems/ezpublish-kernel/pull/2434

This PR is a part of simplifying REST response customization. Custom JSON Generator is no longer needed, as in `services.yml` we rely on kernel's implementation.


konradoboza:
I reverted changes in `composer.json` and `travis.yml` as I was not able to force PlatformUI to get dependency from kernel branch. @andrerom probably "hulk smash" approach will be better in this case... 😈

andrerom:
restarted travis :)

